PR #243, opened 2015-01-15
[#87977] Run bulk order import as a delayed job

drench:
This involved a fairly major refactor of the `OrderImport` model. There are still opportunities for more (some flagged with `TODO`).


app/controllers/order_imports_controller.rb
```diff
@@ -15,23 +15,23 @@ def initialize
...
-    render action: 'show'
+    redirect_to new_facility_order_import_path
   end
 
   private
 
-  def flash_now(key, message)
-    flash.now[key] = truncate(message, length: 2048)
+  def set_flash(key, message)
+    flash[key] = truncate(message, length: 2048)
```
jhanggi: Do we need to worry about this truncation anymore? If not, it might be nice to just go back to plain `flash[:error] = I18n.t(...)` instead of this extracted method.

app/controllers/order_imports_controller.rb
```diff
@@ -44,51 +44,37 @@ def create_order_import!
...
-    I18n.t('controllers.order_imports.create.error', error: exception.message)
+  def deliver_report
+    CsvReportMailer.delay.csv_report_email(report_recipient, report)
+    set_flash(:notice, t("controllers.order_imports.create.job_is_queued", email: @report_recipient))
   end
 
-  def import_success_message
-    I18n.t 'controllers.order_imports.create.success', @import_result.to_h
+  def report_recipient
+    @report_recipient ||= params[:report_recipient] ||= @order_import.creator.email
```
jhanggi: The double `||=` feels really weird. Also, does this handle the empty string `params[:report_recipient]`? Would `@report_recipient ||= params[:report_recipient].presence || @order_import.creator.email` work?

app/controllers/order_imports_controller.rb
```diff
@@ -44,51 +44,37 @@ def create_order_import!
...
+  def process_order_import!
+    raise "Please upload a valid import file" if upload_file.blank?
+    @order_import = create_order_import!
+    @order_import.process_upload!
+    deliver_report
   end
 
-  def import_exception_message(exception)
-    I18n.t('controllers.order_imports.create.error', error: exception.message)
+  def deliver_report
```
jhanggi: Would `queue_report` be a more accurate method name?

spec/app_support/order_row_importer_spec.rb
```diff
@@ -0,0 +1,121 @@
+require "spec_helper"
+
+describe OrderRowImporter do
+  subject { OrderRowImporter.new(row, order_import) }
+  let(:order_import) { OrderImport.new }
+
+  describe "#import" do
+    context "when the fulfillment date is invalid" do
+      it "does not import the row"
```
jhanggi: Looks like a bunch of empty tests here.

spec/models/order_import_spec.rb
```diff
@@ -1,342 +1,376 @@
...
+  let(:product_name) { "Example Item" }
+  let(:quantity) { 1 }
+  let(:order_date) { nucore_format_date(default_order_date) }
+  let(:fulfillment_date) { nucore_format_date(default_fulfilled_date) }
+  let(:note) { "Test note" }
+
+  before(:all) { create_users }
 
   before :each do
     Timecop.freeze(fiscal_year_beginning + 5.days)
```
jhanggi: There's a `:timecop_freeze` metatag option you could use, and then not have to worry about the `Timecop.return` later on. You just need to add a `let(:now)` for defining what time to freeze at.

jhanggi:
Done with my review. Overall, this looks like a pretty good refactoring; obviously there are more places to refactor, but this definitely looks better than it did. Just a few minor comments (and some blank specs).

app/controllers/order_imports_controller.rb
drench: Good catch. I did not mean to put a double `||=` here (or really, anywhere).

drench:

app/models/reports/order_import.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def failure_message_key
+    if @order_import.fail_on_error?
+      "reports.order_import.fail_immediately"
+    else
+      "reports.order_import.fail_continue_on_error"
+    end
+  end
+end
+__END__
```
jhanggi: Is the string below here being used anywhere? If so, I think I'd rather see it in the locales.

(I confess, I've never seen `__END__` before.

app/support/order_row_importer.rb
```diff
@@ -0,0 +1,206 @@
...
+  def validate_account
+    return if user.blank? || product.blank?
+    if account.present?
+      add_error(account.validate_against_product(product, user))
+    else
+      add_error("Can't find account")
+    end
+  end
+
+  def validate_fields
```
jhanggi: Probably not worth it at this point, but we could probably use `ActiveModel::Validations` for this.

app/models/reports/order_import.rb
drench: It's kind of a Perlism. I had some notes after the `__END__` and forgot to delete them.

spec/app_support/order_row_importer_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+      it "does not create an order" do
+        expect { subject.import }.not_to change(Order, :count)
+      end
+
+      it "does not add an order_detail" do
+        expect { subject.import }.not_to change(OrderDetail, :count)
+      end
+    end
+
+    context "when the fulfillment date is invalid" do
```
jhanggi: Lately, I've been liking to explicitly declare what's invalid within the context. That way if somebody else changes something at a higher level, this test doesn't get affected. Also, it makes traveling up the stack to find why it's invalid easier.

spec/app_support/order_row_importer_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+          attributes_for(:service_price_policy, price_group: price_group)
+        )
+      end
+
+      context "and it requires a survey" do
+        before { Service.any_instance.stub(:active_survey?).and_return(true) }
+
+        it_behaves_like "an order was not created"
+
+        it "flags the service as requiring a survey" do
```
jhanggi: We could do a shared example with a parameter:
`it_behaves_like "has error message", /Service requires survey/`.
That could shorten the spec and make it easier to read I think. Just a thought.

spec/app_support/order_row_importer_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+          subject.import
+          expect(subject.errors)
+            .to include_a_string_matching(/Service requires template/)
+        end
+      end
+    end
+
+    context "when the user has an account for the product's facility" do
+      let(:chart_string) { account.account_number }
+      let(:fulfillment_date) { "1/1/1999" }
```
jhanggi: Not sure if we have some other tests around it, but we should validate that something like `1/10/1999` is January 1 and not October 1. Also, we might want some specs around `1/1/99`. Can't remember if that's supposed to be an error or not.

spec/models/order_import_spec.rb
```diff
@@ -1,342 +1,374 @@
...
-    @authable         = FactoryGirl.create(:facility)
-    @facility_account = @authable.facility_accounts.create!(FactoryGirl.attributes_for(:facility_account))
+  let(:row_data) {[ username, account_number, product_name, quantity, order_date, fulfillment_date, note ]}
+  let(:username) { guest.username }
+  let(:account_number) { "111-2222222-33333333-01" }
+  let(:product_name) { "Example Item" }
+  let(:quantity) { 1 }
+  let(:order_date) { nucore_format_date(default_order_date) }
+  let(:fulfillment_date) { nucore_format_date(default_fulfilled_date) }
+  let(:note) { "Test note" }
```
jhanggi: Jesus, this is a lot of setup.

spec/models/order_import_spec.rb
```diff
@@ -1,342 +1,374 @@
...
-    Timecop.return
+  context "validations" do
+    it { should belong_to :creator }
+    it { should belong_to :upload_file }
+    it { should belong_to :error_file }
+    it { should validate_presence_of :upload_file }
+    it { should validate_presence_of :created_by }
   end
 
+  describe "#import_row (low-level behavior)" do
```
jhanggi: Since we have separate specs for `OrderImportRow`, do we need all these specs here now?

app/models/order_import.rb
```diff
@@ -1,142 +1,101 @@
...
+    self.error_file = nil
+    self.error_report = nil
+  end
 
+  def error_mode?
+    @in_error_mode
+  end
 
-    # loop over the order keys
+  def handle_save_clean_orders # TODO refactor and rename
```
jhanggi: Is there anything we can do to make `handle_save_clean_orders` and `handle_save_nothing_error` more inline with each other? One is doing the CSV reading, while the other is delegating out to another method. Definitely makes it hard to follow.

If that's what you meant by "refactor and rename" then we can hold off for a little bit.

drench:
I removed the "low level" section from the `OrderImport` specs and moved reworked versions of most of them to `OrderRowImporter`. (Most, because some of these were redundant). This also allowed `OrderImport#import_row` to become a private method.

drench:
Ready for another look.

db/migrate/20150121202952_add_processed_at_to_order_import.rb
```diff
@@ -0,0 +1,13 @@
+class AddProcessedAtToOrderImport < ActiveRecord::Migration
+  def up
+    add_column :order_imports, :processed_at, :timestamp, after: :created_by
+
+    OrderImport.where("error_file_id IS NOT NULL").each do |order_import|
```
jhanggi: Why are you doing this restriction? Even if an import failed, shouldn't it still be `processed`?

drench: I suppose it could just be `OrderImport.all`, assuming (probably safely) that all existing `OrderImport`s would have been processed when this migration runs.

jhanggi: The migration will run as part of the deploy, so there won't be any unprocessed imports at the time we start this off.

jhanggi:
Overall, looks good. I did some testing and everything seems to be working fine now. The "In process" marker is definitely valuable, but I'm not sold that we need a separate "processed at" column. Usually, I expect it to be within a minute of the created at. We can put it out there and listen to their feedback, though.

PR #244, opened 2015-01-27
[#88161] Anticipate missing data in API responses

drench:
This should more gracefully handle cases where an `order_detail` is has no associated account, or (as unlikely as this may be) where the `order_detail` does not exist.

A related PR for IMSERC is up next.

### DIFF
--- a/app/controllers/api/order_details_controller.rb
+++ b/app/controllers/api/order_details_controller.rb
@@ -1,5 +1,6 @@
 class Api::OrderDetailsController < ApplicationController
   respond_to :json
+  rescue_from ActiveRecord::RecordNotFound, with: :order_detail_not_found
 
   http_basic_authenticate_with name: Settings.api.basic_auth_name, password: Settings.api.basic_auth_password
 
@@ -31,8 +32,13 @@ class Api::OrderDetailsController < ApplicationController
   end
 
   def order_detail
-    @order_detail = OrderDetail.find_by_id(params[:id]) unless defined?(@order_detail)
-    @order_detail
+    @order_detail ||= OrderDetail.find(params[:id])
+  end
+
+  def order_detail_not_found
+    respond_to do |format|
+      format.json { render text: { error: "not found" }.to_json, status: 404 }
+    end
   end
 
   def user_to_hash(user)
--- a/app/controllers/api/order_details_controller.rb
+++ b/app/controllers/api/order_details_controller.rb
@@ -5,48 +5,16 @@ class Api::OrderDetailsController < ApplicationController
   http_basic_authenticate_with name: Settings.api.basic_auth_name, password: Settings.api.basic_auth_password
 
   def show
-    render json: response_to_hash
+    render json: Api::OrderDetail.new(order_detail).to_h
   end
 
   private
-
-  def account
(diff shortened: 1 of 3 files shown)


app/controllers/api/order_details_controller.rb
```diff
@@ -4,22 +4,37 @@ class Api::OrderDetailsController < ApplicationController
...
+      user_to_hash(order_detail.order.user)
+    end
+  end
+
+  def response_to_hash
+    { account: account_to_hash, ordered_for: ordered_for_to_hash }
+  end
+
+  def order_detail
+    @order_detail = OrderDetail.find_by_id(params[:id]) unless defined?(@order_detail)
```
jhanggi: In what case would `@order_detail` not be defined?

jhanggi:
Would it be possible to extract the order detail object json renderer into its own class? So you'd have something like `Api::OrderDetail.new(@order_detail).to_json`?

app/controllers/api/order_details_controller.rb
drench: It will be if a client hits the API with a bad ID. But then maybe we should respond with a 404 if that happens.

jhanggi: If it hits with a bad ID, wouldn't @order_detail be nil and not undefined?

drench: You're right. It's really no different from `@order_detail ||= OrderDetail.find_by_id(params[:id])` then.

As for a 404 vs. a 200 with an empty return, either will have the same result on the client end: no attributes updated (or it will, after https://github.com/tablexi/nucore-imserc/issues/60 goes live).

drench:
Ready for another look.

app/models/api/order_detail.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def to_h
+    { account: account_to_hash, ordered_for: ordered_for_to_hash }
+  end
+
+  private
+
+  def account
+    @account = @order_detail.try(:account) unless defined?(@account)
```
jhanggi: Since we're doing the 404 catching in the controller, do we need to handle the nil case? Also, could this just be `@account ||= @order_detail.account` as opposed to the `defined?`?

app/models/api/order_detail.rb
```diff
@@ -0,0 +1,35 @@
...
+    if account.present?
+      { id: account.id, owner: user_to_hash(account.owner.user) }
+    end
+  end
+
+  def ordered_for_to_hash
+    user_to_hash(@order_detail.order.user)
+  end
+
+  def user_to_hash(user)
```
jhanggi: So we'll still be sending `{ id: nil, name: nil, ...}` if the user is nil? Do we just want to return nil instead?

drench: Is it possible for an `Account` to have no owner, or an `Order` to have no user? I couldn't find either in the dev database.

jhanggi: I think you're right. We probably don't need to worry about the nil user case.

jhanggi:
:+1: I think I'd like to hold off putting this on staging until they've approved the existing feedback items.

PR #245, opened 2015-02-10
[#92893] Allow administrators to manage select global price group membership

drench:
This is specifically for the Cancer Center, but can be applied to any global price group (that is, one with no facility) if `admin_editable` is set.

A separate PR for NU to migrate data from the `cc_pers` table and remove the `nu_cancer_center` extension will be up next. I'd like to get feedback on this one before opening the one for the NU fork if possible.


app/controllers/account_price_group_members_controller.rb
```diff
@@ -1,63 +1,70 @@
...
-  # DELETE /price_group_members/1
-  def destroy
-    @price_group = current_facility.price_groups.find(params[:price_group_id])
-    @account_price_group_member = AccountPriceGroupMember.find(:first, :conditions => { :price_group_id => @price_group.id, :id =>params[:id]} )
-
-    if @account_price_group_member.destroy
-     flash[:notice] = "The account was successfully removed from the Price Group"
+  def create # TODO very similar to UserPriceGroupMembersController#create
+    if price_group_member.save
+      flash[:notice] = I18n.t("controllers.account_price_group_members.create.notice", create_flash_arguments)
```
jhanggi: Could we use something like `controller_name` so we could share this code in the PriceGroupMembersController?

app/controllers/concerns/price_group_members_controller.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def destroy_price_group_member!
+    price_group_member
+    .class
+    .find(:first, conditions: { price_group_id: @price_group.id, id: params[:id] })
+    .destroy
+  end
+
+  def load_price_group_and_ability!
+    @price_group = current_facility.price_groups.find(params[:price_group_id])
```
jhanggi: Does this get called on `new`? Does `new` even exist?

app/controllers/user_price_group_members_controller.rb
```diff
@@ -1,50 +1,55 @@
...
+  def create_flash_arguments
+    {
+      full_name: price_group_member.user.full_name,
+      price_group_name: @price_group.name,
+    }
+  end
+
+  def destroy_user_price_group_member!
+    UserPriceGroupMember
+    .find(:first, conditions: { price_group_id: @price_group.id, id: params[:id] })
```
jhanggi: Can we use more modern `find_by()`?

db/migrate/20150204174650_add_admin_editable_to_price_group.rb
```diff
@@ -0,0 +1,13 @@
+class AddAdminEditableToPriceGroup < ActiveRecord::Migration
+  def up
+    add_column :price_groups, :admin_editable, :boolean, after: :is_internal, null: false, default: true
+
+    (PriceGroup.globals - PriceGroup.cancer_center).each do |price_group|
```
jhanggi: We should probably update the seed file as well for when someone sets up a new environment, base and external are set to false.

lib/ability.rb
```diff
@@ -15,10 +15,18 @@ def initialize(user, resource, controller)
...
-      can :manage, :all
+      if resource.is_a?(PriceGroup)
+        can :manage_members, resource if resource.admin_editable?
+      else
+        can :manage, :all
+      end
       return
     end
 
+    if resource.is_a?(PriceGroup) && !resource.global? && user.manager_of?(resource.facility)
```
jhanggi: This logic seems kind of complicated. Is there a way we can clean that up?

jhanggi:
Code all looks pretty good. Just a few minor comments. I need to run out now, but I'll play around with it tomorrow.

app/controllers/user_price_group_members_controller.rb
drench: Not with this version of Rails, unfortunately.
Until then, I wonder if something like this would be a bit nicer to look at:

```ruby
UserPriceGroupMember.where(id: params[:id], price_group_id: @price_group.id).each(&:destroy)
```

jhanggi: I definitely like that better. I do believe we can use `destroy_all` instead of `each(&:destroy)`.

lib/ability.rb
drench: I considered changing `#manager_of?` to return `false` if `facility` is `nil` but that wouldn't be right, since administrators are managers of global price groups... whatever I've thought of so far is worse than this line, in other words.

drench:
I did further refactoring and updated the seed with the new attribute. The one thing left is the question in `lib/ability.rb`, if there's a decent solution, so I think it's ready for another look and more local testing.

jhanggi:
:+1: I played around with it locally and everything looks good.

PR #247, opened 2015-02-13
[#92708] Trap badly-formatted PO expiration dates

drench:
There's also some refactoring and reformatting that should make Rubocop and Code Climate a bit happier.


jhanggi:
:+1:

PR #248, opened 2015-02-16
[#88306] Remove nils from Arrays before sorting

drench:
This should eliminate the "comparison of Array with Array" exceptions. It was something we were only seeing in the NU fork.

### DIFF
--- a/app/models/reports/instrument_utilization_report.rb
+++ b/app/models/reports/instrument_utilization_report.rb
@@ -15,7 +15,7 @@ class Reports::InstrumentUtilizationReport
   end
 
   def rows
-    @data.sort.map do |key, data_row|
+    sorted_data.map do |key, data_row|
       key + data_row.row_with_percents(@totals)
     end
   end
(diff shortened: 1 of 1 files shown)


jhanggi:
What was causing the nils in the array? Is there a test we should wrap around this fix?

drench:
The source seems to be this returning `nil` when it can't find a `PmuDepartment` for an account: https://github.com/tablexi/nucore-nu/blob/develop/vendor/engines/pmu/lib/pmu/nufs_account_extension.rb#L5

We could have `#pmu_description` return a blank string instead of `nil` too. I like having something in place on this side though if only because the "comparison of Array with Array" exception was a not-too-helpful exception to debug.

jhanggi:
I'd agree that having this check on this side makes sense. Adding the empty string return on the NU side is not a bad idea, but if this addresses the root problem, then I think we'll be fine vs taking on a bunch of changes on NU side.

PR #249, opened 2015-02-16
[#93455] Ensure auto-logout instruments don't complete unpurchased orders

jhanggi:


app/models/order_detail.rb
```diff
@@ -129,6 +129,13 @@ def self.for_facility_url(facility_url)
     details
   end
 
+  def self.purchased_active_reservations
+    where(state: ['new', 'inprocess'])
+      .where('order_status_id IS NOT NULL')
+      .where(reservations: { canceled_at: nil})
```
drench: space after `nil`?

app/models/order_detail.rb
```diff
@@ -129,6 +129,13 @@ def self.for_facility_url(facility_url)
     details
   end
 
+  def self.purchased_active_reservations
+    where(state: ['new', 'inprocess'])
```
drench: The combo of `new` and `inprocess` is a recurring theme in this model. I don't know if you want to deal with DRYing them all up now but I want to at least keep it in mind as a TODO.

app/support/auto_logout.rb
```diff
@@ -0,0 +1,38 @@
+class AutoLogout
```
drench: I realize this class isn't "yours" but there are parts of this extraction that could use cleanup (despite Code Climate giving it an "A").

spec/factories/reservations.rb
```diff
@@ -2,6 +2,10 @@
   factory :reservation do
     reserve_start_at { Time.zone.parse("#{Date.today.to_s} 10:00:00") + 1.day }
     reserve_end_at { reserve_start_at + 1.hour }
+
+    trait :yesterday do
+      reserve_start_at { Time.zone.parse("#{Date.today.to_s} 10:00:00") - 1.day }
```
drench: While this is consistent with the rest of the file, is the `#to_s` necessary?

drench:
Just had a few questions and notes.

app/models/order_detail.rb
jhanggi: Yeah, I agree. That was something I didn't really want to take on with this fix right now, especially as some of them would involve full refactoring of the old `scope`/`condition:` syntax.

app/support/auto_logout.rb
jhanggi: I agree. There's also some shared code with the auto cancel job. That was more than I wanted to take on with this ticket. I think a good time to clean this class up further will be when we address why it's not actually toggling the instrument off and the other tickets Aaron mentioned in the ticket.

jhanggi:
I made a couple of the minor tweaks. The other refactorings I didn't necessarily want to take on right now, since there weren't any tests around this feature to provide a comfortable safety harness. I had already spent a good amount of time trying to track down this issue, and I'm still not 100% positive that this is the root cause. I tried to make things a little better and set up some of the future refactorings.

drench:
:+1:

PR #250, opened 2015-02-24
Add receipt total line for all order details

jdzak:
https://pm.tablexi.com/issues/89150

order-receipt


lib/tasks/demo_seed.rake
```diff
@@ -326,7 +326,7 @@ namespace :demo  do
           :expires_at         => Time.zone.now+1.year,
           :name_on_card       => 'Paul PI',
           :expiration_month   => '10',
-          :expiration_year    => '2014',
+          :expiration_year    => '2020',
```
jhanggi: What if we made it `Date.current.year + 1` or `1.year.from_now.year` so we don't have to revisit this in 5 years.

app/support/order_details/cost.rb
```diff
@@ -0,0 +1,20 @@
+class OrderDetails::Cost
+  include ActionView::Helpers::NumberHelper
+
+  def initialize(order_details)
+    @order_details = order_details
+  end
+
+  def final_total
```
jhanggi: I need to look a little more closely (since it's some metaprogramming), but I think `Order` actually has a total method already. I think it might be better to wrap that, or simply add the `number_to_currency(@order.total)` to the view without the need for this class.

lib/tasks/demo_seed.rake
jdzak: Sure thing. Also, I was planning on using `5.years.from_now.year` so my seeded credit card doesn't expire for a while.

app/support/order_details/cost.rb
jdzak: The one weird thing is, in `ReceiptsController#receipt` it selects order details that are viewable by the acting user. So to get this same behavior in `Order#total_cost` I'll add an optional parameter to optionally filter the order details by an acting user too.

Does that make sense?

```
  def receipt
    @order = Order.find(params[:id])
    raise ActiveRecord::RecordNotFound unless @order.purchased?

    @order_details = @order.order_details.select{|od| od.can_be_viewed_by?(acting_user) }
    raise ActiveRecord::RecordNotFound if @order_details.empty?

    @order_details_cost = OrderDetails::Cost.new(@order_details)

    @accounts = @order_details.collect(&:account).uniq
  end
```

jdzak:
Updated per comments.

app/models/order.rb
```diff
@@ -239,9 +239,13 @@ def update_order_detail_accounts
     end
   end
 
-  def total_cost(order_detail_method)
+  def total_cost(order_detail_method, acting_user = nil)
```
jhanggi: I'm not crazy about this, but the filtering does appear to need to be done (see commit c303e69509c9cff789fe542725fb080591eeb477).  And this is currently duplicating the logic in the controller action.

I'm thinking that it is probably better to just do a sum on `@order_details`'s total. Could we just have `number_to_currency(@order_details.sum { |od| od.total.to_i })` in the view (or set that total in the controller)? Having that extra wrapper class seemed a little overkill.

jdzak: Ya sure, I can do that.

jdzak:
Updated per comments.

jhanggi:
:+1: let's get together later today so I can show you how I merge this on down the chain.

PR #253, opened 2015-02-27
Change problem list pages

jdzak:
https://pm.tablexi.com/issues/47799

This changes the reservations and orders problem list pages.

screen shot 2015-02-28 at 11 03 57 am

screen shot 2015-02-27 at 3 33 03 pm


app/views/facility_orders/show_problems.html.haml
```diff
@@ -10,4 +10,4 @@
 - else
   %p.notice= t(".instructions")
   = render partial: "shared/problem_order_details",
-    locals: { assign_price_policies_path: assign_price_policies_to_problem_orders_facility_orders_path }
+    locals: { assign_price_policies_path: assign_price_policies_to_problem_orders_facility_orders_path, a_reservation_list_page: false }
```
drench: Typo? `a_reservation_list_page` vs. `reservation_list_page`

jdzak: Yep, fixed. Thanks.

app/views/shared/_problem_order_details.html.haml
```diff
@@ -21,8 +23,8 @@
         %td.centered= link_to order_detail.id,
           manage_order_detail_path(order_detail),
           class: "manage-order-detail"
-
-        %td= human_datetime(order_detail.order.ordered_at)
+        %td= problem_order_date(order_detail, local_assigns[:reservation_list_page])
```
jdzak: This was needed because there are two pages that share this same partial (orders and reservations) and we want to display order date for a problem orders, but we want to see actual start time for a problem reservation.

app/controllers/facility_orders_controller.rb
```diff
@@ -86,6 +84,13 @@ def show_problems_path
     show_problems_facility_orders_path
   end
 
+  def problem_orders
```
jhanggi: Why did you you move this method into the controller?

app/helpers/facility_order_status_helper.rb
```diff
@@ -34,13 +34,6 @@ def new_or_in_process_orders
     end
   end
 
-  def problem_orders
```
jhanggi: This might be a little oddly named since it's called `problem_orders` but it actually returns OrderDetails. not necessarily worth changing, but I noticed that since you moved it.

app/controllers/facility_orders_controller.rb
jdzak: Mainly because it's only referenced in this controller.

app/views/shared/_problem_order_details.html.haml
```diff
@@ -1,15 +1,17 @@
...
+        .text-right
+          = link_to t(".assign_price_policies"),
+            assign_price_policies_path,
+            method: :post,
+            class: %w(btn btn-primary)
     %tr
       %th= Order.model_name.human
       %th= OrderDetail.model_name.human
-      %th= Order.human_attribute_name(:ordered_at)
+      %th= problem_order_date_header(local_assigns[:reservation_list_page])
```
jhanggi: This variable name is a little confusing. It almost seems like it's related to pagination.

jdzak: Hmm, how about something like `showing_problem_reservations` or I could be more specific like `show_reservation_start_at`.

Thoughts?

app/helpers/facility_order_status_helper.rb
jdzak: I just changed it so no one is tripped up on this later on.

app/views/shared/_problem_order_details.html.haml
jdzak: K. Went with `show_reservation_start_at`.

jhanggi:
Let's spend a little time together on this tomorrow. Some of the changes break other parts, specifically the ajax'ed tab counts for problem orders.

jdzak:
Fixed ajax issues.

PR #254, opened 2015-03-02
[#87268] Minimum cost calculation

jdzak:
https://pm.tablexi.com/issues/87268

In this I added the subsidised minimum cost calculation and refactored how it was being displayed by moving the calculation markup to a partial.

screen shot 2015-03-02 at 12 45 00 pm

screen shot 2015-03-02 at 12 41 37 pm

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -206,9 +206,6 @@ th.currency {
 .currency-input input {
   width: 7em;
 }
-.total {
-  font-weight: bold;
-}
 
 .pagination {
   margin: 0;
--- a/app/views/price_policies/instrument/_minimum_calculation.html.haml
+++ b/app/views/price_policies/instrument/_minimum_calculation.html.haml
@@ -1,9 +1,5 @@
-- rate = price_policy.minimum_cost
-- subsidy = price_policy.minimum_cost_subsidy
-- cost = price_policy.subsidized_minimum_cost
-
-.rate= number_to_currency rate
+.rate= number_to_currency price_policy.minimum_cost
 
 - if price_policy.has_subsidy?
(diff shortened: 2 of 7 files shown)


app/views/price_policies/instrument/_minimum_calculation.html.haml
```diff
@@ -0,0 +1,9 @@
+- rate = price_policy.minimum_cost
```
jhanggi: What if you pulled these out of this partial and passed them in as local variables. Then /instrument/_rate_calculation.html.haml could share this same partial. Usually I don't like this many assignments within a partial.

app/assets/stylesheets/nucore.scss
```diff
@@ -206,9 +206,6 @@ th.currency {
 .currency-input input {
   width: 7em;
 }
-.total {
```
jhanggi: Can this safely go away? I would expect us to be using this elsewhere.

jdzak: Ya, it can. In [PR 250](https://github.com/tablexi/nucore-open/pull/250/files#diff-c014e377576c262389d42610b44fed6dR209) I added this bold style because I throught it was a unique selector, but it turns out there are other things using it that I don't want to apply the bold style to.

I updated `app/views/orders/receipt.html.haml` to use the `<strong>` tag instead of this style.

app/views/price_policies/instrument/_minimum_calculation.html.haml
jdzak: Sure I can do that.

jhanggi:
I'm merging this in. Be careful with your branch names--this one is referencing the wrong story, and please make sure to include the ticket number in the PR title, not just the description.

PR #256, opened 2015-03-04
[#79851] Complete reservation only instruments

jdzak:
https://pm.tablexi.com/issues/79851


app/models/reservation.rb
```diff
@@ -124,6 +124,7 @@ def end_reservation!
     save!
     # reservation is done, now give the best price
     order_detail.assign_price_policy
+    order_detail.change_status!(OrderStatus.complete.first)
```
jdzak: Make a method on OrderDetail for this.

app/models/reservation.rb
```diff
@@ -124,6 +124,7 @@ def end_reservation!
...
     order_detail.save!
```
jdzak: Check if this is necessary after changing status.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def purchased_active_order_details
+    OrderDetail.purchased_active_reservations
+      .where("reservations.reserve_end_at < ?", Time.zone.now - 12.hours)
+      .readonly(false)
+      .all
+  end
+
+  def non_reservation_order_details
+    OrderDetail.purchased_active_reservations.readonly(false).select { |od| od.product.try(:reservation_only?) }
```
jdzak: Check to see order is complete also make this into a query instead of using the select to filter.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def complete_status
+    @complete_status ||= OrderStatus.find_by_name!('Complete')
```
jdzak: This should be set to complete with a method on OrderDetails too.

app/support/auto_logout.rb
```diff
@@ -23,6 +23,7 @@ def complete_status
   end
 
   def complete_reservation(od)
+    # TODO: I think this logic is the same as Reservation#end_reservation
```
jdzak: Go ahead.

spec/app_support/auto_expire_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+      it 'does not do anything' do
+        expect { action.perform }.not_to change { reservation.reload.actual_end_at }
+        expect(order_detail.state).to eq('new')
+        expect(order_detail.order_status_id).to be_blank
+      end
+    end
+
+    context 'a new reservation only instrument' do
+      let!(:reservation) do
+        start_at = Time.zone.parse("#{Date.today} 9:00:00")
```
jhanggi: I'd use something like `1.hour.ago`.

spec/factories/products.rb
```diff
@@ -85,6 +85,12 @@
       create :instrument_price_policy, price_group: product.facility.price_groups.last, usage_rate: 1, product: product
       product.reload
     end
+
+    trait :reservation_only do
+      after(:create) do |inst, evaluator|
+        inst.relay = FactoryGirl.create(:relay, :manual_control, :instrument => inst)
```
jdzak: reservation_only should not have a relay. control_mechanism should just be nil (manual_control).

spec/factories/relays.rb
```diff
@@ -13,6 +13,9 @@
       type 'RelaySynaccessRevB'
     end
 
+    trait :manual_control do
```
jdzak: Don't need this.

spec/price_policies_controller_shared_examples.rb
```diff
@@ -98,7 +98,7 @@
         end
         it 'should set fields based off the old policy' do
           do_request
-          assigns[:price_policies].map(&:price_group).should == [@price_policy.price_group, @price_group2_policy.price_group]
+          assigns[:price_policies].map(&:price_group).sort.should == [@price_policy.price_group, @price_group2_policy.price_group]
```
jhanggi: If the version of rspec we're using supports it use `match_array` matcher instead of sorting.

jdzak:
Add spec for not ending a reservation in the future.

jdzak:
@jhanggi Comments addressed.

app/models/instrument.rb
```diff
@@ -31,6 +31,13 @@ class Instrument < Product
 
   after_create :set_default_pricing
 
+  # Scopes
+  # --------
+  scope :reservation_only, lambda {
```
jhanggi: Probably should have mentioned this when I saw it earlier. We've been preferring class-method style scopes over `scope`.

jhanggi: At least for new code. There's a lot of old code that we haven't felt has been worth the effort to update.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def purchased_active_order_details
+    OrderDetail.purchased_active_reservations
+      .where("reservations.reserve_end_at < ?", Time.zone.now - 12.hours)
+      .readonly(false)
+      .all
+  end
+
+  def non_reservation_order_details
```
jhanggi: This is kind of a misleading method name. We have a `non_reservation` scope on OrderDetail that's used to mean service or item orders, so I don't love using that terminology for this.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+      od.transaction do
+        expire_reservation(od)
+      end
+    end
+  end
+
+  private
+
+  def order_details
+    purchased_active_order_details | non_reservation_order_details
```
jhanggi: This feels a little weird to me. I feel like we might be conflating two ideas here. We could do both of them in the same rake task, and they would likely share a bit of code, but I think it would make sense to have one that ends reservation-only OrderDetails and one that ends the un-logged-out instruments.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def purchased_active_order_details
+    OrderDetail.purchased_active_reservations
+      .where("reservations.reserve_end_at < ?", Time.zone.now - 12.hours)
```
jhanggi: I think this should restrict to only those that have an `actual_start_at` but missing the `actual_end_at`.

jhanggi: That might make sense to create a new scope on OrderDetail for `running_reservations` or some similar name.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+    OrderDetail.purchased_active_reservations
+      .where("reservations.reserve_end_at < ?", Time.zone.now)
+      .joins(:product)
+      .merge(Instrument.reservation_only)
+      .readonly(false)
+  end
+
+  def expire_reservation(od)
+    od.complete!
+    od.fulfilled_at = od.reservation.reserve_end_at
```
jhanggi: I think this should probably happen before the `complete!` call so it gets saved as part of that call. Otherwise, the fulfilled_at might not get set if the price policy is missing.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+    return unless od.price_policy
+    costs = od.price_policy.calculate_cost_and_subsidy(od.reservation)
```
jhanggi: I would consider looking at `OrderDetail#assign_actual_price` to see if we can reuse that code instead of duplicating it here.

app/support/auto_expire.rb
```diff
@@ -0,0 +1,44 @@
...
+    return if costs.blank?
+    od.actual_cost    = costs[:cost]
+    od.actual_subsidy = costs[:subsidy]
+    od.save!
+  rescue => e
+    STDERR.puts "Error on Order # #{od} - #{e}\n#{e.backtrace.join("\n")}"
```
jhanggi: Can we use `logger.error` (might need `Rails.logger.error`) instead of STDERR?

jhanggi: Or perhaps we should use the new `background_error` notification for this so it bubbles up outside of the log; not sure how common or what might cause these failures.

app/models/instrument.rb
jdzak: Sure, I'll change it.

spec/models/order_detail_spec.rb
```diff
@@ -1666,6 +1666,18 @@ def setup_merge_order
...
   end
 
+  describe '#complete!' do
+    before { order_detail.complete! }
+
+    it 'saved' do
+      expect(order_detail).to_not be_changed
+    end
+
+    it 'sets status to complete' do
```
jhanggi: We might want a test to assert that `fulfilled_at` gets set here.

app/support/auto_expire.rb
jdzak: Yea, I think I picked the name before I fully understood the feature. How does `#reservation_only_order_details` sound instead?

app/support/auto_expire.rb
jdzak: Ok, I'll move `#complete!` to after the cost and subsidy is set.

app/support/auto_expire.rb
jdzak: Yea, I think `background_error` makes sense. If it becomes too loud we can always change it to `Rails.logger.error`.

spec/models/order_detail_spec.rb
jdzak: Sure.

app/support/auto_expire.rb
jdzak: Yea, I think that should work.

app/support/auto_expire.rb
jdzak: I may need to stop by real quick today. I checked out the scopes on OrderDetail and saw a scope called `#in_progress_reservations` which uses `#comfirmed_reservation`.

I'm curious if we can re-use that or if that's different?

app/support/auto_expire.rb
jdzak: Ok, I'll split these up.

jdzak: Call it EndReservationOnly...

jdzak:
@jhanggi Comments addressed

app/support/auto_expire.rb
```diff
@@ -0,0 +1,33 @@
+class AutoExpire
```
jhanggi: Can we call this `AutoExpireReservation` or something a little more descriptive?

app/support/end_reservation_only.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  private
+
+  def order_details
+    reservation_only_order_details
+  end
+
+  def reservation_only_order_details
+    OrderDetail.purchased_active_reservations
+      .merge(Reservation.in_progress)
```
jhanggi: Reservation only reservation will not have an actual_start_at, so I'm not sure this'll capture it.

spec/app_support/end_reservation_only_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+describe EndReservationOnly, :timecop_freeze do
+  let(:now) { DateTime.now.change(hour: 9, min: 31)  }
+
+  let(:action) { described_class.new }
+  let(:order_detail) { reservation.order_detail }
+
+  describe '#perform' do
+    context 'an unpurchased reservation' do
+      let!(:reservation) { create(:setup_reservation, :yesterday, actual_start_at: 1.hour.ago) }
```
jhanggi: A reservation-only reservation will not have `actual_start_at`.

spec/app_support/end_reservation_only_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+
+        create(:purchased_reservation,
+            product: create(:setup_instrument, min_reserve_mins: 1),
+            actual_start_at: start_at,
+            reserve_start_at: start_at,
+            reserve_end_at: end_at)
+      end
+
+      it 'completes reservation' do
+        expect { action.perform }.to change { order_detail.reload.state }.from('new').to('complete')
```
jhanggi: I would add a few more tests around making sure that it sets the fulfilled at date, that it's not a problem reservation, and probably that it gets an actual_cost assigned.

spec/app_support/end_reservation_only_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+    end
+
+    context 'a new reservation only instrument' do
+      let!(:reservation) do
+        start_at = 30.minutes.ago
+        end_at = 1.minute.ago
+
+        create(:purchased_reservation,
+            product: create(:setup_instrument, min_reserve_mins: 1),
```
jhanggi: I would make sure that this instrument explicitly says it's reservation-only (I don't know what the default is).

spec/app_support/auto_expire_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  let(:action) { described_class.new }
+  let(:order_detail) { reservation.order_detail }
+
+  describe '#perform' do
+    context 'a new reservation' do
+      let!(:reservation) { create(:purchased_reservation, :yesterday, actual_start_at: 1.hour.ago) }
+
+      it 'completes the reservation' do
+        expect { action.perform }.to change { order_detail.reload.state }.from('new').to('complete')
```
jhanggi: We should add some additional checks here like setting the fulfilled_at, that it's a problem reservation, it doesn't get actual pricing assigned,

app/support/auto_expire.rb
jdzak: Yea, good call.

app/support/end_reservation_only.rb
jdzak: Hmm, I may need to stop by today for a second. I'm kind of confused  why we need to set actual_end_at time but not actual_start_at. Is that for pricing? 

If so, why can't we use reserve_start_at and reserve_end_at for pricing?

jhanggi: For reservation-only, neither `actual_start_at` nor `actual_end_at` get set. For reservation-only instruments, it is the `reserve_start_at` and `reserve_end_at` that get used for pricing.

spec/app_support/auto_expire_spec.rb
jdzak: Sure

spec/app_support/end_reservation_only_spec.rb
jdzak: Sure

spec/app_support/end_reservation_only_spec.rb
jdzak: Sure

spec/app_support/end_reservation_only_spec.rb
jdzak: Sure

jdzak:
@jhanggi Comments addressed

app/support/auto_expire_reservation.rb
```diff
@@ -0,0 +1,37 @@
...
+      .all
+  end
+
+  def expire_reservation(od)
+    od.complete!
+
+    # OrderDetail#complete! sets fulfilled_at and price policy,
+    # so we have to reset them.
+    od.fulfilled_at = od.reservation.reserve_end_at
+    od.clear_costs
```
jhanggi: Why do we need to do this, and make the method public? Shouldn't this always end up being nil for these reservations since there is no actual_end_at?

lib/tasks/order_details.rake
```diff
@@ -28,6 +9,11 @@ namespace :order_details  do
     AutoLogout.new.perform
   end
 
+  desc "mark order details with reservation only reservations as complete"
+  task :expire_reservation_only => :environment do
```
jhanggi: Can we do this as part of the same `expire_reservations` rake task? That way we don't need to add a new cron entry.

spec/app_support/end_reservation_only_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+        order_detail.reload
+        reservation.reload
+      end
+
+      it 'uses a non-reservation only reservation' do
+        expect(reservation.product).to_not be_reservation_only
+      end
+
+      include_examples 'it does not complete order' do
+        it 'leaves state as new' do
```
jhanggi: Can some of these move into the shared example?

jdzak: Yea, I'll move it in. I thought there would be some case where the state would be something other than 'new', but so far it is always new.

app/support/auto_expire_reservation.rb
jdzak: Yes, you're right. I'll remove the #clear_costs and also make this method private.

lib/tasks/order_details.rake
jdzak: Sure.

jdzak:
@jhanggi Comments addressed.

PR #257, opened 2015-03-05
[#94303] Could not update quantity before using external service

jhanggi:
Using update quantity takes you to update_or_purchase. When leaving to the external service, that becomes the referer, so we would get a 404 when coming back from the service.

This isn't the most elegant solution, but in order to maintain support for #61929 'More options' date reset in cart, we can't do a redirect to show after updating quantities, so this was the simplest.


drench:
:+1:

PR #258, opened 2015-03-05
[#94282] Handle bulk import exceptions and validate header

jdzak:
https://pm.tablexi.com/issues/94282

### DIFF
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -135,9 +135,8 @@ class OrderImport < ActiveRecord::Base
       row_importer = OrderRowImporter.new(row, self)
       row_importer.import
     rescue => e
-      Rails.logger.error("Failed to bulk import: #{upload_file_path}")
-      Rails.logger.error(e.backtrace)
-      ExceptionNotifier.notify_exception(e, data: { message: "Failed to bulk import: #{upload_file_path}" })
+      ActiveSupport::Notifications.instrument('external_error',
+        exception: e, information: "Failed to bulk import: #{upload_file_path}")
       row_importer.add_error("Failed to import row")
       set_error_mode
     end
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -135,8 +135,8 @@ class OrderImport < ActiveRecord::Base
       row_importer = OrderRowImporter.new(row, self)
       row_importer.import
     rescue => e
-      ActiveRecord::Base.logger.error("Failed to bulk import: #{upload_file_path}")
-      ActiveRecord::Base.logger.error(e.backtrace)
+      Rails.logger.error("Failed to bulk import: #{upload_file_path}")
+      Rails.logger.error(e.backtrace)
       ExceptionNotifier.notify_exception(e, data: { message: "Failed to bulk import: #{upload_file_path}" })
       row_importer.add_error("Failed to import row")
       set_error_mode
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -131,8 +131,16 @@ class OrderImport < ActiveRecord::Base
   end
 
   def import_row(row) # TODO refactor
-    row_importer = OrderRowImporter.new(row, self)
-    row_importer.import
+    begin
+      row_importer = OrderRowImporter.new(row, self)
+      row_importer.import
+    rescue => e
+      ActiveRecord::Base.logger.error("Failed to bulk import: #{upload_file_path}")
+      ActiveRecord::Base.logger.error(e.backtrace)
+      ExceptionNotifier.notify_exception(e, :data => {:message => "Failed to bulk import: #{upload_file_path}"})
+      row_importer.add_error("Failed to import row")
+      set_error_mode
+    end
     if row_importer.errors?
       set_error_mode
     else
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -137,7 +137,7 @@ class OrderImport < ActiveRecord::Base
     rescue => e
       ActiveRecord::Base.logger.error("Failed to bulk import: #{upload_file_path}")
       ActiveRecord::Base.logger.error(e.backtrace)
-      ExceptionNotifier.notify_exception(e, :data => {:message => "Failed to bulk import: #{upload_file_path}"})
+      ExceptionNotifier.notify_exception(e, data: { message: "Failed to bulk import: #{upload_file_path}" })
       row_importer.add_error("Failed to import row")
       set_error_mode
     end
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -135,10 +135,9 @@ class OrderImport < ActiveRecord::Base
       row_importer = OrderRowImporter.new(row, self)
       row_importer.import
     rescue => e
-      ActiveSupport::Notifications.instrument('external_error',
+      ActiveSupport::Notifications.instrument('background_error ',
         exception: e, information: "Failed to bulk import: #{upload_file_path}")
       row_importer.add_error("Failed to import row")
-      set_error_mode
     end
     if row_importer.errors?
       set_error_mode
--- a/app/support/order_row_importer.rb
+++ b/app/support/order_row_importer.rb
@@ -167,8 +167,8 @@ class OrderRowImporter
   end
 
   def validate_headers
-    unkown_headers = (@row.headers - HEADERS.values)
-    add_error("Unknown headers: #{unkown_headers.join('/')}") if unkown_headers.present?
+    unknown_headers = (@row.headers - HEADERS.values)
+    add_error("Unknown headers: #{unknown_headers.join(' | ')}") if unknown_headers.present?
   end
 
   def validate_fields
--- a/app/support/order_row_importer.rb
+++ b/app/support/order_row_importer.rb
@@ -45,7 +45,7 @@ class OrderRowImporter
   end
 
   def import
-    add_product_to_order if has_valid_fields?
+    add_product_to_order if has_valid_headers? && has_valid_fields?
   end
 
   def order_date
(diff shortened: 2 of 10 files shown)


app/models/order_import.rb
```diff
@@ -131,8 +131,16 @@ def handle_save_nothing_on_error # TODO refactor and rename
...
   def import_row(row) # TODO refactor
-    row_importer = OrderRowImporter.new(row, self)
-    row_importer.import
+    begin
+      row_importer = OrderRowImporter.new(row, self)
+      row_importer.import
+    rescue => e
+      ActiveRecord::Base.logger.error("Failed to bulk import: #{upload_file_path}")
+      ActiveRecord::Base.logger.error(e.backtrace)
+      ExceptionNotifier.notify_exception(e, :data => {:message => "Failed to bulk import: #{upload_file_path}"})
```
drench: Go with 1.9 hash style

jdzak: Good call.

jdzak:
@drench comments addressed.

app/support/order_row_importer.rb
```diff
@@ -71,16 +71,17 @@ def user
       User.find_by_username(user_field) || User.find_by_email(user_field)
   end
 
+
+  def add_error(message)
+    @errors.add(message) if message.present?
+  end
+
```
jdzak: I made this public so I can call it from OrderImport#import_row

spec/app_support/order_row_importer_spec.rb
```diff
@@ -22,16 +22,19 @@
...
+      "Netid / Email" => username,
+      "Chart String" => chart_string,
+      "Product Name" => product_name,
+      "Quantity" => quantity,
+      "Order Date" => order_date,
+      "Fulfillment Date" => fulfillment_date,
+      "Errors" => errors,
+      "Note" => notes,
+    }
+    CSV::Row.new(ref.keys, ref.values)
```
jdzak: Faking a row with a hash no longer worked because i needed to use CSV::Row#headers to validate headers.

config/environments/test.rb
```diff
@@ -41,6 +41,12 @@
   # config.active_record.mass_assignment_sanitizer = :strict
 
   config.action_mailer.default_url_options = { :host => "localhost", :port => 3000 }
+
+  config.middleware.use ExceptionNotification::Rack,
+    email: {
+      sender_address: "exception@test.com",
+      exception_recipients: ["fake_developer@test.com"]
+    }
```
jdzak: I added this to test that the exception notification was happening.

drench: It's good. Just use one of the safe testing domains like `example.com` (ref: https://www.rfc-editor.org/rfc/rfc2606.txt)

app/models/order_import.rb
```diff
@@ -131,8 +131,16 @@ def handle_save_nothing_on_error # TODO refactor and rename
...
-    row_importer.import
+    begin
+      row_importer = OrderRowImporter.new(row, self)
+      row_importer.import
+    rescue => e
+      ActiveRecord::Base.logger.error("Failed to bulk import: #{upload_file_path}")
+      ActiveRecord::Base.logger.error(e.backtrace)
+      ExceptionNotifier.notify_exception(e, data: { message: "Failed to bulk import: #{upload_file_path}" })
+      row_importer.add_error("Failed to import row")
+      set_error_mode
```
drench: It seems like there's enough going on in this `rescue` to make it extraction-worthy.

spec/models/order_import_spec.rb
```diff
@@ -326,4 +326,32 @@ def generate_import_file(*args)
...
+        expect(import.error_file_content).to include("Failed to import row")
+      end
+
+      it "sets error flags" do
+        expect(import).to be_error_mode
+        expect(import.result).to be_failed
+      end
+
+      it "sends an exception notification" do
+        actual = ActionMailer::Base.deliveries.first.body.raw_source
```
drench: Maybe a more descriptive name like `email_text` or `email_body`?

config/environments/test.rb
jdzak: Sure, I'll use `example.com` because it seems like `test.com` is used for DNS testing.

spec/models/order_import_spec.rb
jdzak: Sounds good.

app/models/order_import.rb
jdzak: Hmm, I feel like this would be a very specific method that wouldn't be able to used anywhere else and this class already needs a decent amount of cognitive overhead to digest all the sub-methods. This is exception logic, so maybe the upside would be that we could move it out of the way of the normal logic.

Thoughts?

jdzak:
Addressed two out of the three comments. The third I just was hoping to get more feedback on.

app/models/order_import.rb
jhanggi: I think a good extraction would be to us ActiveSupport::Notifications. That way the model code isn't dependent on ExceptionNotifier being configured.

I would `ActiveSupport::Notifications.instrument` here, and then have the subscriber do the logging and `notify_exception`. We do this in GS.

app/models/order_import.rb
```diff
@@ -131,8 +131,16 @@ def handle_save_nothing_on_error # TODO refactor and rename
   end
 
   def import_row(row) # TODO refactor
-    row_importer = OrderRowImporter.new(row, self)
-    row_importer.import
+    begin
+      row_importer = OrderRowImporter.new(row, self)
+      row_importer.import
+    rescue => e
+      ActiveRecord::Base.logger.error("Failed to bulk import: #{upload_file_path}")
```
jhanggi: How about the more concise `Rails.logger.error`?

app/support/order_row_importer.rb
```diff
@@ -160,6 +166,11 @@ def validate_account
     end
   end
 
+  def validate_headers
+    unkown_headers = (@row.headers - HEADERS.values)
+    add_error("Unknown headers: #{unkown_headers.join('/')}") if unkown_headers.present?
```
jhanggi: Spelling of `unknown`. Also,  `/` is probably not a good separator because "Netid / Email" is one of the headers.

config/environments/test.rb
```diff
@@ -41,6 +41,12 @@
   # config.active_record.mass_assignment_sanitizer = :strict
 
   config.action_mailer.default_url_options = { :host => "localhost", :port => 3000 }
+
+  config.middleware.use ExceptionNotification::Rack,
```
jhanggi: The test environment should not have ExceptionNotification.

spec/spec_helper.rb
```diff
@@ -63,4 +63,6 @@
     #puts "travelling to #{now}"
     Timecop.travel(now)
   end
+
+  config.before(:each) { ActionMailer::Base.deliveries.clear }
```
jhanggi: I don't think we want to do this before *every* test. I think we should just do it on the ones that need it. I tend to use `expect { do_something }.to change(ActionMailer::Base.deliveries, :count).by(1).

app/models/order_import.rb
```diff
@@ -131,8 +131,16 @@ def handle_save_nothing_on_error # TODO refactor and rename
   end
 
   def import_row(row) # TODO refactor
-    row_importer = OrderRowImporter.new(row, self)
-    row_importer.import
+    begin
+      row_importer = OrderRowImporter.new(row, self)
+      row_importer.import
+    rescue => e
```
jhanggi: Does this get triggered for invalid headers? I don't want to get an email every time there's bad user input.

jdzak: No, it won't send emails for invalid headers since we're validating those now and including them in the error report.

This email is sent in the case the order row importer throws an exception. I figured we want that to happen instead of letting it fail silently with delayed_job.

spec/spec_helper.rb
jdzak: Sounds good.

config/environments/test.rb
jdzak: K

app/support/order_row_importer.rb
jdzak: I'll fix the spelling. Also, I'll switch to `|` as a separator.

app/models/order_import.rb
jdzak: Sure.

jdzak:
@drench @jhanggi comments addressed

app/models/order_import.rb
```diff
@@ -131,8 +131,15 @@ def handle_save_nothing_on_error # TODO refactor and rename
...
+      ActiveSupport::Notifications.instrument('external_error',
```
jhanggi: Let's use something other than `external_error` for the name of this. That name made sense for GS since we were using it for errors relating to using external services/APIs. Maybe `background_error` unless you have something better.

app/models/order_import.rb
```diff
@@ -131,8 +131,15 @@ def handle_save_nothing_on_error # TODO refactor and rename
...
+        exception: e, information: "Failed to bulk import: #{upload_file_path}")
+      row_importer.add_error("Failed to import row")
+      set_error_mode
```
jhanggi: Is this `set_error_mode` necessary since the line above it adds the error, and the next conditional checks for errors?

app/support/order_row_importer.rb
```diff
@@ -160,6 +166,11 @@ def validate_account
     end
   end
 
+  def validate_headers
+    unknown_headers = (@row.headers - HEADERS.values)
```
jhanggi: Do we also need to check that it has all the necessary headers?

config/initializers/notifications.rb
```diff
@@ -0,0 +1,7 @@
+ActiveSupport::Notifications.subscribe('external_error') do |_name, _start, _finish, _id, payload|
```
jhanggi: Should we also add a subscriber for NewRelic?

app/models/order_import.rb
jdzak: Sure.

app/models/order_import.rb
jdzak: No, I'll delete it.

app/support/order_row_importer.rb
jdzak: Sure.

config/initializers/notifications.rb
jdzak: NewRelic is only on the nucore-nu branch. Do you want me to wait until this is merged to the nucore-nu branch before I add that?

jdzak:
@jhanggi @drench comments addressed

PR #259, opened 2015-03-05
[#94276] Pass fulfilled_at into account_is_open when journaling

jhanggi:
Order details that were fulfilled before a chart string expired was not being accepted because we weren't looking at the fulfilled_at. This is a regression in the [NU code](https://github.com/tablexi/nucore-nu/commit/b7b5773fcb1a321e9861ea1c32e4aed3ed52d54f).


drench:
:+1:

PR #261, opened 2015-03-05
[#94276] Fix spec, which was broken on NU

jhanggi:
Yet another attempt to get CI passing.

JournalRow has some extra validations on NU, so we can't just replace with the ValidatorDefault.


drench:
:+1:

PR #263, opened 2015-03-09
[#89149] Added facility filtering to user accounts page

jdzak:
https://pm.tablexi.com/issues/89149

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -118,8 +118,7 @@ class UsersController < ApplicationController
   # GET /facilities/:facility_id/users/:user_id/accounts
   def accounts
     # accounts for this facility
-    facility_accounts = Account.for_facility(current_facility)
-    @account_users = @user.account_users.active.select { |ua| facility_accounts.include?(ua.account) }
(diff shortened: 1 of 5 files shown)


app/controllers/users_controller.rb
```diff
@@ -118,7 +118,8 @@ def reservations
   # GET /facilities/:facility_id/users/:user_id/accounts
   def accounts
     # accounts for this facility
-    @account_users = @user.account_users.active
+    facility_accounts = Account.for_facility(current_facility)
+    @account_users = @user.account_users.active.select { |ua| facility_accounts.include?(ua.account) }
```
jhanggi: Could this be done as `@user.accounts.for_facility(current_facility)`? It looks like the `User#accounts` relation already has the `active` check on it (well, duplicated), but we'l want to make sure that this don't return deleted accounts.

jdzak: Ok, I Changed #for_facility to be a scope on Account and moved the specs to the model. Also, I added a spec to check that deleted accounts are filtered.

jdzak:
@jhanggi Addressed comment.

PR #265, opened 2015-03-18
[#92949] Deactivate relay on auto logout

jdzak:
https://pm.tablexi.com/issues/92949

### DIFF
--- a/app/support/auto_logout.rb
+++ b/app/support/auto_logout.rb
@@ -23,7 +23,8 @@ class AutoLogout
   end
 
   def complete_reservation(od)
-    od.reservation.product.relay.deactivate if od.reservation.can_switch_instrument_off?
+    deactivate_relay = od.reservation.can_switch_instrument_off? && od.reservation.other_reservations_using_relay.count == 0
+    od.reservation.product.relay.deactivate if deactivate_relay
     od.reservation.actual_end_at = Time.zone.now
     od.change_status!(complete_status)
     return unless od.price_policy
--- a/app/support/auto_logout.rb
+++ b/app/support/auto_logout.rb
@@ -23,7 +23,7 @@ class AutoLogout
   end
 
   def complete_reservation(od)
-    deactivate_relay = od.reservation.can_switch_instrument_off? && od.reservation.other_reservations_using_relay.count == 0
+    deactivate_relay = od.reservation.other_reservation_using_relay?
     od.reservation.product.relay.deactivate if deactivate_relay
     od.reservation.actual_end_at = Time.zone.now
     od.change_status!(complete_status)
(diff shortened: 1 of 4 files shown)


jhanggi:
We should add a test that makes sure that it doesn't log out the user if there's a currently running reservation. I think that's handled inside `can_switch_instrument_off?`, but it would be nice to have that check.

jdzak:
@jhanggi I added an spec to check a running reservation isn't killed by auto-logout. Was that what you were asking for?

jdzak:
jhanggi said:
>There need to be two reservations. One that is going to get completed and moved to the problem queue > (I’m not it gets to the problem queue in this PR, but it might in one of your other ones), and another that > is currently running. We need to make sure that calling `complete_reservation` on the former, moves it > to completion, but doesn’t turn off the relay.

jdzak:
@jhanggi Comments addressed.

app/support/auto_logout.rb
```diff
@@ -23,6 +23,8 @@ def complete_status
   end
 
   def complete_reservation(od)
+    deactivate_relay = od.reservation.can_switch_instrument_off? && od.reservation.other_reservations_using_relay.count == 0
```
jhanggi: I think this might be a good candidate to extract into a private method.

jhanggi: I was initially thinking take both halves of the condition, but now I'm thinking just the `Reservation#other_reservation_using_relay?` into it's own method.

jhanggi:
Just one minor comment. Otherwise, looks good.

app/support/auto_logout.rb
jdzak: Sure.

jdzak:
@jhanggi Comments addressed.

jdzak:
One second.. need to fix some failing specs.

jdzak:
All good now.

jhanggi:
After the previous merge, this is now in conflict in `app/support/auto_logout.rb`. Can you clean up that merge.

jdzak:
Fixed the merge conflict. Also, I tweaked the logic in `ReservationSupport#other_reservation_using_relay?` to match the method name.

app/support/auto_logout.rb
```diff
@@ -23,6 +23,8 @@ def complete_status
   end
 
   def complete_reservation(od)
+    deactivate_relay = !od.reservation.other_reservation_using_relay?
```
jhanggi: What would you think of
```
relay = od.reservation.product.relay
relay.deactivate unless od.reservation.other_reservation_using_relay?
```
I think that reads pretty well.

jhanggi:
Just one minor comment. Otherwise, looks good.

app/support/auto_logout.rb
jdzak: Yea, sounds better. I'll switch it.

jdzak:
Ready for review.

PR #266, opened 2015-03-18
[#47799] Show customer which order is for

jdzak:
https://pm.tablexi.com/issues/47799

screen shot 2015-03-17 at 8 26 19 pm
screen shot 2015-03-17 at 8 25 59 pm


jhanggi:
Just noting that this is an update based on client feedback to https://github.com/tablexi/nucore-open/pull/253

PR #267, opened 2015-03-18
[#73235] Allow minutes until auto-logout to be configurable

jdzak:

Edit

View

### DIFF
--- a/app/models/power_relay.rb
+++ b/app/models/power_relay.rb
@@ -4,6 +4,7 @@ module PowerRelay
   def self.included(base)
     ## validations
     base.validates_presence_of :ip, :port, :username, :password
+    base.validates :auto_logout_minutes, :presence => { :if => :auto_logout }
   end
 
   ## instance methods
--- a/app/models/power_relay.rb
+++ b/app/models/power_relay.rb
@@ -1,5 +1,3 @@
-require 'active_support/concern'
-
 # Including class must implement #relay_connection
 # Port refers to the outlet, not the IP port
 module PowerRelay
--- a/app/models/power_relay.rb
+++ b/app/models/power_relay.rb
@@ -1,10 +1,14 @@
+require 'active_support/concern'
+
 # Including class must implement #relay_connection
 # Port refers to the outlet, not the IP port
 module PowerRelay
-  def self.included(base)
+  extend ActiveSupport::Concern
+
+  included do
     ## validations
-    base.validates_presence_of :ip, :port, :username, :password
-    base.validates :auto_logout_minutes, :presence => { :if => :auto_logout }
+    validates_presence_of :ip, :port, :username, :password
+    validates :auto_logout_minutes, presence: { if: :auto_logout }
   end
 
   ## instance methods
--- a/app/support/auto_logout.rb
+++ b/app/support/auto_logout.rb
@@ -3,7 +3,7 @@ class AutoLogout
     order_details.each do |od|
       relay = od.product.relay
       next unless relay.try(:auto_logout) == true
-      next unless auto_logout_time?(od.reservation.try(:reserve_end_at), relay.try(:auto_logout_minutes))
+      next unless auto_logout_time?(od.reservation.reserve_end_at, relay.try(:auto_logout_minutes))
 
       od.transaction do
         complete_reservation(od)
--- a/app/support/auto_logout.rb
+++ b/app/support/auto_logout.rb
@@ -24,7 +24,7 @@ class AutoLogout
 
   def can_auto_logout?(reserve_end_at, auto_logout_minutes)
     return false if reserve_end_at.nil? || auto_logout_minutes.nil?
-    (reserve_end_at + auto_logout_minutes.minutes) < Time.zone.now
+    reserve_end_at < auto_logout_minutes.minutes.ago
   end
 
   def complete_status
--- a/app/support/auto_logout.rb
+++ b/app/support/auto_logout.rb
@@ -1,9 +1,7 @@
 class AutoLogout
   def perform
     order_details.each do |od|
-      relay = od.product.relay
-      next unless relay.try(:auto_logout) == true
-      next unless can_auto_logout?(od.reservation.reserve_end_at, relay.try(:auto_logout_minutes))
+      next unless should_auto_logout?(od)
 
       od.transaction do
         complete_reservation(od)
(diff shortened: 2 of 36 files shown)


app/models/power_relay.rb
```diff
@@ -4,6 +4,7 @@ module PowerRelay
   def self.included(base)
     ## validations
     base.validates_presence_of :ip, :port, :username, :password
+    base.validates :auto_logout_minutes, :presence => { :if => :auto_logout }
```
jhanggi: Use 1.9 hash syntax. I know there's still some residual of the old syntax, but we should change what we can.

We should probably also do a numericality and greater-than-zero validation.

It might also be nice to convert this to `ActiveSupport::Concern` to clean up the syntax a little.

app/support/auto_logout.rb
```diff
@@ -13,11 +15,16 @@ def perform
 
   def order_details
     OrderDetail.purchased_active_reservations.
-      where("reservations.actual_end_at IS NULL AND reserve_end_at < ?", Time.zone.now - 1.hour).
+      where("reservations.actual_end_at IS NULL AND reserve_end_at < ?", Time.zone.now).
```
jhanggi: Should this use the `Reservation.in_progress` with an additional `where('reserve_end_at < ?)`?

app/support/auto_logout.rb
```diff
@@ -13,11 +15,16 @@ def perform
...
       includes(:product).
       readonly(false).all
   end
 
+  def auto_logout_time?(reserve_end_at, auto_logout_minutes)
+    return false if reserve_end_at.nil? || auto_logout_minutes.nil?
+    (reserve_end_at + auto_logout_minutes.minutes) < Time.now
```
jhanggi: What about `reserve_end_at < auto_logout_minutes.minutes.ago`?

app/support/auto_logout.rb
```diff
@@ -1,7 +1,9 @@
 class AutoLogout
   def perform
     order_details.each do |od|
-      next unless od.product.relay.try(:auto_logout) == true
+      relay = od.product.relay
+      next unless relay.try(:auto_logout) == true
+      next unless auto_logout_time?(od.reservation.try(:reserve_end_at), relay.try(:auto_logout_minutes))
```
jhanggi: In what case is `reservation` nil?

app/support/auto_logout.rb
```diff
@@ -13,11 +15,16 @@ def perform
 
   def order_details
     OrderDetail.purchased_active_reservations.
-      where("reservations.actual_end_at IS NULL AND reserve_end_at < ?", Time.zone.now - 1.hour).
+      where("reservations.actual_end_at IS NULL AND reserve_end_at < ?", Time.zone.now).
       includes(:product).
       readonly(false).all
   end
 
+  def auto_logout_time?(reserve_end_at, auto_logout_minutes)
```
jhanggi: Can we make this method name a little more descriptive?

app/support/auto_logout.rb
```diff
@@ -31,6 +38,7 @@ def complete_reservation(od)
     od.actual_cost    = costs[:cost]
     od.actual_subsidy = costs[:subsidy]
     od.save!
+    od.reservation.save!
```
jhanggi: Can we use our `complete!` methods from the other PRs?

app/views/instruments/_instrument_fields.html.haml
```diff
@@ -22,7 +22,11 @@
     = rf.input :port
     = rf.input :username
     = rf.input :password, as: :string
-    = rf.input :auto_logout, :as => :boolean, :label => false, :inline_label => t('instruments.instrument_fields.relay.label.auto_logout')
+    = rf.input :auto_logout do
+      = rf.check_box :auto_logout
```
jhanggi: Can this be `rf.input :auto_logout as: :boolean, label: false`?

db/migrate/20150318015754_add_auto_logout_minutes_to_relays.rb
```diff
@@ -0,0 +1,5 @@
+class AddAutoLogoutMinutesToRelays < ActiveRecord::Migration
+  def change
+    add_column :relays, :auto_logout_minutes, :integer, null: false, default: 60
```
jhanggi: It seems like this should be nullable for relays that don't have auto_logout. It'll seem weird to see that be true and this be 60 in the DB.

spec/app_support/auto_logout_spec.rb
```diff
@@ -19,8 +44,8 @@
 
     it 'does not do anything' do
       expect { action.perform }.not_to change { reservation.reload.actual_end_at }
-      expect(order_detail.state).to eq('new')
-      expect(order_detail.order_status_id).to be_blank
+      expect { action.perform }.not_to change { order_detail.reload.state }
```
jhanggi: This is a little weird to do `action.perform` multiple times. I think we should only be doing that once in the spec.

app/models/power_relay.rb
jdzak: Sure, can do.

app/support/auto_logout.rb
jdzak: None that I know of. I'll remove the `#try` call on it.

app/support/auto_logout.rb
jdzak: Yea it is. I'll change it.

jdzak: Also, for I switched the method name to `Reservation.relay_in_progress` so not to get that confused with reservation-only reservations.

app/support/auto_logout.rb
jdzak: Sure. How does `can_auto_logout?` sound?

app/support/auto_logout.rb
jdzak: Sure, that looks simpler

app/support/auto_logout.rb
jdzak: Yessir.

jdzak: Also, I think `end_reservation!` would work in this case too

app/support/auto_logout.rb
jdzak: Combine autologout check into a single method and maybe call it `should_auto_logout?`

app/views/instruments/_instrument_fields.html.haml
jdzak: I tried, but it keeps adding a label even though I'm specifying `label: false`. That's throwing the style off and pushing the text to the another line.

jhanggi: OK, no problem then.

db/migrate/20150318015754_add_auto_logout_minutes_to_relays.rb
jdzak: Ok.

spec/app_support/auto_logout_spec.rb
jdzak: Yea, I'll change this to use the same shared example that I'm using for #256.

jdzak:
@jhanggi Comments addressed.

jhanggi:
Now we've got some merge conflicts here.

jdzak:
Fixed the conflicts.

app/models/power_relay.rb
```diff
@@ -1,9 +1,14 @@
+require 'active_support/concern'
```
jhanggi: Is this require necessary?

PR #268, opened 2015-03-24
[#94282] Fix bulk upload errors when fields are missing

jdzak:
https://pm.tablexi.com/issues/94282
https://pm.tablexi.com/issues/94770


jdzak:
Found a way to ignore blank lines and lines with only separators but no content.

PR #269, opened 2015-03-26
Fix CI failure on invalid files

jhanggi:
CI suddenly started generating the XLS files with the content type
`application/CDFV2-encrypted`. Cause of this change is unknown, but
this was the easiest way to work around it.


drench:
:+1:

PR #271, opened 2015-03-30
[#92231] Add time to ongoing reservation

jdzak:
https://pm.tablexi.com/issues/92231

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -183,6 +183,8 @@ class ReservationsController < ApplicationController
       return
     end
 
+    @reservation.assign_times_from_params(params[:reservation])
+
     if !current_user.administrator?
       validator = Reservations::DurationChangeValidations.new(@reservation)
       if validator.invalid?
@@ -191,8 +193,6 @@ class ReservationsController < ApplicationController
         return
       end
     end
-
-    @reservation.assign_times_from_params(params[:reservation])
 
     Reservation.transaction do
       begin
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -302,7 +302,7 @@ class ReservationsController < ApplicationController
 
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
-    different_id = params[:id].to_i != @reservation.id
+    params[:id].to_i != @reservation.id ||
     @reservation.actual_end_at ||
     !(@reservation.can_edit? || @reservation.can_customer_edit?)
   end
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -183,16 +183,6 @@ class ReservationsController < ApplicationController
       return
     end
 
-    reservation_params = params[:reservation]
-    if !@reservation.reserve_start_at_editable?
-      reservation_params.tap do |p|
-        p[:reserve_start_date] = @reservation.reserve_start_date
-        p[:reserve_start_hour] = @reservation.reserve_start_hour
-        p[:reserve_start_min] = @reservation.reserve_start_min
-        p[:reserve_start_meridian] = @reservation.reserve_start_meridian
-      end
-    end
-
     @reservation.assign_times_from_params(reservation_params)
 
     validator = Reservations::DurationChangeValidations.new(@reservation)
@@ -280,6 +270,18 @@ class ReservationsController < ApplicationController
   end
 
   private
+
+  def reservation_params
+    reservation_params = params[:reservation]
+    if !@reservation.reserve_start_at_editable?
+      reservation_params.tap do |p|
+        p[:reserve_start_date] = @reservation.reserve_start_date
+        p[:reserve_start_hour] = @reservation.reserve_start_hour
+        p[:reserve_start_min] = @reservation.reserve_start_min
+        p[:reserve_start_meridian] = @reservation.reserve_start_meridian
+      end
+    end
+  end
 
   def switch_instrument_off!
     ReservationInstrumentSwitcher.new(@reservation).switch_off!
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -184,7 +184,6 @@ class ReservationsController < ApplicationController
     end
 
     reservation_params = params[:reservation]
-    binding.pry
     if !@reservation.reserve_start_at_editable?
       reservation_params.tap do |p|
         p[:reserve_start_date] = @reservation.reserve_start_date
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -185,13 +185,11 @@ class ReservationsController < ApplicationController
 
     @reservation.assign_times_from_params(params[:reservation])
 
-    if !current_user.administrator?
-      validator = Reservations::DurationChangeValidations.new(@reservation)
-      if validator.invalid?
-        validator.copy_errors!
-        render :action => "edit"
-        return
-      end
+    validator = Reservations::DurationChangeValidations.new(@reservation)
+    if validator.invalid?
+      validator.copy_errors!
+      render :action => "edit"
+      return
     end
 
     Reservation.transaction do
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -183,7 +183,18 @@ class ReservationsController < ApplicationController
       return
     end
 
-    @reservation.assign_times_from_params(params[:reservation])
+    reservation_params = params[:reservation]
+    binding.pry
+    if !@reservation.reserve_start_at_editable?
+      reservation_params.tap do |p|
+        p[:reserve_start_date] = @reservation.reserve_start_date
+        p[:reserve_start_hour] = @reservation.reserve_start_hour
+        p[:reserve_start_min] = @reservation.reserve_start_min
+        p[:reserve_start_meridian] = @reservation.reserve_start_meridian
+      end
+    end
+
+    @reservation.assign_times_from_params(reservation_params)
 
     validator = Reservations::DurationChangeValidations.new(@reservation)
     if validator.invalid?
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -302,9 +302,12 @@ class ReservationsController < ApplicationController
 
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
+    can_edit = @reservation.can_edit?
+    can_edit &&= @reservation.can_customer_edit? unless current_user.administrator?
     params[:id].to_i != @reservation.id ||
     @reservation.actual_end_at ||
-    !(@reservation.can_edit? || @reservation.can_customer_edit?)
+    @reservation.actual_start_at ||
+    !can_edit
   end
 
   def save_reservation_and_order_detail
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -185,12 +185,7 @@ class ReservationsController < ApplicationController
 
     @reservation.assign_times_from_params(reservation_params)
 
-    validator = Reservations::DurationChangeValidations.new(@reservation)
-    if validator.invalid?
-      set_windows
-      render :action => "edit"
-      return
-    end
+    render_edit and return unless duration_change_valid?
 
     Reservation.transaction do
       begin
@@ -211,8 +206,7 @@ class ReservationsController < ApplicationController
         raise ActiveRecord::Rollback
       end
     end
-    set_windows
-    render :action => "edit"
+    render_edit
   end
 
   # POST /orders/:order_id/order_details/:order_detail_id/reservations/:reservation_id/move
@@ -273,6 +267,7 @@ class ReservationsController < ApplicationController
 
   def reservation_params
     reservation_params = params[:reservation]
+
     if !@reservation.reserve_start_at_editable?
       reservation_params.tap do |p|
         p[:reserve_start_date] = @reservation.reserve_start_date
@@ -281,6 +276,8 @@ class ReservationsController < ApplicationController
         p[:reserve_start_meridian] = @reservation.reserve_start_meridian
       end
     end
+
+    reservation_params
   end
 
   def switch_instrument_off!
@@ -387,5 +384,15 @@ class ReservationsController < ApplicationController
     # 1.week causes a problem with daylight saving week since it's technically longer
     # than a week
     @end_at - @start_at > 8.days
+  end
+
+  def render_edit
+    set_windows
+    render :action => "edit"
+  end
+
+  def duration_change_valid?
+    validator = Reservations::DurationChangeValidations.new(@reservation)
+    validator.valid?
   end
 end
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -187,7 +187,7 @@ class ReservationsController < ApplicationController
 
     validator = Reservations::DurationChangeValidations.new(@reservation)
     if validator.invalid?
-      validator.copy_errors!
+      set_windows
       render :action => "edit"
       return
     end
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -303,8 +303,8 @@ class ReservationsController < ApplicationController
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
     different_id = params[:id].to_i != @reservation.id
-    cannot_edit = current_user.administrator? ? !@reservation.can_edit? : !@reservation.can_customer_edit?
-    different_id || cannot_edit
+    @reservation.actual_end_at ||
+    !(@reservation.can_edit? || @reservation.can_customer_edit?)
   end
 
   def save_reservation_and_order_detail
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -183,6 +183,15 @@ class ReservationsController < ApplicationController
       return
     end
 
+    if !current_user.administrator?
+      validator = Reservations::DurationChangeValidations.new(@reservation)
+      if validator.invalid?
+        validator.copy_errors!
+        render :action => "edit"
+        return
+      end
+    end
+
     @reservation.assign_times_from_params(params[:reservation])
 
     Reservation.transaction do
@@ -295,10 +304,9 @@ class ReservationsController < ApplicationController
 
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
-    params[:id].to_i != @reservation.id ||
-    @reservation.actual_start_at ||
-    @reservation.actual_end_at ||
-    !(@reservation.can_edit? || @reservation.can_customer_edit?)
+    different_id = params[:id].to_i != @reservation.id
+    cannot_edit = current_user.administrator? ? !@reservation.can_edit? : !@reservation.can_customer_edit?
+    different_id || cannot_edit
   end
 
   def save_reservation_and_order_detail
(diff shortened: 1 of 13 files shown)


app/controllers/reservations_controller.rb
```diff
@@ -183,6 +183,15 @@ def update
       return
     end
 
+    if !current_user.administrator?
```
jhanggi: It seems like this should come after the `assign_times_from_params` call. I could be wrong, but without that, it probably won't maintain your changes you made through the interface.

app/controllers/reservations_controller.rb
```diff
@@ -183,6 +183,15 @@ def update
...
+      validator = Reservations::DurationChangeValidations.new(@reservation)
+      if validator.invalid?
+        validator.copy_errors!
```
jhanggi: Can we move this into the validator class itself?

app/controllers/reservations_controller.rb
jhanggi: In the context where this action is used, is this check necessary? Most of the time, won't the administrator will be using one of the other reservation change views (like the order detail edit popup).

app/controllers/reservations_controller.rb
```diff
@@ -295,10 +304,9 @@ def ability_resource
 
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
-    params[:id].to_i != @reservation.id ||
-    @reservation.actual_start_at ||
-    @reservation.actual_end_at ||
-    !(@reservation.can_edit? || @reservation.can_customer_edit?)
+    different_id = params[:id].to_i != @reservation.id
+    cannot_edit = current_user.administrator? ? !@reservation.can_edit? : !@reservation.can_customer_edit?
```
jhanggi: I'm not sure this check needs to change. Also, we want to be careful with how we talk about "administrators". `User#administer?` is asking if the user is a global admin. Usually, we would want to use some other check like `manager_of` or something in the cancan abilities.

app/models/reservation.rb
```diff
@@ -189,7 +189,22 @@ def can_cancel?
   end
 
   def can_customer_edit?
-    !canceled? && !complete? && before_lock_window?
+    !canceled? && !complete? && (reserve_start_at_editable? || reserve_end_at_editable?)
+  end
+
+  def reserve_start_at_editable?
+    Time.zone.now < reserve_start_at && before_lock_window?
```
jhanggi: This feels like it might be redundant. Wouldn't `before_lock_window?` always be true when `Time.zone.now < reserve_start_at`?

app/support/reservations/duration_change_validations.rb
```diff
@@ -0,0 +1,32 @@
...
+  include ActiveModel::Validations
+
+  validate :start_time_not_changed, unless: Proc.new { |r| r.reservation.reserve_start_at_editable? }
+  validate :duration_not_shortened, if: Proc.new { |r| Time.zone.now >= r.reservation.reserve_start_at }
+
+  def initialize(reservation)
+    @reservation = reservation
+  end
+
+  def reservation
```
jhanggi: Not a big deal, but I'm partial to `attr_reader` for this.

app/support/reservations/duration_change_validations.rb
```diff
@@ -0,0 +1,32 @@
...
+    @reservation
+  end
+
+  def copy_errors!
```
jhanggi: Do we have access to `after_validation` that we could hook this method into? I'd like to see this happen automatically after running the validations. Maybe overriding `valid?` with a call to `super` in there.

app/support/reservations/duration_change_validations.rb
```diff
@@ -0,0 +1,32 @@
...
+    errors.each do |e|
+      reservation.errors.add(e)
+    end
+  end
+
+  def start_time_not_changed
+    if reservation.reserve_start_at_changed?
+      errors.add(:reserve_start_at, "cannot change once the reservation has started")
```
jhanggi: Can we put the text into I18n?

app/support/reservations/duration_change_validations.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def copy_errors!
+    errors.each do |e|
+      reservation.errors.add(e)
+    end
+  end
+
+  def start_time_not_changed
+    if reservation.reserve_start_at_changed?
```
jhanggi: Would it make sense to move the check we're doing in the Proc into here? Would that be easier to read?

app/support/reservations/duration_change_validations.rb
```diff
@@ -0,0 +1,32 @@
...
+      errors.add(:reserve_start_at, "cannot change once the reservation has started")
+    end
+  end
+
+  def duration_not_shortened
```
jhanggi: Same notes as the other method.

app/views/reservations/_reservation_fields.html.haml
```diff
@@ -1,8 +1,20 @@
...
+      .controls
+        = f.object.reserve_start_date
   .control-group
     .controls
       .string.optional.control-label &nbsp;
-      = time_select f, :reserve_start, minute_step: @instrument.reserve_interval
+      - if start_editable
+        = time_select f, :reserve_start, minute_step: @instrument.reserve_interval
+      - else
+        .controls
```
jhanggi: Can we have `.controls` inside of `.controls`?

spec/app_support/reservations/duration_change_validations_spec.rb
```diff
@@ -0,0 +1,91 @@
+require 'spec_helper'
+
+describe Reservations::DurationChangeValidations do
+
+  subject(:validator) { described_class.new(reservation) }
+  let(:reservation) { create :setup_reservation }
```
jhanggi: Can these tests be run with `build_stubbed`?

spec/app_support/reservations/duration_change_validations_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+
+    context "with an ongoing reseration" do
+      before do
+        reservation.assign_attributes(
+          reserve_start_at: 30.minutes.ago,
+          reserve_end_at: 30.minutes.from_now)
+      end
+
+      it { expect(validator).to be_invalid }
```
jhanggi: It's probably worth checking that the `errors` contains the right thing. That way we're not returning false positives when some other validation is causing it to be invalid.

spec/controllers/reservations_controller_spec.rb
```diff
@@ -603,7 +603,8 @@
       context 'next reservation is between 5 minutes' do
         let(:next_reservation) do
           Reservation.new :reserve_start_at => Time.zone.parse('2013-08-15 12:02'),
-                          :reserve_end_at   => Time.zone.parse('2013-08-15 12:17')
+                          :reserve_end_at   => Time.zone.parse('2013-08-15 12:17'),
```
jhanggi: Can we start switching to 1.9 hash syntax as we're working on new stuff? No need for a full-out campaign, but we should change it where we're touching the code.

app/controllers/reservations_controller.rb
jdzak: Yea, you're right about `assign_times_from_params`, I'll fix that.

Also, I'll remove the admin check for the validation. Sorry, I'm still kind of unfamiliar with how each role edits different details, but that makes sense.

app/controllers/reservations_controller.rb
jdzak: Sure.

app/controllers/reservations_controller.rb
jdzak: I removed  the check for `@reservation.actual_start_at` because if a reservation has been started, then it will have `@reservation.actual_start_at` which would make `invalid_for_update?` return true.

I can probably re-add the check for `@reservation.actual_end_at` because we want to keep that.

For the abilities check, maybe we can work through this together because I want to make sure I do it right.

app/models/reservation.rb
jdzak: I don't think so because if the lock window is 15 minutes before the reservation, then this condition would be false for the 15 minute block before the reservation starts. Right?

app/support/reservations/duration_change_validations.rb
jdzak: Me too, I'll change it.

app/support/reservations/duration_change_validations.rb
jdzak: Yep, i'll change it.

app/support/reservations/duration_change_validations.rb
jdzak: Sure.

spec/controllers/reservations_controller_spec.rb
jdzak: Sure.

spec/app_support/reservations/duration_change_validations_spec.rb
jdzak: Sure.

spec/app_support/reservations/duration_change_validations_spec.rb
jdzak: I tried but I got the error:
```
     Failure/Error: let(:reservation) { build_stubbed :setup_reservation }
     ActiveRecord::InvalidForeignKey:
       Mysql2::Error: Cannot add or update a child row: a foreign key constraint fails (`nucore-open_test`.`schedules`, CONSTRAINT `fk_schedules_facility` FOREIGN KEY (`facility_id`) REFERENCES `facilities` (`id`)): INSERT INTO `schedules` (`created_at`, `facility_id`, `name`, `updated_at`) VALUES ('2015-09-11 14:30:00', 1001, 'Schedule 1', '2015-09-11 14:30:00')
     # ./spec/factories/products.rb:81:in `block (3 levels) in <top (required)>'
     # ./spec/app_support/reservations/duration_change_validations_spec.rb:6:in `block (2 levels) in <top (required)>'
     # ./spec/app_support/reservations/duration_change_validations_spec.rb:5:in `block (2 levels) in <top (required)>'
     # ./spec/app_support/reservations/duration_change_validations_spec.rb:42:in `block (3 levels) in <top (required)>'
```

app/views/reservations/_reservation_fields.html.haml
jdzak: I'm not sure, I've been experimenting with the styles because I can't get it to vertically align how i want. I may need some help with this too.

app/support/reservations/duration_change_validations.rb
jdzak: Yea, I was on the fence about this. Having the proc with the if/unless feels nice because we can scroll through the list of validations and see the preconditions for when each validation is fired.

IDK, I don't feel strongly, so if you prefer the other way I can change it.

jdzak:
Ready for review. 

There's one failing test which I'm still trying to debug, but other than that it's ready to go.

jdzak:
Btw the failing test is `rspec spec/controllers/reservations_controller_spec.rb:820` if you were curious in taking a look at it.

app/controllers/reservations_controller.rb
```diff
@@ -295,8 +302,7 @@ def ability_resource
 
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
-    params[:id].to_i != @reservation.id ||
-    @reservation.actual_start_at ||
+    different_id = params[:id].to_i != @reservation.id
```
jhanggi: Where does `different_id` get used? Should this have a `||` at the end of the line?

app/support/reservations/duration_change_validations.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  def copy_errors!
+    errors.each do |field, message|
+      reservation.errors.add(field, message)
+    end
+  end
+
+  def valid?(context = nil)
+    super(context).tap do
+      copy_errors!
```
jhanggi: Could we use the `after_validation` callback for this?

app/views/reservations/_reservation_fields.html.haml
```diff
@@ -1,8 +1,15 @@
 .datetime-block
-  = f.input :reserve_start_date, :input_html => { :class => 'datepicker' }
+  - start_disabled = f.object.persisted? && !f.object.reserve_start_at_editable?
+  = f.input :reserve_start_date, :input_html => { :class => 'datepicker', disabled: start_disabled }
   .control-group
     .controls
       .string.optional.control-label &nbsp;
-      = time_select f, :reserve_start, minute_step: @instrument.reserve_interval
+      = time_select f, :reserve_start, { minute_step: @instrument.reserve_interval }, disabled: start_disabled
+      - if start_disabled
```
jhanggi: Is this really necessary? I would expect it to simply not update the start at field if it doesn't get the parameters.

spec/controllers/reservations_controller_spec.rb
```diff
@@ -618,7 +619,8 @@
       context 'next reservation is on a 5 minute, but with seconds' do
         let(:next_reservation) do
           Reservation.new :reserve_start_at => Time.zone.parse('2013-08-15 12:05:30'),
-                          :reserve_end_at   => Time.zone.parse('2013-08-15 12:20:30')
+                          :reserve_end_at   => Time.zone.parse('2013-08-15 12:20:30'),
```
jdzak: Let me convert these to the 1.9 hash syntax also.

app/views/reservations/_reservation_fields.html.haml
jdzak: Initially I tried to make it work without the hidden fields, but I was running into a issues because `Reservation#assign_times_from_params` will [set Reservation#reserve_start_at and Reservation#reserve_end_at to nil so the dates are set correctly](https://github.com/tablexi/nucore-open/blob/master/app/support/reservations/date_support.rb#L177-L186). Alternatively, I could probably fill in reserve_start_at and reserve_end_at in the controller if they don't exist in the params so we can remove the hidden fields from the page.

app/controllers/reservations_controller.rb
jdzak: Whoops, let me fix both of those issues.

app/support/reservations/duration_change_validations.rb
jdzak: Ya, I'll change it.

jdzak:
Ready for re-review.

app/controllers/reservations_controller.rb
```diff
@@ -295,10 +302,12 @@ def ability_resource
 
   # TODO you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
+    can_edit = @reservation.can_edit?
+    can_edit &&= @reservation.can_customer_edit? unless current_user.administrator?
     params[:id].to_i != @reservation.id ||
-    @reservation.actual_start_at ||
     @reservation.actual_end_at ||
-    !(@reservation.can_edit? || @reservation.can_customer_edit?)
+    @reservation.actual_start_at ||
```
jhanggi: I believe this is now preventing me from editing a started reservation.

jhanggi:
I removed that one line to try to do some further testing.

If I try to shorten the time, I get the "Reserve end at cannot be shortened once the reservation has started" message, but now the duration is set to the shortened time and is disabled. I also see a similar thing when my lengthened reservation conflicts with another reservation.

jdzak:
Weird. It wasn't doing that before. I was tinkering with
`invalid_for_update?` recently, I wonder if I screwed something up with
that. I'll take a look at it tonight.

jdzak:
Alright, fixed the issue. I reverted some of the `ReservationController#invalid_for_update?` code and forgot to re-test with relay reservations. Also, I'm not enabling/disabling the start/end time fields based on the original `Reservation` record and not the un-persisted user entered values.

jdzak:
Also, Todd said we should enforce the lock window before a reservation has started, so I added code for that.

jhanggi:
It's 1:08pm. I "Create & Start" a reservation beginning at 1:10pm.

Current behavior:  For the next two minutes, I have the ability to change the reservation start time. Once 1:10 hits, it locks the start_at.

Expected behavior: Once the reservation has actually started, I should not be able to edit the start time.

jhanggi:
If the start time is locked, I can edit the hidden fields (not a problem). When I submit, it is changing the reserve_start_at of the reservation. If the start at is locked, we should throw away any of the start at fields in the controller.

jdzak:
Alright, fixed those two issues. Ready for re-review.

app/controllers/reservations_controller.rb
```diff
@@ -183,7 +183,24 @@ def update
       return
     end
 
-    @reservation.assign_times_from_params(params[:reservation])
+    reservation_params = params[:reservation]
```
jhanggi: What if we took this entire chunk and extracted it into a `reservation_params` method?

app/controllers/reservations_controller.rb
```diff
@@ -183,7 +183,24 @@ def update
...
+        p[:reserve_start_hour] = @reservation.reserve_start_hour
+        p[:reserve_start_min] = @reservation.reserve_start_min
+        p[:reserve_start_meridian] = @reservation.reserve_start_meridian
+      end
+    end
+
+    @reservation.assign_times_from_params(reservation_params)
+
+    validator = Reservations::DurationChangeValidations.new(@reservation)
+    if validator.invalid?
```
jhanggi: This might also be a candidate for extracting the method.

app/helpers/reservations_helper.rb
```diff
@@ -27,4 +27,18 @@ def reservation_actions(reservation)
   def reservation_view_edit_link(reservation)
     ReservationUserActionPresenter.new(self, reservation).view_edit_link
   end
+
+  def start_time_disabled(reservation)
```
jhanggi: Since these methods return booleans, perhaps they should end with `?`.

app/models/reservation.rb
```diff
@@ -189,13 +189,32 @@ def can_cancel?
...
+
+  def reserve_start_at_editable?
+    before_lock_window? && actual_start_at.blank?
+  end
+
+  def reserve_end_at_editable?
+    outside_lock_window? && Time.zone.now <= reserve_end_at && next_duration_available? && actual_end_at.blank?
+  end
+
+  def next_duration_available?
```
jhanggi: Is this at all related to `ReservationFinder#next_reservation` in Product::SchedulingSupport? It's a really confusing method, but it might do something similar to what you're looking for.

We should make sure that you also can't extend your reservation outside of the schedule rules. E.g. There's a schedule rule for 9-5, and you have a reservation from 4-5. You shouldn't be able to extend beyond that.

app/views/reservations/_reservation_fields.html.haml
jhanggi: Does this relate at all to the `reservation_params.tap` up in the controller? With that, could we get rid of this?

jdzak: Yep.

app/controllers/reservations_controller.rb
jdzak: I like, that'll look a lot cleaner.

app/helpers/reservations_helper.rb
jdzak: Sure.

app/models/reservation.rb
jdzak: Yea, I'll switch to use that instead and add a spec for the end of day situation.

app/controllers/reservations_controller.rb
jhanggi: On thing that might be nice is to make a `render_edit` method that does `set_window` and `render :edit`. Then you could use one method here and at the bottom of `update`.

So this would become something like
```ruby
render_update and return unless duration_change_valid?
```
then
```ruby
def duration_change_valid?
   Reservations::DurationChangeValidations.new(@reservation).valid?
end
```

It might be worth reworking `invalid_for_update?` to be a positive. Not a big deal, I generally prefer positive method names than negative. Unless that leads you to try something like `unless/else`. That way lies the path to madness.

jdzak: LOL. Makes sense, I'll change this.

jdzak:
Ready for re-review.

jhanggi:
:+1: It might be interesting to extract some of the `can_customer_edit?` and `reserve_start_at_editable?` type methods into a policy/permissions class, but I'm not sure if it's worth the effort at this point, but if we end up having to play more with this feature, it might be worth considering.

Also, we'll wait until we get the relay features released to prod before we merge this down.

jhanggi:
I ran this on CI, and it has some failures.

jdzak:
Fixed the spec failures. It was failing because I added some code to check if `actual_start_at` exists on the reservation, but accidentally removed the check to see if `reserve_start_at check` exists.

PR #272, opened 2015-04-03
Fix other reservation using relay scope

jdzak:
Fixes issue not filtering by active reservations when checking for other reservation using relay.


jhanggi:
Can you add a spec to make sure that if there is a problem reservation that it still allows triggering the relay? This test should fail before this change and pass afterwards.

jdzak:
The specs passed.

PR #273, opened 2015-04-07
[#73235] Label changes per Aaron's feedback

jdzak:
From Aaron,
> On the Instrument Details tab, please change the following fields: "Automatic Logout?" change to "Auto-Relay Shutoff?" (both on "display" and "edit" screens). "Automatic Logout After Minutes" change to "Auto-Relay Shutoff After X Minutes" (just on "display" screen). This is more accurate, since NUCore doesn't really log users out.

### DIFF
--- a/config/locales/en.models.yml
+++ b/config/locales/en.models.yml
@@ -235,8 +235,8 @@ en:
         port: Relay Port
         username: Relay Username
         password: Relay Password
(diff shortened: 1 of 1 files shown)


jdzak:
Also, I sent Aaron a screenshot of this change and he said it looks good.

PR #274, opened 2015-04-07
[#73235] Consider schedules when counting running relays

jdzak:


app/models/product.rb
```diff
@@ -3,6 +3,7 @@ class Product < ActiveRecord::Base
   belongs_to :facility
   belongs_to :initial_order_status, :class_name => 'OrderStatus'
   belongs_to :facility_account
+  belongs_to :schedule
```
jhanggi: This is defined already in Instrument through `Products::SchedulingSupport`.

PR #275, opened 2015-04-14
[#95854] Validate notes in bulk import

jdzak:
https://pm.tablexi.com/issues/95854

### DIFF
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -121,7 +121,7 @@ class OrderImport < ActiveRecord::Base
             result.successes += 1
           end
         end
-      rescue Exception => e
+      rescue => e
         set_error_mode
         result.failures += 1
         self.error_report += "Unable to open CSV File: #{e.message}"
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -121,13 +121,13 @@ class OrderImport < ActiveRecord::Base
             result.successes += 1
           end
         end
-
-        raise ActiveRecord::Rollback if result.failed?
-      rescue ArgumentError => e
+      rescue Exception => e
         set_error_mode
         result.failures += 1
         self.error_report += "Unable to open CSV File: #{e.message}"
       end
+
+      raise ActiveRecord::Rollback if result.failed?
     end
 
     if result.succeeded?
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -127,8 +127,6 @@ class OrderImport < ActiveRecord::Base
         set_error_mode
         result.failures += 1
         self.error_report += "Unable to open CSV File: #{e.message}"
-        ActiveSupport::Notifications.instrument('background_error',
-          exception: e, information: "Failed to bulk import: #{upload_file_path}")
       end
     end
 
--- a/app/models/order_import.rb
+++ b/app/models/order_import.rb
@@ -110,18 +110,26 @@ class OrderImport < ActiveRecord::Base
 
   def handle_save_nothing_on_error # TODO refactor and rename
     Order.transaction do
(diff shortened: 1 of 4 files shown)


app/support/order_row_importer.rb
```diff
@@ -223,4 +224,8 @@ def validate_product_is_importable
   def validate_user
     add_error("Invalid username or email") if user.blank?
   end
+
+  def validate_notes
+    add_error("Invalid note, 100 characters max") if note.present? && note.size > 100
```
jhanggi: I'm not a huge fan of having `100` here. It's kind of a magical number, and it duplicates the limit in the database and the validation that's already on OrderDetail.

jhanggi:
While this solves the immediate issue, it doesn't really solve the bigger concern I have which is that we've had several errors like this, and I think we need a more comprehensive fix so that we don't run into more of these issues in the future.

Is there a way we can rescue exceptions on a row by row level, and add the exception message to the error report? Maybe it should notify us as well so we know that there's a problem, but we might not need to fix it immediately if it errors gracefully.

app/support/order_row_importer.rb
jdzak: I tried to avoid the duplication by using the order details created in `#add_product_to_order` (see below) but it seems ackward because ItemAdder will `create!` the order details instead of building them. I can't seem to work around this without some major refactoring.

Maybe we can DRY up the note validation by making a constant for the max size, but I the larger issue of missing validations will still exist. Thoughts?

```
def add_product_to_order
  begin
    @order_details = order.add(product, quantity, note: note)
  rescue ActiveRecord::RecordInvalid => e
    add_error(e.message)
    order.destroy
    return
  end

  purchase_order! unless order.purchased?
  backdate_order_details_to_complete!
end
```

jdzak: Also, it looks like this code is causing specs to fail.. so maybe this workaround doesn't even work.

jhanggi: I'm not sure the `order.destroy` is necessary, but I think the idea makes sense.

jdzak:
Alright, now we're showing errors if the Order or OrderDetail models fail validation. Also we're failing gracefully and showing an error if the CSV fails to open because it's invalid.

Ready for re-review.

app/models/order_import.rb
```diff
@@ -110,18 +110,24 @@ def handle_save_clean_orders # TODO refactor and rename
...
+            result.failures += 1
+          else
+            result.successes += 1
+          end
         end
-      end
 
-      raise ActiveRecord::Rollback if result.failed?
+        raise ActiveRecord::Rollback if result.failed?
+      rescue ArgumentError => e
```
jhanggi: Do we want to rescue more than just ArgumentError? I could see us wanting to catch everything (except Rollback, but that could possibly be taken out of the begin block).

jhanggi:
Just my one comment about what errors we're catching. Otherwise, looking pretty good.

jdzak:
Ready for re-review.

k8dxf

app/models/order_import.rb
```diff
@@ -110,15 +110,21 @@ def handle_save_clean_orders # TODO refactor and rename
...
+          row_importer = import_row(row)
+          self.error_report += row_importer.row_with_errors.to_csv
+
+          if row_importer.errors?
+            result.failures += 1
+          else
+            result.successes += 1
+          end
         end
+      rescue Exception => e
```
jhanggi: We should probably just `rescue => e`. That'll just catch `StandardError`. It's usually not a good idea to `rescue Exception` as `Exception` includes a lot of things like interrupts.

jdzak: Good call. I'll fix it.

jdzak:
Fixed comments. Ready for re-review.

PR #276, opened 2015-04-17
[#92949] Consider admin reservations when checking for running reservations

jdzak:
Previously we didn't consider admin reservations when looking for running reservations. This changes the logic to now consider admin reservations.

https://pm.tablexi.com/issues/92949

### DIFF
--- a/app/support/reservations/relay_support.rb
+++ b/app/support/reservations/relay_support.rb
@@ -32,8 +32,9 @@ module Reservations::RelaySupport
 
   def other_reservations_using_relay
     order_detail.reservation.product.schedule.reservations
-      .active_without_admin
+      .active_with_admin
       .relay_in_progress
+      .where(order_details: { state: ['new', 'inprocess', nil]})
       .not_this_reservation(self)
   end
 
--- a/app/support/reservations/relay_support.rb
+++ b/app/support/reservations/relay_support.rb
@@ -32,7 +32,7 @@ module Reservations::RelaySupport
 
   def other_reservations_using_relay
     order_detail.reservation.product.schedule.reservations
-      .active_with_admin
+      .active
       .relay_in_progress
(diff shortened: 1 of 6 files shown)


app/support/reservations/relay_support.rb
```diff
@@ -32,8 +32,9 @@ def can_kill_power?
 
   def other_reservations_using_relay
     order_detail.reservation.product.schedule.reservations
-      .active_without_admin
+      .active_with_admin
```
jhanggi: Do we actually now need the `active_with_admin` and `active_without_admin` methods anymore? Can we go back to just `active` like we had before?

spec/factories/reservations.rb
```diff
@@ -6,6 +6,19 @@
     trait :yesterday do
       reserve_start_at { Time.zone.parse("#{Date.today} 10:00:00") - 1.day }
     end
+
+    trait :later_yesterday do
+      reserve_start_at { Time.zone.parse("#{Date.today} 10:00:00") - 1.day + 1.hour }
+    end
+
+    trait :running do
+      after(:create) do |reservation|
```
jhanggi: Why do we need this in an `after(:create)` block? Can we not just set these as normal factory methods?

```
reserve_start_at { 30.minutes.ago }
reserve_end_at { 30.minutes.from_now }
actual_start_at { 30.minutes.ago }
```

jdzak: Hmm, I'm not sure. I'll switch to normal factory methods.

app/support/reservations/relay_support.rb
jdzak: Ya, i think we can. I'll change it.

jdzak:
Fixed comments. Ready for re-review.

PR #279, opened 2015-04-29
[#50175] Restrict global price group chart string changes

jdzak:
This is work in progress PR just to get comments.

https://pm.tablexi.com/issues/50175

### DIFF
--- a/app/controllers/concerns/price_group_members_controller.rb
+++ b/app/controllers/concerns/price_group_members_controller.rb
@@ -13,7 +13,9 @@ module PriceGroupMembersController
     load_and_authorize_resource
   end
 
-  def new; end
+  def new
+    @price_group_ability.authorize! :create, PriceGroupMember
+  end
 
   def initialize
     @active_tab = "admin_facility"
@@ -21,6 +23,8 @@ module PriceGroupMembersController
   end
 
   def create
+    @price_group_ability.authorize! :create, PriceGroupMember
+
     if price_group_member.save
       set_flash(:notice, :create, create_flash_arguments)
     else
@@ -30,6 +34,8 @@ module PriceGroupMembersController
   end
 
   def destroy
+    @price_group_ability.authorize! :destroy, price_group_member
+
     if destroy_price_group_member!
       set_flash(:notice, :destroy)
     else
--- a/app/controllers/concerns/price_group_members_controller.rb
+++ b/app/controllers/concerns/price_group_members_controller.rb
@@ -7,15 +7,14 @@ module PriceGroupMembersController
     before_filter :check_acting_as
     before_filter :init_current_facility
     before_filter :load_price_group_and_ability!
+    before_filter :authorize_price_group_member, only: [:new, :create, :destroy]
 
     layout "two_column"
 
     load_and_authorize_resource
   end
 
-  def new
-    @price_group_ability.authorize! :create, PriceGroupMember
-  end
+  def new; end
 
   def initialize
     @active_tab = "admin_facility"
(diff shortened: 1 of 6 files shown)


jdzak:
Alright, I cleaned this up. It should be ready for review now.

app/controllers/concerns/price_group_members_controller.rb
```diff
@@ -56,4 +57,8 @@ def load_price_group_and_ability!
     @price_group = current_facility.price_groups.find(params[:price_group_id])
     @price_group_ability = Ability.new(current_user, @price_group, self)
   end
+
+  def authorize_price_group_member
+    @price_group_ability.authorize! :create, PriceGroupMember
```
jhanggi: While it's likely fine in this scenario, you're calling this before filter on `destroy` as well, which is technically a different permission. Could you use `action_name`?

spec/controllers/account_price_group_members_controller_spec.rb
```diff
@@ -3,30 +3,64 @@
...
+      end
+    end
+  end
 
   context "new" do
     before :each do
       @method=:get
       @action=:new
-      @params={ :facility_id => @authable.url_name, :price_group_id => @price_group.id }
+      @account=create_nufs_account_with_owner
```
jhanggi: Can we clean up some of this formatting and add spaces around the `=`? I'd rather change the whole file to match our organizational styles than continue to adhere to older stylings for consistency. (But I'd only worry about the files that you're currently touching).

spec/controllers/account_price_group_members_controller_spec.rb
```diff
@@ -38,20 +72,23 @@
     before :each do
       @method=:post
       @action=:create
-      account=create_nufs_account_with_owner
-      @params={ :facility_id => @authable.url_name, :price_group_id => @price_group.id, :account_id => account.id }
+      @account=create_nufs_account_with_owner
+      @params={ :facility_id => @authable.url_name, :price_group_id => price_group.id, :account_id => @account.id }
```
jhanggi: Can we switch to 1.9 syntax?

jhanggi:
Overall, looks pretty good. I do have a question:

Should a director or whoever be able to click in and see the users/payment sources within the cancer center? Currently, they're able to see the list of CC payment sources, but not the list of users. Should we restrict the show view as well?

app/controllers/concerns/price_group_members_controller.rb
jdzak: Whoops, I'll fix that.

spec/controllers/account_price_group_members_controller_spec.rb
jdzak: Sure.

spec/controllers/account_price_group_members_controller_spec.rb
jdzak: Yeppers.

jdzak:
Ready for re-review.

jdzak:
Also, regarding your question... I'm double checking with Todd but I believe there were scenarios where it was useful to allow admin, facility admins, and directors to see the payment sources for any price group.

jdzak:
Straight from the horses mouth (Todd is the horse)...

> Yes, any role with rights to the Price group view now should be able to
> view-only the list of Cancer Center users. Likewise, for the other global
> price group, they should be able to view-only the payment sources.
>
> For local price groups, they should be able to edit/view all.

jhanggi:
:+1: 

I want to wait until we get a stable release of the relay pieces before we merge this down.

PR #282, opened 2015-05-08
[#96853] Use SQL for query instead of loading all

jhanggi:
OrderDetail.need_journal was loading ~3500 records. When there are
several order details on the Orders#show view, this got very slow.

This pushes the lookup for existence into SQL.


drench:
LGTM. My only question is do we have any specs that touch this?

PR #283, opened 2015-05-12
[#72141] Export account transactions

jdzak:
https://pm.tablexi.com/issues/73141

### DIFF
--- a/app/models/reports/account_transactions_report.rb
+++ b/app/models/reports/account_transactions_report.rb
@@ -68,24 +68,24 @@ class Reports::AccountTransactionsReport
   private
 
   def order_detail_cost(order_detail)
-    order_detail.send(:extend, PriceDisplayment)
+    order_detail.extend(PriceDisplayment)
     order_detail.display_cost
   end
 
   def order_detail_subsidy(order_detail)
-    order_detail.send(:extend, PriceDisplayment)
+    order_detail.extend(PriceDisplayment)
     order_detail.display_subsidy
   end
 
   def order_detail_total(order_detail)
-    order_detail.send(:extend, PriceDisplayment)
+    order_detail.extend(PriceDisplayment)
     order_detail.display_total
   end
 
   def order_detail_quantity(order_detail)
     out = ''
     unless order_detail.problem?
-      order_detail.extend PriceDisplayment
+      order_detail.extend(PriceDisplayment)
       out = order_detail.display_quantity.value
     end
     out
--- a/app/models/reports/account_transactions_report.rb
+++ b/app/models/reports/account_transactions_report.rb
@@ -1,0 +1,65 @@
+require 'csv'
+
+class Reports::AccountTransactionsReport
+  include ApplicationHelper
+  include ERB::Util
+
+  def initialize(order_details, options = {})
+    @order_details = order_details
+    @date_range_field = options[:date_range_field] || 'fulfilled_at'
+  end
+
+  def to_csv
+    report = []
+
+    report << CSV.generate_line(headers)
+
+    @order_details.each do |od|
+      report << CSV.generate_line(build_row(od))
+    end
+
+    report.join
+  end
+
+  private
+
+  def headers
+    [
+      Order.model_name.human,
+      OrderDetail.model_name.human,
+      OrderDetail.human_attribute_name(@date_range_field),
+      Facility.model_name.human,
+      OrderDetail.human_attribute_name('description'),
+      OrderDetail.human_attribute_name('quantity'),
+      OrderDetail.human_attribute_name('user'),
+      OrderDetail.human_attribute_name('cost'),
+      OrderDetail.human_attribute_name('order_status')
+    ]
+  end
+
+  def build_row(order_detail)
+    [
+      order_detail.order.id,
+      order_detail.id,
+      format_usa_date(order_detail.send(:"#{@date_range_field}")),
+      order_detail.order.facility,
+      order_detail_description(order_detail),
+      order_detail_quantity(order_detail),
+      order_detail.order.user.full_name,
+      order_detail_total(order_detail),
+      order_detail.order_status
+    ]
+  end
+
+  private
+
+  def order_detail_total(order_detail)
+    order_detail.send(:extend, PriceDisplayment)
+    order_detail.display_total
+  end
+
+  def order_detail_quantity(order_detail)
+    order_detail.extend PriceDisplayment
+    order_detail.quantity_display.value
+  end
+end
--- a/app/models/reports/account_transactions_report.rb
+++ b/app/models/reports/account_transactions_report.rb
@@ -30,22 +30,36 @@ class Reports::AccountTransactionsReport
       OrderDetail.human_attribute_name(@date_range_field),
       Facility.model_name.human,
       OrderDetail.human_attribute_name('description'),
+      Reservation.human_attribute_name('reserve_start_at'),
+      Reservation.human_attribute_name('reserve_end_at'),
+      Reservation.human_attribute_name('actual_start_at'),
+      Reservation.human_attribute_name('actual_end_at'),
       OrderDetail.human_attribute_name('quantity'),
       OrderDetail.human_attribute_name('user'),
       OrderDetail.human_attribute_name('cost'),
+      OrderDetail.human_attribute_name('subsidy'),
+      OrderDetail.human_attribute_name('total'),
       OrderDetail.human_attribute_name('order_status')
     ]
   end
 
   def build_row(order_detail)
+    reservation = order_detail.reservation
+
     [
       order_detail.order.id,
       order_detail.id,
       format_usa_date(order_detail.send(:"#{@date_range_field}")),
       order_detail.order.facility,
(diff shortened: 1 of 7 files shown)


app/models/reports/account_transactions_report.rb
```diff
@@ -0,0 +1,65 @@
...
+
+  private
+
+  def order_detail_total(order_detail)
+    order_detail.send(:extend, PriceDisplayment)
+    order_detail.display_total
+  end
+
+  def order_detail_quantity(order_detail)
+    order_detail.extend PriceDisplayment
```
jhanggi: We should be consistent with how we DCI.

app/models/reports/account_transactions_report.rb
```diff
@@ -0,0 +1,65 @@
...
+      OrderDetail.human_attribute_name('cost'),
+      OrderDetail.human_attribute_name('order_status')
+    ]
+  end
+
+  def build_row(order_detail)
+    [
+      order_detail.order.id,
+      order_detail.id,
+      format_usa_date(order_detail.send(:"#{@date_range_field}")),
```
jhanggi: Is `@date_range_field` whitelisted anywhere?

lib/transaction_search.rb
```diff
@@ -41,7 +41,10 @@ def define_search_method(new_method_name, old_method_name)
         do_search(@search_fields)
         add_optimizations
         sort_and_paginate
-        render :layout => @layout if @layout
+        respond_with(@order_details) do |format|
+          format.html { render layout: @layout if @layout }
+          format.csv { render text: Reports::AccountTransactionsReport.new(@order_details, date_range_field: @date_range_field).to_csv }
```
jhanggi: Do we need to set any additional headers? See `CSVHelper#set_csv_headers`. On that note, it might be nice get rid of the hack in that file where we use `CSVHelper::CSV` in order to support multiple ruby 1.8.7 and 1.9.2+. That could be a separate PR, but I would support it as part of this ticket.

app/support/price_displayment.rb
```diff
@@ -54,18 +54,34 @@ def actual_cost?
...
+
+    def output
+      type == :timeinput ? time_quantity_html : value
+    end
+  end
+
+  def quantity_display
     if reservation.try(:actual_duration_mins) && reservation.actual_duration_mins > 0
-      content_tag :span, reservation.actual_duration_mins, :class => 'timeinput'
+      QuantityDisplay.new(reservation.actual_duration_mins, :timeinput)
```
jhanggi: I'm assuming you're planning to use this in other displays as well. Otherwise, is it necessary to pass `:timeinput` in?

app/models/reports/account_transactions_report.rb
jdzak: Now it does. :)

app/support/price_displayment.rb
jdzak: I can't remember why I decided to include that. I'll remove it.

lib/transaction_search.rb
jdzak: It looks like these headers are to get IE8 to download csv files. I'm kind of curious if they are still needed or if rails includes this now. I'll do a test with IE8.

jdzak:
Ready for re-review.

app/models/reports/account_transactions_report.rb
```diff
@@ -0,0 +1,95 @@
...
+
+  private
+
+  def order_detail_cost(order_detail)
+    order_detail.extend(PriceDisplayment)
+    order_detail.display_cost
+  end
+
+  def order_detail_subsidy(order_detail)
+    order_detail.extend(PriceDisplayment)
```
jhanggi: Seems like we're doing this multiple times to a single order_detail. Would it be clearer to just do this once in `build_row` and then directly call the methods on `order_detail` as opposed to going through a private method here?

jdzak: Ya, I'll DRY it up.

app/models/reports/account_transactions_report.rb
```diff
@@ -0,0 +1,95 @@
...
+  end
+
+  def order_detail_total(order_detail)
+    order_detail.extend(PriceDisplayment)
+    order_detail.display_total
+  end
+
+  def order_detail_duration(order_detail)
+    out = ''
+    unless order_detail.problem?
```
jhanggi: I think this might read a little clearer is we did a simple if/else:

```
if order_detail.problem? 
  ''
else
  order_detail.display_quantity.value
end
```

Also, the `display_quantity.value` feels a little strange. My initial thought was to implement `to_s` on `QuantityDisplay`, but I'm wondering if that class is really necessary.

jhanggi:
There are also a bunch of test failures on CI.

app/models/reports/account_transactions_report.rb
jdzak: Hmm, I think I can get rid of the `QuantityDisplay` class and so something like below. What do you think?

```
def display_quantity(html = false)
  if reservation.try(:actual_duration_mins) && reservation.actual_duration_mins > 0
    val = reservation.actual_duration_mins
    html ? time_wrap(val) : val 
  elsif reservation.try(:duration_mins)
    val = reservation.duration_mins
    html ? time_wrap(val) : val
  elsif quantity_as_time?
    val = quantity
    html ? time_wrap(val) : val
  else
    quantity
  end
end
```

jhanggi: I guess I like that less. I don't like the extra ternaries inside the conditionals. I'm also not a big fan of the parameter. I'd rather have separate methods.

Maybe the QuantityDisplay object is the way to go, and we just need a `to_s` or some other type of method.

jdzak: How about a `html` method instead of `to_s` (see below)? If possible, I'd like to avoid overriding `to_s`.

```
  class QuantityDisplay < Struct.new(:value)
    def html
      value
    end
  end

  class TimeQuantityDisplay
    include ActionView::Helpers::TagHelper

    def html
      content_tag :span, value, :class => 'timeinput'
    end
  end

  def display_quantity
    if reservation.try(:actual_duration_mins) && reservation.actual_duration_mins > 0
      TimeQuantityDisplay.new(reservation.actual_duration_mins)
    elsif reservation.try(:duration_mins)
      TimeQuantityDisplay.new(reservation.duration_mins)
    elsif quantity_as_time?
      TimeQuantityDisplay.new(quantity)
    else
      QuantityDisplay.new(quantity)
    end
  end

  def wrapped_quantity
    display_quantity.html
  end
```

jhanggi: I think I like that better, but what I don't like, and why I was suggesting overriding `to_s` is `.display_quanity.value`. I would expect to just be able to call `display_quantity`.

jdzak: Ah I see. So how about something like this?

```
  class QuantityDisplay < Struct.new(:value)
    def html
      value
    end
  end

  class TimeQuantityDisplay
    include ActionView::Helpers::TagHelper

    def html
      content_tag :span, value, :class => 'timeinput'
    end
  end

  def build_quantity_presenter
    if reservation.try(:actual_duration_mins) && reservation.actual_duration_mins > 0
      TimeQuantityDisplay.new(reservation.actual_duration_mins)
    elsif reservation.try(:duration_mins)
      TimeQuantityDisplay.new(reservation.duration_mins)
    elsif quantity_as_time?
      TimeQuantityDisplay.new(quantity)
    else
      QuantityDisplay.new(quantity)
    end
  end

  def display_quantity
    build_quantity_presenter.value
  end

  def wrapped_quantity
    build_quantity_presenter.html
  end
```

jhanggi: I think that works.

jdzak:
Ready for re-review.

Also, I tested with IE8 and it works fine without the CSVHelper code.

spec/lib/transaction_history_search_spec.rb
```diff
@@ -44,6 +44,8 @@ def all_order_details_with_search
 
     # fake signing in as staff
     @controller.session_user = @staff
+    @controller.response = ActionDispatch::Response.new
+    @controller.request = ActionDispatch::Request.new('rack.input' => StringIO.new, content_type: 'html')
```
jhanggi: What does this do? Why is it necessary?

jhanggi:
Overall looks pretty good. I'd like to hold off on merging this down until we get #288 in (since it actually makes some fixes to issues currently on staging) and released.

spec/lib/transaction_history_search_spec.rb
jdzak: When I added the `respond_to` in `TransactionSearch:: ClassMethods#define_search_method`   it causing the specs the fail with the error below. I traced the error into activesupport where it was calling request.negotiate_mime.

```
NoMethodError: undefined method `negotiate_mime' for nil:NilClass

Stack trace:
./lib/transaction_search.rb:44:in `block in define_search_method'
./spec/lib/transaction_history_search_spec.rb:146:in `block (4 levels) in <top (required)>'
```

PR #287, opened 2015-05-21
[#96992] Exclude reconciled orders when completing

jdzak:
https://pm.tablexi.com/issues/96992

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -207,6 +207,10 @@ class OrderDetail < ActiveRecord::Base
     joins("INNER JOIN relays ON relays.instrument_id = products.id")
   end
 
+  def self.unreconciled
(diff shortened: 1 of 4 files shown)


app/models/order_detail.rb
```diff
@@ -207,6 +207,10 @@ def self.joins_relay
     joins("INNER JOIN relays ON relays.instrument_id = products.id")
   end
 
+  def self.unreconciled
+    where("order_details.state <> ?", 'reconciled')
```
drench: Minor style thing (not a showstopper): go double quotes all around.

drench:
:+1: minus the one OCD bit.

app/models/order_detail.rb
jdzak: cool, i'll switch it

PR #288, opened 2015-05-23
[#94617] Add ability to enter end time during reservation creation

drench:
https://pm.tablexi.com/issues/94617
This adds reservation end time fields that update the duration and vice-versa. @jhanggi I went a bit deeper into the clockpunch guts than I'd have liked. Setting the visible duration field and firing `change()` got me into an infinite loop, so this is setting the hidden field on its own.

### DIFF
--- a/app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
+++ b/app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
@@ -40,36 +40,40 @@ class window.ReservationTimeFieldAdjustor
       @$form.find('[name="reservation[duration_mins]_display"]')
 
   _durationChangeCallback: =>
-    durationMinutes = @timeParser.to_minutes(@$durationDisplayField.val())
+    durationMinutes = @_durationMinutes()
     if durationMinutes % @reserveInterval == 0
       @reserveEnd
         .setDateTime(@reserveStart.getDateTime().addMinutes(durationMinutes))
 
   _getDuration: -> @reserveEnd.getDateTime() - @reserveStart.getDateTime()
 
-  _minimumDuration: -> @reserveInterval * 60 * 1000
+  _durationMinutes: -> @timeParser.to_minutes(@$durationDisplayField.val())
 
   _reserveEndChangeCallback: =>
     duration = @_getDuration()
 
     if duration < 0
-      duration = @_minimumDuration()
-      @reserveStart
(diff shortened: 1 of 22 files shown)


app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
```diff
@@ -0,0 +1,105 @@
...
+  hour: ->
+    hour = parseInt(@$hourField.val(), 10)
+    hour += 12 if @$meridianField.val() == "PM"
+    hour
+
+  minute: -> parseInt(@$minuteField.val(), 10)
+
+  setDateTime: (dateTime) ->
+    @$dateField.val(dateTime.toString("M/d/yyyy"))
+    if dateTime.getHours() > 12
```
jhanggi: This handles noon, but does not handle midnight. I'll make some updates.

jhanggi:
@jdzak could you review this tomorrow since I did some refactoring.

jdzak:
So far I've noticed two things:

1. when reserving an instrument the default duration doesn't seem to be getting filled in properly. For example, I tried reserving an instrument and the duration is set to zero instead of the 1 minute.
2. when beginning a reservation I'm getting the error, `undefined local variable or method 'yea' for #<Instrument:0x007fcaf665b210>`

jdzak:
"Uncaught ReferenceError: reserveInterval is not defined" when editing a started reservation

jhanggi:
Ready for another look.

jdzak:
Looks good. :+1:

jdzak:
I am having an issue with new reservations where the default duration isn't getting set properly. I haven't figured out why yet, but something I noticed.

jdzak:
Looks like the `reserve_end_at_min` is set to the same as the `reserve_start_at_min` and then the javascript's `ReservationTimeFieldAdjustor#setDurationFields` method sets the duration to zero.

jhanggi:
What are the instrument options you have set up? min/max/interval

jdzak:
screen shot 2015-05-28 at 12 48 27 pm

jhanggi:
Updated. With a reservation interval of 1 minute, it was rounding the start and end times to 5 minutes, putting them at the same time. Now I'm rounding to the precision of the interval.

jdzak:
That fixed it. Thanks.

PR #290, opened 2015-05-28
[#94617] Fix specs testing rounding of reservations to nearest interval

jhanggi:


spec/controllers/reservations_controller_spec.rb
```diff
@@ -601,19 +601,40 @@
...
 
         it 'should default the duration mins to minimum duration' do
           assigns(:reservation).duration_mins.should == 15
         end
+
+        describe 'and the instrument has a one minute interval' do
+          let(:reserve_interval) { 1 }
+
+          it 'should not do any additional rounding' do
+            expect(assigns(:reservation).reserve_start_min).to eq(2)
```
jdzak: Why is `reserve_start_min` set to 2? I'd expect it to be 1.

jhanggi: It's rounding '2013-08-15 12:02' to the nearest 1 minute.

jhanggi: If it were 12:07, this would round to 7, the next one would round to 10, and the 15-minute-interval would round to 15.

jdzak: Oh whoops, for some reason I thought I was looking at duration. Looks good.

PR #292, opened 2015-05-29
Hide users tab for BaseRate and ExternalRate

jdzak:
Responding to Aaron's feedback.

https://pm.tablexi.com/issues/50175

### DIFF
--- a/lib/ability.rb
+++ b/lib/ability.rb
@@ -28,7 +28,7 @@ class Ability
         can :manage, [AccountPriceGroupMember, UserPriceGroupMember]
       end
 
-      if resource.global? && (user.user_roles.map(&:role) & UserRole.facility_roles).any? && resource.admin_editable?
+      if user_has_facility_role?(user) && editable_global_group?(resource)
         can :read, UserPriceGroupMember
       end
     end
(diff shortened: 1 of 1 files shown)


lib/ability.rb
```diff
@@ -28,7 +28,7 @@ def initialize(user, resource, controller)
         can :manage, [AccountPriceGroupMember, UserPriceGroupMember]
       end
 
-      if resource.global? && (user.user_roles.map(&:role) & UserRole.facility_roles).any?
+      if resource.global? && (user.user_roles.map(&:role) & UserRole.facility_roles).any? && resource.admin_editable?
```
jhanggi: Can we extract this conditional into a method with a name specifying what it's looking at?

jdzak: How about something like `user_has_facility_role?` and `cancer_center_price_group?`?

jhanggi: what about `editable_global_group?` or something like that? I don't like using cancer center as part of the name since it's specific to northwestern (although, we do end up using that string elsewhere, but I'd like to move away from it).

jdzak:
Alright. Ready for re-review.

PR #293, opened 2015-06-01
Configure Circle CI

jhanggi:
If we squash these commits down, make sure we keep some of the commit messages regarding why we removed the validations.


drench:
:+1:

jdzak:
Looks good here too.

PR #294, opened 2015-06-08
[#94617] Fix ability to create admin reservations

jhanggi:
As part of "#94617 Add ability to enter end time during reservation creation"
a change was made adding a couple extraneous fields to the facility admin
reservation form. As we had decided that the change should only affect
the user-side, and not the facility admin side, I am reverting the changes to
this view.

https://github.com/tablexi/nucore-open/pull/288


drench:
:+1:

PR #296, opened 2015-06-10
[#96852] Add export buttons and background export

jdzak:
https://pm.tablexi.com/issues/96852

### DIFF
--- a/app/assets/javascripts/app/account_transaction_report.coffee
+++ b/app/assets/javascripts/app/account_transaction_report.coffee
@@ -32,4 +32,4 @@ class AccountTransactionReport
       $emailField.prop('disabled', true)
 
 $ ->
-  window.report = new AccountTransactionReport($('.js--exportSearchResults')).init()
+  window.transactionReport = new AccountTransactionReport($('.js--exportSearchResults')).init()
--- a/app/assets/javascripts/app/account_transaction_report.coffee
+++ b/app/assets/javascripts/app/account_transaction_report.coffee
@@ -1,0 +1,26 @@
+class AccountTransactionReport
+  constructor: (@$element) ->
+    return unless @$element.length > 0
+    @registerExportAllHandler()
+
+  registerExportAllHandler: ->
+    $('#js--exportAll').click (event) => @exportAllClicked(event)
+
+  exportAllClicked: (event) ->
+    newTo = prompt 'Have the report emailed to this address:', $('#js--toEmail').val()
+
+    if newTo
+      url = event.target.href
+      $.post(url, { to_email: newTo }, ->
+        $status = $('#js--exportStatus')
+        $status.text("Successfully sent report")
+        $status.show()
+        setTimeout( ->
+          $status.fadeOut()
+        , 5000);
+      )
+
+    event.preventDefault()
+
+$ ->
+  window.report = new AccountTransactionReport($('#table_billing'))
--- a/app/assets/javascripts/app/account_transaction_report.coffee
+++ b/app/assets/javascripts/app/account_transaction_report.coffee
@@ -1,26 +1,29 @@
 class AccountTransactionReport
   constructor: (@$element) ->
-    return unless @$element.length > 0
-    @registerExportAllHandler()
 
-  registerExportAllHandler: ->
-    $('#js--exportAll').click (event) => @exportAllClicked(event)
+  init: ->
+    @$element.click @exportAllClicked
 
   exportAllClicked: (event) ->
-    newTo = prompt 'Have the report emailed to this address:', $('#js--toEmail').val()
-
-    if newTo
-      url = event.target.href
-      $.post(url, { to_email: newTo }, ->
-        $status = $('#js--exportStatus')
-        $status.text("Successfully sent report")
-        $status.show()
-        setTimeout( ->
-          $status.fadeOut()
-        , 5000);
-      )
-
     event.preventDefault()
 
+    $form = $($(event.target).data('form'))
+    $emailField = $form.find('[name=email]')
+    defaultEmail = $emailField.val()
+
+    toAddress = prompt 'Have the report emailed to this address:', defaultEmail
+
+    if toAddress
+      $emailField.val(toAddress)
+      $form.find('[name=format]').val('csv').prop('disabled', false)
+      $emailField.prop('disabled', false)
+
+      $form.submit()
+
+      # since the submit doesn't reload the page when you download the CSV, we need
+      # to reset format so the normal search submit works properly
+      $form.find('[name=format]').val(null).prop('disabled', true)
+      $emailField.prop('disabled', true)
+
 $ ->
-  window.report = new AccountTransactionReport($('#table_billing'))
+  window.report = new AccountTransactionReport($('.js--exportSearchResults')).init()
--- a/app/assets/javascripts/app/account_transaction_report.coffee
+++ b/app/assets/javascripts/app/account_transaction_report.coffee
@@ -32,4 +32,4 @@ class AccountTransactionReport
       $emailField.prop('disabled', true)
 
 $ ->
-  window.transactionReport = new AccountTransactionReport($('.js--exportSearchResults')).init()
+  new AccountTransactionReport($('.js--exportSearchResults')).init()
--- a/app/assets/javascripts/app/account_transaction_report.coffee
+++ b/app/assets/javascripts/app/account_transaction_report.coffee
@@ -18,7 +18,13 @@ class AccountTransactionReport
       $form.find('[name=format]').val('csv').prop('disabled', false)
       $emailField.prop('disabled', false)
 
-      $form.submit()
+      # Do an ajax request so we don't need to re-render this search page
+      $.ajax(
+        type: $form.attr('method'),
+        url: $form.attr('action'),
+        data: $form.serialize()
+      ).success (responseText) ->
+        Flash.info(responseText)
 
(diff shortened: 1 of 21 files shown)


app/assets/javascripts/app/account_transaction_report.coffee
```diff
@@ -0,0 +1,26 @@
+class AccountTransactionReport
+  constructor: (@$element) ->
+    return unless @$element.length > 0
+    @registerExportAllHandler()
+
+  registerExportAllHandler: ->
+    $('#js--exportAll').click (event) => @exportAllClicked(event)
```
jhanggi: It seems a little weird to test on seeing if there is a `#table_billing`, then using `#js--exportAll` for the event. It almost seems like it would be cleaner to just do everything on `#js--exportAll`.

jdzak: Whoops. Yea, initially I meant to do something like `$('#table_billing').find('#js--exportAll')`, but that probably isn't even worth it since the `js--exportAll` is unique.

jhanggi: Here, I believe you can just do `$('#js--exportAll').click @exportAllClicked`. The function can be passed directly in to the click handler.

app/mailers/account_transaction_report_mailer.rb
```diff
@@ -0,0 +1,12 @@
+class AccountTransactionReportMailer < BaseMailer
```
jhanggi: Instead of creating a separate mailer, can we just use the existing `CsvReportMailer` and pass the report in?

app/mailers/account_transaction_report_mailer.rb
```diff
@@ -0,0 +1,12 @@
...
+  def csv_report_email(to_address, order_detail_ids, date_range_field)
+    report = Reports::AccountTransactionsReport.new(
+      OrderDetail.where(id: order_detail_ids),
+      date_range_field: date_range_field)
+
+    attachments['transaction_report.csv'] = report.to_csv
+    mail(to: to_address, subject: "Transaction Export") do |format|
```
jhanggi: The strings here should be I18ned.

app/mailers/account_transaction_report_mailer.rb
```diff
@@ -0,0 +1,12 @@
...
+      format.text { render(text: '') }
```
jhanggi: This text should probably not be blank.

app/mailers/base_mailer.rb
```diff
@@ -0,0 +1,8 @@
+class BaseMailer < ActionMailer::Base
```
jhanggi: Good extraction.

app/views/shared/transactions/_table_inside.html.haml
```diff
@@ -1,3 +1,13 @@
+#js--exportStatus.text-center.notice{ style: 'display: none' }
+%div
+  .pull-right
+    %ul#control.options.inline
+      %li
+        = link_to t('reports.export.export'), format: :csv
```
jhanggi: Do we want both of these links? I think we just want one export.

lib/transaction_search.rb
```diff
@@ -41,9 +41,10 @@ def define_search_method(new_method_name, old_method_name)
         do_search(@search_fields)
         add_optimizations
         sort_and_paginate
+        email_raw_export(params[:to_email], account_transaction_report) if params[:to_email]
         respond_to do |format|
           format.html { render layout: @layout if @layout }
-          format.csv { render text: Reports::AccountTransactionsReport.new(@order_details, date_range_field: @date_range_field).to_csv }
+          format.csv { render text: account_transaction_report.to_csv }
```
jhanggi: I think there should not be the ability to directly access the CSV version. The backgrounding should probably happen here within the `format.csv`. You can use `current_user.email` as a fallback if `params[:to_email]` doesn't exist.

jhanggi:
A couple big problems:

* The links are visible on the Send Notifications, Move Transactions, Create Journal, etc. I believe we only want the export available on the All Transaction view. If you try to use the links on those other views, they don't work because the POST doesn't exist.
* When using the Export Raw, the filters are not submitted. The export should respect the filters you have selected. I'm fine with it only using the filters that had been selected as of the last submit.

app/assets/javascripts/app/account_transaction_report.coffee
```diff
@@ -0,0 +1,26 @@
...
+
+  exportAllClicked: (event) ->
+    newTo = prompt 'Have the report emailed to this address:', $('#js--toEmail').val()
+
+    if newTo
+      url = event.target.href
+      $.post(url, { to_email: newTo }, ->
```
jhanggi: I'm inclined to use GET for this, but I could see arguments for POST.

jhanggi: For consistency's sake, I think we should do an actual submit instead of AJAX similar to what we do for the export in the reports tab.

jhanggi:
Done with my review.

app/mailers/account_transaction_report_mailer.rb
jdzak: Let me that again. Initially I tried doing that, but there was an issue when delayed_job tried to de-serializing the report.

app/views/shared/transactions/_table_inside.html.haml
jdzak: I'll remove the non-backgrounded one. Not sure why I left that one it.

app/assets/javascripts/app/account_transaction_report.coffee
jdzak: K, I'll see how that works.

jhanggi:
@jdzak I haven't touched the mailer/backgrounding yet, but I did do some clean up of the code. Can you do a little bit of testing and make sure the synchronous download is working as expected?

app/assets/javascripts/app/account_transaction_report.coffee
```diff
@@ -0,0 +1,35 @@
...
+      ).success (responseText) ->
+        Flash.info(responseText)
+
+      # since the submit doesn't reload the page when you download the CSV, we need
+      # to reset format so the normal search submit works properly
+      $form.find('[name=format]').val(null).prop('disabled', true)
+      $emailField.prop('disabled', true)
+
+$ ->
+  window.report = new AccountTransactionReport($('.js--exportSearchResults')).init()
```
drench: Do we use `window.report` anywhere?

app/support/minutes_to_time_formatter.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  private
+
+  def hours
+    (minutes / 60).floor
+  end
+
+  def padded_minutes
+    (minutes % 60).to_s.rjust(2, '0')
```
drench: Neat. I haven't seen `rjust` used much (I tend to run immediately toward `sprintf`/`format` and not look back).

drench:
LGTM. Just had one question.

app/assets/javascripts/app/flash.coffee
```diff
@@ -0,0 +1,28 @@
...
+  flash: (level, message) ->
+    # existing flashes
+    @location_selector.find(".alert").remove()
+
+    flash = $("<p>#{message}</p>").addClass('alert').addClass("alert-#{level}")
+    @location_selector.append(flash)
+
+    setTimeout ->
+      flash.fadeOut('slow').remove()
+    , 10000
```
jdzak: Good call on formalizing js flash messages.

jdzak:
Looks good to me. Thanks for finishing this.

jdzak:
Also, I tested the synchronous report requests and that seemed to work the same as the xhr report request.

jhanggi:
The ajax vs not should end up looking the same. The xhr is useful so we don't need to do the additional render of the view.

app/assets/javascripts/app/account_transaction_report.coffee
jhanggi: That's a good question for @jdzak since he's the one that originally wrote that.

jdzak: Yea we do. Let me fix this.

jhanggi:
Per discussion we had last week, we decided that we are going to default the start date to the beginning of the previous month on the exportable searches. This will prevent the default being to export for _all_ time. People could still set it way in the past and get everything, but that won't be the default.

I'm still a little concerned with how this will perform on large datasets. I know we're just grabbing the IDs in the initial search, but it could be a slow operation, but at least we're not rendering as well. I played a bit with ways to try to serialize the ActiveRecord::Relation, but couldn't come up with a decent solution in a reasonable amount of time.

drench:
Minor question about the JS global. Otherwise :+1:

spec/models/reports/account_transaction_report_spec.rb
```diff
@@ -5,7 +5,8 @@
 
   describe '#to_csv' do
     context 'with no order details' do
-      let(:order_details) { [] }
+      # TODO use `.none` in Rails 4
+      let(:order_details) { OrderDetail.where("1 = 0") }
```
jdzak: Hah, clever.

jdzak:
Looks fine to me. Dan has a good point about the `window.transactionReport` assignment. I copied it from another coffeescript report file, but I don't think that variable is actually used anywhere.

PR #297, opened 2015-06-12
[#94617] Do not run some reservation JS on admin reservations

jhanggi:
The admin reservation new/edit pages were not showing the calendar
because of JS errors caused by referencing variables that don’t exist
on those pages.

The ReservationTimeChecker error has been happening for a while, but
has no visible effect, just an error to the console.

### DIFF
--- a/app/assets/javascripts/app/reservation_time_checker.coffee
+++ b/app/assets/javascripts/app/reservation_time_checker.coffee
@@ -1,8 +1,12 @@
 class window.ReservationTimeChecker
   constructor: (@selector, @alertId = 'duration-alert')->
-    @initAlert()
-    @respondToChange()
+    if @validPage()
+      @initAlert()
+      @respondToChange()
 
+  validPage: ->
+    # These variable are not set on the admin reservation pages
+    reserveInterval? && reserveMinimum? && reserveMaximum?
 
   duration: ->
     parser = new TimeParser()
--- a/app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
+++ b/app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
@@ -86,5 +86,7 @@ class window.ReservationTimeFieldAdjustor
     )
 
 $ ->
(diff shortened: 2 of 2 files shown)


drench:
:+1:

app/assets/javascripts/app/reservation_time_checker.coffee
```diff
@@ -1,8 +1,12 @@
...
   constructor: (@selector, @alertId = 'duration-alert')->
-    @initAlert()
-    @respondToChange()
+    if @validPage()
+      @initAlert()
+      @respondToChange()
 
+  validPage: ->
+    # These variable are not set on the admin reservation pages
+    reserveInterval? && reserveMinimum? && reserveMaximum?
```
jdzak: Would it make sense to change the method `validPage` to `adminReservationPage?`? 

So you'd have something like...

```coffeescript
class window.ReservationTimeChecker
  constructor: (@selector, @alertId = 'duration-alert')->
    unless @adminReservationPage()
      @initAlert()
      @respondToChange()

  adminReservationPage: ->
    reserveInterval? && reserveMinimum? && reserveMaximum?
```

jhanggi: If I wanted to call it `adminReservationPage`, I would need to invert the logic to be `!reserveInterval? || !reserveMinimum? || !reserveMaximum?` which just seems gross. We want to run the code on the regular user-facing reservation page, and something like `regularReservationPage` seemed weird as well.

jdzak: Hmm, would `customerReservationPage` work?

jhanggi: I actually don't like being that specific because I'm not really testing whether I'm on the customer reservation page, I'm actually testing if the page I'm on has set up the proper variables and environment in order for this code to function.

jdzak: Ah ok, i see now.

PR #298, opened 2015-06-16
[#75325] Add training requests

drench:
This should allow users to request training on restricted instruments, and give senior staff and above the ability to view those requests, and grant or ignore (delete) them.

There's a lot here that's just cleanup, and some (like a few of the traits in the `User` factory) might not be used yet but I expect to use soon.


app/controllers/instrument_request_approval_controller.rb
```diff
@@ -0,0 +1,20 @@
+class InstrumentRequestApprovalController < ApplicationController
+  before_filter :load_instrument
+
+  def new; end
+
+  def create
+    TrainingRequest.create!(user: current_user, product: @instrument)
```
jhanggi: I'm assuming this should never be invalid, hence why we would want to see exceptions if it's not?

app/controllers/instrument_request_approval_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+      "controllers.instrument_request_approval.create.notice",
+      instrument: @instrument,
+    )
+    redirect_to facility_path(current_facility)
+  end
+
+  private
+
+  def load_instrument
+    @instrument = Instrument.find_by_url_name(params[:instrument_id])
```
jhanggi: Should this be `find_by_url_name!` so it raises a 404?

app/controllers/instrument_request_approval_controller.rb
```diff
@@ -0,0 +1,20 @@
+class InstrumentRequestApprovalController < ApplicationController
+  before_filter :load_instrument
+
+  def new; end
+
+  def create
```
jhanggi: Do we want to add the paths in a comment like we do on most other controllers?

app/controllers/instruments_controller.rb
```diff
@@ -19,47 +19,57 @@ def index
     @instruments.each { |i| @reservations[i.id] = i.reservations.upcoming[0..2]}
   end
 
-  # GET /instruments/1
+  # GET /facilities/:facility_id/instruments/:instrument_id
   def show
```
jhanggi: This method makes me cry. I'm not sure if it's worth taking the time now, but it would be nice to extract a lot of this logic into its own class.

app/controllers/training_requests_controller.rb
```diff
@@ -0,0 +1,32 @@
+class TrainingRequestsController < ApplicationController
+  admin_tab :index
+
+  before_filter :authenticate_user!
+  before_filter :check_acting_as
+
+  load_and_authorize_resource except: :index
```
jhanggi: Why are we excluding index?

app/models/facility.rb
```diff
@@ -29,6 +29,7 @@ def accounts
   has_many :order_imports, dependent: :destroy
   has_many :orders, :conditions => 'ordered_at IS NOT NULL'
   has_many :facility_accounts
+  has_many :training_requests, finder_sql: proc { "SELECT training_requests.* FROM training_requests INNER JOIN products ON products.id = training_requests.product_id WHERE products.facility_id = #{self.id}" }
```
jhanggi: Does `has_many :training_requests, through: :products` not work?

app/views/admin/shared/_sidenav_product.html.haml
```diff
@@ -1,5 +1,9 @@
...
-  = tab "Items (#{current_facility.items.count})", facility_items_path(current_facility), sidenav_tab == 'items'
-  = tab "Bundles (#{current_facility.bundles.count})", facility_bundles_path(current_facility), sidenav_tab == 'bundles'
+  = tab "#{Instrument.model_name.pluralize} (#{current_facility.instruments.count})", facility_instruments_path(current_facility), sidenav_tab == Instrument.model_name.plural
+  = tab "#{Service.model_name.pluralize} (#{current_facility.services.count})", facility_services_path(current_facility), sidenav_tab == Service.model_name.plural
+  = tab "#{Item.model_name.pluralize} (#{current_facility.items.count})", facility_items_path(current_facility), sidenav_tab == Item.model_name.plural
+  = tab "#{Bundle.model_name.pluralize} (#{current_facility.bundles.count})", facility_bundles_path(current_facility), sidenav_tab == Bundle.model_name.plural
+
+- if current_ability.can?(:index, TrainingRequest)
+  %ul.sidebar-nav
+    = tab "#{TrainingRequest.model_name.titleize.pluralize} (#{current_facility.training_requests.count})", facility_training_requests_path(current_facility) , sidenav_tab == TrainingRequest.model_name.plural
```
jhanggi: do we need the `titleize`, or can we just set the model name to "Training Request" in the locales?

app/models/product_user_observer.rb
```diff
@@ -0,0 +1,15 @@
+class ProductUserObserver < ActiveRecord::Observer
```
jhanggi: I've become less of a fan of observers like this over time. From what I've seen they tend to be almost more trouble than callbacks, and even harder to realize that they're there.

What would you think of a `ProductUserCreator` class that handles this, and use that inside the ProductUserController#create action (which confusingly seems to be the `new` method)?

app/models/product_user_observer.rb
```diff
@@ -0,0 +1,15 @@
...
+    training_request = get_training_request(product_user) || return
+    product_user.update_attribute(:requested_at, training_request.created_at)
+    training_request.destroy
+  end
+
+  private
+
+  def get_training_request(product_user)
+    TrainingRequest
+      .where(user_id: product_user.user_id, product_id: product_user.product_id)
```
jhanggi: I wish we were on rails 4+ so we could just use `find_by(xxx)`.

jhanggi: Actually, should we just do `destroy_all` just in case there ever get to be more than one? I see we have a uniqueness constraint.

config/locales/en.models.yml
```diff
@@ -267,3 +270,7 @@ en:
         remittance_information: Bill To
       nufs_account:
         account_number: Payment Source Number
+      training_request:
+        user: Name
+        product: Instrument
```
jhanggi: Is it only instruments that can have a training request list?

jhanggi:
Done with my code review. Now onto actually looking at how it works.

app/views/instrument_request_approval/new.html.haml
```diff
@@ -0,0 +1,4 @@
+= t_model_error(@instrument.class, :requires_approval, instrument: @instrument)
```
jhanggi: We should probably include the current facility as a header on this page. Not sure if that's `content_for` or not.

app/controllers/instrument_request_approval_controller.rb
```diff
@@ -0,0 +1,20 @@
+class InstrumentRequestApprovalController < ApplicationController
```
jhanggi: Now that I see what this controller is doing, the name seems odd. To me, this  seems like it should be doing the approval, not the request.

app/views/admin/shared/_sidenav_product.html.haml
```diff
@@ -1,5 +1,9 @@
...
-  = tab "Services (#{current_facility.services.count})", facility_services_path(current_facility), sidenav_tab == 'services'
-  = tab "Items (#{current_facility.items.count})", facility_items_path(current_facility), sidenav_tab == 'items'
-  = tab "Bundles (#{current_facility.bundles.count})", facility_bundles_path(current_facility), sidenav_tab == 'bundles'
+  = tab "#{Instrument.model_name.pluralize} (#{current_facility.instruments.count})", facility_instruments_path(current_facility), sidenav_tab == Instrument.model_name.plural
+  = tab "#{Service.model_name.pluralize} (#{current_facility.services.count})", facility_services_path(current_facility), sidenav_tab == Service.model_name.plural
+  = tab "#{Item.model_name.pluralize} (#{current_facility.items.count})", facility_items_path(current_facility), sidenav_tab == Item.model_name.plural
+  = tab "#{Bundle.model_name.pluralize} (#{current_facility.bundles.count})", facility_bundles_path(current_facility), sidenav_tab == Bundle.model_name.plural
+
+- if current_ability.can?(:index, TrainingRequest)
+  %ul.sidebar-nav
```
jhanggi: What about using an `%li.divider` instead of two rounded boxes touching?

jhanggi:
Everything seems to be working pretty well. Just had a few more comments as I used the feature.

config/locales/en.models.yml
drench: Right: items can too. Just go with "Product" then?

app/controllers/instrument_request_approval_controller.rb
drench: More like `RequestProductApprovalController`?

app/controllers/instruments_controller.rb
drench: Yes. I added a `TODO` to `ProductsCommonController#show` (which is similar).

config/locales/en.models.yml
drench: All of the product types, really, though I don't know if it's possible through the UI to require approval for Bundles. There are specs around restricted Bundles (which I updated) but is that something that happens?

drench:
This is ready for another look. There's still a lot of refactoring potential, particularly in the various product controller specs, and of course the product controller itself. But the large pieces should be there, such as applying this to product types other than `Instruments`, removing the observer, and the UI changes.

config/locales/en.models.yml
jhanggi: It doesn't look like bundles are ever restricted, however it might be possible to have a restricted item within a bundle; I'm not sure how that's handled.

```
Product.where(requires_approval: true).active_plus_hidden.count(group: 'type')
   (3.1ms)  SELECT COUNT(*) AS count_all, type AS type FROM "PRODUCTS" WHERE "PRODUCTS"."REQUIRES_APPROVAL" = 1 AND "PRODUCTS"."IS_ARCHIVED" = 0 GROUP BY type
 => {"Instrument"=>252, "Item"=>7, "Service"=>52}
```

I don't think we need to worry about it.

app/models/product.rb
```diff
@@ -11,6 +11,7 @@ class Product < ActiveRecord::Base
   has_many   :product_accessories, conditions: { deleted_at: nil }, dependent: :destroy
   has_many   :accessories, :through => :product_accessories, :class_name => 'Product'
   has_many   :price_policies
+  has_many   :training_requests
```
jhanggi: Should we have a `dependent: :destroy` here?

app/models/product_user.rb
```diff
@@ -2,9 +2,14 @@ class ProductUser < ActiveRecord::Base
...
 
-  before_create lambda { |pu| pu.approved_at = Time.zone.now }
-end
\ No newline at end of file
+  validates_numericality_of :product_id, :user_id, :approved_by, only_integer: true
+  validates_uniqueness_of :user_id, scope: :product_id, message: "is already approved"
+
+  before_create -> (product_user) { product_user.approved_at = Time.zone.now }
+
+  def training_request
```
jhanggi: Maybe I just haven't seen it yet, but where does this get used? And since we're destroying the training request once the user is approved, wouldn't this always be empty?

app/models/product_user_creator.rb
```diff
@@ -0,0 +1,19 @@
+class ProductUserCreator
+  def self.create(user:, product:, approver:)
+    product_user = ProductUser.new(
+      product: product,
+      user: user,
+      approved_by: approver.id,
+      approved_at: Time.zone.now,
+    )
+
+    product_user.save && manage_training_request(product_user)
```
jhanggi: I think it might be a good idea to wrap this in a transaction. So if the training request deletion fails, so does everything.

spec/controllers/items_controller_spec.rb
```diff
@@ -81,10 +76,11 @@
...
     end
 
-    context "restricted item" do
+    context "when the item requires approval" do
       before :each do
-        @item.update_attributes(:requires_approval => true)
+        @item.update_attributes(requires_approval: true)
       end
+
       it "should show a notice if you're not approved" do
```
jhanggi: Should we add tests in here to make sure you're shown the option of requesting training?

spec/models/product_user_creator_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "spec_helper"
+
+describe ProductUserCreator do
+  subject do
+    -> { ProductUserCreator.create(user: user, product: product, approver: approver) }
```
jhanggi: This is an interesting idiom. Where'd you get the idea?

I feel like it might be easier to understand to actually define a method here called `create_product_user`, that you can use when you're testing the side effects, and you can keep `let(:product_user) { create_product_user }` as an alias when you're looking for actual references to the object.

app/controllers/training_requests_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def flash_arguments
+    @flash_arguments ||= {
+      user: @training_request.user.to_s,
+      product: @training_request.product.to_s,
+    }
+  end
+
+  def load_product
+    @product = Product.find_by_url_name!(params[:product_id])
```
jhanggi: This should probably be restricted to the current facility. Right now, I can load a restricted instrument from a different facility by changing the URL.

jhanggi: And probably should be restricted to non-archived products.

app/controllers/training_requests_controller.rb
```diff
@@ -0,0 +1,54 @@
+class TrainingRequestsController < ApplicationController
+  admin_tab :index
+
+  before_filter :authenticate_user!
+  before_filter :check_acting_as
+
+  load_and_authorize_resource
+
+  layout "two_column"
```
jhanggi: I'm not sure this should be `two_column`. The alignment looks a little funky at wider windows.

jhanggi: For new/create the alignment looks funky, but it's the right thing for index.

app/models/product_user.rb
drench: It will exist for a very brief time when we create a new `ProductUser`, but for most purposes it's going to be nothing. It's used in `ProductUserCreator.manage_training_request` so it might make more sense to move this code there.

jhanggi: Yeah, I think that makes sense. It feels weird here, where in pretty much every case it's going to be empty.

spec/models/product_user_creator_spec.rb
drench: I don't remember where I picked it up but it's something I used a bit on another project, mainly because it made things like `it { is.expected_to change(Something, :method).from(old).to(new) }` possible. But that's not something we'll use here yet anyway because we need to upgrade `rspec`…

spec/controllers/items_controller_spec.rb
drench: Good catch. Yet another reason to refactor/consolidate the various product controller specs.

app/controllers/training_requests_controller.rb
```diff
@@ -0,0 +1,58 @@
...
+      user: @training_request.user.to_s,
+      product: @training_request.product.to_s,
+    }
+  end
+
+  def load_product
+    @product =
+      Product
+      .active
+      .where(facility_id: current_facility.id)
```
jhanggi: Could this be `current_facility.products.active.find_by_url_name!(params[:product_id])`?

app/models/product_user_creator.rb
```diff
@@ -0,0 +1,21 @@
...
+  def self.create(user:, product:, approver:)
+    product_user = ProductUser.new(
+      product: product,
+      user: user,
+      approved_by: approver.id,
+      approved_at: Time.zone.now,
+    )
+
+    product_user.transaction do
+      product_user.save && manage_training_request(product_user)
```
jhanggi: In order for the transaction to work, wouldn't the `save` and `destroy` need to be the bang versions? Otherwise they just return true/false, which I don't think `transaction` responds to.

jhanggi:
We're getting pretty close here. Just a couple minor comments.

app/models/product_user_creator.rb
drench: Doesn't look like `#destroy!` exists in Rails 3.x. Guess I'm going with `training_request.destroy || raise ActiveRecord::Rollback`.

drench:
I added commits for those last 2 issues.

PR #299, opened 2015-06-19
[#98767] Reservation duration rounding for display

drench:
Current behavior truncates seconds to whole minutes when displaying reservation durations. This rounds them instead.

I added specs initially to document existing behavior, and updated after as I made changes to `date_support.rb`.

* `#actual_duration_mins` has a minimum of 1 minute, unless `@actual_duration_mins` is set. In that case, it's rounded, so it could be zero, as it could be with the previous `#to_i`. Is that what we want? I think `@actual_duration_mins` would always be an integer (or really, a string representation of an integer) so maybe this is fine.
* `#duration_mins` has no 1 minute minimum, whether calculated or set from `@duration_mins`, much like the above. Is that what we want?
* There is a `#duration_value` method that also does a `#to_i` conversion but I left it alone because I don't see it used anywhere. Unless I missed something, I'd rather just remove the method.


app/support/reservations/date_support.rb
```diff
@@ -73,9 +73,9 @@ def duration_unit
 
   def duration_mins
     if @duration_mins
-      @duration_mins.to_i
+      @duration_mins.to_f.round
     elsif reserve_end_at and reserve_start_at
-      ((reserve_end_at - reserve_start_at) / 60).floor
+      ((reserve_end_at - reserve_start_at) / 60).round
```
jhanggi: We actually do want to truncate, not round. It needs to match the logic that we're using for price calculations. See what we're doing in InstrumentPricePolicyCalculations. We want to truncate start and end, and then subtract.

Not sure if it's worth the refactoring of the price calculations to use `duration_mins` or whether that would be too dangerous a change.

drench:
Duration minutes should now match what we're doing in cost calculations. I reworked the specs to check for seconds truncation.  I don't know how appropriate it was to bring `InstrumentPricePolicyCalculations` into it, but I wanted to check against it to see if the results were the same.

spec/app_support/reservations/rendering_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+          expect(reservation.actuals_string).to eq("No actual times recorded")
+        end
+      end
+    end
+  end
+
+  describe "#as_calendar_object" do
+    let(:actual_start_at) { DateTime.new(2015, 8, 1, 14, 15, 16) }
+    let(:actual_end_at) { DateTime.new(2015, 8, 1, 15, 16, 17) }
+    let(:title) { "Admin\nReservation" }
```
jhanggi: Any specific reason for including a new line in here?

spec/app_support/reservations/date_support_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+            let(:actual_end_at) { Time.zone.parse(timestring) }
+            let(:price_policy_calculation) do
+              DummyPricePolicy.new.estimate_cost_and_subsidy(actual_start_at, actual_end_at)
+            end
+
+            it "returns #{expected_minutes}" do
+              expect(reservation.actual_duration_mins).to eq(expected_minutes)
+            end
+
+            it "matches InstrumentPricePolicyCalculations' cost at $1 per minute" do
```
jhanggi: I think it was a smart idea to write these tests to make sure we're matching the price calculations. I'm not sure it's worth keeping those tests around, though. It clutters up this test, and I'm not sure it'll be clear on down the road why these tests exist.

jhanggi:
Looks good. Some nice refactoring in there and good additional test coverage.

spec/app_support/reservations/rendering_spec.rb
drench: Maybe not the best explanation, but I wrote the test to match what the code does. I don't know the reasoning behind it looks like it originated here: 06bb0685f5b254f8ddbd7cfd917dcd7e022fb7ac. Switching this to a space didn't seem to make a difference in how the tooltips render.

drench:
WIP test removed. Ready for another go.

PR #300, opened 2015-06-26
Fix spec that is broken when merging into the NU branch.

jhanggi:
The account needed to be added to one of the products within the bundle,
not the bundle itself.

https://github.com/tablexi/nucore-open/pull/298/files#diff-a4df8e7f76722c53491b648a90dfe540L85


drench:
:tada: PR number :100:x3
:+1:

PR #301, opened 2015-06-29
[#81326] Clarify report start/end date fields

drench:


drench:
screen shot 2015-06-29 at 3 37 00 pm

jhanggi:
This view looks good, and seems to work all right when you shrink the browser, although the breaks aren't perfect, I think they're good enough.

Only problem I noticed is on the Instrument utilization report page, because it's sharing a partial, it comes out funky. I'm thinking we take the "Fulfilled" from the start date label there and move it before the between. So we have "Fulfilled between [Start Date] and [End Date]". What do you think?

instrumentutilization

app/views/reports/_date_range_fields.html.haml
```diff
@@ -1,3 +1,6 @@
+%li
+  .field-clarifier= t("preposition.between")
 %li= form.input :date_start, label: t("reports.fields.#{controller_name}.date_start", default: t('reports.fields.date_start')), input_html: { value: human_datetime(@date_start, date_only: true), class: 'datepicker' }
-
+%li
+  .field-clarifier= t("support.array.two_words_connector")
```
jdzak: Neat. I didn't know this existed.

drench: I felt a little weird using it, since there's no `Array` involved. Still I was hoping something similar existed for "between" but I didn't find it.

drench:
Reworked to reword. I think it looks better, but this page could use more help to support smaller screens.

jhanggi:
I actually just played with it in the iOS simulator. We don't do much real responsiveness so it's already zoomed out pretty far, and it actually looks ok (not amazing, but usable) in both iPad and iPhone in portrait mode and landscape mode.

ios simulator screen shot jun 30 2015 11 22 40 am


ios simulator screen shot jun 30 2015 11 22 14 am

PR #303, opened 2015-07-02
Instrument pricing documentation

jhanggi:
I wrote this up as something to send to UIC for clarification. Double check my calculations and make sure they match with how you think the model works.


mlineen:
:+1:

PR #306, opened 2015-07-07
Make paperclip configurable to support S3

drench:
This moves the hardcoded paperclip settings into `settings/`. I've included instructions in the README on configuring the app to use S3.

It also includes a rake task that should migrate all `StoredFile`s on the filesystem to S3, to be run after configuring the system to use S3: `bundle exec rake paperclip:move_to_s3`. We should save the output of this run to track down any files it was unable to migrate.

Something to watch out for: I tried upgrading the paperclip gem to 4.3, but it seems to break journaling here: https://github.com/tablexi/nucore-open/blob/master/spec/models/journal_spec.rb#L170

### DIFF
--- a/app/models/stored_file.rb
+++ b/app/models/stored_file.rb
@@ -6,10 +6,7 @@ class StoredFile < ActiveRecord::Base
   validates_presence_of   :product_id,      :if => lambda {|o| o.file_type == 'info' || o.file_type == 'template'}
   validates_presence_of   :order_detail_id, :if => lambda {|o| o.file_type == 'template_result' || o.file_type == 'sample_result'}
   validates_inclusion_of  :file_type, :in => %w(info template template_result sample_result import_error import_upload)
-  has_attached_file       :file,
-                          :storage => :filesystem,
-                          :url => "#{ENV['RAILS_RELATIVE_URL_ROOT']}/:attachment/:id_partition/:style/:safe_filename",
-                          :path => ":rails_root/public/:attachment/:id_partition/:style/:safe_filename"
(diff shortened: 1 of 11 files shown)


app/models/stored_file.rb
```diff
@@ -6,10 +6,7 @@ class StoredFile < ActiveRecord::Base
   validates_presence_of   :product_id,      :if => lambda {|o| o.file_type == 'info' || o.file_type == 'template'}
   validates_presence_of   :order_detail_id, :if => lambda {|o| o.file_type == 'template_result' || o.file_type == 'sample_result'}
   validates_inclusion_of  :file_type, :in => %w(info template template_result sample_result import_error import_upload)
-  has_attached_file       :file,
-                          :storage => :filesystem,
-                          :url => "#{ENV['RAILS_RELATIVE_URL_ROOT']}/:attachment/:id_partition/:style/:safe_filename",
-                          :path => ":rails_root/public/:attachment/:id_partition/:style/:safe_filename"
+  has_attached_file       :file, Settings.paperclip.to_hash
```
jhanggi: Do we need to put the `url`/`path` setting somewhere?

jhanggi: n/m I see it below now.

jhanggi:
It looks like Journal also `has_attached_file`. Do we need to migrate those files as well?

drench:
I updated `Journal` to use the same paperclip settings as `StoredFile`, and added journal attachments to the S3 migration task.

I also put in a suggested `:class` component to the S3 path to avoid name collisions but left `settings.yml` alone for backward compatibility.

jhanggi:
:+1:

PR #307, opened 2015-07-13
Bring nucore-open and UICore closer

drench:
* Updates gem versions (except `delayed_job`) to bring the 3 projects closer in line with each other
* Uses the `add_account_for_user` helper in specs wherever possible
* Sets up `AccountPriceGroupMember`s in specs (mostly) instead of `UserPriceGroupMember`s


jhanggi:
:+1: I think we can probably squash this to 2 commits: 1 for gem versions and 1 for the spec updates. Can you then do a PR against both NU and UIC that uses these updates?

PR #309, opened 2015-07-15
Update gems from NU and UIC forks

drench:
Here are a couple more version bumps from NU missed in the previous update, and a `delayed_job` major version update from UIC that we should exercise before merging into NU.


jhanggi:
:+1:

PR #310, opened 2015-07-16
[#25081] Redirect attachment downloads to private, expiring S3 URLs

drench:
To get the S3 URLs to be temporary *and* force download I borrowed @cenavarro's workaround as seen here https://github.com/tablexi/chidry/pull/422.

Downloads should continue to work when using the local filesystem, though the redirected-to-URLs won't be private or temporary (that is, business as usual).

Need to run the following rake tasks:
```
rake paperclip:mime:update_from_file_extensions
rake paperclip:move_to_s3
```


app/models/journal.rb
```diff
@@ -179,15 +176,10 @@ def status_string
     elsif is_successful? == false
```
jhanggi: Would it be worth dropping the `is_` on this method as well?

app/models/concerns/downloadable_file.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  included do
+    has_attached_file :file, Settings.paperclip.to_hash
+
+    # TODO Limit attachment types for safe uploads
+    do_not_validate_attachment_file_type :file
+  end
+
+  def download_url
+    file.send(:directory).files.get_url(
```
jhanggi: Does this still work if you're using file storage?

app/views/facility_journals/show.html.haml
```diff
@@ -16,62 +16,58 @@
...
-  %p.notice= t('.notice')
+  %p.notice= t(".notice")
 - else
-  = form_tag facility_journal_reconcile_path(current_facility, @journal), :method => :post do
+  = form_tag facility_journal_reconcile_path(current_facility, @journal), method: :post do
     %table.table.table-striped.table-hover
       %thead
-        - is_submittable = @journal.is_successful? && !@journal.is_reconciled?
-        - if is_submittable
+        - if @journal.submittable?
```
jhanggi: In the case of the `else` statement of `reconciled`, do we end up doing a lot of extra lookups because it runs the find every time this is called?

jhanggi:
Looks like we don't expose the original order imports. Can you double check that and make sure we don't need an endpoint for that?

app/models/journal.rb
drench: I considered it. It would take a migration since it's a column name.

drench:
I'm not aware of a place where we provide a download of the original order import, though I could see adding a column for it in Import History. It looks like there's room for it.

app/controllers/order_imports_controller.rb
```diff
@@ -28,6 +28,14 @@ def create
     redirect_to new_facility_order_import_path
   end
 
+  def error_report
+    if order_import.error_file_present?
+      redirect_to order_import.error_file_download_url
+    else
+      render_404
```
jhanggi: Do we want to call this directly, or `raise ActiveRecord::RecordNotFound`? What do we do elsewhere?

app/models/concerns/downloadable_file.rb
```diff
@@ -0,0 +1,28 @@
...
+  included do
+    has_attached_file :file, Settings.paperclip.to_hash
+
+    # TODO Limit attachment types for safe uploads
+    do_not_validate_attachment_file_type :file
+  end
+
+  def download_url
+    if fog?
+      file.send(:directory).files.get_url(
```
jhanggi: Can we add some comments about what this is doing, and where this solution came from? Just because it feels a little weird.

config/routes.rb
```diff
@@ -191,7 +191,11 @@
       end
     end
 
-    resources :order_imports, :only => [ :new, :create ]
+    resources :order_imports, only: [:new, :create] do
+      member do
+        get "error_report", to: "order_imports#error_report"
```
jhanggi: This might look a little cleaner as a one-liner `get "error_report", to: "order_imports#error_report", on: :member`.

spec/controllers/facility_journals_controller_spec.rb
```diff
@@ -116,8 +116,9 @@ def create_order_details
           assigns[:journal].created_by.should == @admin.id
         end
 
-        it 'should have an is_successful value of true' do
-          assigns[:journal].is_successful? == true
+        it "has an is_successful value of true" do
+          expect(assigns[:journal].is_successful?).to be true
+          expect(assigns[:journal]).to be_successful
```
jhanggi: Is there a reason to do both here? Feels redundant.

jhanggi:
Looks pretty good. I want to do a little testing on Monday before we pull this in.

spec/controllers/facility_journals_controller_spec.rb
drench: It is. I wanted to have both tests until we completely phase out `is_successful`.

app/views/facility_journals/show.html.haml
```diff
@@ -16,62 +16,59 @@
       = f.input :status_string
   .span4
     .well
-      - if Settings.financial.journal_format.xls
-        %p=link_to 'Download XLS File', @journal.file.url
-      - if Settings.financial.journal_format.xml
-        %p=link_to 'Download XML File', facility_journal_path(current_facility, @journal, :format => 'xml')
-      - if Settings.financial.journal_format.csv
-        %p=link_to 'Download CSV File', facility_journal_path(current_facility, @journal, :format => 'csv')
+      - %i(xls xml csv).each do |format|
```
jhanggi: I think we also need this on `facility_journals/index` if there's a pending journal.

jhanggi:
There are still quite a few links that still point directly to the file. If you do a grep for `file.url`, you should find most of them.

app/controllers/order_imports_controller.rb
```diff
@@ -28,6 +28,14 @@ def create
     redirect_to new_facility_order_import_path
   end
 
+  def error_report
+    if order_import.error_file_present?
+      redirect_to order_import.error_file_download_url
```
jhanggi: This is a little strange. When I use Fog and download this file, it's being saved as `error_report.txt`, but the filename in S3 and everything is `.csv`. Does that happen to you? It might just be a weird thing with chrome?

drench: It looks like Paperclip does its own content type detection if we don't specify it, and CSVs were seen as `text/plain`. Firefox retains the `.csv` extension, but I guess when Chrome sees `text/plain` it changes the extension. I've got a patch that should set the content type header correctly now.

config/initializers/mime_types.rb
```diff
@@ -3,3 +3,5 @@
 # Add new mime types for use in respond_to blocks:
 # Mime::Type.register "text/richtext", :rtf
 # Mime::Type.register_alias "text/html", :iphone
+
+Mime::Type.register "application/vnd.ms-excel", :xls
```
jhanggi: Would registering "text/csv", :csv here fix the paperclip issue?

app/views/facility_journals/index.html.haml
```diff
@@ -29,12 +29,7 @@
             %tr
               %td
                 %ul.unstyled
-                  - if Settings.financial.journal_format.xls
-                    %li=link_to 'XLS File', pending_journal.file.url
-                  - if Settings.financial.journal_format.xml
-                    %li=link_to 'XML File', facility_journal_path(current_facility, pending_journal, format: 'xml')
-                  - if Settings.financial.journal_format.csv
-                    %li=link_to 'CSV File', facility_journal_path(current_facility, pending_journal, format: 'csv')
+                  = render "downloads", facility: current_facility, journal: pending_journal
```
jhanggi: Using the same locales here results in too much text in the column. I think we want to leave this cell as just "XML", "XLS", "CSV". The other view should have the full "Download XXX file".

app/views/products/_show.html.haml
```diff
@@ -22,5 +22,5 @@
     %h3 Documentation
     %ul.unstyled
       - docs.each do |stored_file|
-        %li.document= link_to stored_file.name, stored_file.file.url
+        %li.document= link_to stored_file.name, stored_file.download_url
```
jhanggi: We should probably put a bunch of these stored files behind a redirector. Otherwise we're going to run into the issue of users sitting on the page for a while and having the link expire.

jhanggi: The tricky thing with that is also ensuring that permissions work properly. We don't want someone being able to change the URL by hand and get access to a file they shouldn't have access to.

app/models/order_import.rb
```diff
@@ -196,6 +198,7 @@ def set_error_mode
   def store_error_report
     self.error_file = StoredFile.new(
       file: StringIO.new(self.error_report),
+      file_content_type: "text/csv",
```
jhanggi: Looks like existing files are stored with the wrong mime-type: 

```
StoredFile.where(name: 'error_report.csv').group(:file_content_type).count
   (4.0ms)  SELECT COUNT(*) AS count_all, file_content_type AS file_content_type FROM "STORED_FILES" WHERE "STORED_FILES"."NAME" = 'error_report.csv' GROUP BY file_content_type
 => {"text/plain"=>268, "text/x-c++"=>14, "text/x-pascal"=>4, "text/x-java"=>6}
```

We should probably have a rake task or migration that cleans these up so they get downloaded as the correct thing.

app/views/facility_accounts/show_statement.html.haml
```diff
@@ -38,7 +38,7 @@
                 %li.indented= link_to 'View Order Form', od.external_service_receiver.show_url, :target => '_blank'
               - if !od.stored_files.template_result.empty?
                 - od.stored_files.template_result.each do |stored_file|
-                  %li.indented= link_to 'View Order File', stored_file.file.url
+                  %li.indented= link_to 'View Order File', stored_file.download_url
```
drench: Is this view obsolete? I see where we display statements as PDFs, but not as HTML.

app/views/facility_orders/_order_file_icon.html.haml
```diff
@@ -4,5 +4,5 @@
       %i.icon-file-text{:data => { :toggle => 'tooltip' }, :title => t('order_details.show.link.view.order_form') }
 - od.stored_files.template_result.each do |stored_file|
   .inline-icons
-    = link_to stored_file.file.url do
+    = link_to template_results_facility_order_order_detail_path(current_facility, od.order, od, stored_file) do
```
jhanggi: Did you create a helper for this?

app/views/order_details/order_file.html.haml
```diff
@@ -22,7 +22,8 @@
         %td
           %p
             %b Download and complete the template
-          %p.document= link_to "#{@order_detail.product.name} &mdash; Order Form Template".html_safe, @order_detail.product.stored_files.template.first.file.url
+          %p.document
+            = link_to "#{@order_detail.product.name} &mdash; Order Form Template".html_safe, product_file_path(@order_detail.product.stored_files.template.first)
```
jhanggi: This gives me a 403 when I try to access it as a regular user.

app/views/orders/_service_desc.html.haml
```diff
@@ -19,5 +19,5 @@
     - if order_detail.stored_files.template_result.empty?
       = link_to 'Upload Order Form', order_order_detail_order_file_path(order_detail.order, order_detail), :class => 'btn'
     - else
-      = link_to('Uploaded Order Form', order_detail.stored_files.template_result.first.file.url)
+      = link_to "Uploaded Order Form", order_detail_first_template_result_path(order_detail)
```
jhanggi: I think this might need to go through a different controller action. Where this goes (`OrderManagement::OrderDetailsController#template_results`) only loads purchased orders, so I'm getting a 404 on `Couldn't find Order with id=40 [WHERE `orders`.`facility_id` = 1 AND (ordered_at IS NOT NULL)]`

app/views/order_details/_result_file.html.haml
```diff
@@ -1,4 +1,4 @@
 %li
-  = link_to "Results File #{result_file_counter + 1}", result_file.file.url
+  = link_to "Results File #{result_file_counter + 1}", sample_result_path(result_file)
```
jhanggi: I'm getting a 403 as a normal user here as well.

app/views/facility_accounts/show_statement.html.haml
jhanggi: I think it is obsolete, but we should probably leave it in anyways. Although, it looks like you triggered a couple exceptions when trying to view this on stage. We might want to fix that up at some point.

app/views/orders/_service_desc.html.haml
drench: I moved `template_results` and `sample_results` to the `OrderDetailsController` which should avoid both the 404s and 403s on these.

app/views/order_details/order_file.html.haml
drench: I'd like to change the route so it's not scoped under facility management. Regular users would see the link when scoped to an order, but that wouldn't make sense in the context of an admin managing a facility. Any suggestions?

app/models/order_import.rb
drench: Added a task called `paperclip:mime:update_from_file_extensions`. It should be able to run before or after the S3 migration.

app/controllers/file_uploads_controller.rb
```diff
@@ -23,6 +23,20 @@ def upload
     @file    = @product.stored_files.new(:file_type => params[:file_type])
   end
 
+  # GET /facilities/:facility_id/:product/:product_id/files/:file_type/:id
+  def download
```
jhanggi: I think if you change the before_filters to:
```
load_and_authorize_resource :class => StoredFile, :except => [:uploader_create, :download]
```
That should solve your 403 problem (I see no reason why any of these files need to be authorized).

jhanggi:
Found a few more issues. I'd like to get this fixed up and out soon, but I'm wondering if it would be worth spending some time writing some specs, particularly around permissions. There seem to be quite a few issues going back and forth.

I wish we had capybara set up to do some integration tests, but I'm not 100% sure it's worth spending the time to get it up and running.

jhanggi:
In addition to the 404 I'm getting when trying to access a template_result from my cart, I'm also getting a 403 when trying to download a sample_result as a normal user from "My Orders".

jhanggi:
This looks good now. I'm thinking I'll wait to merge this in until we get some other pieces released to production.

PR #312, opened 2015-07-20
[#98882 102019] Reservation time validation updates

drench:
I did some cleanup as I was investigating, but the two primary issues this should address are:

* Being able to change reservation times while its cart is open: cee3913f433671ccc455347c949b5a5e3c4c97f0
* Taking a reservation whose start time is currently outside a lock window, and changing it to a time within a lock window: 5a2c6eb4c3e513a36809e5ec5d4971a184254c7e


jhanggi:
:+1:

PR #313, opened 2015-07-27
[#76059] Abbreviate weekday name to avoid wrapping

drench:
Before timeline nav before

After timeline nav after

### DIFF
--- a/app/views/facility_reservations/timeline.html.haml
+++ b/app/views/facility_reservations/timeline.html.haml
@@ -16,7 +16,7 @@
       = link_to "&laquo;".html_safe,
         timeline_facility_reservations_path(current_facility, date: format_usa_date(@display_date - 1.day)), id: "reservation_left"
       #reservation_date
-        = l(@display_date.to_date, format: :timeline_navigation)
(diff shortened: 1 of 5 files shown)


app/views/facility_reservations/timeline.html.haml
```diff
@@ -16,7 +16,7 @@
       = link_to "&laquo;".html_safe,
         timeline_facility_reservations_path(current_facility, date: format_usa_date(@display_date - 1.day)), id: "reservation_left"
       #reservation_date
-        = human_date_with_day(@display_date)
+        = l(@display_date.to_date, format: :timeline_navigation)
```
jhanggi: Just curious, would it be possible to pull the `#to_date` a little further up so `@display_date` is always already a date?

jhanggi:
One potential refactoring, but otherwise looks good.

drench:
Refactored. I also reformatted the public timeline partial to bring it closer to the timeline partial. We may be able to extract common pieces of these at some point, given their similarities.

jhanggi:
:+1:

PR #314, opened 2015-07-28
[#102824] Use has_review_period? to determine Billing tab URL

drench:
This is specifically for UI Core where there is no review period, which means the notifications is a 404.

A big TODO that remains: extract and refactor the first 20 lines of this view.


jhanggi:
:+1:

PR #315, opened 2015-07-29
[#102917] Use delayed job for email sending

jhanggi:
This should ideally help address some timeouts that are happening with large order lists for Send Notifications.

I will not be surprised if there are some test failures from checking that emails were indeed sent.

### DIFF
--- a/app/controllers/facility_notifications_controller.rb
+++ b/app/controllers/facility_notifications_controller.rb
@@ -107,7 +107,7 @@ private
   def notify_accounts
     @accounts_to_notify.each do |account, facility|
       account.notify_users.each do |u|
-        Notifier.delay.review_orders(:user => u, :facility => facility, :account => account)
+        Notifier.delay.review_orders(user: u, facility: facility, account: account)
       end
     end
   end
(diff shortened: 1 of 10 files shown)


app/controllers/facility_notifications_controller.rb
```diff
@@ -107,7 +107,7 @@ def mark_as_reviewed
   def notify_accounts
     @accounts_to_notify.each do |account, facility|
       account.notify_users.each do |u|
-        Notifier.review_orders(:user => u, :facility => facility, :account => account).deliver
+        Notifier.delay.review_orders(:user => u, :facility => facility, :account => account)
```
drench: Have you run into any trouble with deserialization? I remember it being a problem at one point with `AccountTransactionReportMailer.csv_report_email`, and we worked around that by passing in just IDs. But that was also an older `delayed_job` gem version…

jhanggi: When it takes basic ActiveRecord objects, it seems to be fine. I am going to change the one mailer that we send `self` back (since that one's already a background job anyways).

jhanggi:
Made a couple more changes. Test should pass now.

drench:
:+1:

PR #316, opened 2015-07-30
Document overriding of settings for S3 storage

jhanggi:


config/settings.yml
```diff
@@ -92,6 +92,8 @@ feature:
   suspend_accounts_on: true
   product_specific_contacts_on: true
 
+# This is possible to be overridden in settings.local.yml if your fork is using S3, so
```
drench: How about "This may be overridden in settings.local.yml…" ?

drench:
Just one wording suggestion. Otherwise :+1:

PR #319, opened 2015-08-04
[#103115] Fix error on next_available not found for admin reservation

jhanggi:
`facility_reservations#new (NoMethodError) "undefined method `reserve_interval' for nil:NilClass"`

When @instrument.next_available_reservation is nil


drench:
:+1:

PR #320, opened 2015-08-06
[#98857] Use blank for ownerless account owner names

drench:
Though this shouldn't happen going forward, there are some ownerless accounts in UICore, causing the order detail modal to not load.

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -13,16 +13,12 @@ module AccountsHelper
       [
         account.to_s,
         account.id,
-        { 'data-account-owner' => account_owner_name(account) },
+        { 'data-account-owner' => account.owner_user_name },
       ]
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/helpers/accounts_helper.rb
```diff
@@ -21,4 +21,8 @@ def available_accounts_array
   def available_accounts_options
     options_for_select(available_accounts_array, @order_detail.account_id)
   end
+
+  def account_owner_name(account)
+    account.owner_user.try(:name) || ""
```
jhanggi: Would we be better off if we changed `Account#owner_user` to return some sort of null object?

drench:
As discussed, we went with replacing to all calls to `Account#owner_user.name`.

jhanggi:
:+1: Feel free to go ahead to squash this into master and merge this down to uic.

PR #321, opened 2015-08-14
[#102019] Reservation backdating updates

drench:
This includes some new specs and refactoring I did while investigating but the meat is in these two commits:

* 5e2c1402f4fabd5982eb2506b6fcab840f83b446
  We were reverting the reservation start date/time params silently even when editable by an administrator and in a cart. By allowing these params through in this case, admins can backdate and non-admins will see a validation error if they try.

* d505efba063c81ca56f885025e34498ae697902f
  Even with the backdated params allowed through, we were denying changes even if the order was not yet complete. This allows them through if the reservation is still in a cart.

### DIFF
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -46,7 +46,7 @@ class FacilityReservationsController < ApplicationController
 
     raise ActiveRecord::RecordNotFound unless @reservation == Reservation.find(params[:id])
     set_windows
-    unless @reservation.admin_editable? || @reservation.can_edit_actuals?
+    if @reservation.locked?
       return redirect_to facility_order_order_detail_reservation_path(current_facility, @order, @order_detail, @reservation)
     end
   end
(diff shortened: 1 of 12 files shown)


app/controllers/facility_reservations_controller.rb
```diff
@@ -46,7 +46,7 @@ def edit
 
     raise ActiveRecord::RecordNotFound unless @reservation == Reservation.find(params[:id])
     set_windows
-    unless @reservation.can_edit? || @reservation.can_edit_actuals?
+    unless @reservation.admin_editable? || @reservation.can_edit_actuals?
```
jhanggi: Can we extract this `||` check into a method since it's also used down below? I would see if there's a way we can phrase the positivity/negativity of the name to flip the two  `unless`es into `if`s.

app/controllers/facility_reservations_controller.rb
```diff
@@ -197,11 +199,13 @@ def show_problems_path
   private
 
   def reserve_changed?
-    @reservation.can_edit? && @reservation.changes.any? { |k,v| k == 'reserve_start_at' || k == 'reserve_end_at' }
+    @reservation.admin_editable? &&
```
jhanggi: What does the `can_edit?` => `admin_editable?` change do?

jhanggi: Ah, just a name change. Makes sense.

jhanggi:
Just my one comment about seeing if we can flip some `unless`es to make the logic easier to understand. Otherwise, looks good.

drench:
Updated. How does `Reservation#locked?` sound?

jhanggi:
Works for me. :+1: Let's get this into UIC as well.

PR #322, opened 2015-08-17
[#103870] Make training requests an optional feature

drench:
Default is on, but when off it should act as it did before https://github.com/tablexi/nucore-open/pull/298 and display a message to the user about contacting the facility for approval.

### DIFF
--- a/app/controllers/products_common_controller.rb
+++ b/app/controllers/products_common_controller.rb
@@ -77,11 +77,16 @@ class ProductsCommonController < ApplicationController
 
     # is the user approved?
     if add_to_cart && !@product.can_be_used_by?(acting_user) && !session_user_can_override_restrictions?(@product)
-      if TrainingRequest.submitted?(session_user, @product)
-        flash[:notice] = t_model_error(Product, "already_requested_access", product: @product)
-        return redirect_to facility_path(current_facility)
+      if SettingsHelper.feature_on?(:training_requests)
+        if TrainingRequest.submitted?(session_user, @product)
+          flash[:notice] = t_model_error(Product, "already_requested_access", product: @product)
+          return redirect_to facility_path(current_facility)
+        else
+          return redirect_to new_facility_product_training_request_path(current_facility, @product)
+        end
       else
-        return redirect_to new_facility_product_training_request_path(current_facility, @product)
+        add_to_cart = false
+        @error = "requires_approval"
       end
     end
 
--- a/config/locales/en.models.yml
+++ b/config/locales/en.models.yml
@@ -55,6 +55,11 @@ en:
           requires_approval: The %{product} requires approval to reserve. Would you like to be contacted by the core for access or training?
         instrument:
           not_available: The %{instrument} instrument is currently unavailable for reservation online.
+          requires_approval_html: |
+            The %{instrument} instrument requires approval to reserve;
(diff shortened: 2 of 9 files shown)


config/locales/en.models.yml
```diff
@@ -55,6 +55,11 @@ en:
           requires_approval: The %{product} requires approval to reserve. Would you like to be contacted by the core for access or training?
         instrument:
           not_available: The %{instrument} instrument is currently unavailable for reservation online.
+          requires_approval_html: |
+            The %{instrument} instrument requires approval to reserve;
+            please contact the facility for further information:<br/><br/>
+            %{facility}<br/>
+            <a href=\"mailto:%{email}\">%{email}</a>
```
drench: This is a reformatted revert of 110cf0ed0cf5377276c5965b1824fbe49203063c

app/controllers/products_common_controller.rb
```diff
@@ -77,11 +77,16 @@ def show
 
     # is the user approved?
     if add_to_cart && !@product.can_be_used_by?(acting_user) && !session_user_can_override_restrictions?(@product)
-      if TrainingRequest.submitted?(session_user, @product)
-        flash[:notice] = t_model_error(Product, "already_requested_access", product: @product)
-        return redirect_to facility_path(current_facility)
+      if SettingsHelper.feature_on?(:training_requests)
```
jhanggi: Is there a way we can share this code with the instruments controller? It looks like it's the same other than the `@instrument` vs `@product`

spec/controllers/bundles_controller_spec.rb
```diff
@@ -93,9 +93,29 @@
...
-          assert_redirected_to(new_facility_product_training_request_path(facility, bundle))
+        around(:each) do |example|
+          saved_setting = SettingsHelper.feature_on?(:training_requests)
+          SettingsHelper.enable_feature(:training_requests, training_requests)
+          example.run
+          SettingsHelper.enable_feature(:training_requests, saved_setting)
+        end
+
+        context "if the training request feature is enabled" do
+          let(:training_requests) { true }
```
jhanggi: Can we make this variable something more like `training_requests_enabled` to make it a little more readable in the `around`?

spec/controllers/items_controller_spec.rb
```diff
@@ -89,9 +89,29 @@
           do_request
         end
 
-        it "gives the user the option to submit a request for approval" do
-          expect(assigns[:add_to_cart]).to be_blank
-          assert_redirected_to(new_facility_product_training_request_path(facility, item))
+        around(:each) do |example|
```
jhanggi: Is there a way we can come up with a better technique than duplicating this everywhere? We have a `include_context "feature enabled", :create_users`, in one place, but I'm not sure that's the best option. Is there something we could do with metadata?

```
context "with training requests", feature_setting: { training_requests: true } do
```
or something similar; that was just the first thing that popped in my mind.

jhanggi:
Overall, looks pretty good; just a few minor thoughts about how to clean things up a bit.

jhanggi:
:+1:

app/controllers/products_common_controller.rb
drench: I'm working with @danhodos and @yoon on refactoring this. It's significant enough to deserve its own PR.

PR #326, opened 2015-08-20
[#60971 60973] Add more message types to the Notices tab

drench:
This takes the existing "Notices" tab in the header and expands it to include:
* Disputed Orders
* Problem Orders
* Problem Reservations
* Training Requests

The items (and total count) should reflect only the messages appropriate for the current user in the current facility context. That is, if the user is not allowed to manage training requests, they will not appear in the notifications menu even if some exist.

If there are no messages for a given category, it is not included in the drop-down. If there are no messages of any type for the user and context, the "Notices" tab should not appear at all.

Each menu item is a link to a page to view and manage the appropriate type. "Disputed Orders" is new, and is under the Billing tab. It should show all disputed order types, both reservation and non-reservation.


app/controllers/facilities_controller.rb
```diff
@@ -113,6 +114,14 @@ def transactions_with_search
     paginate_order_details
   end
 
+  # GET /facilities/:facility_id/disputed_orders
+  def disputed_orders_with_search
```
jhanggi: One question I have is do we care about having this new page or could we just have two options in the dropdown, one for reservations and one for orders?

app/models/facility.rb
```diff
@@ -101,8 +102,20 @@ def to_s
     "#{name} (#{abbreviation})"
   end
 
+  def problem_order_details
```
jhanggi: I think this naming scheme is kind of confusing. I would expect `problem_order_details` to be all order details, not to exclude anything. I was actually fine with using `reservations` and `non_reservations`.

app/models/facility.rb
```diff
@@ -101,8 +102,20 @@ def to_s
...
+    complete_problem_order_details.non_reservations
```
jhanggi: And then I was thinking `complete_problem_order_details` would be ALL problem order details.

app/models/message_notifier.rb
```diff
@@ -0,0 +1,90 @@
...
+  def message_count
+    notifications.count +
+    order_details_in_dispute.count +
+    problem_order_details.count +
+    problem_reservation_order_details.count +
+    training_requests.count
+  end
+
+  def messages?
+    notifications? ||
```
jhanggi: I'm not sure I'm convinced that the extra predicate method wrapping `any?` on each of these is adding much value. If we're not wrapping the `count`s above, maybe we don't wrap them here either.

app/views/facilities/disputed_orders.html.haml
```diff
@@ -0,0 +1,12 @@
...
+- content_for :h1 do
+  = current_facility
+
+- content_for :h2 do
+  %h2= t(".head")
+
+= content_for :top_block do
+  = render "shared/transactions/top", tab: "disputed_orders"
+
+= render "shared/transactions/table" if @order_details.present?
```
jhanggi: We should show something if order details is empty. Take a look at what we do for some of the other search pages.

app/views/shared/_header.html.haml
```diff
@@ -18,9 +18,9 @@
...
               %li= link_to t('pages.cart') + " (#{current_cart.order_details.sum(:quantity)})", :cart
               %li.divider-vertical
-              - if current_user.operator? || current_user.administrator?
-                - n_count=current_user.notifications.active.count
-                - if n_count > 0
-                  %li= link_to t('pages.notices') + " (#{n_count})", notifications_path
-                  %li.divider-vertical
+              - notifier = MessageNotifier.new(current_user, current_ability, current_facility)
+              - if notifier.messages?
+                = render "shared/message_notifier", notifier: notifier
```
jhanggi: Is there a reason you chose to put the notifier setup and initial check outside of the partial? It seems like it would be more consolidated if everything were in the partial.

app/views/shared/_message_notifier.html.haml
```diff
@@ -0,0 +1,33 @@
+%li
+  %ul.nav.pull-right
+    %li.dropdown.pull-right
+      = link_to "Notices (#{notifier.message_count})",
```
jhanggi: I18n.

app/views/shared/_message_notifier.html.haml
```diff
@@ -0,0 +1,33 @@
...
+        class: "dropdown-toggle",
+        data: { toggle: "dropdown" }
+
+      %ul.dropdown-menu
+        - if notifier.order_details_in_dispute?
+          %li
+            = link_to "#{t("message_notifier.order_details_in_dispute")} (#{notifier.order_details_in_dispute.count})",
+              facility_disputed_orders_path(current_facility)
+
+        - if notifier.notifications?
```
jhanggi: I think it might be good to try to come up with a different name for how this is used in this interface. This is kind of confusing.

app/models/message_notifier.rb
jhanggi: What if this method was just `message_count > 0` and `message_count` was memoized? It looks like you handle any of the permissions there as well.

It may be premature optimization, but what if you memoized each individual count? I think that would minimize the number of queries run (one `count(*)` for each notification type).

app/views/shared/_message_notifier.html.haml
```diff
@@ -0,0 +1,33 @@
+%li
+  %ul.nav.pull-right
+    %li.dropdown.pull-right
+      = link_to "Notices (#{notifier.message_count})",
+        :notifications,
+        class: "dropdown-toggle",
+        data: { toggle: "dropdown" }
+
+      %ul.dropdown-menu
+        - if notifier.order_details_in_dispute?
```
jhanggi: I had another thought: what if `notifier` responded to `each` and yields an object for each notification type that has notifications. It might mean that the notifier needs to know about `link_to` and the URL helpers, but it could clean up the view code quite a bit. It sort of turns the notifier object into more of a presenter.

```
%ul.dropdown-menu
  - notifier.each do |notifications|
    %li= notifications.link
```

spec/models/message_notifier_spec.rb
```diff
@@ -0,0 +1,204 @@
...
+  let(:user) { create(:user) }
+
+  let(:disputed_orders_visible?) { false }
+  let(:manage_training_requests?) { false }
+  let(:notifications_visible?) { false }
+  let(:show_problem_orders?) { false }
+  let(:show_problem_reservations?) { false }
+
+  before(:each) do
+    subject.stub(:disputed_orders_visible?).and_return(disputed_orders_visible?)
```
jhanggi: Take a look at https://github.com/ryanb/cancan/wiki/Testing-Abilities#controller-testing

It might be nicer and fuller testing if you create a stub `Ability` object vs stubbing the private methods on the MessageNotifier. If you do that, I would almost move the `ability.can?`s into the `return [] unless @ability.can?(...)`. One less level of indirection.

jhanggi:
This is a good start. I think there's some additional refactoring that can be done.

One other thought: I wonder if we should call the notifications out a little more. Make them orange, or use the bootstrap badge styling. Something so they're a little more prominent against the cart.

Also, I don't really care, but if I'm an admin, I can see the notices when I'm in the customer view looking down the list of products, but then it goes away once I'm in my cart.

app/models/facility.rb
drench: I went back and forth on this. On the front end, there are "problem orders" and "problem reservations" and I went with similar naming here. But `reservations` and `non_reservations` make more sense, internally.

app/views/facilities/disputed_orders.html.haml
drench: The `alert` in this partial catches that case https://github.com/tablexi/nucore-open/blob/60971-60973-notifications_menu/app/views/shared/transactions/_top.html.haml#L7 (and I added a translation for it).

But it's still potentially confusing: if your messages menu says you have disputed orders, and you follow the link, there's a chance you won't see any orders listed if the disputed orders are all outside the default search date range.

drench:
Major refactor, with some renaming. The iterator suggestion made me look at this in a different way. The `count` memoization is here, and the class became more of a presenter.

There's more I'm wondering about: the public interface could be smaller, but then there seem to be too many private methods already. It also does a lot of slinging around of the `controller` object that makes me feel there's a way to reduce if not eliminate that.

The notices changing when viewing the cart is an odd case in that, even though we could derive the current facility through the order, the URL isn't scoped to a facility so `current_facility` isn't set in that case.

It's ready for another look.

app/controllers/facilities_controller.rb
```diff
@@ -113,6 +114,14 @@ def transactions_with_search
...
+    set_default_start_date
+    @export_enabled = true
```
jhanggi: I'm not sure there's any value in having the export enabled for this view.

app/controllers/facilities_controller.rb
```diff
@@ -113,6 +114,14 @@ def transactions_with_search
     paginate_order_details
   end
 
+  # GET /facilities/:facility_id/disputed_orders
+  def disputed_orders_with_search
+    set_default_start_date
```
jhanggi: I think we can remove this call, which should address your concern about the confusion of maybe not seeing an order if it's outside the range. That was mainly there on the main one so CSV exports don't include _everything_ by default.

app/models/message_summarizer.rb
```diff
@@ -0,0 +1,184 @@
...
+    ((user.operator? || user.administrator?) && user.notifications.active.count)
+  end
+
+  def order_details_in_dispute_count
+    if facility && ability.can?(:disputed_orders, Facility)
+      facility.order_details_in_dispute.count
+    end
+  end
+
+  def problem_order_details_count
```
jhanggi: It feels to me like these pieces should be part of the individual MessageSummary classes, so each of those would have a `count` method.

app/models/message_summarizer.rb
```diff
@@ -0,0 +1,184 @@
+class MessageSummarizer
+  class MessageSummary
```
jhanggi: I feel like this and the subclasses have gotten long enough where they should probably be moved to separate files. It seems weird that the actual class doesn't start until line 89.

app/models/message_summarizer.rb
```diff
@@ -0,0 +1,184 @@
...
+      notifications,
+      order_details_in_dispute,
+      problem_order_details,
+      problem_reservation_order_details,
+      training_requests,
+    ].select(&:any?)
+  end
+
+  def notifications_count
+    ((user.operator? || user.administrator?) && user.notifications.active.count)
```
jhanggi: Do we have an Ability for notifications?

app/models/order_detail.rb
```diff
@@ -129,6 +129,13 @@ def self.for_facility_url(facility_url)
     details
   end
 
+  def self.in_dispute
```
jhanggi: hooray getting rid of the old syntax.

app/views/shared/_message_summary.html.haml
```diff
@@ -0,0 +1,14 @@
+- summarizer = MessageSummarizer.new(controller)
+- if summarizer.messages?
+  %li
+    %ul.nav.pull-right
+      %li.dropdown.pull-right
+        = link_to "Notices (#{summarizer.message_count})",
+          :notifications,
```
jhanggi: Should this actually be linking to anything? The notifications link will show up in the dropdown, correct?

config/locales/en.yml
```diff
@@ -305,6 +311,8 @@ en:
       no_problem_orders: There are no problem orders.
 
   facility_reservations:
+    disputed:
+      no_disputed_reservations: There are no Disputed reservations
```
jhanggi: Probably leave this uncapitalized for consistency.

spec/helpers/transaction_history_helper_spec.rb
```diff
@@ -0,0 +1,51 @@
+require "spec_helper"
+
+describe TransactionHistoryHelper do
+  describe "#chosen_field"
```
jhanggi: Is there anything that's supposed to be inside this describe block.

spec/models/message_summarizer_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+
+    context "and the user can manage training requests" do
+      before { ability.can(:manage, TrainingRequest) }
+
+      it_behaves_like "there is one overall message"
+
+      it { expect(subject.training_requests.count).to eq(1) }
+
+      it "returns only training requests when iterating" do
+        subject.each do |message_summary|
```
jhanggi: Once you've implemented `each`, if you can probably `include Enumerable` and get the `first` message for free.

spec/models/message_summarizer_spec.rb
```diff
@@ -0,0 +1,226 @@
...
+          expect(message_summary)
+            .to be_kind_of(MessageSummarizer::TrainingRequests)
```
jhanggi: I think I'd like to see some deeper testing, like seeing that the message ends up as "Training Requests (1)".

jhanggi:
Cool. I think we're getting closer.

app/models/message_summarizer.rb
drench: Doesn't look like it. I derived this from https://github.com/tablexi/nucore-open/blob/master/app/views/shared/_header.html.haml#L21 (since we're replacing it) but really it ought to be an `Ability`.

app/views/shared/_message_summary.html.haml
drench: Right. Clicking the "link" opens the drop-down. It's probably confusing to use "notices"/"notifications" to mean both the specific `Notification` and the collection of all types of message. I'm calling them, collectively, "messages" internally.

spec/helpers/transaction_history_helper_spec.rb
drench: Intentionally left blank. This whole file is tangential: when I started adding Disputed Orders to the Billing section I ran across the transaction search helper, and noticed there were no specs. So I wrote some, and left some pending.

drench:
This is ready for another look. The previous iteration had more in its public interface than we needed. As I wrote in the commit for 8339cf7, the view that uses this only needs to:
* know if there are any messages
* know how many total messages there are
* iterate through each message type
* get a labeled link for each message type

So that's the interface, and that's what's tested. Though it felt weird creating several subclasses that were 100% private.

app/models/message_summarizer.rb
```diff
@@ -0,0 +1,52 @@
+class MessageSummarizer
+  include Enumerable
+
+  def initialize(controller)
+    @controller = controller
+  end
+
+  def each
```
jhanggi: Could we `delegate :each, to: :message_summaries`?

app/models/message_summarizer/message_summary.rb
```diff
@@ -0,0 +1,33 @@
...
+    count > 0
+  end
+
+  def count
+    @count ||= allowed? ? get_count : 0
+  end
+
+  def link
+    controller.view_context.link_to(label, path)
+  end
```
jhanggi: It might be worth adding the methods with `NotImplementedError`s for the methods we expect to be overridden.

app/models/message_summarizer/order_Details_in_dispute_summary.rb
```diff
@@ -0,0 +1,19 @@
+class MessageSummarizer::OrderDetailsInDisputeSummary < MessageSummarizer::MessageSummary
```
jhanggi: Looks like the filename has a capital D.

app/models/message_summarizer/problem_order_details_summary.rb
```diff
@@ -0,0 +1,19 @@
...
+  def allowed?
+    facility && ability.can?(:show_problems, Order)
+  end
+
+  def get_count
+    facility.problem_non_reservation_order_details.count
+  end
+
+  def l18n_key
+    "message_summarizer.problem_order_details"
```
jhanggi: Since they're all under `message_summarizer`, would it make sense to pull that up into the superclass?

jhanggi:
Well done. Just a couple very minor comments. :+1:

app/models/message_summarizer/problem_order_details_summary.rb
drench: So close: `NotificationsSummary`'s key is the exception with `pages.notices`.

PR #328, opened 2015-08-21
[#119048] Restore unavailability display on timelines

drench:
Since `@display_date` wasn't being used as a `Time`, #313 changed `@display_date` to a `Date`. Except it was: `ScheduleRule.unavailable_for_date` takes a `day` argument and expecting it to behave like a `Time`. The reserve start/end times were no longer being generated correctly, and we lost the greyed-out areas on timelines showing off-hours unavailability.

This patch explicitly converts `day` into a `Time` so the hour and minute manipulation work even if it's a `Date`, though it should work either way.

### DIFF
--- a/app/models/schedule_rule.rb
+++ b/app/models/schedule_rule.rb
@@ -31,8 +31,8 @@ class ScheduleRule < ActiveRecord::Base
     rules.each do |rule|
       res = Reservation.new({
         :product => instrument,
-        :reserve_start_at => day.dup.change(:hour => rule.start_hour, :min => rule.start_min),
-        :reserve_end_at => day.dup.change(:hour => rule.end_hour, :min => rule.end_min),
+        :reserve_start_at => day.to_time.change(:hour => rule.start_hour, :min => rule.start_min),
+        :reserve_end_at => day.to_time.change(:hour => rule.end_hour, :min => rule.end_min),
         :blackout => true
         })
       reservations << res
--- a/spec/models/schedule_rule_spec.rb
+++ b/spec/models/schedule_rule_spec.rb
@@ -1,7 +1,6 @@
 require 'spec_helper'
 
 describe ScheduleRule do
-
   it "should create using factory" do
     @facility   = FactoryGirl.create(:facility)
     @facility_account = @facility.facility_accounts.create(FactoryGirl.attributes_for(:facility_account))
@@ -59,7 +58,7 @@ describe ScheduleRule do
       @rule       = @instrument.schedule_rules.build(FactoryGirl.attributes_for(:schedule_rule))
       @rule.includes_datetime(DateTime.new(1981, 9, 15, 12, 0, 0)).should == true
     end
-    
+
     it "should not recognize non-inclusive datetimes" do
       @facility   = FactoryGirl.create(:facility)
       @facility_account = @facility.facility_accounts.create(FactoryGirl.attributes_for(:facility_account))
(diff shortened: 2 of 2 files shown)


spec/models/schedule_rule_spec.rb
```diff
@@ -12,6 +15,52 @@
...
+          expect(reservations.size).to eq(2)
+          reservations.each do |reservation|
+            expect(reservation).to be_kind_of(Reservation)
+            expect(reservation).not_to be_persisted
+          end
+        end
+
+        it "reserves midnight to 9 AM as unavailable" do
+          expect(reservations.first.reserve_start_at)
+            .to eq(day.to_time.change(hour: 0, min: 0))
```
jhanggi: I wonder if there's a little bit better approach for this test. This seems to be testing the actual implementation a little too closely. Maybe a string comparison?

Not a big deal if you don't like that as much. I'm not against this current implementation.

spec/models/schedule_rule_spec.rb
```diff
@@ -12,6 +15,52 @@
...
+      end
+
+      let(:reservations) { ScheduleRule.unavailable_for_date(instrument, day) }
+
+      shared_examples_for "it generates reservations to cover unavailability" do
+        it "returns two dummy reservations" do
+          expect(reservations.size).to eq(2)
+          reservations.each do |reservation|
+            expect(reservation).to be_kind_of(Reservation)
+            expect(reservation).not_to be_persisted
```
jhanggi: Should we also test `to be_blackout`? (I think there's a predicate method).

jhanggi: Do we have the rspec feature: `expect(reservations).to all( be_kind_of(Reservation )`)?

spec/models/schedule_rule_spec.rb
```diff
@@ -1,6 +1,9 @@
 require 'spec_helper'
 
 describe ScheduleRule do
+  let(:facility) { create(:facility) }
+  let(:facility_account) { facility.facility_accounts.create(attributes_for(:facility_account)) }
+  let(:instrument) { create(:instrument, facility: facility, facility_account: facility_account) }
 
   it "should create using factory" do
     @facility   = FactoryGirl.create(:facility)
```
jhanggi: Not sure if it's worth cleaning up a bit to do `@faciltiy = facility`, etc.

app/models/schedule_rule.rb
```diff
@@ -31,8 +31,8 @@ def self.unavailable_for_date(instrument, day)
     rules.each do |rule|
       res = Reservation.new({
         :product => instrument,
-        :reserve_start_at => day.dup.change(:hour => rule.start_hour, :min => rule.start_min),
-        :reserve_end_at => day.dup.change(:hour => rule.end_hour, :min => rule.end_min),
+        :reserve_start_at => day.to_time.change(:hour => rule.start_hour, :min => rule.start_min),
```
jhanggi: Do you have a rubocop setup to update the hash syntax? If so, should we update this file?

jhanggi:
Fix looks good. Just a couple minor comments on the specs.

spec/models/schedule_rule_spec.rb
drench: Doesn't appear we do.

spec/models/schedule_rule_spec.rb
drench: I know it. There's 400 more lines in this spec I'd like to clean up.

app/models/schedule_rule.rb
drench: I really wanted to update this method, but I like that looking at this diff emphasizes the change is just from `dup` to `to_time`. Now that we have some tests on this method, I'm more comfortable with bigger changes.

PR #329, opened 2015-08-24
Reload routes when using the feature flag metadata helper

jhanggi:
Without the route reloading, if the flag was originally false, routes would not be available even after enabling the feature.

This should address the test failures on UIC.


drench:
:+1:

PR #330, opened 2015-08-24
[#103850] Feature flag for public daily view

jhanggi:
UIC does not want this feature enabled as they have concerns with
hidden instruments on shared schedules.

The test will fail in UIC until #329 is merged.


drench:
:+1:

PR #331, opened 2015-08-25
[#119338] Ensure locales/override/*.yml is loaded

jhanggi:
In UIC, the override/en.yml was not being added to the load path. This
also adds tests to ensure the overrides are being loaded properly.


drench:
:+1:

PR #333, opened 2015-08-26
[#119372] Do not error on instrument validations when reservation interval is 0

jhanggi:
To reproduce:
* Go to Instruments#new
* In Reservation Restrictions, Leave the Interval (minutes) blank
* Put a value into either Minimum (minutes) or Maximum (minutes)
* Submit the form

The error was happening because reserve_interval was nil.

### DIFF
--- a/spec/models/instrument_spec.rb
+++ b/spec/models/instrument_spec.rb
@@ -25,56 +25,79 @@ describe Instrument do
   end
 
   describe 'min/max reservation is a multiple of reservation interval' do
+    subject { instrument }
+    let(:reserve_interval) { 5 }
+    let(:min_reserve_mins) { 1 }
+    let(:max_reserve_mins) { 1000 }
 
-    it 'ensures that the minimum reservation time is a multiple of the reservation interval' do
-      instrument.reserve_interval = 5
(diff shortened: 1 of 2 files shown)


spec/models/instrument_spec.rb
```diff
@@ -24,41 +24,58 @@
...
+    end
+
+    it 'does not error when minimum reservation is set, but reservation interval is not' do
+      instrument.reserve_interval = nil
+      instrument.min_reserve_mins = 30
+      expect { instrument.valid? }.not_to raise_error
+      expect(instrument.errors[:reserve_interval]).to be_present
+    end
+
+    it 'does not error when maximum reservation is set, but reserve_interval is not' do
```
drench: Super minor, but it took me a bit to notice how this spec was different from the previous one. I'd consider doing this more in a setup-in-`context` style but then I'd feel the need to rework this whole file.

drench:
One note on the spec style but that's a take or leave. :+1:

jhanggi:
I've cleaned up the specs a bit, but only in that area. I'm leaving the rest of the file alone.

drench:
:+1:

PR #334, opened 2015-08-26
[#102969] Allow skipping monitoring of daemons

jhanggi:
Allow passing a command line argument that disables the daemon monitoring.
This is useful if you are using something external like eye to do your
monitoring.

This also moves the daemon pids into tmp/pids


jhanggi:
@phoolish there will be following PRs for nu and uic

phoolish:
:+1:

PR #327, opened 2015-08-21
[#102917] Send notification optimizations

jhanggi:
Based on https://github.com/tablexi/nucore-nu/pull/80

Here are the benchmarks found when running the NU branch on Oracle with 90 accounts.

```
original
  4.260000   0.340000   4.600000 ( 11.907166)
  3.730000   0.380000   4.110000 ( 11.669257)
  3.670000   0.340000   4.010000 (  9.287855)
  3.780000   0.280000   4.060000 ( 10.854380)
  3.930000   0.300000   4.230000 ( 10.596839)
  3.940000   0.300000   4.240000 ( 10.105375)
  4.100000   0.320000   4.420000 ( 11.470097)
  3.960000   0.300000   4.260000 ( 10.153493)
  4.210000   0.330000   4.540000 ( 14.617991)
  3.920000   0.290000   4.210000 ( 10.630960)
order_details.each, no includes
  2.000000   0.160000   2.160000 (  4.928323)
  1.870000   0.140000   2.010000 (  5.722727)
  1.820000   0.130000   1.950000 (  5.672192)
  1.850000   0.130000   1.980000 (  5.050869)
  1.990000   0.150000   2.140000 (  7.732272)
  1.940000   0.140000   2.080000 (  5.147758)
  2.000000   0.150000   2.150000 (  5.993348)
  1.920000   0.140000   2.060000 (  5.821812)
  1.980000   0.140000   2.120000 (  5.354489)
  1.950000   0.160000   2.110000 (  5.342445)
order_details.each, includes
  1.240000   0.090000   1.330000 (  3.946294)
  1.240000   0.090000   1.330000 (  3.251158)
  1.430000   0.100000   1.530000 (  3.968313)
  1.200000   0.090000   1.290000 (  4.353429)
  1.330000   0.100000   1.430000 (  4.299328)
  1.220000   0.090000   1.310000 (  3.494067)
  1.240000   0.100000   1.340000 (  3.889009)
  1.130000   0.130000   1.260000 (  3.726788)
  1.320000   0.110000   1.430000 (  3.942863)
  1.180000   0.090000   1.270000 (  3.899770)
order_details.update_all, no includes
  0.420000   0.050000   0.470000 (  2.087882)
  0.290000   0.030000   0.320000 (  1.347384)
  0.300000   0.040000   0.340000 (  2.127723)
  0.300000   0.030000   0.330000 (  1.831687)
  0.290000   0.040000   0.330000 (  1.652789)
[...]

PR #338, opened 2015-08-31
[#102917] Send emails in send notification

jhanggi:
This is a regression from #327. I was a little too confident in the
testing support and it turns out there weren’t tests around actually
sending the emails.

### DIFF
--- a/spec/controllers/facility_notifications_controller_spec.rb
+++ b/spec/controllers/facility_notifications_controller_spec.rb
@@ -97,7 +97,7 @@ describe FacilityNotificationsController do
           maybe_grant_always_sign_in(:admin)
         end
 
-        it 'should send emails to the two accounts accounts' do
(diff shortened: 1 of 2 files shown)


spec/controllers/facility_notifications_controller_spec.rb
```diff
@@ -93,6 +97,10 @@ def self.it_should_404_for_zero_day_review
           maybe_grant_always_sign_in(:admin)
         end
 
+        it 'should send emails to the two accounts accounts' do
```
drench: I'd nix the "should" here.

drench:
:+1:, with one spec wording suggestion.

spec/controllers/facility_notifications_controller_spec.rb
jhanggi: Good catch. I'll fix before I merge.

PR #339, opened 2015-08-31
[#60973] Ensure notices tab style matches the nav

drench:
When clicked, the "Notices" tab's style would revert to the bootstrap default, ignoring what might be defined in nucore-overrides.scss.

It should have no real effect on this fork, but it's noticeable with NU's and UIC's styles.


jhanggi:
:+1:

PR #340, opened 2015-09-02
[#95795] Avoid numeric overflow on price policy create/edit

drench:
I did a lot of cleanup and refactoring while working through this one, but the fix itself is in this commit: 6fcbc03895ebf25ae2c4ea77d9fb496d270f351d

What it boiled down to is this:

Let's say I'm creating or editing an `Instrument` price policy (and not any other `Product` type). I set the base rate as 10.00 per hour. I then check the "Can Purchase?" box next to the External Rate, but leave the Rate Per Hour blank, and then click "Save Rules". This fails, and I'm sent back to the price policy edit screen, where the system, using the `display_usage_rate` helper, has filled in the base rate field from `params`. It made two assumptions: that it's a number, and that it's a rate per minute. Since it's a string, multiplication turns it into "10.0010.0010.0010.0010.00…" and if I attempt to save with that rate, we get the numeric overflow exception.

But since it's already an hourly rate, and we want to display it as an hourly rate, the conversion doesn't seem necessary in the first place, so this change removes the multiplier.

### DIFF
--- a/app/controllers/price_policies_controller.rb
+++ b/app/controllers/price_policies_controller.rb
@@ -101,13 +101,11 @@ class PricePoliciesController < ApplicationController
     @price_policies = @product.price_policies.for_date(@start_date)
     raise ActiveRecord::RecordNotFound if @price_policies.none?
 
-    if ActiveRecord::Base.transaction do
-        raise ActiveRecord::Rollback unless @price_policies.all?(&:destroy)
-        flash[:notice] = "Price Rules were successfully removed"
-        redirect_to facility_product_price_policies_path
-      end
+    if PricePolicyUpdater.destroy_all!(@price_policies, @start_date)
+      flash[:notice] = I18n.t("controllers.price_policies.destroy.success")
+      redirect_to facility_product_price_policies_path
     else
-      flash[:error] = "An error was encountered while trying to remove the Price Rules"
+      flash[:error] = I18n.t("controllers.price_policies.destroy.failure")
       redirect_to facility_product_price_policies_path
     end
   end
--- a/app/controllers/price_policies_controller.rb
+++ b/app/controllers/price_policies_controller.rb
@@ -71,15 +71,13 @@ class PricePoliciesController < ApplicationController
   # DELETE /facilities/:facility_id/{product_type}/:product_id/price_policies/:id
   def destroy
     return remove_active_policy_warning if @start_date <= Date.today
-    load_price_policies_for_start_date!
 
-    if PricePolicyUpdater.destroy_all!(@price_policies, @start_date)
+    if PricePolicyUpdater.destroy_all_for_product!(@product, @start_date)
       flash[:notice] = I18n.t("controllers.price_policies.destroy.success")
-      redirect_to facility_product_price_policies_path
     else
       flash[:error] = I18n.t("controllers.price_policies.destroy.failure")
-      redirect_to facility_product_price_policies_path
     end
+    redirect_to facility_product_price_policies_path
   end
 
   private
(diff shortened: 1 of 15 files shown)


app/controllers/price_policies_controller.rb
```diff
@@ -6,193 +6,176 @@ class PricePoliciesController < ApplicationController
...
-    render 'price_policies/edit'
+    render "price_policies/edit"
   end
 
-  # PUT /price_policies/1
-  def update
-    @start_date = start_date_from_params
+  # DELETE /facilities/:facility_id/{product_type}/:product_id/price_policies/:id
+  def destroy
+    return remove_active_policy_warning if @start_date <= Date.today
```
jhanggi: Can we call this method something like `remove_active_policy_warning_and_redirect` or something at implies that it does everything that needs to happen in a response?

app/services/price_policy_updater.rb
```diff
@@ -0,0 +1,54 @@
...
+  def initialize(price_policies, start_date, expire_date = nil, params = nil)
+    @price_policies = price_policies
+    @start_date = start_date
+    @expire_date = expire_date
+    @params = params
+  end
+
+  def destroy_all!
+    ActiveRecord::Base.transaction do
+      @price_policies.all?(&:destroy) || raise(ActiveRecord::Rollback)
```
jhanggi: Is `@price_policies` an array at this point, or is there a reason we can't use `destroy_all`?

app/controllers/price_policies_controller.rb
```diff
@@ -6,193 +6,176 @@ class PricePoliciesController < ApplicationController
...
   end
 
-  # DELETE /price_policies/2010-01-01
-  def destroy
-    @start_date     = start_date_from_params
-    unless @start_date > Date.today
-      # force the user to really think about what they're doing, but tell them how to do it if they really want.
-      flash[:error] = I18n.t('controllers.price_policies.errors.remove_active_policy')
-      return redirect_to facility_product_price_policies_path
+  def base_new_policies_on_most_recent
```
jhanggi: This seems like something that could be moved into `PricePolicyBuilder`.

app/controllers/price_policies_controller.rb
```diff
@@ -6,193 +6,176 @@ class PricePoliciesController < ApplicationController
...
+    @max_expire_date = PricePolicy.generate_expire_date(@start_date)
+  end
 
   def start_date_from_params
-    start_date=params[:id] || params[:start_date]
-    return unless start_date
-    format=start_date.include?('/') ? "%m/%d/%Y" : "%Y-%m-%d"
-    Date.strptime(start_date, format)
+    start_date = params[:id] || params[:start_date] || return
+    format = start_date.include?("/") ? :usa : :ymd
```
jhanggi: Where does the `ymd` side happen?

drench: Good point: I don't think this is generally possible given the UI will not even let me type a `-` character into that field. I'll take it out.

drench: Strike that: if the `start_date` comes from `params[:id]`, then it's in YYYY-MM-DD format. This still could be written more clearly.

drench:
Ready for another look.

app/controllers/price_policies_controller.rb
```diff
@@ -6,193 +6,152 @@ class PricePoliciesController < ApplicationController
...
 
+  def product_var
+    @product_var ||= model_name.gsub(/PricePolicy\z/, "").downcase
+  end
 
-  #
-  # Override CanCan's find -- it won't properly search by zoned date
-  def init_price_policy
-    product_var="@#{model_name.gsub('PricePolicy', '').downcase}"
+  def remove_active_policy_warning_and_redirect
```
jhanggi: This method name implies that it's removing something. Maybe just prepend with `flash_` or something?

jhanggi:
The controller is still pretty complicated and confusing, but this is definitely an improvement. I really hate that policies are keyed by the start date. :+1:

PR #341, opened 2015-09-02
[#119594] Order users and accounts in Price Groups views

jhanggi:
There was no ordering applied previously. This also fixes the query so
it doesn’t load all of the members into an array.

### DIFF
--- a/app/controllers/price_groups_controller.rb
+++ b/app/controllers/price_groups_controller.rb
@@ -32,7 +32,11 @@ class PriceGroupsController < ApplicationController
       raise ActiveRecord::RecordNotFound
     end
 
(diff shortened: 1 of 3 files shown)


app/controllers/price_groups_controller.rb
```diff
@@ -95,8 +103,8 @@ def destroy
 
   private
 
-  def get_paginated_group_members(method)
-    @price_group.public_send(method).paginate(page: params[:page], per_page: 10)
+  def paginate(relation)
+    relation.paginate(page: params[:page], per_page: 10)
```
drench: So much nicer.

drench:
:+1:

PR #342, opened 2015-09-03
[#119628] Do not validate media type for DownloadableFiles

jhanggi:
Paperclip gets really fussy when you try to upload a file and the MIME
type it detects does not match the file extension. This was happening
occasionally on NU for order uploads, and started happening for UIC.

This disables paperclip’s validations.

Already done in NU on https://github.com/tablexi/nucore-nu/pull/79

I’ve done this as a hotfix on UIC to verify that this fixes their issue.


drench:
:+1:

PR #344, opened 2015-09-04
Convert accessory scaling decorators to use SimpleDelegator

jhanggi:
I was getting warnings from rspec 2.99 when doing assertions like
`expect(order_detail).to be_complete`, but felt like this was a
significant enough change to warrant its own PR.

> Matching with be_persisted on an object that doesn't respond to
`persisted?` is deprecated. Use `respond_to_missing?` or `respond_to?`
on your object instead.

Using SimpleDelegator seems to be the simplest solution because it will
handle `respond_to?`.


drench:
`SimpleDelegator` > `method_missing` for sure.
:+1: (after the Circle issue's cleared up)

PR #345, opened 2015-09-04
[#98770] Change statement PDF filename

jhanggi:
This is pulled up from https://github.com/tablexi/nucore-nu/pull/88

It adds additional information like the facility into statement PDF filenames and allows overriding via locales.


drench:
:+1: after clearing up the Circle issue

PR #346, opened 2015-09-04
Use alphabetic sequnce for PriceGroup name

jhanggi:
We were getting an odd test failure where we’re looking for price
groups in a specific order. It was failing when that order spanned
“Price Group 999” to “Price Group 1000” because “1000” comes before
“999” alphabetically.


drench:
:+1:

PR #347, opened 2015-09-04
Upgrade rspec to 2.99

jhanggi:
This is a transition to 3.X. After running the tests we should see no deprecation errors.

There will be an associated PR on NU as well.


jhanggi:
Transpec didn't hit all the files.

drench:
How do you feel about `be_truthy`/`be_falsy` vs. `be true`/`be false`? I did a quick `ruby -pie` search & replace and it remained green except one `destroy` check (which I changed): https://github.com/tablexi/nucore-open/compare/rspec-upgrade-299...rspec-strict-true-false?expand=1

Other than that question, :+1:.

jhanggi:
I don't have a strong preference either way. I'll go ahead and pull it in to this branch.

jhanggi:
I'm also going to leave this open while I get the NU branch settled up.

PR #349, opened 2015-09-09
[#60973] Show admin notifications when in admin context

drench:
When problem/disputed orders or training requests exist, these notifications should only appear in the nav tab when the user is in a manager context.

Merged order notifications should still appear in the nav tab in any context.

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -1,5 +1,5 @@
 class NotificationsController < ApplicationController
-  admin_tab     :all
+  customer_tab  :all
   before_filter :authenticate_user!
   before_filter :check_acting_as
   before_filter :check_notifications
--- a/app/models/message_summarizer.rb
+++ b/app/models/message_summarizer.rb
@@ -7,6 +7,10 @@ class MessageSummarizer
     @controller = controller
   end
 
+  def has_visible_tab?
+    manager_context? || notifications.any?
+  end
+
   def message_count
     @message_count ||= message_summaries.sum(&:count)
   end
@@ -16,6 +20,10 @@ class MessageSummarizer
   end
 
   private
+
+  def manager_context?
+    @controller.current_facility && @controller.admin_tab?
+  end
 
   def notifications
     @notifications ||= NotificationsSummary.new(@controller)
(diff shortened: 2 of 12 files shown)


jhanggi:
Did you hear back from Aaron? I would vastly prefer to either show _none_ of the notifications in the non-admin context or _all_ of the notifications. From an interface perspective, it feels inconsistent, and from a code perspective Doing the `manager_context?` check within each individual notification type kind of bugs me.

jhanggi:
If we do have to go that route, what would you think of making what's inside the message types `show_outside_of_manager_context?` that defaults to false and then you just override it in the merge order notifications case, since that's the exception to the rule.

drench:
I don't have a reply yet. I wrote this before realizing we might want all-or-nothing and asking for clarification. I'm happy to get feedback on in the mean time.

jhanggi:
Let's wait until we get feedback. I'd rather go all or nothing.

drench:
Updated to function as discussed:
* If the user has merged order notifications should appear in all contexts, as they always have.
* Other types of notifications should appear only when in a manager context.
* In a non-manager context, the "Notices" tab will only appear in the top nav if there are merged order notifications.
* When in manager context, the "Notices" tab should always appear in the nav; if there are no notifications, it should read "Notices (0)".

jhanggi:
Found an issue. If I set up an order that has a merge order, if I leave the manager context, I get an error. Looks like if there are Notifications, we try to look up the other messages as well.

```
NoMethodError - undefined method `order_details_in_dispute' for nil:NilClass:
  app/models/message_summarizer/order_details_in_dispute_summary.rb:9:in `get_count'
  app/models/message_summarizer/message_summary.rb:13:in `count'
  app/models/message_summarizer/message_summary.rb:9:in `any?'
  app/models/message_summarizer.rb:60:in `message_summaries'
  app/models/message_summarizer.rb:29:in `message_count'
  app/models/message_summarizer.rb:11:in `messages?'
  app/views/shared/_message_summary.html.haml:4:in `_app_views_shared__message_summary_html_haml__3145284826613540623_70255182056180'
```

jhanggi:
Is this ready for another look?

drench:
It's ready for another look. The only issue I have is I'm ambivalent about the `#visible?` method, at least the name. For example, training requests could be "visible" but if there are none, the link won't appear in the drop-down. This is the behavior we want, but is `#visible?` the right name?

drench:
:+1: for 1130207dae960d5c7dd4dba57db9e9941c0799f9

PR #350, opened 2015-09-10
[#94618] Show most recently used facilities near the top of the index

drench:
If a user is logged in and has made at least one order, the facilities index page will list the user's five most recently ordered-from facilities at the top, followed by a complete facility listing:

screen shot 2015-09-10 at 1 51 43 pm

If there user is not logged in or the user has made no orders, the page will look as before, with all facilities listed.

This includes template cleanup/i18n and conversion of `@facilities` from an `Array` into a relation.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -157,4 +157,11 @@ class User < ActiveRecord::Base
   def to_s
     full_name
   end
+
+  def recently_used_facilities(limit = 5)
+    @recently_used_facilities ||= Hash.new do |hash, key|
+      hash[key] = Facility.where(id: orders.pluck(:facility_id).uniq).limit(limit)
+    end
+    @recently_used_facilities[limit]
+  end
 end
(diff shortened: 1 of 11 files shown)


app/models/user.rb
```diff
@@ -157,4 +157,11 @@ def full_name
   def to_s
     full_name
   end
+
+  def recently_used_facilities(limit = 5)
+    @recently_used_facilities ||= Hash.new do |hash, key|
+      hash[key] = Facility.where(id: orders.pluck(:facility_id).uniq).limit(limit)
```
drench: I imagine there's a better way to write this. Open to suggestions.

drench:
After reducing the text size of the facility links, here's what it would look like without the top 5:
screen shot 2015-09-10 at 4 27 24 pm

And here's what it would look like with the top 5:
screen shot 2015-09-10 at 4 27 40 pm

PR #351, opened 2015-09-13
[#87870] Add notes to mailed receipts

drench:
If the user includes notes when ordering on behalf of another user, this should include them in the emailed receipts, both HTML and text versions.

This patch also includes some overdue haml cleanup, localization and specs.


app/views/notifier/order_receipt.html.haml
```diff
@@ -1,29 +1,30 @@
...
 %table.table.table-striped.table-hover
   %thead
     %tr
-      %th Order #
-      %th Product
-      %th Quantity
-      %th Status
+      %th= t(".th.order_number")
+      %th= t(".th.product")
+      %th= t(".th.quantity")
```
jhanggi: Would some of these be better as `OrderDetail.human_attribute_name(:quantity)` as opposed to specifically targeting this view?

app/views/notifier/order_receipt.html.haml
```diff
@@ -31,19 +32,23 @@
         %td= od
         - if od.reservation
           %td
-            = ((od.bundle ? "#{h od.bundle} &mdash; " : '') + h(od.product)).html_safe
+            = ((od.bundle ? "#{od.bundle} &mdash; " : "") + h(od.product)).html_safe
```
jhanggi: What is this `h()` doing?

config/locales/en.yml
```diff
@@ -796,7 +797,22 @@ en:
...
+        ordered_by: Ordered By
+      order_details: Order Details
       outro: "All prices are estimates. Actual cost is assigned when the order is complete."
+      reservation: Reservation
+      th:
+        order_number: "Order #"
+        product: Product
+        quantity: Quantity
+        status: Status
+        price: "%{text}Price"
```
jhanggi: Why is there no space here?

drench: This is for the text version of the receipt. The way it was working was the "text" piece coming out of `order_estimated_or_actual_text` might be a prefix like "Estimated", or an empty string. The extra space wasn't so much a problem in the html version (though I like that it's gone in this patch).

app/views/notifier/order_receipt.html.haml
drench: This is a confusing one. Looking again, I think what it needs to be is:
`(od.bundle ? "#{h od.bundle} &mdash; ".html_safe : od.product)`

I'll update and test later when I'm off the train.

jhanggi: There's a method in ApplicationHelper, `order_detail_description` which seems to have the same goal. We might be able to reuse that (although it could probably use a little cleanup as well).

app/models/product.rb
```diff
@@ -124,7 +124,7 @@ def to_param
   end
 
   def to_s
-    name.present? ? name.html_safe : ''
+    name.presence || ""
```
drench: The `#html_safe` was introduced here 273e94102ea3b40ae38bb5b838c526667454c383 to address double-escaping in a receipt view that doesn't seem to exist anymore.

config/locales/en.yml
```diff
@@ -796,6 +797,8 @@ en:
     order_receipt:
       subject: "NU Core Order Receipt"
       intro: "Thank you for your order on the NU Core website."
+      label:
+        ordered_at: Ordered at
```
drench: This is here to preserve the wording on the mailed receipts before I localized them. Everywhere else it looks like `OrderDetail#ordered_at` humanizes to "Ordered Date".

drench:

drench:
Merge conflicts resolved.

PR #352, opened 2015-09-14
[#120317] Restore in-progress reservation update

drench:
This should correct the issue where `ReservatrionsController#update` on a started `Reservation` would fail due to the existing start date/time parameters being blank.

The key difference was restoring the start date/time params when `!@reservation.admin_editable? || !@reservation.reserve_start_at_editable?` rather than `!@reservation.admin_editable? && !@reservation.reserve_start_at_editable?`.


jhanggi:
:+1:

PR #354, opened 2015-09-18
[#121201] Hide Dispute tab if there is no review period

jhanggi:
UIC does not have a review period, so there is no way to dispute, so there's no need for the tab.


drench:
:+1: though take or leave a couple potential `TODO`s: I18n the labels, and maybe changing the `current_user.manager_of?` into `Ability` checks (we have a `:show_problems` ability, but not one for disputed orders yet).

PR #355, opened 2015-09-18
[#60973] Fix color of "Notices (0)"

jhanggi:
It's not obvious with the open color scheme, but on NU, the text for
"Notices (0)" in the header was gray as opposed to the white that it's
supposed to be.

nucore-notice


drench:
:+1:

PR #356, opened 2015-09-18
[#102970] Add Whenever to manage cronjobs

jhanggi:
I feel like it would be good to have some documentation about the cron jobs, or even better, how to make sure you have everything running you need in production. I feel like that may be job for the next implementation.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -10,6 +10,7 @@ gem 'foreigner',        '1.1.1'
 
 ## deployment
 gem 'capistrano',       '2.15.4'
+gem 'whenever', require: false
 
 ## auth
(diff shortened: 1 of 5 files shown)


drench:
I'm going to blame the spec failure on https://github.com/tablexi/nucore-open/blob/whenever/config/application.rb#L21
Do we really want to autoload `/config`? The tests and the app itself seem to work all right without it…

jhanggi:
Removed `/config` from the autoload path, and it seems like we're all good now. PR soon for the NU branch, and one for the UIC branch.

drench:
:+1:

PR #357, opened 2015-09-21
[#120379] Improve performance of all transaction view

jhanggi:
The loading of the search options, especially when running cross-facility
has gotten fairly slow. This makes some changes to how some of those
options are loaded to speed things up.

While testing against UIC, we discovered they're running MySQL 5.1, so some
of the optimizations aren't as good as they are on 5.6+.

##### Cross Facility view:

MySQL version | Original | Optimized
----|----|----
5.1 | ~2.75s | 1.4s
5.6 | ~1.5s  | 1s

##### Single facility:

MySQL version | Original | Optimized
----|----|----
5.1 | ~1.5s | 1.4s
5.6 | .9s | .7s

And from testing on staging (which is 5.1), I saw similar scale improvements:

Facility view | Original | Optimized
----|----|----
Cross | 2s | .75ms
Single | .8s | .75ms

I also tested against Oracle for single facilities, and saw slight, but
similar improvements.

### DIFF
--- a/app/inputs/transaction_chosen_input.rb
+++ b/app/inputs/transaction_chosen_input.rb
@@ -5,10 +5,10 @@ class TransactionChosenInput < SimpleForm::Inputs::Base #CollectionSelectInput
     options[:label_method] ||= :name
     options[:value_method] ||= :id
 
-    search_fields[attribute_name] = [collection_items.first.send(options[:value_method].to_sym)] if collection_items.size == 1
+    search_fields[attribute_name] = [collection_items.first.send(options[:value_method].to_sym)] if collection_items.to_a.size == 1
 
     select_options = {:multiple => true, :"data-placeholder" => placeholder_label }
-    select_options.merge!({:disabled => :disabled}) unless collection_items && collection_items.size > 1
+    select_options.merge!({:disabled => :disabled}) unless collection_items.to_a.size > 1
 
     template.select_tag(attribute_name, option_data, select_options).html_safe
   end
--- a/app/inputs/transaction_chosen_input.rb
+++ b/app/inputs/transaction_chosen_input.rb
@@ -26,7 +26,7 @@ class TransactionChosenInput < SimpleForm::Inputs::Base #CollectionSelectInput
   private
 
   def collection_items
-    template.instance_variable_get("@#{attribute_name}").to_a
+    template.instance_variable_get("@#{attribute_name}")
   end
 
   def model_label
--- a/app/inputs/transaction_chosen_input.rb
+++ b/app/inputs/transaction_chosen_input.rb
@@ -5,10 +5,15 @@ class TransactionChosenInput < SimpleForm::Inputs::Base #CollectionSelectInput
     options[:label_method] ||= :name
     options[:value_method] ||= :id
 
-    search_fields[attribute_name] = [collection_items.first.send(options[:value_method].to_sym)] if collection_items.to_a.size == 1
+    # By calling `to_a` here, it fetches the AR::Relation into objects. Since that will
+    # happen later anyways, by doing it up front, we prevent extra `count(*)` queries,
+    # which were slowing down the page.
+    collection_size = collection_items.to_a.size
+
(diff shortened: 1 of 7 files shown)


jhanggi:
I've cherry-picked this into a UIC branch that I released to their staging environment for testing. I'd still like to hear any feedback or at least get a thumbs up.

app/inputs/transaction_chosen_input.rb
```diff
@@ -5,10 +5,10 @@ def input
     options[:label_method] ||= :name
     options[:value_method] ||= :id
 
-    search_fields[attribute_name] = [collection_items.first.send(options[:value_method].to_sym)] if collection_items.size == 1
+    search_fields[attribute_name] = [collection_items.first.send(options[:value_method].to_sym)] if collection_items.to_a.size == 1
 
     select_options = {:multiple => true, :"data-placeholder" => placeholder_label }
-    select_options.merge!({:disabled => :disabled}) unless collection_items && collection_items.size > 1
+    select_options.merge!({:disabled => :disabled}) unless collection_items.to_a.size > 1
```
drench: Would we gain anything by extracting `collection_items.to_a.size`, since it's in here twice?

drench: Even if there's no performance benefit, I'd like to see it extracted if only to comment on why it converts to `Array`; I could easily see myself forgetting in the future why it's there and removing that `#to_a` while refactoring.

app/models/account.rb
```diff
@@ -120,9 +121,9 @@ def <=>(obj)
     account_number <=> obj.account_number
   end
 
-  def owner_user
-    self.owner.user if owner
-  end
+  # def owner_user
+  #   self.owner.user if owner
+  # end
```
drench: Remove before merge

app/models/order_detail.rb
```diff
@@ -108,11 +108,12 @@ def self.for_facility(facility)
...
-  def self.for_facility_id(facility_id=nil)
-    details = joins(:order)
+  def self.for_facility_id(facility_id = nil)
+    details = scoped
 
-    unless facility_id.nil?
-      details = details.where(:orders => { :facility_id => facility_id})
+    if facility_id.present?
+      details = joins(:order)
+      details = details.where(orders: { :facility_id => facility_id })
```
drench: Chain?

drench:
Just a few minor comments. There are style update possibilities here and there too, but as far as query changes go it looks all right to me.

app/inputs/transaction_chosen_input.rb
jhanggi: We don't really get any added bonus for extracting it; the important part is that it's loading the AR::Relation. But I definitely see your point. I'll extract it.

app/models/account.rb
jhanggi: Good catch.

jhanggi:
Updated.

app/inputs/transaction_chosen_input.rb
```diff
@@ -21,7 +26,7 @@ def label
   private
 
   def collection_items
-    template.instance_variable_get("@#{attribute_name}")
+    template.instance_variable_get("@#{attribute_name}").to_a
```
drench: Does the `#to_a` need to be here given line 11? Or does it maybe not need to be on line 11, and be here instead.

drench:
One new question.

app/inputs/transaction_chosen_input.rb
jhanggi: I was going back and forth by adding the `to_a` here, and then using `collection_items.size` throughout the `input` method instead of the `collection_size` variable. I eventually settled on leaving it as a variable, but guess I forgot to remove this one.

jhanggi:
Updated to remove the extra `to_a`

drench:
:+1: (pending Circle result)

jhanggi:
There was an odd NufsAccount ordering issue that appears to stem from alphabetization of descriptions, so I changed it from going a..z...aa...zz to aaaaaaa...aaaaaaaz...aaaaaaza...zzzzzzz. The tests are now passing.

drench:
:+1:

PR #358, opened 2015-09-24
Remove --silent from rake tasks created via whenever/cron

jhanggi:
This will prevent the —silent option from being included in cron tasks created by whenever. That makes sure that any output is sent via email as configured within the crontab.


drench:
:+1:

PR #359, opened 2015-09-24
[#60973] Update Facility staff-related permissions

drench:
This should bring the Notices drop-down permissions in line with what users may access through other parts of the UI.

The notices drop-down uses `Ability` permissions to determine what to show the user, while the corresponding links in the UI are *not* using `Ability` checks. So we ended up with inconsistencies, like hiding the problem reservations tab from Staff and Senior Staff, while showing problem reservations in the notices drop-down.

This came down to replacing the `:manage` (anything goes) permissions to `Reservation` and `Order` objects for Staff (that is, `#operator_of?(current_facility)`) with specific grants.

### DIFF
--- a/app/views/admin/shared/_tabnav_order.html.haml
+++ b/app/views/admin/shared/_tabnav_order.html.haml
@@ -2,7 +2,7 @@
   = display_tab "New / In Process", facility_orders_path, action: :index
   - if current_ability.can?(:show_problems, Order)
     = display_tab "Problem Orders", show_problems_facility_orders_path, action: :show_problems
-  - if SettingsHelper.has_review_period? && current_ability.can?(:manage, Facility)
+  - if SettingsHelper.has_review_period? && current_ability.can?(:disputed, Order)
     = display_tab "Disputed", disputed_facility_orders_path, action: :disputed
   - if current_user.manager_of?(current_facility)
     = tab "Add New", new_facility_order_import_path(current_facility), (controller.controller_name == 'facility_orders_import')
--- a/app/views/admin/shared/_tabnav_order.html.haml
+++ b/app/views/admin/shared/_tabnav_order.html.haml
@@ -1,6 +1,8 @@
 %ul.nav.nav-tabs.js-tab-counts
   = display_tab "New / In Process", facility_orders_path, action: :index
-  - if current_user.manager_of?(current_facility)
+  - if current_ability.can?(:show_problems, Order)
     = display_tab "Problem Orders", show_problems_facility_orders_path, action: :show_problems
-    = display_tab "Disputed", disputed_facility_orders_path, action: :disputed if SettingsHelper.has_review_period?
+  - if SettingsHelper.has_review_period? && current_ability.can?(:manage, Facility)
+    = display_tab "Disputed", disputed_facility_orders_path, action: :disputed
+  - if current_user.manager_of?(current_facility)
     = tab "Add New", new_facility_order_import_path(current_facility), (controller.controller_name == 'facility_orders_import')
--- a/app/views/admin/shared/_tabnav_reservation.html.haml
+++ b/app/views/admin/shared/_tabnav_reservation.html.haml
@@ -3,5 +3,5 @@
   = display_tab "New / In Process", facility_reservations_path, action: :index
   - if current_ability.can?(:show_problems, Reservation)
     = display_tab "Problem Reservations", show_problems_facility_reservations_path, action: :show_problems
-  - if SettingsHelper.has_review_period? && current_ability.can?(:manage, Facility)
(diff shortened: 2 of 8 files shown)


lib/ability.rb
```diff
@@ -86,12 +113,13 @@ def initialize(user, resource, controller)
 
       if user.facility_director_of?(resource)
         can [ :activate, :deactivate ], ExternalService
+        can :disputed, [Order, Reservation]
```
jhanggi: You've added this ability, but it looks like you're testing against `:manage`. Should we be using `disputed`?

app/views/admin/shared/_tabnav_reservation.html.haml
```diff
@@ -1,6 +1,7 @@
 %ul.nav.nav-tabs
   = tab "Daily View", timeline_facility_reservations_path(current_facility), tab_active?(:timeline)
   = display_tab "New / In Process", facility_reservations_path, action: :index
-  - if current_user.manager_of?(current_facility)
+  - if current_ability.can?(:show_problems, Reservation)
     = display_tab "Problem Reservations", show_problems_facility_reservations_path, action: :show_problems
-    = display_tab "Disputed", disputed_facility_reservations_path, action: :disputed if SettingsHelper.has_review_period?
+  - if SettingsHelper.has_review_period? && current_ability.can?(:manage, Facility)
```
jhanggi: Do we need this same check in `_tabnav_order.html.haml`?

drench: Good catch. That's the pattern I need to remember: if `Reservation` needs a change, `Order` probably does too, and vice-versa.

lib/ability.rb
drench: Yes. I had gone with `:manage` instead of `:disputed` when it looked like it was redundant in 6ec2da2, but I just changed it to check `:disputed` (including in the drop-down, though only against `Order`).

app/views/admin/shared/_tabnav_order.html.haml
```diff
@@ -1,6 +1,8 @@
 %ul.nav.nav-tabs.js-tab-counts
   = display_tab "New / In Process", facility_orders_path, action: :index
-  - if current_user.manager_of?(current_facility)
+  - if current_ability.can?(:show_problems, Order)
     = display_tab "Problem Orders", show_problems_facility_orders_path, action: :show_problems
-    = display_tab "Disputed", disputed_facility_orders_path, action: :disputed if SettingsHelper.has_review_period?
+  - if SettingsHelper.has_review_period? && current_ability.can?(:disputed, Order)
+    = display_tab "Disputed", disputed_facility_orders_path, action: :disputed
+  - if current_user.manager_of?(current_facility)
```
drench: I left this with the existing non-`Ability` check, but I consider it a `TODO`: the `OrderImportsController` authorizes against `Order`, so we'll have to think about what restricting (or making an `authorize_resource` exception for) `#new` would mean here.

drench:
Updated & ready for another look.

spec/lib/ability_spec.rb
```diff
@@ -110,6 +119,7 @@
 
     it_behaves_like "it can manage training requests"
     it_behaves_like "it can read notifications"
+    it_behaves_like "it cannot access problem reservations"
```
jhanggi: This seems weird. A facility director should be able to access problem reservations, and it looks like they can through the interface, so I'm not sure what this test is doing.

lib/ability.rb
```diff
@@ -63,10 +63,37 @@ def initialize(user, resource, controller)
...
+        can [
+          :assign_price_policies_to_problem_orders,
+          :create,
+          :edit,
+          :edit_admin,
+          :index,
+          :show,
+          :timeline,
+          :update,
+          :update_admin,
```
jhanggi: Looks like you're missing `:tab_counts` for Reservations.

spec/lib/ability_spec.rb
drench: This one's wrong. The spec wasn't using a `Facility` as its subject resource, so the grants (in the test) weren't right. The next commit I'm pushing up should take care of it (and reflect what we see in the UI for this case).

lib/ability.rb
drench: Ah: missed it because the spec isn't there (but it is for Orders).

drench:
Updated. I did more manual testing and have not run into any more inconsistencies between what appears in the Notices drop-down and what's visible and accessible through the tab nav.

* Facility Staff and Facility Senior Staff have access to training requests
* Facility Directors and admins have the above Staff abilities as well as access to problem/disputed reservations and orders
* Unprivileged users see none of the above

jhanggi:
:+1:

PR #360, opened 2015-09-24
Upgrade devise to 3.5

jhanggi:
In anticipation of Rails upgrade


config/settings.local.yml.template
```diff
@@ -1,5 +1,8 @@
 # This will override what is in settings.yml, which is committed to version control.
 
+devise:
+  secret_key:
```
drench: In a comment here, or maybe better in `README.md`, we should include a note about setting `secret_key`or the app won't start.

jhanggi: Actually, what do you think of adding something to `settings/development.yml` like I did for `test.yml`. That way it's one less step you need to deal with when doing initial setup, but something you would still need to configure when doing a stage/prod deploy?

drench: Sounds good. The less complicated the setup the better.

jhanggi:
Updated. Not going to merge this is until the next stage release. While I think the odds of it creating problems are very slim, if it does create a problem, it might be tricky.

drench:
:+1:

PR #361, opened 2015-09-25
[#121286] Only run cron jobs on a single server

jhanggi:
It feels a little weird to put capistrano-based functionality in the
open branch, but it will still work fine without capistrano, and `:db`
is the standard role for a single server (usually the one used for
running db:migrate).

PR for NU coming soon.


drench:
:+1: if https://github.com/tablexi/nucore-nu/issues/95 is :+1:

PR #363, opened 2015-09-25
Remove unneeded precompile asset files

jhanggi:
We’ve been running without these in NU for a while.


drench:
:+1:
I'm sure there are more things like this to uncover. Ideally every time we touch a file in any *Core fork we ought to diff it against the others.

PR #364, opened 2015-09-28
[#121327] Facility Admin/Directors were unable to delete admin reservations

jhanggi:
Related to refactor of abilities in #359 and #362.


drench:
:+1:

PR #366, opened 2015-09-29
[#121342] Restore external users' ability to log in

jhanggi:
The recent upgrade of the Devise LDAP adapter changed its behavior so
external users would no longer fall back to the
database_authenticatable strategy.


drench:
:+1:

PR #367, opened 2015-09-29
Clean up environments/production.rb

jhanggi:
To be followed by cleanup to NU & UIC’s stage and prod environment
files.


drench:
:+1:

PR #368, opened 2015-09-30
[#94618] Homepage fixes

jhanggi:
* Return facilities list to alphabetical order
* Use acting_user as source for most recent facilities
* Order recent facilities by name
* Ensure orders are searched by ordered_at for use in recent facilities
* Add index for the homepage ordering

### DIFF
--- a/spec/factories/facilities.rb
+++ b/spec/factories/facilities.rb
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :facility do
-    sequence(:name) { |n| "Facility#{n}" }
+    sequence(:name, "AAAAAAAA") { |n| "Facility#{n}" }
     sequence(:email) { |n| "facility-#{n}@example.com" }
     sequence(:abbreviation) { |n| "FA#{n}" }
     short_description 'Short Description'
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -23,12 +23,12 @@ class FacilitiesController < ApplicationController
 
   def set_recently_used_facilities
     @recently_used_facilities =
-      current_user.present? && current_user.recently_used_facilities.sorted
+      acting_user.present? && acting_user.recently_used_facilities.sorted
   end
 
   # GET /facilities
   def index
-    @facilities = Facility.active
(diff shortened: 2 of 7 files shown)


spec/factories/facilities.rb
```diff
@@ -1,6 +1,6 @@
 FactoryGirl.define do
   factory :facility do
-    sequence(:name) { |n| "Facility#{n}" }
+    sequence(:name, "AAAAAAAA") { |n| "Facility#{n}" }
```
drench: We're doing this enough that I wish we had something like `alphabetical_sequence(:name)`.

drench:
:+1:

mlineen:
Why not use `{ |n| "Facility#{n.to_s.rjust(10,'0')}" }`

jhanggi:
I think that's uglier to read, but we could use `sequence(:name, "00000000001")`, which does lead to easier to grok names while debugging.

jhanggi:
And I was playing around, and using that as the starting point, I have an idea for a PR for FactoryGirl to natively (or via monkey patch) support `alphabetic_sequences`.

jhanggi:
Not quite ready for PR: https://github.com/tablexi/nucore-open/compare/master...alphabetic_sequence

mlineen:
Does `sequence(:name, "Facility_0001")` without a block effectively do what we want? 
```
> 'Facility_0001'.next
 => "Facility_0002"
```

jhanggi:
Unfortunately, not.

```
> "Facility_9999".next
 => "Facility_10000"
```

The only thing I can think of that might work is a custom iterator that adds the extra character on the right of the string, e.g.:
```
8, 9, 90, 91, .., 98, 99, 990
A, B, ..., Y, Z, ZA, .., ZY, ZZ, ZZA
```

I think that's almost more confusing if you're in debugging, especially in the numerical version.

I'm open to other suggestions. This has come up multiple times across multiple projects in the past couple weeks, so I do have a vested interest in finding the right solution.

mlineen:
I guess what I was advocating is in tests using enough leading zeros so that you never run into the alphabetical issue which is basically what your code is doing, but I'm not creating the safety net of warning if you wrap.

PR #369, opened 2015-10-01
Remove "Disputed Orders" from billing tab sidenav if no review period

jhanggi:


drench:
Interesting branch name. Was that some tab completion gone weird? :+1:

jhanggi:
Yeah. Didn't catch it until I had already issued the PR. oops.

PR #370, opened 2015-10-02
[#121396] Sending Notifications for more than 1000 OD causes Oracle error

jhanggi:
Oracle does not allow more than 1000 items in a `WHERE IN` clause. This
breaks it up into 1000 item AR::Relations and loops over those as
necessary.

Unfortunately, this makes the code more complicated, but it's about as simple as I can make it while dealing with this limitation. It will still use one giant AR relation for MySQL.

Once this goes through CI (I have the branch on NU as well), I'd say we can remove the 1000 item spec (which is obviously slow).

### DIFF
--- a/spec/app_support/notification_sender_spec.rb
+++ b/spec/app_support/notification_sender_spec.rb
@@ -1,0 +1,57 @@
+require "rails_helper"
+
+RSpec.describe NotificationSender, :aggregate_failures do
+  let(:user) { create(:user) }
+
+  let(:facility) { create(:facility) }
+
+  let(:facility_account) do
+    facility.facility_accounts.create(attributes_for(:facility_account))
+  end
+
+  let(:item) do
+    facility
+    .items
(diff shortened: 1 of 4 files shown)


spec/app_support/notification_sender_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    facility.facility_accounts.create(attributes_for(:facility_account))
+  end
+
+  let(:item) do
+    facility
+    .items
+    .create(attributes_for(:item, facility_account_id: facility_account.id))
+  end
+
+  let(:account) { create(:nufs_account, :account_users_attributes => account_users_attributes_hash(:user => user), :facility_id => facility.id) }
```
drench: Hashrockets. Also, considering how often we use this pattern, I feel like we could replace the `account_users_attributes:` with transient attributes (the `:setup_account` factory seems to be almost there).

drench:
One style comment but otherwise :+1:

PR #371, opened 2015-10-03
WIP refactor of common Product controller #show actions

drench:
This is the result of a major refactor/extraction of the `#show` actions for `Product` controllers I worked on with @danhodos and @yoon. Though it's been about 6 weeks since we did this, we haven't otherwise touched the code we extracted in `master` (maybe because it's kind of radioactive). In other words, despite its age, the logic shouldn't have changed in the mean time.

This may not be ready for a merge but I wanted to put this put out there before it gets stale and see if it's worth polishing up.

### DIFF
--- a/app/controllers/products_common_controller.rb
+++ b/app/controllers/products_common_controller.rb
@@ -41,60 +41,19 @@ class ProductsCommonController < ApplicationController
   # GET /facilities/:facility_id/(services|items|bundles)/:(service|item|bundle)_id
   # TODO InstrumentsController#show has a lot in common; refactor/extract/consolidate
   def show
-    assert_product_is_accessible!
-    add_to_cart = true
-    @login_required = false
+    @show_product = ShowProduct.new(self, @product, acting_user, session_user, acting_as?)
+    @add_to_cart = @show_product.able_to_add_to_cart?
+    @login_required = @show_product.login_required
+    @error = @show_product.error
 
-    # does the product have active price policies?
-    unless @product.available_for_purchase?
-      add_to_cart = false
-      @error = "not_available"
+    if @show_product.redirect.present?
+      flash[:notice] = @show_product.error if @show_product.error
+      redirect_to @show_product.redirect
+    else
+      flash.now[:notice] = @show_product.error if @show_product.error
+      @active_tab = "home"
+      render layout: "application"
     end
-
-    # is user logged in?
-    if add_to_cart && acting_user.blank?
-      @login_required = true
-      add_to_cart = false
-    end
-
-    # when ordering on behalf of, does the staff have permissions for this facility?
-    if add_to_cart && acting_as? && !session_user.operator_of?(@product.facility)
-      add_to_cart = false
-      @error = "not_authorized_acting_as"
-    end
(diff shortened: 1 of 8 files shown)


config/locales/en.models.yml
```diff
@@ -54,35 +54,35 @@ en:
...
           no_accounts: "Sorry, but we could not find a valid payment source that you can use to reserve this instrument"
-          no_price_groups: "You are not in a price group that may reserve instrument %{instrument_name}; please contact the facility."
+          not_in_price_group: "You are not in a price group that may reserve instrument %{name}; please contact the facility."
           not_authorized_to_order_on_behalf: You are not authorized to order instruments from this facility on behalf of a user.
           not_interval: "must be a multiple of the interval (%{reserve_interval})"
           max_less_than_min: "must be greater than or equal to minimum reservation minutes"
         item:
           not_available: This item is currently unavailable for purchase online.
-          requires_approval: This item requires approval to purchase; please contact the facility.
+          requires_approval_html: This item requires approval to purchase; please contact the facility.
```
jhanggi: Why the switch to `_html`?

app/models/show_product.rb
```diff
@@ -0,0 +1,125 @@
+class ShowProduct
+  attr_reader :acting_user, :controller, :error, :login_required, :redirect, :session_user
+
+  def initialize(controller, product, acting_user, session_user, acting_as)
```
jhanggi: Since you're passing a reference to the controller in, you could probably `delegate :acting_user, :session_user, :acting_as?, to: :controller`, and then you wouldn't need as many parameters and instance variables.

app/controllers/products_common_controller.rb
```diff
@@ -41,60 +41,19 @@ def index
...
-          return redirect_to facility_path(current_facility)
-        else
-          return redirect_to new_facility_product_training_request_path(current_facility, @product)
-        end
-      else
-        add_to_cart = false
-        @error = "requires_approval"
-      end
+    @show_product = ShowProduct.new(self, @product, acting_user, session_user, acting_as?)
+    @add_to_cart = @show_product.able_to_add_to_cart?
```
jhanggi: I'm guessing setting these as instance variables instead of just using them directly is to maintain compatibility with the specs?

jhanggi:
I think this is a really good first pass at cleaning these two controller actions up. I'm a fan of de-duping a lot of the logic. There's definitely more to be done. One thought I had is that the 404 logic and asserting that you're logged in could probably be left in the controller. I think the check against `archived?` could raise a RecordNotFound, perhaps moved into `init_product`? And once you're passed that check, it seems like everything then requires you to be logged in. Then you use this new class for the authorization, similar to what you're doing now.

Good work. I think it's worth pursuing this. I'd rather not merge this just yet, but if it looks like we'll be needing to change these areas, then we can merge it in to avoid conflicts, and then continue the refactoring in other PRs.

drench:
With the Rubocop sweep and Rails upgrade and other changes this has gotten behind. I'd like to revive it at some point but it should be a fresh PR.

PR #372, opened 2015-10-09
Update Guardfile to run javascript and vendor specs

drench:
* Guard should pick up changes under `/vendor`
* Guard should run JavaScript tests
* The specs related to `time_formatter.coffee` should now be in its expected location
* We're not using turnip or capybara (yet) so those sections are gone


jhanggi:
:+1:

PR #373, opened 2015-10-13
Ensure NotificationHelper tests get run with review period disabled

jhanggi:
If the review period feature is off, then the test fails.


drench:
:+1:

PR #375, opened 2015-10-15
Operator role cleanup

drench:
This removes the `#operator?` and `#manager?` methods from `Role`. The first wasn't correct and wasn't really necessary, while the other wasn't used at all. This replaces the one remaining use of `#operator?` with a helper method.

Also, senior staff did not have `:read` ability on `Notifications` yet staff did. The change isn't the scoping we discussed but that seemed to be leading toward larger `MessageSummarizer` changes (or at least its specs).


app/helpers/application_helper.rb
```diff
@@ -8,6 +8,14 @@ def app_name
     #NUCore.app_name
   end
 
+  def app_name_link_text
```
drench: Could this replace `#app_name`?

lib/ability.rb
```diff
@@ -70,6 +69,8 @@ def initialize(user, resource, controller)
           UserPriceGroupMember,
         ]
 
+        can :read, Notification
```
jhanggi: This actually changes the behavior of Notifications somewhat. I could probably talked in to the new behavior, but there are a couple inconsistencies.

It used to be if you had Notifications, then they were always visible even when you weren't in an admin context, because they're scoped to the user. Now we're only showing the Notifications while you're in a facility admin context, however we're still only scoping the notifications to the user, so if the user has admin rights on multiple facilities, and they have notifications on one facility, they'll still appear while in the context of another facility.

app/helpers/application_helper.rb
jhanggi: Probably. Although, I'd almost consider dropping `app_admin_name` altogether and just always say "NU Core" in the header. 

@nutodd, what do you think?

We noticed some inconsistencies with how senior staff was treated. They were not considered `operator?`, and actually neither were global admins. The only place this method is used is to determine whether to display "NU Core" or "NU Core Admin" in the header. I didn't even realize that was a thing since I'm 99% of the time logged in as an admin. Do you think we could drop the logic?

drench: @nutodd: I'm inclined to go with @jhanggi's suggestion to use "NU Core" throughout (it would certainly simplify the code) but I'd like to get your input before going ahead.

lib/ability.rb
```diff
@@ -34,7 +34,7 @@ def initialize(user, resource, controller)
     end
 
     can :list, Facility if user.facilities.size > 0 and controller.is_a?(FacilitiesController)
-    can :read, Notification if user.operator? || user.administrator?
+    can :read, Notification if user.notifications.active.any?
```
drench: While I left the facility operator `:read` in, this puts in the behavior we discussed earlier: if a user has notifications, the user may read them.

app/helpers/application_helper.rb
```diff
@@ -8,6 +8,14 @@ def app_name
...
+    if current_ability.can?(:act_as, current_facility)
```
jhanggi: If you're an admin and are in the front-end (e.g. placing an order), both ordering on behalf of and as yourself, you'll see "NU Core Admin". I'm thinking this might not be the correct logic. Although, it becomes moot if we're able to drop the logic.

jhanggi:
Other than the app title logic, this looks good to go.

drench:
Updated to use `app_name` ("NU Core") everywhere, per @nutodd.

jhanggi:
:+1: Hooray for deleting code!

PR #376, opened 2015-10-16
[#121571] Fix link in "needs approval" email

jhanggi:
As part of the new order detail management popup we removed the old
actions for editing an order detail. This points to the proper place,
also adding a redirect for the now-non-existent path.


drench:
:+1:

PR #377, opened 2015-10-23
Ability to run db:migrate from scratch

jhanggi:
This is part of my slow progress to get a functional Oracle Docker image running. Oracle has a limit on the length of index names, so a straight `db:schema:load` would not work.

And, when NUCore was brought in to git, there was a pre-existing schema which was not captured in the migrations, thus `rake db:migrate` would not work from a blank database.

I tried two techniques in order to get `rake db:migrate` working and
decided on this one.

The first:
* Find the first commit’s schema.rb and put that in an initial migration.
* Delete all migrations up to that point
* Fix any issues caused by missing data

That was fairly simple, but it lost much of the history of the project,
so I tried another process.

* Find the first commit's schema.rb and put that in an intial migration.
* Run `rake db:drop db:create db:migrate` and as I found errors, correct
the original migration accordingly.
* Fix issues caused by missing data or API changes in the migrations

They are similar, but I'm happier with this result.

* It gives a better history of the project
* Index names are more in line with Rails standards

Notes on the changes:
* Some columns have changed ordering within the tables -- not an issue
for new deploys, but could result in weirdness on existing development
commits when working off of existing databases
* Timestamps for some tables now have `null: false`. 
* The new version of foreigner uses 1.9 hash syntax, so `add_foreign_key`
in `schema.rb` will be using the new syntax while the rest of schema.rb
uses the `=>` syntax.


db/migrate/20120501184422_remove_nucs.rb
```diff
@@ -1,22 +0,0 @@
-class RemoveNucs < ActiveRecord::Migration
```
jhanggi: The `nucs_` table creation migrations exist within the nucs engine in the NU branch. That's why I removed them from the initial schema and removed this migration.

db/migrate/20110216164102_add_facilities_journal_mask_column.rb
```diff
@@ -1,6 +1,6 @@
 class AddFacilitiesJournalMaskColumn < ActiveRecord::Migration
   def self.up
-    add_column :facilities, :journal_mask, :string, :limit => 50, :null => true, :after => 
+    add_column :facilities, :journal_mask, :string, :limit => 50, :null => true #TODO why? , :after =>
```
drench: TIL on top of a TODO: Rails 3.2's `rake notes` does not look for annotations under `db/` but Rails 4 does.

drench:
LGTM, though I only tried it with MySQL. I haven't gotten a NU Core Open running against Oracle (yet).

jhanggi:
I was pretty much able to change my database.yml to point at the oracle docker container (which I had started from NU). I think I might have had to bundle again with the oracle group, but it was surprisingly straightforward to test. Pulling this down into NU was a little trickier to test, but wasn't too bad.

drench:
Ran for me with no problems when pointing my nucore-open instance at the Docker container I've used with nucore-nu. :+1:

PR #378, opened 2015-10-23
More powerful rake task for clearing oracle databases

jhanggi:
I’ve had bad luck with `rake db:drop db:create` on oracle. We had
`db:oracle_drop`, but that parsed the schema.rb file for table names,
so when dealing with a database that didn’t match the schema.rb, it
might not clear everything out. This drops every table and sequence. So
it’s very dangerous, hence why it’s only allowed to run in dev & test
modes, and while it’s suffixed with _severe.

### DIFF
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -11,9 +11,11 @@ namespace :db do
   desc "Drop database for current RAILS_ENV"
   task :oracle_drop => :environment do
     next unless db_allow_task?
-    config= Rails.configuration.database_configuration[Rails.env]
+    config = Rails.configuration.database_configuration[Rails.env]
     connect_string = "#{config["username"]}/#{config["password"]}@#{config["database"]}"
-    system "bash -ic 'cd #{Rails.root}/db && ./generate_drops.sh | sqlplus #{connect_string}'"
+    Dir.chdir(Rails.root.join("db")) do
+      system "./generate_drops.sh | sqlplus #{Shellwords.shellescape(connect_string)}"
+    end
   end
 
   desc "DANGER! Drops all user tables and sequences"
--- a/lib/tasks/db.rake
+++ b/lib/tasks/db.rake
@@ -1,10 +1,49 @@
 namespace :db do
+  def db_allow_task?
+    if Rails.env.test? || Rails.env.development?
+      true
+    else
+      puts "Only allowed in development and test mode"
+      false
+    end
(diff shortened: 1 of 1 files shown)


lib/tasks/db.rake
```diff
@@ -1,10 +1,49 @@
...
+      false
+    end
+  end
 
   desc "Drop database for current RAILS_ENV"
   task :oracle_drop => :environment do
+    next unless db_allow_task?
     config= Rails.configuration.database_configuration[Rails.env]
     connect_string = "#{config["username"]}/#{config["password"]}@#{config["database"]}"
     system "bash -ic 'cd #{Rails.root}/db && ./generate_drops.sh | sqlplus #{connect_string}'"
```
drench: I realize this code has been here a long time but that `cd` would fail if there were any spaces in `Rails.root`, and I could see `connect_string` breaking it, depending on the password.

jhanggi: Do we just need to wrap some quotes around the directories? My bash fu is not the greatest.

drench: I was thinking more like:

```ruby
Dir.chdir(Rails.root.join("db")) do
  system "./generate_drops.sh | sqlplus #{Shellwords.shellescape(connect_string)}"
end
```

I'm not sure the `bash -ic` needs to be there, since I would expect the environment to be set up correctly already when running this from the command line.

jhanggi:
Good suggestion, @drench. Updated.

drench:
After running this "deep clean" `rake db:reset` is failing due to several index and FK names in `schema.rb` being > 30 chars. Did #377 work because some of lingering indexes that now have to be recreated, and can't?

jhanggi:
I think db:reset does a db:schema:load, which won't work on oracle out of
the box because of the index lengths. For oracle, you need to do a
db:migrate to get schema in the proper state.

drench:
So is this the place for shortening those index names, or a separate PR? We ought to be able to `db:schema:load` on Oracle too.

jhanggi:
#377 might have been a good place since it was dealing with the schema itself, but I don't think this PR is the right spot. 

I'm also not even really sure the best way to go about cleaning up the index names. I'd really like something that gets rid of the `sys_xxx` indexes in NU/UIC. Definitely up for ideas, though.

drench:
:+1: and we'll deal with the index names elsewhere.

PR #379, opened 2015-10-27
Update oracle instructions for support on El Capitan

jhanggi:


drench:
Not on El Capitan yet to test this completely, but LGTM. I was able to run `sqlplus` afterward.

It seems kind of funny that overriding libraries with env vars is considered insecure, but rewriting binaries to do the same is all right (if I'm reading `fix_oralib.rb` correctly).

jhanggi:
That might need to be run as `sudo`. I'm not sure: I just realized I have instantclient installed within my home directory `~/opt/oracle/instantclient`.

When you upgrade, let me know how it goes and update the instructions accordingly.

PR #381, opened 2015-10-28
Fix migration from scratch issue

jhanggi:
When running the migration for a new project, the `change_column
:products, :facility_account_id, :integer, null: false` was failing
because there was a foreign key attached to it. Oddly, it was only
failing when run on the stage server database; it worked fine locally.

This changes the order of operations so the index doesn’t exist until
the column is in the state it needs to be. It also changes from
explicit sql to `add_foreign_key`.

### DIFF
--- a/db/migrate/20100701183019_create_facility_accounts.rb
+++ b/db/migrate/20100701183019_create_facility_accounts.rb
@@ -7,18 +7,18 @@ class CreateFacilityAccounts < ActiveRecord::Migration
       t.integer    :created_by,     :null => false
       t.datetime   :created_at,     :null => false
     end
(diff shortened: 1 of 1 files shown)


drench:
Migration from scratch worked for me on MySQL 5.6 but not Oracle. :+1: if the Oracle cleanup is separate (which I think it is).

jhanggi:
What happened on oracle? It went fine for me.

jhanggi:
Merging this for now. There's some additional oracle cleanup that needs to happen (indexes are lost because add_foreign_key does not create an index in oracle, while it does in mysql).

PR #382, opened 2015-10-30
Gemfile updates

drench:
This is a combination of updates, mostly coming from https://github.com/tablexi/nucore-uic
So far I only bumped the mysql2 version to match UIC, though we could consider trying 0.4.x.

I have branches ready to go for the other 3 forks but would like to see this one merged first.


jhanggi:
:+1: Are there any restrictions on the 0.4.x mysql2 like it's only for rails 4+?

drench:
That's what I understand. There was some incompatibility corrected recently https://github.com/rails/rails/pull/21536

PR #383, opened 2015-11-09
[#98883] Move payment sources from Billing to Users

drench:
This is the first PR of 3 expected for this feature (or PR 2 of 4 if you include #375).

The primary user-visible change is the move of Payment Sources from the Billing to the Users tab (or `admin_billing` to `admin_users` internally). The rest is cleanup, refactoring, and laying groundwork for the next PRs that will:

* Add a new global user role called "Account Manager" that will have access to a cross-facility `admin_users` tab but not `admin_billing`
* Add a new admin UI for granting and revoking the Account Manager role, as well as other global roles.


app/controllers/search_controller.rb
```diff
@@ -4,19 +4,54 @@ class SearchController < ApplicationController
...
+    @query_conditions ||=
+      [condition_sql, search_term, search_term, search_term, search_term]
+  end
+
+  def search_term
+    @search_term ||= generate_multipart_like_search_term(params[:search_term])
+  end
+
+  def load_users
+    user_relation = User.where(query_conditions)
```
jhanggi: Would it be worth extracting a `UserFinder` or `UserQuery` service object? Get some of this out of the controller?

app/controllers/users_controller.rb
```diff
@@ -28,9 +28,12 @@ def initialize
   def index
     @new_user = User.find_by_id(params[:user])
 
-    @users = User.with_recent_orders(current_facility)
-                  .order(:last_name, :first_name)
-                  .paginate(:page => params[:page])
+    @users =
+      if all_facility?
+        User.scoped
```
jhanggi: Would it be better to have the `with_recent_orders` scope allow the AllFacility, and not apply the `for_facility` scope?

app/views/facility_accounts/show_statement_list.html.haml
```diff
@@ -1,29 +1,29 @@
...
         %th
-        %th= t('.th.sent_at')
-        %th= t('.th.sent_by')
-        %th.currency= t('.th.orders')
+        %th= t(".th.sent_at")
+        %th= t(".th.sent_by")
+        %th.currency= t(".th.orders")
     %tbody
-      - @statements.each do |s|
+      - StatementPresenter.wrap(@statements).each do |statement|
```
jhanggi: I like this change. I'm just curious how this is related to the moving of the tab?

app/views/shared/nav/_administration.html.haml
```diff
@@ -0,0 +1,7 @@
+- if !acting_as? && current_user.try(:administrator?)
```
jhanggi: What's the case where you are on this page but not logged in?

lib/nav_tab/link.rb
```diff
@@ -0,0 +1,30 @@
+class NavTab::Link
+  include ActionView::Helpers::UrlHelper
+
+  attr_reader :subnav, :text, :url
+
+  def initialize(tab: nil, text: nil, url: nil, subnav: nil, cross_facility: false)
+    @tab = tab.present? && tab.to_s
```
jhanggi: I think `tab.presence` will give you the same thing, nil if empty/blank and the value otherwise.

lib/nav_tab/link_collection.rb
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  private
+
+  def accounts
+    NavTab::Link.new(tab: :accounts, text: t_my(Account), url: accounts_path)
+  end
+
+  def admin_billing
+    ability.can?(:manage_billing, facility) &&
```
jhanggi: This idiom feels a little esoteric. Would
```ruby
if ability.can?(:manage_billing, facility)
  NavTab::Link.new(tab: :admin_billing, url: billing_tab_landing_path)
end
```

be clearer? It's one more line, but it's a little more obvious, I think.

lib/nav_tab/link.rb
```diff
@@ -0,0 +1,30 @@
+class NavTab::Link
```
jhanggi: Do these really belong under `lib`? It seems like they're somewhat of presenters.

I see the module NavTab is under lib, which is, I suspect, why you put it here. I would prefer if we starting moving application code out of `lib` and into `app`, be it `app/services` or `app/presenters`. 

On a similar note, what would be your thoughts on moving things out of `app/support` and into `app/services` or `app/xxx/concerns` (e.g. `app/support/products/relay_support` -> `app/models/concerns/products/relay_support`. Where it really bugs me is the `spec/app_support` directory feels really awkward (I confess, I think that was me that started that, and I now regret my decision).

Not something that needs to be done all at once, but something we do as we touch pieces of code.

jhanggi:
while doing initial testing:

http://localhost:3000/facilities/example/accounts/1/account_users/user_search (adding a user to a payment source) has the Billing sidebar.

jhanggi:
Also adding a new payment source http://localhost:3000/facilities/example/accounts/new_account_user_search

jhanggi:
I started testing as a "Billing Administrator" (and changing the feature flags to match UIC's). Before, all they had was a "Billing" tab all the way to the right by the search box. Now they have that plus "Orders", "Reservations", "Billing". The orders and reservations tabs don't have anything in them. We should hide those from the billing admin.

Now, with the "Payment Sources" link removed from the billing sidebar, there is no way for a billing admin to access the payment sources now (in uic, they can't add/remove, but you can see transactions and membership).

Would it be possible for a "Billing Administrator" to still see the link in the sidebar? That would probably make it so UIC does not see any interface changes.

app/views/facility_accounts/show_statement_list.html.haml
drench: I was cleaning up partials related to moving the billing tab, and this got caught in that. But no it's not directly related.

lib/nav_tab/link_collection.rb
drench: Reformatted. Funny how the lines added up: RuboCop dinged this class for now being 101 lines long! (It prefers 100 or less)

app/views/shared/nav/_administration.html.haml
drench: This partial gets included everywhere. `layouts/application` renders `shared/nav`, which then renders `shared/nav/administration`.

After further refactoring using the `NavTab` presenters (not in this branch), it's reduced down to 2 lines and probably not worth its own partial.

drench:
The billing sidenavs I missed before should be user sidenavs with 3831d4f3b263f257357aba4733b94ae2e7dec8f7.

app/services/user_finder.rb
```diff
@@ -0,0 +1,44 @@
...
+        OR
+          LOWER(username) LIKE ?
+        OR
+          LOWER(CONCAT(first_name, last_name)) LIKE ?
+      )
+    SQL
+  end
+
+  def query_conditions
+    [condition_sql, @search_term, @search_term, @search_term, @search_term]
```
jhanggi: One thing that I like to do when I have more than a single `?` in my query is to use named parameters. So the query would be `LOWER(first_name) LIKE :term OR LOWER(last_name) LIKE :term ...`, and then you would query with `User.where(condition_sql, term: @search_term)`. Saves you from needing to repeat the same param over and over again.

app/views/admin/shared/_sidenav_billing.html.haml
```diff
@@ -11,8 +11,9 @@
     = tab t("facilities.disputed_orders.head"), facility_disputed_orders_path, (sidenav_tab == "disputed_orders")
   %li.divider
   = tab t('facility_accounts.accounts_receivable.head'), facility_accounts_receivable_path, (sidenav_tab == 'accounts_receivable')
-  %li.divider
-  = tab Account.model_name.human.pluralize, facility_accounts_path, (sidenav_tab == 'accounts')
+  - if current_user.billing_administrator?
```
drench: I've got mixed feeling about this (and the same change in `vendor/engines/c2po/app/views/admin/shared/_sidenav_billing.html.haml`). I would have liked to make this an ability check against `Account` but too many managerial roles have `:manage` ability on `Account` and we need this link to show up for Billing Administrators only.

There's also the issue that once clicking through, it takes the user to an `admin_users` tab context rather than `admin_billing`.

drench:
This should be ready for another look.

app/models/facility.rb
```diff
@@ -49,6 +49,11 @@ def accounts
   scope :active, :conditions => { :is_active => true }
   scope :sorted, order: :name
 
+  def self.cross_facility
```
jhanggi: Should we standardize on `all_facility` vs `cross_facility`? I don't have a preference, but it seems like we should be using the same thing everywhere.

app/presenters/nav_tab/link_collection.rb
```diff
@@ -0,0 +1,123 @@
...
+  end
+
+  private
+
+  def accounts
+    NavTab::Link.new(tab: :accounts, text: t_my(Account), url: accounts_path)
+  end
+
+  def admin_billing
+    if !facility.all_facility? && ability.can?(:manage_billing, facility)
```
jhanggi: Would it be worth extracting

```ruby
def single_facility?
  !facility.all_facility?
end
```

It's a bunch of places in this class and would probably be easier to read.

app/models/facility.rb
drench: I lean toward `cross_facility`. `Facility.all_facility` would feel too close to `Facility.all` to me.

app/views/admin/shared/_sidenav_billing.html.haml
jhanggi: Could we add a condition the the accounts controllers to switch which sidebar displays based on whether you're in billing context or users context?

app/models/facility.rb
jhanggi: That works for me.

app/views/affiliates/index.html.haml
```diff
@@ -1,22 +1,26 @@
...
+        %td
+          = link_to t("affiliates.remove.label"),
+            affiliate_path(affiliate),
+            confirm: t("affiliates.remove.confirm", name: affiliate.name),
+            method: :delete
         %td.indented_lots
           = affiliate.name
-          %span.indented== (#{link_to('Edit', edit_affiliate_path(affiliate))})
+          %span.indented
+            == (#{link_to(t("affiliates.edit"), edit_affiliate_path(affiliate))})
```
jhanggi: Seems like we don't use the `==` very often. I actually had to look up what it meant. Would
`= %[#{link_to(t("affiliates.edit"), edit_affiliate_path(affiliate))}]` be more obvious (although there are already so many parentheses and curly braces, it's pretty confusing as is).

drench: I extracted the links in this partial to a helper.

jhanggi:
If you click into an order, the Orders, Reservations, and Billing tabs show back up.

app/presenters/nav_tab/link_collection.rb
drench: Done, though I feel like we should do a bigger refactor related to all the `cross_facility` vs. `all_facility` changes. I'd be useful to have `#cross_facility?` and `#single_facility?`, and have `current_facility` set to a null-object `NilFacility` when outside a facility context.

drench:
* 9754a5a should resolve the issue where a billing admin would see facility-specific admin tabs for Orders and Reservations. 
* 13630cc sets up a feature flag to determine where Payment Sources lives: under Users or Billing. This included some new sidebar logic in the partials under `/app/views/facility_accounts` though I left those that looked either obsolete or not Billing Administrator-related alone (like `user_search.html.haml` and `search.html.haml`)

jhanggi:
:+1: Cool. I think we're good now.

PR #384, opened 2015-11-09
Create Billing Admin in demo seed

jhanggi:
Two commits, first one is some rubocop autocorrects including single->double quotes and 1.9 hash syntax. Second commit is the actual change.


drench:
Are you using the RuboCop config from starter? It's not catching all the single quoting. We may be overdue in committing `.rubocop.yml` and `.rubocop-txi.yml` to this project. Aside from that, :+1:

PR #385, opened 2015-11-10
[#121941] Fix display of schedule rules when server is not in local time

jhanggi:
If the server time does not match the application time (what’s set in
`config/application.rb` `config.time_zone`, then schedule rules would
be shifted by the offset between the app time and the server time.

There are still a couple issues that are related to when the browser time doesn't match the application time we'll need to address. I've opened tickets for those in redmine in the open project.

### DIFF
--- a/app/controllers/timelineable.rb
+++ b/app/controllers/timelineable.rb
@@ -6,13 +6,13 @@ module Timelineable
   end
 
   def timeline
-    @display_date = display_date_as_time.to_date
+    @display_date = display_date_as_time
     @schedules = current_facility.schedules.active.order(:name)
   end
 
   private
 
   def display_date_as_time
-    (parse_usa_date(params[:date]) if params[:date]) || Time.zone.now
+    parse_usa_date(params[:date]) || Time.current.beginning_of_day
   end
 end
--- a/app/controllers/timelineable.rb
+++ b/app/controllers/timelineable.rb
@@ -6,7 +6,7 @@ module Timelineable
   end
 
   def timeline
-    @display_date = display_date_as_time
+    @display_datetime = display_date_as_time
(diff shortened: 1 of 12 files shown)


app/controllers/timelineable.rb
```diff
@@ -6,13 +6,13 @@ module Timelineable
   end
 
   def timeline
-    @display_date = display_date_as_time.to_date
+    @display_date = display_date_as_time
```
drench: I got burned in the past assuming `@display_date` was a `Date`, so we made it so. If it can't be a `Date` how would you feel about giving it a new name?

drench:
I had just the one comment on `@display_date`.

app/controllers/timelineable.rb
jhanggi: I'm open to giving it a new name, but I can't think of anything good. Any ideas?

Other than the I18n formatting, where did it get you?

drench: Mistaking it for a `Date` value was the root of https://github.com/tablexi/nucore-open/pull/328, where the unavailability display disappeared from timelines.

jhanggi: Fair. Do you have a better idea for a name? It's basically the current time or midnight at the beginning of the specified date.

drench: Maybe `@display_date_as_time`? I'd really like to have the word `time` in there somewhere.

jhanggi:
Updated. I went with `@display_datetime`.

drench:
There are a couple `FacilityReservationsController` specs failing because they're expecting values for `assigns[:display_date]`. :+1: after cleaning that up.

drench:
:+1:

PR #386, opened 2015-11-11
Add rubocop configuration

jhanggi:
I’m thinking we should just run rubocop as we’re working on new files; I don’t think it’s worth a big push on its own.

If we’re going to be doing rubocop cleanup, I’d like to see it always be its own commit when possible to make it easier to see what’s actually changes (especially in small PRs).

### DIFF
--- a/.rubocop-txi.yml
+++ b/.rubocop-txi.yml
@@ -3,7 +3,6 @@ AllCops:
     - "db/schema.rb" # You can't touch this
     - ".bundle/**/*" # Auto-generated
     - "bin/**/*"     # Auto-generated
-    - "vendor/**/*"  # We cannot solve the world's problems
   RunRailsCops: true
 
 Lint/HandleExceptions:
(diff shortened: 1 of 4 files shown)


.rubocop-txi.yml
```diff
@@ -0,0 +1,53 @@
+AllCops:
+  Exclude:
+    - "db/schema.rb" # You can't touch this
+    - ".bundle/**/*" # Auto-generated
+    - "bin/**/*"     # Auto-generated
+    - "vendor/**/*"  # We cannot solve the world's problems
```
drench: For this project, I think we want to include `vendor/**/*`.

drench:
We should probably add `rubocop` to the `Gemfile` as well.

.rubocop-txi.yml
jhanggi: Good call. I'll do that.

jhanggi:
Updated. (I was thinking about running rubocop over Gemfile, but wanted to keep the changes readable).

drench:
:+1:

PR #388, opened 2015-11-12
Suppress training request notifications when the feature is off

drench:
I ran into an exception while running with training requests off. If a training request exists, it would try to put a link in the drop-down whether the feature was enabled or not. But creating that link fails because the route to it only exists when the feature is enabled.

With this, if the feature is off, the notifications for it are off too.


jhanggi:
:+1:

PR #389, opened 2015-11-13
[#103993] Payments model

jhanggi:
I’m going to be using this table for tracking payments via credit card
in NU. Matt and I have been discussing more a more full-featured
payment tracking system down the road, so this starts down that path.

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -7,7 +7,7 @@ class Payment < ActiveRecord::Base
     @@valid_sources ||= [:check]
   end
 
-  validate :source, presence: true, inclusion: { in: valid_sources }
-  validate :account, :amount, presence: true
+  validates :source, presence: true, inclusion: { in: valid_sources }
+  validates :account, :amount, presence: true
 
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -1,0 +1,13 @@
+class Payment < ActiveRecord::Base
+  belongs_to :account, inverse_of: :payments
+  belongs_to :statement, inverse_of: :payments
+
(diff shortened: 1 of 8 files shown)


jhanggi:
Fixed a typo I had in the validations (`validate` vs `validates`) and added a `paid_by` user.

jhanggi:
@drench and @benlinton could y'all take a look at this PR? I'm working off of it inside of NU, so I'd like to get it merged soon.

drench:
:+1: and I see what you mean about `master` being protected. It might not work for this project.

app/models/payment.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  # Add additional sources in an engine with Payment.valid_sources << :new_source
+  def self.valid_sources
+    @@valid_sources ||= [:check]
+  end
+
+  validates :source, presence: true, inclusion: { in: valid_sources }
+  validates :account, :amount, presence: true
+
+end
```
benlinton: Do we only create a `payment` when the check is presented or when it's been processed?  If presented, we'll have to account for invalid payments.

benlinton:
:+1:

app/models/payment.rb
benlinton: Updated my comment.

jhanggi: I hadn't really thought of the case where the check is invalid or bounces or whatever. I would probably assume we would add it when the check is deposited. I mainly used checks as an example; this is mainly going to be used imminently for credit card payments, which we will only record on capture success. 

We're going to ignore chargebacks and refunds for now. We may give Payments a status at some later time if we expand this feature.

PR #394, opened 2015-11-17
More granular extension points in Facility forms

jhanggi:
This is a proof of concept for one idea I had about how to manage extension points in the views. I'm working on how to handle adding a merchant id per facility for credit card processing and would like it to work nicely with the c2po engine.

The nice thing it does is gets rid of the full overridden views for Facility#manage and Facility#edit and allows us to inject partials from the engine.

I'll have another PR that presents a similar, but slightly different API for this same extension point.


jhanggi:
CI failure here is because I moved the translation key. I'll deal with it if we decide if we like this or the other option (or if we decide we like neither, then it goes away).

app/presenters/facility_management.rb
```diff
@@ -0,0 +1,11 @@
...
+
+  def self.manage_partials
+    @@manage_partials ||= []
+  end
+
+  def self.edit_partials
+    @@edit_partials ||= []
+  end
+
+end
```
benlinton: I'd prefer we name this `FacilityPresenter`

benlinton:
I like this concept.  One thought:

Will we want the ability to granularly order the forms fields?  In that case should we just add all of the form fields as partials into the `manage_partials` method (i.e. class variable) and then provide some sort of optional `after: "my_partial"` hook to explicitly define order when injecting new partials?

jhanggi:
Closing in favor of going down the road of #396.

PR #395, opened 2015-11-17
Extend facility edit forms for payment types

jhanggi:
This is an alternative for #394 with a slightly different API tying thing piece specifically to the account types being added by the engine.

This is a proof of concept for one idea I had about how to manage extension points in the views. I'm working on how to handle adding a merchant id per facility for credit card processing and would like it to work nicely with the c2po engine.

If we go down this route, I might consider creating a migration that changes the name of the columns from `accept_cc` and `accept_po` to more explicit `accept_credit_card` and `accept_purchase_order` to get rid of the `short_name` option.

With either option, I think there are ways to get rid of even more of the views being overridden by this engine.


jhanggi:
In this branch, I removed the translation key that we're testing, hence the CI failure. As with #394, it'll be easy to deal with if we decide on one of the options.

app/presenters/facility_management.rb
```diff
@@ -0,0 +1,14 @@
+class FacilityManagement
```
jhanggi: I just realized there's `AccountManager`, which would probably be a better home for this (with some refactoring on it).

vendor/engines/c2po/lib/c2po.rb
```diff
@@ -9,17 +10,21 @@ class Engine < Rails::Engine
 
       # make this engine's views override the main app's views
       paths = ActionController::Base.view_paths.to_a
-      index = paths.find_index { |p| p.to_s.include? 'c2po' }
+      index = paths.find_index { |p| p.to_s.include? "c2po" }
       paths.unshift paths.delete_at(index)
       ActionController::Base.view_paths = paths
+
+      FacilityManagement.register_payment_type(CreditCardAccount, short_name: "cc")
+      FacilityManagement.register_payment_type(PurchaseOrderAccount, short_name: "po")
```
benlinton: for shortname i'd prefer `credit_card` and `purchase_order`, the few extra bytes are worth the improved readability

benlinton: Oh i see, the abbreviated versions were used before this PR was created.

app/presenters/facility_management.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def self.register_payment_type(type, options = {})
+    options[:short_name] ||= type.name.underscore
+    payment_types[type] = options
+  end
+
+  private
+
+  def self.payment_types
+    @@payment_types ||= {}
```
benlinton: Will we ever care about having control over the order of payment types?  I assume, no.

benlinton:
:+1:

jhanggi:
Closing in favor of #396

PR #396, opened 2015-11-17
Use "ViewHooks" for injecting partials

jhanggi:
This is a more generic take on what I was doing on #394. This makes it pretty easy to add hooks where ever we end up needing them. There's definitely some refactoring still to be done, and I'm not 100% on the names.

Let me know your thoughts. So far, I think this may be my favorite of the three.


app/helpers/view_hook_helper.rb
```diff
@@ -0,0 +1,13 @@
+module ViewHookHelper
+
+  # Look up the current path and render the view hooks for a specific placement
+  def view_hook(placement, args = {})
+    # Take from translation helper shortcut (e.g. `t(".locale")`), this takes the
+    # current view path, and converts slashes to dots, and the underscored partial as well
+    # "facilities/manage" => "facilities.manage"
+    # "facilities/_facility_fields" => "facilities/facility_fields"
```
benlinton: `"facilities/_facility_fields" => "facilities.facility_fields"`

app/helpers/view_hook_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+    path = @virtual_path.gsub(%r{/_?}, ".")
+    ViewHook.render_placement(path,  placement, self, args)
```
benlinton: Extra space.

app/helpers/view_hook_helper.rb
```diff
@@ -0,0 +1,13 @@
+module ViewHookHelper
+
+  # Look up the current path and render the view hooks for a specific placement
+  def view_hook(placement, args = {})
```
drench: When I'm looking through some haml, the word "render" sticks out to flag something external. I like how `#view_hook` as a name works as a companion to `#add_hook`, but something like `#render_view_hook` might have a more clear association for me.

app/helpers/view_hook_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+    # Take from translation helper shortcut (e.g. `t(".locale")`), this takes the
```
drench: Taken(?)

vendor/engines/c2po/app/views/facilities/_c2po_facility_fields.html.haml
```diff
@@ -0,0 +1,4 @@
+.well
+  %p= t('facilities.facility_fields.payment_options')
+  = f.input :accepts_po, :as => :boolean, :label => false, :inline_label => t('facilities.facility_fields.accepts_po')
+  = f.input :accepts_cc, :as => :boolean, :label => false, :inline_label => t('facilities.facility_fields.accepts_cc')
```
drench: I know this was pulled straight from the old partial but this could use style updating (quoting, hashrockets) before a merge.

vendor/engines/c2po/app/views/facilities/_c2po_manage.html.haml
```diff
@@ -0,0 +1,2 @@
+= f.input :accepts_po, :label => t('facilities.facility_fields.accepts_po')
+= f.input :accepts_cc, :label => t('facilities.facility_fields.accepts_cc')
```
drench: Same style update as above.

app/helpers/view_hook_helper.rb
jhanggi: Yeah, I'm definitely not sold on all the names I've come up with so far. This PR (and the other two) are really more an exploration of the concept. There's obviously cleanup and some refactoring left to do, but I didn't want to spend too much time on that since some of these spikes will just be thrown away.

drench:
I like this the best of the 3, particularly because it's more general. I think there are already other places I would want to apply this. `ViewHook` looks good. I'm not sure I have any refactoring suggestions, outside method renaming.

benlinton:
It's probably not worth the effort.  But it would be kind of amazing if we replaced `f.input` with `f.input_with_hook`.  And then we could append new inputs with `ViewHook.add(partial: "middle_name_form", after_input: "first_name", ...)`.

benlinton:
I like this approach the best out of the 3 as well.

jhanggi:
Closing for now, but I'll re-open it when it's ready for review again.

jhanggi:
I think this is ready for review. There are definitely other places this technique could be used, but for now I focused on the c2po engine. I did some reorganization to the view folder, so some of that might be a little tricky to read.

jhanggi:
Any feedback on this? I'd like to get this in so I can start using it in nu.

app/presenters/view_hook.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  class << self
+    def instance
+      @@instance ||= new
+    end
+
+    delegate :add_hook, :render_view_hook, to: :instance
+  end
+
+  # Best used throught `render_view_hook` helper method
```
drench: Typo "throught"

spec/presenters/view_hook_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    it "finds by string" do
+      expect(view_hook.find("view1", "placement")).to eq(["partial"])
+    end
+  end
+
+  describe "render_view_hook" do
+    let(:context) { double }
+    before do
+      view_hook.add_hook("view1", "placement1", "partial")
+      view_hook.add_hook("view1", "placement1", "partial2")
```
drench: Maybe add a test to make sure the dots-to-underscore/slash translation is doing what we expect?

drench:
Looks good. A couple minor comments.

benlinton:
:+1: (after dan's feedback is resolved)

PR #397, opened 2015-11-18
[#98883] Add Account Manager user role

drench:
This adds a new global user role called Account Manager, which has access to the User administration tab in a cross-facility context, which is something that no other roles (even global admins) have.

Previously many of the navigation tabs and links used `Role` checks to determine display. The changes here should all use `Ability` checks instead. This does mean we are now explicitly taking away abilities from the global administrator role, but it shouldn't change the behavior this role has seen all along. For example, the cross-facility Billing tab had a condition to make it visible only to Billing Administrators, which also meant it was not visible to global admins. Converting this to an `Ability` check meant removing permissions from global admins, which until now had an across-the-board `can(:manage, :all)`. The `Ability` restrictions were also the reason behind the spec change in 29bfcd06bcf85839cbd779993a7d8557a1d27fe5; it's something that probably should have been tested as a Billing Administrator all along.

This change also refactors the global admin tabs to use the `NavTab` presenters brought in with #383.

A PR in the near future will implement a UI for adding and removing users to and from this new role.


app/presenters/nav_tab/global_link_collection.rb
```diff
@@ -0,0 +1,43 @@
...
+  private
+
+  def admin_cross_facility_billing
+    if can?(:manage_billing, Facility.cross_facility)
+      global_tab(:admin_billing, facility_transactions_path("all"))
+    end
+  end
+
+  def admin_cross_facility_users
+    if can?(:manage_accounts, Facility.cross_facility) && can?(:manage_users, Facility.cross_facility)
```
jhanggi: Is there ever a case where someone would have only one of these abilities and need to be able to get into this tab?

lib/role.rb
```diff
@@ -4,11 +4,17 @@
 # convenience methods for testing the role status of users.
 module Role
 
+  def grant_role!(role, facility: nil)
```
jhanggi: Do we need this method? It doesn't look like this gets used anywhere, and we have We have `UserRole.grant`, which should probably be doing the valid roles check.

jhanggi: One other problem with that method is it allows you to create duplicate role assignments. E.g. a user could end up with the Facility Director role twice.

jhanggi: (I'm fine holding off on that fix until the next PR)

jhanggi:
Code is looking pretty solid. I'll try to do a round of testing later this afternoon.

app/presenters/nav_tab/global_link_collection.rb
drench: Currently I don't think so. Account Managers have both abilities, but that's all I see. We grant `:manage` to `Facility` if a user is a `manager_of?` a facility, but I don't expect anyone to be a manager of `Facility.cross_facility`.

That said, this could be `||` rather than `&&`.

lib/role.rb
drench: Yep I missed taking this out. I used it in a WIP that I ended up rebasing away. I do like the interface of this method more: `user.grant_role!(:facility_staff, facility: facility)` beats `UserRole.grant(user, FACILITY_STAFF, facility)` for me.

`UserRole.grant` mostly seems like a way to write `UserRole.create!` using fewer characters, and as you mentioned doesn't do anything to prevent duplicate role assignments. It looks like `UserRole.grant` is barely used outside specs and seeds anyway. I'm inclined to throw out both methods and add a validation.

app/views/admin/shared/_sidenav_users.html.haml
```diff
@@ -1,5 +1,5 @@
 %ul.sidebar-nav
-  - if current_ability.can?(:administer, User)
+  - if current_ability.can?(:manage_users, current_facility)
```
jhanggi: I'm not sure if this is the correct check. If I'm senior staff or staff, right now, they see the well (`.sidebar-nav`), but it's empty.

A different issue, but once I click into a specific user, and go to the the "Payment Sources" tab, I see a "Add Payment Source" button, but clicking on it gives me a 403.

jhanggi:
We should probably lock down which tabs an Account Manager sees when looking under a user. They probably don't need to see Orders, Reservations, or Access List.

Also, in testing, I found a couple things that don't match the ticket:
* An account administrator should be able to create a new Chart String account
* An account administrator should not be able to suspend/activate Accounts
* Under recent payment sources, currently it shows existing POs and CCs, and I can manage membership. I'm not sure if this is correct behavior or not.

drench:
All the above issues (and a few more) should be cleaned up now:

* When viewing user detail, an Account Manager should no longer have access to Orders, Reservations, or Access List or see those tabs.
* An Account Manager should be able to CRUD Chart Strings (cross-facility) but NOT CC or PO (single facility) accounts.
* An Account Manager should not be able to suspend/activate Accounts
* When viewing the Payment Sources index in a cross-facility context (`/facilities/all/accounts`), it should only list cross-facility Accounts (`NufsAccount`s)
* Staff and Senior Staff should see and be able to use the "Users" link in the Users sidenav (`app/views/shared/_sidenav_users.html.haml`)
* Only Account Managers should see the cross-facility Users tab under any context. Not even global administrators should ever see this tab.

One issue leftover from before that this does not address:
* After creating a new chart string, the user lands on `/facilities/all/accounts/user/:user_id` to show the list of this user's accounts. But it's in a Billing, not User context (with the billing sidenav).

app/controllers/facilities_controller.rb
```diff
@@ -36,7 +36,8 @@ def index
 
   # GET /facilities/:facility_id
   def show
-    raise ActiveRecord::RecordNotFound unless current_facility && current_facility.is_active?
+    return redirect_to(facilities_path) if current_facility.try(:cross_facility?)
```
drench: This line and the next are two spots where I would have liked to have `current_facility` set to a null object `Facility` when there is no facility.

lib/ability.rb
```diff
@@ -19,6 +19,10 @@ def initialize(user, resource, controller)
         can :manage, AccountPriceGroupMember
       else
         can :manage, :all
+        cannot [:manage_accounts, :manage_billing], Facility.cross_facility
+        if resource.blank? || resource.is_a?(Facility) && resource.cross_facility?
```
drench: This is another spot that might be more clear with a null object. The resource might be blank if `current_facility` is nil.

drench:
A note on the "show" page that users see after a chart string create,`/facilities/:facility_id/accounts/user/:user_id`: this area seems outdated and maybe obsolete? It looks like `/facilities/:facility_id/users/:user_id/accounts` is where the users should land instead.

app/controllers/facility_accounts_controller.rb
```diff
@@ -52,7 +52,10 @@ def initialize
 
   # GET /facilties/:facility_id/accounts
   def index
-    @accounts = Account.has_orders_for_facility(current_facility).paginate(:page => params[:page])
+    accounts = Account.with_orders_for_facility(current_facility)
+    accounts = accounts.where(facility_id: nil) if current_facility.cross_facility?
```
drench: This is the only place we use `.with_orders_for_facility` (which is why I felt safe changing the name). And since it's used nowhere else I considered making the method do what I'm doing here, to filter out facility-specific accounts when the facility is "all". But would that method name still make sense if it did that? I'm not sure what I'd call it.

jhanggi: I think this is fine. I'm not sure a better way. It's already a little funky because `ids_with_orders` adds some extra filtering if it's `single_facility?`

app/views/admin/shared/_sidenav_users.html.haml
```diff
@@ -1,5 +1,5 @@
 %ul.sidebar-nav
-  - if current_ability.can?(:administer, User)
+  - if current_ability.can?(:manage, User)
```
jhanggi: This has some interesting behavior. If I'm a director, when I go to the users's tab, I this this item highlighted. But as soon as I go to either "bulk email" or "payment sources", this item disappears.

app/presenters/nav_tab/global_link_collection.rb
```diff
@@ -0,0 +1,43 @@
...
+  private
+
+  def admin_cross_facility_billing
+    if can?(:manage_billing, Facility.cross_facility)
+      global_tab(:admin_billing, facility_transactions_path("all"))
+    end
+  end
+
+  def admin_cross_facility_users
+    if can?(:manage_accounts, Facility.cross_facility) || can?(:manage, User)
```
jhanggi: I'm seeing both this and the cross-facility Billing as a director while inside a normal facility. When I click I get a 403.

jhanggi:
Found a few more issues with Account managers:
Under "Users" when I search for a user, the results have "Order For". I would expect for it not to be there, and the link to 403, which it currently does not.

http://localhost:3000/accounts/8/statements is 403ing. This is linked to from http://localhost:3000/facilities/all/users/5/accounts (obviously the IDs will be different for you).

http://localhost:3000/facilities/all/accounts/1/members "Remove Access" 403s, as does the "Add Access" button.

app/presenters/nav_tab/global_link_collection.rb
drench: Definitely the `can?(:manage, User)` is wrong here, since it changes depending on facility context, and we don't want that in this case. The ability needs to be: can it manage users in a cross-facility context. Current context doesn't matter. That's why I went with `can?(:manage_accounts, Facility.cross_facility)` and I think I'll have to do something similar for users. But I'm open to ideas.

drench:
* 5bdd8fb0ec321da58d62aa067b725b0d51b98939 should address the add and remove access 403s (these functions should both work for Account Managers)
* 5f56edffc589abe05d2e36d1d7b3b4a9703d06ce should remove the user switch ("order for") links when in the cross-facility context

I'm still working through the permission issues for the cross-facility Users tab.

drench:
Here's an Account Manager scenario with a 403 and I don't know how it should play out (`id`s are my database of course):

* Begin at `/facilities/all/users`
* Click on a user link
* Arrive at `/facilities/all/users/2`
* Click the "Payment Sources" tab in the horizontal nav (somewhat confusing given there's a different link with the same label in the vertical nav on the left, but that's another issue)
* Arrive at `/facilities/all/users/2/accounts`
* Click on a payment source link
* Arrive at `/accounts/5/statements` which is a 403

The things I'm wondering:
* Should `/facilities/all/users/2/accounts` list all accounts, even facility-specific accounts? My hunch is that it should.
* The partial that generates the payment source links in `/facilities/all/users/2/accounts` (`app/views/shared/_accounts_table.html.haml`) creates links differently depending on context:
  - if this is a single facility and the system uses statements, the link goes to `facility_account_statement_path`
  - if not, it goes to `account_statements_path`; this is the path that is giving the 403 above.

  Do we want to allow Account Managers to take this path (which looks like it leads to more 403s to chase down), or do we need a 3rd option, maybe pointing at `/facilities/all/accounts/1`, which would keep the user within the context of the cross-facility "Users" tab?

drench:
Continuing from the previous comment, there are already 3 options: if the user cannot `:manage` the listed `Account`, there is no link at all, just text.

drench:
This last set of commits should address the issues we discussed before Thanksgiving. Here's the diff between Wednesday and now: https://github.com/tablexi/nucore-open/compare/a8c4643fe3b2802001f4bd0ea79abc71bfb107f0...c9e37a2259b20d0f9231596002578831e5732da0

This reduces the display possibilities in `/app/views/shared/_accounts_table.html.haml` to three:
  - If the user cannot `:manage` the account, it displays its name only. There is just a label with no link in this case.
  - If the user is in a cross-facility context, it links to `facility_account_path`. This option should be what Account Managers see.
  - Otherwise, it links to `facility_account_orders_path`. This is what global administrators, facility directors and facility administrators should see.

`facility_account_orders_path` replaces the "recent" option in `FacilityAccountsController#show_statement`. This was kind of a confusing action, since it was listing purchased `OrderDetail`s, not `Statements`, and did not seem to be limited to "recent" orders. Its replacement should display the same information, but orders it by recently-ordered first. The new partial is based on `app/views/facility_accounts/show_statement.html.haml`, with cleanup. I went with adding an `OrderDetailPresenter`. I don't know if it would make sense to combine this with the decorator in #405; this felt more like a presenter than an decorator.

I've tried this out as an account manager, facility director, global administrator, facility staff, and (using UICore feature flags) billing administrator and things seem all right (no links leading to 403s or strange contexts).

app/controllers/facility_accounts_controller.rb
```diff
@@ -177,9 +185,6 @@ def show_statement
       when 'list'
```
jhanggi: Minor: now that we've gotten rid of `recent`, we can probably just make this into a plain if/else condition rather than a case.

app/models/order_detail.rb
```diff
@@ -167,6 +167,10 @@ def self.all_movable
     .where("order_details.state NOT IN('canceled', 'reconciled')")
   end
 
+  def self.purchased
```
jhanggi: This scope already exists on OrderDetail (albeit with a terrible name): `scope :ordered, joins(:order).merge(Order.purchased)`. I much prefer `purchased`. I would vote for changing the name of the scope and then setting up `ordered` as an alias.

It would probably be too big of a hassle at this point, but we could give a deprecation warning when using `ordered` over `purchased` and then go clean up the usages.

app/presenters/order_detail_presenter.rb
```diff
@@ -0,0 +1,64 @@
+class OrderDetailPresenter < SimpleDelegator
+
+  include ActionView::Helpers::NumberHelper
+  include DateHelper
+
+  delegate :admin_editable?, to: :reservation, prefix: true
+  delegate :name, to: :order_status, prefix: true
```
jhanggi: This seems to be doing the same thing that #405 is doing. I think it would be nice to converge on a single thing. The "presenter"/"decorator" distinction is always a little weird. Theoretically you could do a couple layers of wrapping, but that seems overkill.

Looking at OrderStatus, we might not even need this because it already defines `to_s` as `name`. So in HAML, `= order_detail.order_status` should render out to the status's name already.

app/presenters/order_detail_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+  end
+
+  def description_as_html
+    [bundle, product].compact.map do |description|
+      ERB::Util.html_escape(description)
+    end.join(" &mdash; ").html_safe
+  end
+
+  def edit_reservation_path
+    edit_facility_order_order_detail_reservation_path(facility, order, id, reservation)
```
jhanggi: Just curious, where is this class getting the route helpers?

app/presenters/order_detail_presenter.rb
```diff
@@ -0,0 +1,64 @@
...
+
+  private
+
+  def format_as_currency(value)
+    value.present? ? number_to_currency(value) : ""
+  end
+
+  # Is a fulfilled order detail nearing the end of the 90 day reconcile period?
+  # Returns true if it is 60+ days fulfilled, false otherwise
+  def reconcile_warning?
```
jhanggi: Can we make the application helper use this method so we don't have the logic duplicated in two places?

jhanggi:
As an Account Manager, if I go to "Payment Sources", and go into a Credit Card account, there is a "Statements" tab. That 403s.

jhanggi:
Out of scope for this PR, but I wanted to mention it: when creating an account, the tab stops are kind of nuts. From type, the next stop is description, and from number, focus goes the "NUCore" link in the header.

jhanggi:
We're getting close.

drench:
On this:
> As an Account Manager, if I go to "Payment Sources", and go into a Credit Card account, there is a "Statements" tab. That 403s.

I would expect the list of accounts in that context to only be cross-facility. Do you have a cross-facility credit card account, and is that valid?

jhanggi:
Ah, you're right. I do have a cross-facility CreditCardAccount. The demo seed doesn't set the facility for the credit card it creates. Not a huge deal, but we should probably clean that up.

app/presenters/order_detail_presenter.rb
drench: Good question, and the answer was a surprise to me. The `OrderDetail` model includes `NotificationSubject`, which then includes `url_helpers`, which feels like something a decorator/presenter could make unnecessary.

I'll add the `include` explicitly here in the presenter, since I don't want to count on that sticking around.

app/presenters/order_detail_presenter.rb
drench: Done in 7d9cbbf180bf5d95cce29828e7ef85829c0cd70c.

drench:
With 791a5ecc4a3518b4737b83f5caa4004d71b8773a, the demo seed will set a `Facility` when creating the credit card account. We could probably still use validations though I don't believe there's a way to create a CC or PO account without a facility through the UI.

drench:
This should be ready for another look. @benlinton I'd still like to see how we should handle the `OrderDetail` presenter/decorator merge, since I agree we should use one or the other but probably not both.

I have a branch to implement the global user role UI, but it builds off this so I will open a PR for that after this one is squashed and closed.

benlinton:
Let's use your presenter, and kill my decorator.

app/controllers/statements_controller.rb
```diff
@@ -3,7 +3,6 @@ class StatementsController < ApplicationController
   before_filter :authenticate_user!
   before_filter :check_acting_as
   before_filter :init_account
-  before_filter :init_statement, :except => :index
```
jhanggi: I think we still need the `init_statement`, just not the `except: :index`.

jhanggi: Or is `StatementsController#show` even a valid path? I can't quite figure out how to get to it other than direct url manipulation.

app/helpers/application_helper.rb
```diff
@@ -17,7 +17,7 @@ def order_detail_description(order_detail) # TODO: deprecate in favor of "_as_(h
     order_detail_description_as_html(order_detail)
   end
 
-  def order_detail_description_as_html(order_detail)
+  def order_detail_description_as_html(order_detail) # TODO: deprecate in favor of OrderDetailPresenter#description_as_html
```
jhanggi: Can this code be replaced with a direct call to `OrderDetailPresenter#description_as_html`?

app/controllers/statements_controller.rb
drench: Good catch; I misread that as an `:only`.

app/controllers/facility_accounts_controller.rb
```diff
@@ -171,22 +179,22 @@ def accounts_receivable
...
+      action = "show_statement_list"
+      @statements =
+        current_facility
+        .statements
+        .where(account_id: @account.id)
+        .paginate(page: params[:page])
+    else
+      action = "show_statement"
+      @statement = Statement.find(params[:statement_id])
+      @order_details = @statement.order_details
```
jhanggi: Can you add `.paginate(:page => params[:page])` onto this PR? I need it in the credit card branch, but I'd rather avoid the conflicts.

app/controllers/statements_controller.rb
drench: But yes the `statements#show` route seems obsolete given we have `facility_accounts#show_statement`.

drench:
Comments addressed.

jhanggi:
:+1: Looks good.

PR #398, opened 2015-11-20
Add Strong Params to Facilities

jhanggi:
I’m going to need to protect a couple columns on Facility for the
credit card processing, so I figured I’d bring in strong_params.
Getting it across the board will be a big undertaking (probably will do
it with the Rails 4 upgrade), but this starts laying the groundwork.

Two commits. First one is the actual change, second one is rubocop.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 ## base
 gem 'rails',            '3.2.22'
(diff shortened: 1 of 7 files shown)


drench:
Ambitious and looks good. I had one question on the permitted params array.

drench:
:+1:

PR #399, opened 2015-11-20
Remove unused actions

jhanggi:
I noticed these while I was cleaning up something else, and realized
these aren’t used anywhere (plus they’re broken—there’s no `product`
layout, and hasn’t been since the initial fork in github).

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -1,7 +1,7 @@
 class FacilitiesController < ApplicationController
 
   customer_tab :index, :list, :show
-  admin_tab :edit, :manage, :schedule, :update, :agenda, :transactions,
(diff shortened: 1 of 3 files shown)


drench:
Great to see unused code go away. And I think there's more! I've been looking at `lib/ability.rb` a lot lately and noticed `can [ :schedule, :agenda, :list, :show ], Facility` in there.

jhanggi:
Good call. I'll take care of that tomorrow. I'll just need to be careful
because there is an Instruments#schedule action that is legit. But agenda
can go.

jhanggi:
I rebased to clean up the merge conflicts, and I removed the `:schedule` and `:agenda` from the ability file.

drench:
:+1:

PR #400, opened 2015-11-20
[#122107] Add account and owner to transaction CSV export

jhanggi:
Based on https://github.com/tablexi/nucore-uic/pull/39, this also adds the account description (for NU) and the owner so it matches more closely to the html view.


drench:
:+1:

PR #401, opened 2015-11-23
Fix spec broken when not using statements

jhanggi:
UIC does not have statements enabled so the route generated here does not exist.


drench:
:+1:

PR #402, opened 2015-11-23
[#122135] Avoid general report error when missing owner

jhanggi:


app/controllers/general_reports_controller.rb
```diff
@@ -12,7 +12,10 @@ def account
 
 
   def account_owner
-    render_report(2, 'Name') {|od| format_username od.account.owner.user }
+    render_report(2, 'Name') do |od|
+      # Space at beginning it intentional to bubble it to thee top of the list
```
drench: A couple typos: "it intentional" and "thee top"

drench:
We could probably use a decorator or presenter on `Account` but I'm OK leaving it as a `TODO`. :+1: after typo cleanup.

PR #403, opened 2015-11-23
Split accounts engine boilerplate

benlinton:
Includes the new `split_accounts` engine, related models, factories, and specs.  

Also updated `spec/factories/nufs_accounts.rb` which may have unwanted effects in downstream repositories, so please review that factory extra carefully.


spec/factories/nufs_accounts.rb
```diff
@@ -1,46 +1,44 @@
 overridable_factory :nufs_account do
-  sequence(:account_number) do |n|
-    "9#{'%02d' % n}-7777777" # fund3-dept7
+
+  transient do
+    owner { create(:user) }
   end
 
+  sequence(:account_number) { |n| "9#{'%02d' % n}-7777777" } # fund3-dept7
```
benlinton: I think this may have been modified in one of Jason's recent PRs.

spec/factories/nufs_accounts.rb
```diff
@@ -7,10 +12,19 @@
   expires_at { Time.zone.now + 1.month }
   created_by 0
 
+  # TODO: add the owner as part of the default so the factory is valid.
+  trait :with_owner do
```
jhanggi: Take a look at https://github.com/tablexi/nucore-nu/pull/104/files I cherry-picked up the commit (5363f3ee15f567729dfdf47e9616778c6a720fa6) that was relevant to open. This needs to be outside the `overridable_factory` section, otherwise it'll cause problem downstream.

spec/factories/nufs_accounts.rb
```diff
@@ -7,10 +12,19 @@
...
+    # Every account must have an account_user "owner" in order for the account
+    # to be valid in rails. And foreign key constraints require that each
+    # account_user has an account inserted before the account_user is inserted.
+    callback(:after_build, :before_create) do |split_account, evaluator|
```
jhanggi: Is this actually a `split_account`, or is it just a normal `NufsAccount`?

vendor/engines/split_accounts/app/models/concerns/account_extension.rb
```diff
@@ -0,0 +1,11 @@
+module Concerns
```
jhanggi: I would put this under the SplitAccounts namespace. And the `Concerns` doesn't really need to be part of the namespacing, at least in my opinion.

vendor/engines/split_accounts/app/models/split.rb
```diff
@@ -0,0 +1,19 @@
+class Split < ActiveRecord::Base
```
jhanggi: Part of me wonders if this should have a more explicit name. Either `AccountSplits` or put it under the engine namespace `SplitAccounts::Split`.

vendor/engines/split_accounts/app/models/split_account.rb
```diff
@@ -0,0 +1,28 @@
+class SplitAccount < Account
+
+  has_many :splits, foreign_key: :parent_split_account_id, inverse_of: :parent_split_account
```
jhanggi: Should we also have `has_many :child_accounts, through: :splits`?

vendor/engines/split_accounts/app/models/split_account.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  validate :valid_percent_total
+  validate :one_split_has_extra_penny
+
+  def valid_percent_total
+    unless percent_total == 100
```
jhanggi: This is a style thing of mine, but I prefer not to use `unless` except in one liners where it's easier to read than an `if !`. I would probably change this to `if percent_total != 100`.

vendor/engines/split_accounts/app/models/split_account.rb
```diff
@@ -0,0 +1,28 @@
...
+      errors.add(:splits, "percent total must equal 100")
```
jhanggi: Can you move the string to the locales? I think you can do something like `errors.add(:splits, :percent_total`, and then in the locales, set `active_record.errors.splits.percent_total` (I might not have exactly right) in the `en.yml`.

vendor/engines/split_accounts/app/models/split_account.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def one_split_has_extra_penny
+    unless extra_penny_count == 1
+      errors.add(:splits, "can have only one with extra penny")
+    end
+  end
+
+  def extra_penny_count
+    splits.select(&:extra_penny).size
```
jhanggi: When using boolean fields, activerecord should also give you the `?` version, so `extra_penny?` might be a little clearer.

vendor/engines/split_accounts/spec/factories/split_accounts.rb
```diff
@@ -0,0 +1,45 @@
+FactoryGirl.define do
+  factory :split_account do
```
jhanggi: Would it be possible to use `factory :split_acount, parent: :nufs_account` so you don't need to define quite so much here?

vendor/engines/split_accounts/spec/factories/split_accounts.rb
```diff
@@ -0,0 +1,45 @@
...
+      callback(:after_build, :before_create) do |split_account, evalutor|
+        split_account.splits << build(:split, percent: 33.33, extra_penny: true, parent_split_account: split_account)
+        split_account.splits << build(:split, percent: 33.33, extra_penny: false, parent_split_account: split_account)
+        split_account.splits << build(:split, percent: 33.34, extra_penny: false, parent_split_account: split_account)
+      end
+    end
+
+    # Leave this at the bottom of the factory.
+    # Add a split if none exist and if transient `without_splits` is false.
+    callback(:after_build, :before_create) do |split_account, evaluator|
```
jhanggi: Do we want to allow a "split account" to only have one split? That seems a little weird.

vendor/engines/split_accounts/spec/models/split_account_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    context "when splits do not total 100 and one split has extra_penny" do
+      let(:split_account) do
+        build(:split_account, without_splits: true).tap do |split_account|
+          split_account.splits << build(:split, percent: 50, extra_penny: true, parent_split_account: split_account)
+          split_account.splits << build(:split, percent: 49.99, extra_penny: false, parent_split_account: split_account)
+        end
+      end
+
+      it "is invalid" do
+        expect(split_account).to_not be_valid
```
jhanggi: Whenever I'm testing the invalidity of a model, I like to also confirm it's invalid for the right reason. So in addition to this, I would usually check `expect(split_account.errors).to include(:split)`.

Also, I think we've standardized on `not_to` over `to_not`.

vendor/engines/split_accounts/spec/models/split_account_spec.rb
```diff
@@ -0,0 +1,102 @@
...
+    context "when splits total 100 and no splits have extra_penny" do
+      let(:split_account) do
+        build(:split_account, without_splits: true).tap do |split_account|
+          split_account.splits << build(:split, percent: 50, extra_penny: false, parent_split_account: split_account)
+          split_account.splits << build(:split, percent: 50, extra_penny: false, parent_split_account: split_account)
+        end
+      end
+
+      it "is invalid" do
+        expect(split_account).to_not be_valid
```
jhanggi: Same here, I would make sure it's invalid for the right reason.

vendor/engines/split_accounts/app/models/split.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  belongs_to :parent_split_account, class_name: "SplitAccount", foreign_key: :parent_split_account_id, inverse_of: :splits
+  belongs_to :subaccount, class_name: "Account", foreign_key: :subaccount_id, inverse_of: :parent_splits
+
+  scope :with_extra_penny, -> { where(extra_penny: true) }
```
jhanggi: We've been mostly standardizing on the `def self.with_extra_penny` version of scopes.

jhanggi:
Done with my review. Overall, it's a good start., just had a few comments.

vendor/engines/split_accounts/app/models/concerns/account_extension.rb
benlinton: Can removing `module Concerns` sometimes mess up rails magic automatic-path-resolving-from-class-name?

vendor/engines/split_accounts/app/models/split.rb
benlinton: I could go for `SplitAccounts::Split`.  I started off with `AccountSplit` and it got confusing with the joins.

vendor/engines/split_accounts/app/models/split.rb
benlinton: We should ideally go with scopes because they get some extra functionality that class methods don't (e.g. scopes are always chainable).  http://blog.plataformatec.com.br/2013/02/active-record-scopes-vs-class-methods/

vendor/engines/split_accounts/app/models/split_account.rb
benlinton: Yeah

vendor/engines/split_accounts/app/models/split_account.rb
benlinton: I usually follow https://github.com/bbatsov/ruby-style-guide for if/unless.  But I don't feel strongly about it.  I'll switch to `if not`

vendor/engines/split_accounts/app/models/split_account.rb
benlinton: Sure.

vendor/engines/split_accounts/app/models/split_account.rb
benlinton: Good idea.

vendor/engines/split_accounts/spec/factories/split_accounts.rb
benlinton: I think so, if we ever do some ajax-y UI stuff, allowing this will come in handy.

spec/factories/nufs_accounts.rb
benlinton: `nufs_account`

vendor/engines/split_accounts/spec/factories/split_accounts.rb
benlinton: I was thinking it would be a good idea to create a `account` factory that is the parent of `nufs_account` and `split_account`.  But I figured it was outside the scope of this PR.

vendor/engines/split_accounts/spec/models/split_account_spec.rb
benlinton: I can conform to the standard, no problem.  But why `not_to`?  I feel like it's easier to scan for and append to `to_not`.

benlinton: I like your recommended approach for confirming the errors and think I'm going to use it from now on.  Do you ever take this a step further and add `expect(split_account.errors[:splits]).to include(I18n.t("activerecord.errors.models.split_account.attributes.splits.percent_total"))` ?

benlinton:
Made most of the recommended updates.  Ready for a re-review.

vendor/engines/split_accounts/app/models/split.rb
jhanggi: Yeah, you might be right. The fact that scopes are always chainable is nice. I don't think we use the extensions anywhere.

The one nice thing about converting over to the class methods is that it becomes an indicator of how recent it is and it's clear that it's a rails 3 scope and doesn't use the old rails 2 syntax. But that's not a very strong argument. I'm fine leaving this as `scope`.

vendor/engines/split_accounts/app/models/concerns/account_extension.rb
jhanggi: I don't think that it will in this case, but I'm not sure. Even if it's within a `Concerns` namespace, it should probably still be in the `SplitAccounts` namespace.

vendor/engines/split_accounts/app/models/split.rb
jhanggi: I like `SplitAccounts::Split` if things are still clean with that.

vendor/engines/split_accounts/spec/factories/split_accounts.rb
jhanggi: Yeah, that sounds like a good idea to me, but that's fine if it's a change for later.

db/schema.rb
```diff
@@ -494,6 +494,16 @@
...
   add_index "schedules", ["facility_id"], :name => "i_schedules_facility_id"
 
+  create_table "splits", :force => true do |t|
+    t.integer "parent_split_account_id",                               :null => false
+    t.integer "subaccount_id",                                         :null => false
+    t.decimal "percent",                 :precision => 6, :scale => 3, :null => false
+    t.boolean "extra_penny",                                           :null => false
+  end
+
+  add_index "splits", ["parent_split_account_id"], :name => "index_splits_on_parent_split_account_id"
```
jhanggi: We have foreigner installed in this project, so we've been doing foreign keys as well as indexes when possible to keep referential integrity. We might want to add those here as well (I'd be fine if that's a separate PR).

db/schema.rb
```diff
@@ -510,6 +510,16 @@
 
   add_index "schedules", ["facility_id"], :name => "i_schedules_facility_id"
 
+  create_table "splits", :force => true do |t|
```
jhanggi: I thought usually when you namespaced a model, it would use the namespace in the table name, otherwise you need to explicitly declare `self.table_name =` in the model.

It might be nice to namespace this table.

jhanggi:
I just have one comment about the table name. Otherwise, looks good.

PR #404, opened 2015-11-23
Treat the full_title as an empty string when blank

drench:
This should prevent titles like " - NU Core" slipping through. It's something I noticed on `/facilities/:facility:id/accounts/user/:user_id` but could be elsewhere.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -9,7 +9,7 @@ module ApplicationHelper
   end
 
   def html_title(title=nil)
-    full_title = title.nil? ? "" : "#{title} - "
+    full_title = title.blank? ? "" : "#{title} - "
     (full_title + app_name).html_safe
   end
 


jhanggi:
The code change looks fine, but what page are you seeing this on? I'm not seeing it anywhere, and I can't find that route.

drench:
This is where it's defined in the routes: https://github.com/tablexi/nucore-open/blob/a1735e3a873584a8b1a603959f07e332dcf265a1/config/routes.rb#L21

In the current master branch you can arrive at that page after successful account creation.

Begin here as a global admin: `/facilities/example/accounts/new_account_user_search`. Search for an existing user and select it. That takes you to `/facilities/example/accounts/new?owner_user_id=:id`. Create a new chart string. On success it redirects you to `user_accounts_path`. This route and page look neglected if not obsolete so I changed the redirect in #397 to `facility_user_accounts_path`.

But circling back… the page title is " - NU Core" and with this patch it reads "NU Core".

jhanggi:
:+1:

PR #405, opened 2015-11-24
Create order_detail_decorator

benlinton:
I ran into a bug where the `app/views/facility_orders/index.html.haml` view crashed when `order_status` wasn't available on a newly seeded database. 

**Updated:** I removed the `order_detail_decorator` to make way for @drench's `presenter`.

### DIFF
--- a/app/views/facility_orders/index.html.haml
+++ b/app/views/facility_orders/index.html.haml
@@ -36,8 +36,8 @@
             %td= order.user.full_name
             %td.centered= od.quantity
             = render :partial => 'shared/order_detail_cell', :locals => { :od => od }
-            %td= od.assigned_user.nil? ? '' : od.assigned_user.full_name
-            %td= od.order_status.name
+            %td= od.assigned_user
+            %td= od.order_status
             %td.currency
               - od.send(:extend, PriceDisplayment)
               = od.wrapped_total
--- a/lib/tasks/demo_seed.rake
+++ b/lib/tasks/demo_seed.rake
@@ -402,17 +402,20 @@ namespace :demo do
         created_at = (ordered_at - (60 * rand(60) + 1))
         group_id   = (o.order_details.collect { |od| od.group_id || 0 }.max || 0) + 1
         product.bundle_products.each do |bp|
-          od = OrderDetail.create!(created_by: o.user.id,
-                                   order_id: o.id,
-                                   product_id: bp.product_id,
-                                   actual_cost: rand(2),
-                                   actual_subsidy: 0,
-                                   estimated_cost: rand(2),
-                                   estimated_subsidy: 0,
-                                   quantity: bp.quantity,
-                                   created_at: created_at,
-                                   bundle_product_id: product.id,
-                                   group_id: group_id)
+          od = OrderDetail.create!(
+            created_by: o.user.id,
+            order_id: o.id,
+            product_id: bp.product_id,
+            actual_cost: rand(2),
+            actual_subsidy: 0,
+            estimated_cost: rand(2),
+            estimated_subsidy: 0,
+            quantity: bp.quantity,
+            created_at: created_at,
+            bundle_product_id: product.id,
+            group_id: group_id,
+            order_status_id: bp.product.initial_order_status.id,
+          )
           od.account = account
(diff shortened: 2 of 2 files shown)


jhanggi:
Do we need to fix something in the seeds? I like the approach you took, but in this instance, I'm a little hesitant to silently eat the problem. If we're missing an OrderStatus, that's a serious problem. I'm actually a little surprised we don't have a foreign key constraint on `order_details.order_status_id`.

spec/decorators/order_detail_decorator_spec.rb
```diff
@@ -0,0 +1,49 @@
+require "rails_helper"
+
+RSpec.describe OrderDetailDecorator, type: :decorator do
+  let(:decorated) { OrderDetailDecorator.new(order_detail) }
+
+  describe "delegates to assigned_user" do
+    let(:order_detail) do
+      OrderDetail.new.tap do |order_detail|
```
jhanggi: You could actually use `object_double(OrderDetail.new, assigned_user: :assigned_user)`. I think you could also use `instance_double`. I don't fully understand the difference.

benlinton:
I'm not sure if this is by design or by mistake, but `OrderDetail` model has many optional foreign keys:

* parent_order_detail_id
* price_policy_id
* assigned_user_id
* response_set_id
* account_id
* dispute_by_id
* order_status_id
* group_id
* bundle_product_id
* statement_id
* product_accessory_id

I'm not sure if I have enough project knowledge to say which of these should be non-null/strict foreign keys.  Is this a can of worms outside the scope of this PR?

benlinton:
Also, outside the scope of this PR.  But I would like to move the `demo` seeds into the `db/seeds` directory via the `seedbank` gem.  https://github.com/james2m/seedbank

spec/decorators/order_detail_decorator_spec.rb
benlinton: Good tip!  Thanks

benlinton:
Ready for re-review.  I updated the seeds without messing with foreign keys yet.  I'd like to clean up the seeds and foreign keys in another PR, after I get the `split_account` stuff mostly in working order.

jhanggi:
I believe all of those foreign keys should be nullable with the exception of order_status_id.

app/decorators/order_detail_decorator.rb
```diff
@@ -0,0 +1,6 @@
+class OrderDetailDecorator < SimpleDelegator
+
+  delegate :full_name, to: :assigned_user, prefix: :assigned_user, allow_nil: true
+  delegate :name, to: :order_status, prefix: :order_status, allow_nil: true
```
jhanggi: Looking at OrderStatus, we might not even need this because it already defines to_s as name. So in HAML, = order_detail.order_status should render out to the status's name already and handle the nil case as well.

app/decorators/order_detail_decorator.rb
```diff
@@ -0,0 +1,6 @@
+class OrderDetailDecorator < SimpleDelegator
+
+  delegate :full_name, to: :assigned_user, prefix: :assigned_user, allow_nil: true
```
jhanggi: `User#to_s` is defined as `full_name`, so you might be able to get away with the same thing here where we just say `od.assigned_user` and it gives us the full name or `""` if it's nil.

app/decorators/order_detail_decorator.rb
benlinton: Yeah, I guess we could use `to_s`.  I noticed `to_s` and thought it would be less clear of what was going on in the view.  But now that I think about it some more, I guess it's not really a code small to use the overridden `to_s` and forego the decorator.  Do you think we'll need this decorator for OrderDetail in the future, I assumed yes since the model is so fat.  But haven't taken the time to verify.

jhanggi: I think it might be worth having a conversation with Dan since he has a similar--but more full featured-- presenter in #397.

benlinton:
I removed the order_detail_decorator to make way for @drench's presenter.

jhanggi:
:+1:

PR #406, opened 2015-12-01
[#103993] Pull up changes for credit cards

jhanggi:
These are changes I made in the NU branch that should probably exist in open/master as well.


drench:
:+1:

PR #407, opened 2015-12-07
[#98883] Add UI for managing global user roles

drench:
This adds a UI, available only to Global Administrators, for granting and revoking global user roles.
The UI is based on the one in place for facility user roles.

Notes:

* A user may not grant or revoke roles for itself. It would be easy to lock yourself out, and would complicate the controller actions.
* A user may have multiple global roles, though it might result in some odd behavior. I've done some testing, but it can use more.
  - In UIC, there are Global Admins who are also Billing Administrators. This combination seems to work all right with UIC's feature flags but I don't know if I've missed something.
  - The new Account Manager role is still incompatible with the Global Administrator role, in that the tabs for the Global Administrator will take precedence. Still, the various functions allowed by both roles should work, even if it's difficult or impossible to reach them through navigation. Still, the UI will not stop you from granting this role combination. If we need to make this combination work, we need to discuss further.
* This adds some validations on `UserRole` that might make existing data technically invalid, in that we have some duplicate `UserRole`s and nonexistent role names in production. Since granting/removing roles always involves removing all global roles first, I don't expect this to be a problem.

### DIFF
--- a/app/assets/javascripts/app/user_roles.js.coffee
+++ b/app/assets/javascripts/app/user_roles.js.coffee
@@ -1,0 +1,1 @@
+$ -> $("select[multiple].js--chosen").chosen()
--- a/app/assets/javascripts/app/user_roles.js.coffee
+++ b/app/assets/javascripts/app/user_roles.js.coffee
@@ -1,1 +1,0 @@
-$ -> $("select[multiple].js--chosen").chosen()
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -1,0 +1,19 @@
+class GlobalUserRolesController < GlobalSettingsController
+
+  def index
+    @users = UserPresenter.wrap(User.with_global_roles)
+  end
+
+  def destroy
+    user = User.find(params[:id])
+    if (user == current_user)
+      flash[:error] = "nope"
+    else
+      user.user_roles.global.each(&:destroy)
+      flash[:notice] = "yep"
+    end
+
+    redirect_to global_user_roles_url
+  end
+
+end
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -30,32 +30,28 @@ class GlobalUserRolesController < GlobalSettingsController
     if @user == current_user
       flash[:error] = translate("self_not_allowed", action: "change")
     else
-      assign_global_role!
-      flash[:notice] =
-        translate("update.success", user: @user.username, role: role_name)
+      assign_global_roles!
+      flash[:notice] = translate("update.success", user: @user.username)
     end
   rescue ActiveRecord::RecordInvalid
-    flash[:error] =
-      translate("update.failure", user: @user.username, role: role_name)
+    flash[:error] = translate("update.failure", user: @user.username)
   ensure
     redirect_to global_user_roles_url
   end
 
   private
 
-  def assign_global_role!
+  def assign_global_roles!
     @user.transaction do
       @user.user_roles.global.destroy_all
-      UserRole.create!(user: @user, role: role_name)
+      params[:roles].each do |role_name|
+        UserRole.create!(user: @user, role: role_name)
+      end
     end
   end
 
   def load_user
     @user = User.find(params[:id])
-  end
-
-  def role_name
-    @role_name ||= params[:user_role][:role]
   end
 
   def translate(key, arguments = {})
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -65,10 +65,6 @@ class GlobalUserRolesController < GlobalSettingsController
     @user = User.find(params[:id])
   end
 
-  def roles_from_params
-    params[:roles].presence || []
-  end
-
   def translate(key, arguments = {})
     I18n.t("global_user_roles.#{key}", arguments)
   end
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -20,7 +20,28 @@ class GlobalUserRolesController < GlobalSettingsController
     redirect_to global_user_roles_url
   end
 
+  def edit
+    @user = UserPresenter.new(User.find(params[:id]))
+  end
+
   def search
+  end
+
+  def update
+    user = User.find(params[:id])
+    if user == current_user
+      flash[:error] = t("global_user_roles.update.self_not_allowed")
+    else
+      role_name = params[:user_role][:role]
+      user.transaction do
+        user.user_roles.global.each(&:destroy)
+        UserRole.create!(user: user, role: role_name)
+      end
+      flash[:notice] =
+        t("global_user_roles.update.success", user: user.username, role: role_name)
+    end
+
+    redirect_to global_user_roles_url
   end
 
 end
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -6,14 +6,21 @@ class GlobalUserRolesController < GlobalSettingsController
 
   def destroy
     user = User.find(params[:id])
-    if (user == current_user)
-      flash[:error] = "nope"
+    if user == current_user
+      flash[:error] = t("global_user_roles.destroy.self_not_allowed")
     else
       user.user_roles.global.each(&:destroy)
-      flash[:notice] = "yep"
+      if user.user_roles.global.empty?
+        flash[:notice] = t("global_user_roles.destroy.success", username: user.username)
+      else
+        flash[:error] = t("global_user_roles.destroy.failure", username: user.username)
+      end
     end
 
     redirect_to global_user_roles_url
+  end
+
+  def search
   end
 
 end
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -1,47 +1,64 @@
 class GlobalUserRolesController < GlobalSettingsController
 
+  before_filter :load_user, only: [:destroy, :edit, :update]
+
   def index
     @users = UserPresenter.wrap(User.with_global_roles)
   end
 
   def destroy
-    user = User.find(params[:id])
-    if user == current_user
-      flash[:error] = t("global_user_roles.destroy.self_not_allowed")
+    case
+    when @user == current_user
+      flash[:error] = translate("self_not_allowed", action: "remove")
+    when @user.user_roles.global.destroy_all
+      flash[:notice] = translate("destroy.success", user: @user.username)
     else
-      user.user_roles.global.each(&:destroy)
-      if user.user_roles.global.empty?
-        flash[:notice] = t("global_user_roles.destroy.success", username: user.username)
-      else
-        flash[:error] = t("global_user_roles.destroy.failure", username: user.username)
-      end
+      flash[:error] = translate("destroy.failure", user: @user.username)
     end
 
     redirect_to global_user_roles_url
   end
 
   def edit
-    @user = UserPresenter.new(User.find(params[:id]))
+    @user = UserPresenter.new(@user)
   end
 
   def search
   end
 
   def update
-    user = User.find(params[:id])
-    if user == current_user
-      flash[:error] = t("global_user_roles.update.self_not_allowed")
+    if @user == current_user
+      flash[:error] = translate("self_not_allowed", action: "change")
     else
-      role_name = params[:user_role][:role]
-      user.transaction do
-        user.user_roles.global.each(&:destroy)
-        UserRole.create!(user: user, role: role_name)
-      end
+      assign_global_role!
       flash[:notice] =
-        t("global_user_roles.update.success", user: user.username, role: role_name)
+        translate("update.success", user: @user.username, role: role_name)
     end
-
+  rescue ActiveRecord::RecordInvalid
+    flash[:error] =
+      translate("update.failure", user: @user.username, role: role_name)
+  ensure
     redirect_to global_user_roles_url
   end
 
+  private
+
+  def assign_global_role!
+    @user.transaction do
+      @user.user_roles.global.destroy_all
+      UserRole.create!(user: @user, role: role_name)
+    end
+  end
+
+  def load_user
+    @user = User.find(params[:id])
+  end
+
+  def role_name
+    @role_name ||= params[:user_role][:role]
+  end
+
+  def translate(key, arguments = {})
+    I18n.t("global_user_roles.#{key}", arguments)
+  end
 end
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -7,15 +7,7 @@ class GlobalUserRolesController < GlobalSettingsController
   end
 
   def destroy
-    case
-    when @user == current_user
-      flash[:error] = translate("self_not_allowed", action: "remove")
-    when @user.user_roles.global.destroy_all
-      flash[:notice] = translate("destroy.success", user: @user.username)
-    else
-      flash[:error] = translate("destroy.failure", user: @user.username)
-    end
-
+    destroy_all_global_roles
     redirect_to global_user_roles_url
   end
 
@@ -32,7 +24,10 @@ class GlobalUserRolesController < GlobalSettingsController
   end
 
   def update
-    if @user == current_user
+    case
+    when params[:roles].blank?
+      destroy_all_global_roles
+    when @user == current_user
       flash[:error] = translate("self_not_allowed", action: "change")
     else
       assign_global_roles!
@@ -55,8 +50,23 @@ class GlobalUserRolesController < GlobalSettingsController
     end
   end
 
+  def destroy_all_global_roles
+    case
+    when @user == current_user
+      flash[:error] = translate("self_not_allowed", action: "remove")
+    when @user.user_roles.global.destroy_all
+      flash[:notice] = translate("destroy.success", user: @user.username)
+    else
+      flash[:error] = translate("destroy.failure", user: @user.username)
+    end
+  end
+
   def load_user
     @user = User.find(params[:id])
+  end
+
+  def roles_from_params
+    params[:roles].presence || []
   end
 
   def translate(key, arguments = {})
--- a/app/controllers/global_user_roles_controller.rb
+++ b/app/controllers/global_user_roles_controller.rb
@@ -20,7 +20,12 @@ class GlobalUserRolesController < GlobalSettingsController
   end
 
   def edit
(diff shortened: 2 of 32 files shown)


app/models/user_role.rb
```diff
@@ -30,8 +30,16 @@ def self.facility_management_roles
...
+    account_manager + administrator + billing_administrator
+  end
+
   def self.valid_roles
-    account_manager + administrator + billing_administrator + facility_roles
+    global_roles + facility_roles
+  end
+
+  def self.global
+    where(facility_id: nil)
```
jhanggi: Would this be better as `where(role: global_roles)`? I just know there are some weird cases in UIC where there is an `Administrator` role with an `facility_id`. I believe these were added by hand, but it seems like the role name should take precedence over `facility_id: nil`. I could be swayed by a good argument, though.

app/models/user_role.rb
```diff
@@ -49,4 +57,17 @@ def self.grant(user, role, facility = nil)
...
+    scope: [:facility_id, :user_id],
+    message: I18n.t("activerecord.errors.models.user_role.duplicate")
+  validates_with UserRoleFacilityValidator
+
+  def facility_role?
+    self.class.facility_roles.include?(role)
+  end
+
+  def global_role?
+    !facility_role?
```
jhanggi: (somewhat related to my last comment above), it seems like we should use the same logic for `global_role?` as the scope `global`

app/presenters/user_presenter.rb
```diff
@@ -0,0 +1,31 @@
+class UserPresenter < SimpleDelegator
+  include ActionView::Helpers::FormOptionsHelper
+
+  delegate :global, to: :user_roles, prefix: true
+
+  def self.wrap(users)
+    users.map { |user| new(user) }
```
jhanggi: I'm not sure it's going to be an issue here, but it's sometimes nice to have these kind of things be lazy. Since `users` is likely an AR::Relation, it's lazy, it would be nice to keep this wrapper function lazy as well so it doesn't hit the database until it absolutely has to.

app/presenters/user_presenter.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  def global_role_list
+    global_roles.join(", ")
+  end
+
+  def global_role_select_options
+    options_for_select(UserRole.global_roles, selected: user_roles.map(&:role))
+  end
+
+  def name_last_comma_first
```
jhanggi: Why are we displaying last, first? I don't think we do that anywhere else on the site.

app/presenters/user_presenter.rb
```diff
@@ -0,0 +1,31 @@
+class UserPresenter < SimpleDelegator
```
jhanggi: One way I like to think about decorators/presenters is that they're tied to a specific context. For example, this one is a wrapper around User for displaying of roles, but many of these methods would not be useful in other contexts. Something to continue, but this could be `UserUserRolePresenter` or possibly organize by the thing being decorated: `User::GlobalRolePresenter` or something along those lines. I don't feel strongly, but it's something to consider.

app/views/global_user_roles/edit.html.haml
```diff
@@ -0,0 +1,27 @@
...
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_global", sidenav_tab: "global_user_roles"
+
+%h2= t("pages.global_user_roles")
+
+= form_for(:user_role, url: global_user_role_path(@user.id), method: :put) do |form|
+  = form.error_messages
+  %ul.form
+    %li
```
jhanggi: This is nitpicky, and I realize we have it as well on the facility staff UI, but can we get rid of the `li` or at least the bullets?

app/views/global_user_roles/edit.html.haml
```diff
@@ -0,0 +1,27 @@
+= content_for :h1 do
```
jhanggi: Is there a way we can share a good chunk of the form with `map_users`?

app/views/search/_global_user_role_link.html.haml
```diff
@@ -0,0 +1 @@
+= link_to user.name_last_comma_first, edit_global_user_role_path(user.id)
```
jhanggi: I don't see this partial being used anywhere. Is it?

spec/controllers/affiliates_controller_spec.rb
```diff
@@ -5,7 +5,7 @@
   render_views
 
   before(:all) { create_users }
-
+  before(:each) { @authable = create(:facility) }
```
jhanggi: Where did this become necessary? Doesn't look like you're touching this controller at all.

spec/controllers/global_user_roles_controller_spec.rb
```diff
@@ -0,0 +1,142 @@
+require "rails_helper"
+
+RSpec.describe GlobalUserRolesController do
+  shared_examples_for "the user must log in" do
```
jhanggi: Should we consider pulling these up into a more global shared_examples?

spec/presenters/user_presenter_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+  shared_examples_for "it has multiple global roles" do
+    it { expect(global_role_list).to eq("Administrator, Billing Administrator") }
+    it { expect(global_role_select_options).to include('selected="selected">Administrator') }
+    it { expect(global_role_select_options).to include('selected="selected">Billing Administrator') }
+  end
+
+  context "when the user has no global roles" do
+    let(:user) { create(:user) }
+
+    it_behaves_like "it has no global roles"
```
jhanggi: When we're only using each of the shared_examples once, it almost becomes a little harder to understand the spec because you have to looks in multiple places, and the wording is kind of weird "when the user has no global roles it has no global roles". sort of becomes a tautology. I'd be inclined towards putting the specs directly within the individual contexts.

app/views/global_user_roles/edit.html.haml
```diff
@@ -0,0 +1,27 @@
...
+      = text_field_tag :name, @user.full_name, disabled: true
+    %li
+      = label_tag :email, nil, class: "require"
+      = text_field_tag :email, @user.email, disabled: true
+    %li
+      = form.label :roles, nil, class: "require"
+      = select_tag :roles, @user.global_role_select_options, multiple: true
+
+  %ul.inline
+    %li= form.submit t(".submit"), class: "btn"
```
jhanggi: I would expect this to be disabled if submitting the form is going to generate an error because I can't edit myself.

app/views/global_user_roles/edit.html.haml
```diff
@@ -0,0 +1,27 @@
...
+      = text_field_tag :username, @user.username, disabled: true
+    %li
+      = label_tag :name, nil, class: "require"
+      = text_field_tag :name, @user.full_name, disabled: true
+    %li
+      = label_tag :email, nil, class: "require"
+      = text_field_tag :email, @user.email, disabled: true
+    %li
+      = form.label :roles, nil, class: "require"
+      = select_tag :roles, @user.global_role_select_options, multiple: true
```
jhanggi: Sort of a matter of taste, but I don't love multi-select boxes like this. Is there a way we can easily make it a series of checkboxes or chosen? I think I'd be more inclined towards checkboxes.

jhanggi: If I deselect all roles and submit, I get an error "undefined method `each' for nil:NilClass" on `params[:roles].each` inside `assign_global_roles!`

app/views/search/_global_user_role_link.html.haml
jhanggi: Now I see. Inside `user_search_results.html.haml` it does a dynamic partial render (which is probably not super safe, but probably not worth fixing right now...).

config/routes.rb
```diff
@@ -274,11 +274,16 @@
     end
   end
 
+  resources :global_user_roles do
+    collection do
```
jhanggi: This can also be done on one line as `get "search", on: :collection`. Not a big deal, though.

app/views/global_user_roles/search.html.haml
```diff
@@ -0,0 +1,20 @@
...
+= form_tag user_search_results_path, id: "ajax_form" do
+  %p.instructions
+    = t(".instructions.search")
+    - if SettingsHelper.feature_on?(:create_users)
+      = t(".instructions.add_user")
+
+  = text_field_tag :search_term, nil, size: 30, class: "search-query"
+  = submit_tag "Search", class: "btn"
+  = hidden_field_tag :search_type, "global_user_role"
+  = hidden_field_tag :create_link,
```
jhanggi: Could this just always be `"true"` here and let the `create_users` partial handle checking the feature flag?

jhanggi: And if not, could we use `SettingsHelper.feature_on?(:create_users).to_s`?

app/models/user_role.rb
drench: I feel like users with `Administrator` with a non-NULL `facility_id` should either become Facility Administrators or Global Administrators. The users who are also Billing Administrators (with no facility) I'm guessing were probably intended to be Global Administrators, at least. There are only a few so maybe we can get clarification from UIC on them.

app/presenters/user_presenter.rb
drench: I wondered about this too. I don't see it being a problem in how it's being used so far (in `GlobalUserRolesController` and `UserFinder`), but I'm sure it's something we'll run into as we do more decorating/presenting in this project. Do you know a good pattern for this? I'd consider putting it in a base presenter class.

spec/controllers/affiliates_controller_spec.rb
drench: It's part of 202f349d8262a2a2f0c414996754567ec1f84aae. Specs like this using `controller_spec_helper.rb` indirectly create Facility Staff user roles, and if `@authable` isn't defined, the `UserRole`s are created without a Facility. These would no longer pass validation, and the specs fail.

spec/controllers/global_user_roles_controller_spec.rb
drench: Sounds good to me since it's a step toward phasing out `controller_spec_helper.rb`.

app/views/global_user_roles/edit.html.haml
drench: A user should only end up in this situation by editing the URL manually, so I lean toward the controller not allowing the user to be here at all.

app/models/user_role.rb
drench: Having said that, we're defining roles by name otherwise so it makes sense to query by name here.

app/views/global_user_roles/edit.html.haml
drench: I think we could, though one hitch is we only allow one facility role per user, so the role `<select>` would be different (maybe *very* different if we go with checkboxes for the global UI).

app/views/global_user_roles/search.html.haml
drench: The `_create_new_user` partial checks the hidden field AND the feature flag, so setting this to `true` here in all cases is fine, if maybe misleading. I'd like to leave the feature check here to make the intent more explicit.
Interestingly `hidden_field_tag` takes care of the stringification so no `#to_s` necessary.

app/views/global_user_roles/edit.html.haml
drench: I got the `map_users` partial closer to `global_user_roles/edit` here 9cf4b08534ec6212e0241065a2db02ab6ea2f223 but did not try to extract common parts. If we go with allowing multiple facility roles for a user, nearly the entire form would be the same. Want to leave this as a `TODO`?

app/views/global_user_roles/edit.html.haml
jhanggi: I think that's fine. I just discovered it because I was able to access the form while editing myself.

spec/controllers/affiliates_controller_spec.rb
jhanggi: Gotcha. Makes sense.

app/views/global_user_roles/edit.html.haml
jhanggi: I think that's fine leaving it as a TODO for now.

app/presenters/user_presenter.rb
jhanggi: I would think you would return some sort of lazy object that would only instantiate the `map` when some other enumerable was called on it. Something like this. Not tested at all, and probably not worth bringing in right now.

```ruby
class DecoratedCollection
  attr_reader :collection, :decorator

  delegate :each, to: :collection
  include Enumerable

  def initialize(collection, decorator)
    @collection = collection
    @decorator = decorator
  end

  def wrapped_collection
    @wrapped_collection ||= collection.map { |obj| decorator.new(obj) }
  end
end
```
This is not unlike what Draper does: https://github.com/drapergem/draper/blob/master/lib/draper/collection_decorator.rb

app/controllers/global_user_roles_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  def search
+  end
+
+  def update
+    case
+    when params[:roles].blank?
+      destroy_all_global_roles
+    when @user == current_user
```
jhanggi: I'm guessing we won't get to this point because it's prevented by the `edit` rules, but should this check happen before the `params[:roles].blank?` case so a user _definitely_ can't remove their own permissions.

app/controllers/global_user_roles_controller.rb
```diff
@@ -0,0 +1,75 @@
...
+      @user = UserPresenter.new(@user)
+    end
+  end
+
+  def search
+  end
+
+  def update
+    case
+    when params[:roles].blank?
```
jhanggi: Is there a reason this needs to be a special case here vs something like `params.fetch(:roles, []).each` or `Array(params[:roles]).each` below?

jhanggi: Or your `roles_from_params` method you already have, which is functionally equivalent to these.

jhanggi:
The create new user link that pops up tries to take you to http://localhost:3000/facilities/all/users/new, which is inaccessible (403) as a global admin.

app/controllers/global_user_roles_controller.rb
drench: The `blank?` case is there because updating while selecting no global roles is equivalent to destroying all global roles. It's mainly there so the user will see a more appropriate `flash` afterward. That is, if I deselect all global user roles and save successfully, I will see the `destroy.success` message instead of `update.success`.

`roles_from_params` was an accidental commit of something I had in a WIP. I'm removing it.

drench:
Yes, `/facilities/all/users/new` lives under the cross-facility Users tab where only Account Managers are allowed. Should this be allowed, even for Account Managers? I'm now wondering if user creation should live only under facility-specific management. What do you think?

jhanggi:
Account Managers need to be able to add users, so it will need to be available in the cross-facility context. If it's overly-difficult to get this cross-facility user adding for global admins, I might be ok if you can't create users in the global roles context, only modify their roles.

app/views/global_user_roles/search.html.haml
```diff
@@ -0,0 +1,18 @@
+= content_for :h1 do
+  = t("pages.global_settings")
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_global", sidenav_tab: "global_user_roles"
+
+%h2= t(".h2")
+= form_tag user_search_results_path, id: "ajax_form" do
```
drench: Just want to note for future cleanup I ran into several very similar forms (and this PR is responsible for adding one of them):

* [app/views/account_users/user_search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/account_users/user_search.html.haml)
* [app/views/facility_account_users/user_search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/facility_account_users/user_search.html.haml)
* [app/views/facility_accounts/new_account_user_search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/facility_accounts/new_account_user_search.html.haml)
* [app/views/facility_accounts/search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/facility_accounts/search.html.haml)
* [app/views/facility_accounts/user_search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/facility_accounts/user_search.html.haml)
* [app/views/facility_users/search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/facility_users/search.html.haml)
* [app/views/global_user_roles/search.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/global_user_roles/search.html.haml)
* [app/views/product_users/new.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/product_users/new.html.haml)
* [app/views/user_price_group_members/new.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/user_price_group_members/new.html.haml)
* [app/views/users/_search_form.html.haml](https://github.com/tablexi/nucore-open/tree/21b9c1f28f769060d018848c98edda99798a6c47/app/views/users/_search_form.html.haml)

drench:
This should be ready for another lookover and testing round.

I set up the global user role selection to use chosen (it looked nicer to me and was easier to implement than checkboxes). Global admins should no longer see the create user link in a cross-facility/no-facility context. I'd like to allow it, but I'm not sure how we can do this easily without opening up the cross-facility "Users" tab to them.

app/assets/javascripts/app/user_roles.js.coffee
```diff
@@ -0,0 +1 @@
+$ -> $("select[multiple].js--chosen").chosen()
```
jhanggi: Seems like the filename could be a little more generic since it will technically run on every page. And I think we could even just leave it as `.js--chosen` without the `select[multiple]`.

Now's probably not the time to clean up all the very specific instantiations of chosen (e.g. `$('#product_add').chosen()` in manage_orders.coffee and add the class, but something we could think about doing.

drench:
Autofill on Chrome should be disabled, and any element with the class `.js--chosen` should have chosen applied (currently it's only on one field in this new form for editing global user roles).

jhanggi:
:+1:

PR #409, opened 2015-12-07
Make account owner trait share-able among subclasses

jhanggi:
I’m doing this so I can use the `with_owner` trait on purchase orders and credit cards easily within  my credit card branch.

### DIFF
--- a/spec/factories/accounts.rb
+++ b/spec/factories/accounts.rb
@@ -1,0 +1,14 @@
+FactoryGirl.define do
+  trait :with_account_owner do
+    transient do
+      owner { FactoryGirl.create(:user) }
+    end
+
+    # Every account must have an account_user "owner" in order for the account
+    # to be valid in rails. And foreign key constraints require that each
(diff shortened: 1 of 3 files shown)


spec/factories/accounts.rb
```diff
@@ -0,0 +1,14 @@
+FactoryGirl.define do
+  trait :with_account_owner do
```
drench: Interesting having a factory with only a trait. It works like I'd hoped in that you cannot create generic `Accounts` but the trait is there in the subclasses. Neat.

drench:
:+1:

spec/factories/accounts.rb
jhanggi: Technically, the trait is global, so you could apply it to any other factory instantiation now. However, that would clearly break if you tried to `FactoryGirl.create(:user, :with_account_owner)`, but because it's global, that's why I called it `:with_account_owner` instead of the shorter `:with_owner`.

PR #410, opened 2015-12-09
[#120455] Split accounts engine payment source

benlinton:
Support creating, updating, and showing a `split_account` account type as a payment source.

Installed the `nested_form_fields` gem to easily add multiple `splits` on a new `split_account`.  

The `new` Account views now use a set of nav tabs rather than a select dropdown to choose account type. This means we no longer need to load an generic `Account` object for the `new` action, and can get rid of hacks to support using `Account.new` for the `new` form.  We can now load the specific subclassed account_type (i.e. `NufsAccount`) for new and create which seems like a cleaner approach.

Since multiple engines are now adding account types, converted the `AccountManager:: GLOBAL_ACCOUNT_CLASSES`-like constants to instance variables in the new `AccountConfig` class that are append-able by multiple engines.  Deprecated `app/support/account_manager.rb` in favor of putting account_type related methods in the `AccountConfig` class.

Move `app/support/account_builder.rb` to services folder and refactored it.  Inherit the `AccountBuilder` service object within engines that add new account types.  Began removing coupled code and tests from the main app that actually belong in engines.

I also started refactoring additional legacy code, but removed it from this PR in order to keep the PR more focused/shorter.


app/controllers/facility_accounts_controller.rb
```diff
@@ -1,35 +1,12 @@
...
-      # be verbose with failures. Too many tasks (#29563, #31873) need it
-      begin
-        account.set_expires_at!
-        account.errors.add(:base, I18n.t('controllers.facility_accounts.create.expires_at_missing')) unless account.expires_at
-      rescue AccountNumberFormatError => e
-        account.expires_at = Time.zone.now # Prevent expires_at missing message
-      rescue ValidatorError => e
-        account.errors.add(:base, e.message)
-      end
-    end
```
benlinton: I attempted to remove `configure_new_account` in favor of using the new `app/services/account_builder.rb`.  But I'm not sure I successfully supported all use cases.  Please review extra carefully.  Thanks!

app/controllers/facility_accounts_controller.rb
```diff
@@ -39,9 +16,9 @@ def configure_new_account(account)
   before_filter :init_current_facility
   before_filter :init_account
 
-  authorize_resource :class => Account
+  authorize_resource class: Account
 
-  before_filter :check_billing_access, :only => billing_access_checked_actions
+  before_filter :check_billing_access, only: check_billing_access_actions
```
benlinton: `check_billing_access_actions` is now populated by a class variable so multiple engines can concat it.

app/controllers/facility_accounts_controller.rb
```diff
@@ -79,13 +73,12 @@ def edit
 
   # PUT /facilities/:facility_id/accounts/:id
   def update
-    class_params = account_class_params
-
     if @account.is_a?(AffiliateAccount)
-      class_params[:affiliate_other] = nil if class_params[:affiliate_id] != Affiliate.OTHER.id.to_s
+      @account.affiliate_other = nil if params[:affiliate_id] != Affiliate.OTHER.id.to_s
     end
```
benlinton: Can someone explain how affiliates work exactly?  For example, `affiliate_other` vs `affiliate_id`.  And why are affiliates only supported on facility accounts?  

I think I successfully mimicked the existing affiliate functionality, but I can't with certainty that I fully understand affiliates.

app/controllers/facility_accounts_controller.rb
```diff
@@ -93,22 +86,6 @@ def update
...
   end
```
benlinton: TODO: start using strong params here.

app/models/account.rb
```diff
@@ -38,60 +38,82 @@ def price_groups
...
-      )
-      # set (new?) role
-      @account_user.user_role = new_role
-      # set creation information
-      @account_user.created_by = session_user.id
+  # Returns an array of subclassed Account objects.
+  # Engines can concat to this on initialization.
+  def self.account_types
+    @@account_types ||= [NufsAccount]
+  end
```
benlinton: I started using class variables so multiple engines can easily append to this list.  Ideally I'll convert this list of objects to a list of class name strings before merging to master.

app/models/account.rb
```diff
@@ -38,60 +38,82 @@ def price_groups
...
+  # TODO: rename to scoped_to_single_facility?
+  def limited_to_single_facility?
+    facility_account_types.map(&:to_s).include?(self.name)
   end
 
-  def self.cross_facility?
+  # TODO: shouldn't this use `global_account_types.include?(self.name)` instead?
+  def cross_facility?
     !limited_to_single_facility?
   end
```
benlinton: I kept the existing functionality so far, but shouldn't this use `global_account_types.include?(self.name)` instead?

app/models/account.rb
```diff
@@ -38,60 +38,82 @@ def price_groups
...
+  # Engines can concat to this on initialization.
+  def self.facility_account_types
+    @@facility_account_types ||= []
+  end
 
-    return @account_user
+  # Returns an array of subclassed Account objects availalbe for statements.
+  def self.statement_account_types
+    facility_account_types
   end
```
benlinton: Why are `facility_account_types` equivalent to `statement_account_types`?

app/models/account.rb
```diff
@@ -38,60 +38,82 @@ def price_groups
...
+    account_types.size > 1
+  end
+
+  def self.using_statements?
+    statement_account_types.present?
+  end
+
+  def self.using_affiliates?
+    using_statements?
+  end
```
benlinton: There's a good chance we can deprecate these methods (`using_statements?`, `using_affiliates?`, etc.) or move them out of the Account model.  They do not feel like they belong in the Account model.

app/models/account.rb
```diff
@@ -277,6 +299,39 @@ def description_to_s
...
+      @account_user.created_by = session_user.id
+
+      self.account_users << @account_user
+
+      raise ActiveRecord::Rollback unless self.save
+    end
+
+    return @account_user
+  end
+
```
benlinton: I simply moved this chunk of code, it hasn't been modified.

app/services/account_builder.rb
```diff
@@ -0,0 +1,93 @@
...
+    account.created_by = current_user.id
+    account
+  end
+
+  # Set the faciltiy for an account.
+  # Returns the modified account object.
+  def build_facility(account, facility)
+    account.facility_id = facility.id if facility.present?
+    account
+  end
```
benlinton: TODO: add another check to see if this account_type can even have a facility assigned.

app/services/account_builder.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  # TODO: figure out how to refactor existing expires_at code. See both
+  # `configure_new_account` and `set_expires_at!`. Those methods should get
+  # deprecated.
+  def build_nufs_account_expires_at(account, params)
+    account.set_expires_at!
+    account
+  end
```
benlinton: Should we build the expiration date using account_type-specific AccountBuilder methods.  Or instead should every Account subclass support a `set_expires_at` method?

app/services/account_builder.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  # Returns a valid subclassed Account object; unless account_type is invalid,
+  # then raises an AccountController::RoutingErro exception.
+  def valid_account_type!(account_type, facility)
+    account_types = Account.creatable_account_types_for_facility(facility).map(&:to_s)
+    return account_type if account_types.include?(account_type)
+    raise ActionController::RoutingError, "invalid account_type: #{account_type}"
+  end
+
+end
```
benlinton: I originally started off having engines `prepend` this class and then call `super`, but this approach seemed a lot cleaner and solved allowing engines to extend the account builder (rather than adding a lot of extension bloat to controllers).

app/support/order_uncanceler.rb
```diff
@@ -31,10 +31,10 @@ def uncancel_to_complete(order_detail)
...
         order_detail.save!
-        puts "#{order_detail} was uncanceled"
+        Rails.logger.info "OrderDetail #{order_detail} was uncanceled"
         true
       rescue => e
-        puts "Could not save #{order_detail} because #{e.message}"
+        Rails.logger.info "Could not save OrderDetail #{order_detail} because #{e.message}"
         raise ActiveRecord::Rollback
       end
     end
```
benlinton: Was there a reason we were using `puts` here?  If `Rails.logger` is acceptable, that'll prevent the specs from printing a bunch of `puts` output.

app/views/facility_accounts/_account_type_tabs.html.haml
```diff
@@ -0,0 +1,6 @@
+- if available_account_types.present? && available_account_types.size > 1
+  %ul.nav.nav-tabs
+    - available_account_types.each do |account_type|
+      %li{class: ("active" if current_account_type == account_type)}
+        = link_to new_facility_account_path(owner_user_id: owner_user.id, account_type: account_type) do
+          = account_type.constantize.model_name.human
```
benlinton: Ditched the `account_type` dropdown for a set of nav tabs.  We might have to revisit the `account.js` once I better understand how affiliates factor in.

lib/account_manager.rb
```diff
@@ -1,54 +1,32 @@
...
 
   module Overridable
     def valid_account_types
-      [ NufsAccount ]
+      Account.account_types
     end
   end
+
   include Overridable
 end
```
benlinton: Began deprecating this file.  Eventually it will go away in favor of a service object, using the STI parent class, and/or etc.

vendor/engines/c2po/app/services/c2po/account_builder_extension.rb
```diff
@@ -0,0 +1,35 @@
...
+    def build_credit_card_account_expires_at(account, params)
+      account.expires_at = Date.civil(account.expiration_year.to_i, account.expiration_month.to_i).end_of_month.end_of_day
+      acount
+    end
+
+    # Sets `expires_at` for PurchaseOrderAccount only.
+    def build_purchase_order_account_expires_at(account, params)
+      account.expires_at = parse_usa_date(account.expires_at).try(:end_of_day)
+      account
+    end
```
benlinton: I __think__ that I captured the `expires_at` setters for both credit card and purchase orders, but I'm not 100% sure.  Please review.

vendor/engines/split_accounts/app/models/split_accounts/account_extension.rb
```diff
@@ -1,11 +1,13 @@
...
-  module AccountExtensions
+  module AccountExtension
 
     extend ActiveSupport::Concern
 
     included do
       has_many :parent_splits, class_name: "SplitAccounts::Split", foreign_key: :subaccount_id, inverse_of: :subaccount
       has_many :parent_split_accounts, through: :parent_splits
+
+      scope :excluding_split_accounts, -> { where("type != ?", "SplitAccounts::Split") }
```
benlinton: Normally with rails 4 I would use `where.not(...)`.  For rails 3.2 will this work, or should I use Arel to better support Oracle?

vendor/engines/split_accounts/app/services/split_accounts/account_builder_extension.rb
```diff
@@ -0,0 +1,22 @@
...
+      account = build_split_account_expires_at(account, params)
+      account
+    end
+
+
+    # Sets `expires_at` for SplitAccounts::SplitAccount only.
+    def build_split_account_expires_at(account, params)
+      account.expires_at = Time.zone.now + 50.years
+      account
+    end
```
benlinton: Call `set_expires_at` if we end up supporting that for SplitAccount model.

vendor/engines/split_accounts/app/views/split_accounts/facility_accounts/_show_splits.html.haml
```diff
@@ -0,0 +1,17 @@
...
+      %tr
+        %th Account
+        %th Percent
+        %th Extra Penny
+    %tbody
+      - account.splits.each do |split|
+        %tr
+          %td= split.subaccount
+          %td= split.percent
+          %td= split.extra_penny?
```
benlinton: TODO: use locales.

app/models/account.rb
```diff
@@ -38,60 +38,82 @@ def price_groups
...
+  # TODO: rename to scoped_to_single_facility?
   def self.limited_to_single_facility?
-    AccountManager::FACILITY_ACCOUNT_CLASSES.include? self.name
+    facility_account_types.map(&:to_s).include?(self.name)
   end
 
+  # TODO: shouldn't this use `global_account_types.include?(self.name)` instead?
   def self.cross_facility?
     !limited_to_single_facility?
   end
```
benlinton: I kept the existing functionality so far, but shouldn't this use `global_account_types.include?(self.name)` instead?

app/models/account.rb
```diff
@@ -26,6 +26,7 @@ def price_groups
   accepts_nested_attributes_for :account_users
 
   scope :active, lambda {{ :conditions => ['expires_at > ? AND suspended_at IS NULL', Time.zone.now] }}
+  scope :filter_global_account_types, -> { where("accounts.type in (:global)", { global: global_account_types.map(&:to_s) }) }
```
jhanggi: I feel like `filter_global_account_types` almost implies that it filters them out. What would you think of simply `:global`? Also, is the `to_s` necessary? I would expect AR to take care of that for you.

benlinton:
Once this gets initially reviewed and we all think it's heading the the right direction.  I'll beef up the tests around the new AccountBuilder and other refactored code.

app/models/account.rb
jhanggi: They just happen to be equivalent by coincidence. They will not be equivalent for Dartmouth. I would consider making it a separate value.

jhanggi: What would you think of a `Account::Config` object? So you would do things like `Account.config.facility_types << :credit_card` in your configurations. It seems like we're adding a whole lot of functionality onto Account.

app/models/account.rb
jhanggi: Probably, but there could probably be a few things done to enforce mutual exclusivity. Theoretically, a mischievous developer could add an account type to both lists which could cause come confusion. We probably don't need to worry about that, but it's something to think about. One option would be to maintain a list of all valid account types and a separate list of either global or scoped types, and then compute the other by `valid_account_types - [global_types|scoped_types]`.

app/models/account.rb
```diff
@@ -88,10 +103,14 @@ def self.for_facility(facility)
     accounts = scoped
 
     if facility.single_facility?
-      accounts = accounts.where("accounts.type in (:allow_all) or (accounts.type in (:limit_one) and accounts.facility_id = :facility)",
-            {:allow_all => AccountManager::GLOBAL_ACCOUNT_CLASSES,
-              :limit_one => AccountManager::FACILITY_ACCOUNT_CLASSES,
-              :facility => facility})
+      accounts = accounts.where(
+        "accounts.type in (:allow_all) or (accounts.type in (:limit_one) and accounts.facility_id = :facility)",
+        {
```
jhanggi: While we're cleaning up the syntax, the curly braces should be able to go away, too.

app/models/account.rb
```diff
@@ -110,6 +129,13 @@ def self.with_orders_for_facility(facility)
     where(id: ids_with_orders(facility))
   end
 
+  # The subclassed Account objects will be cross facility by default; override
+  # this method if the subclassed Account object is always scoped to a
+  # particular facility.
+  def facility
```
jhanggi: Perhaps the comment should say something about overriding using `belongs_to :facility`. That's how we use do in CC/POs.

app/models/nufs_account.rb
```diff
@@ -1,5 +1,6 @@
 class NufsAccount < Account
-  validates_uniqueness_of :account_number, :message => "already exists"
+  validates_uniqueness_of :account_number, message: "already exists"
+  validates_presence_of :expires_at
```
jhanggi: What prompted adding this validation check? Looks to be always be the case, but just curious the impetus.

app/support/order_uncanceler.rb
jhanggi: I'm not sure the reason. `Rails.logger` should be fine. Where does this class get used? If it's within a rake task, we'll want to make sure it still prints out there.

lib/account_manager.rb
jhanggi: Dig it.

spec/controllers/facility_account_users_controller_spec.rb
```diff
@@ -115,8 +115,8 @@
       context 'with a missing owner' do
 
         before :each do
-          @acount_user = @account.owner
-          AccountUser.delete(@acount_user.id)
+          @account_user = @account.owner
```
jhanggi: Good find.

app/views/facility_accounts/account_fields/_nufs_account.html.haml
```diff
@@ -1 +1,6 @@
+- is_new ||= false
```
jhanggi: `||=` with booleans always feel dangerous. In partials I like to use `local_assigns[:is_new]` that'll be nil if it's not set, otherwise it'll be the value you passed in.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_purchase_order_account.html.haml
```diff
@@ -1,9 +1,9 @@
-= f.input :account_number, :label => t('facility_accounts.purchase_orders.account_number'), :required => true
+- is_new ||= false
```
jhanggi: Is `is_new` being used in this partial?

vendor/engines/c2po/app/views/facility_accounts/account_fields/_purchase_order_account.html.haml
```diff
@@ -1,9 +1,9 @@
...
+= f.input :description, required: true
 
-= f.input :description, :required => true
+= render partial: 'facility_accounts/account_fields/affiliate', locals: { f: f }
+= render partial: 'facility_accounts/account_fields/remittance_information', locals: { f: f }
 
-= render :partial => 'facility_accounts/account_fields/affiliate', :locals => { :f => f }
-
-= render :partial => 'facility_accounts/account_fields/remittance_information', :locals => { :f => f }
+= f.input :formatted_expires_at, required: true, input_html: { class: 'datepicker' }, label: Account.human_attribute_name(:expires_at)
```
jhanggi: Is there a reason you reordered these fields?

app/services/account_builder.rb
```diff
@@ -0,0 +1,93 @@
+# Builds an Account object given params.
+# This class supports extensions from engines via the `build_subclass` method.
+class AccountBuilder
```
jhanggi: I like how you've pulled more into this class. I have a few thoughts.

I'm not sure I'm crazy about injecting more methods into this class and dynamically calling them (the `build_nufs_account` etc). I'd actually be slightly more inclined towards creating subclasses and using a template method for `build_subclass`. You could do a factory method here so `AccountBuilder.for(account_class)` that dynamically sets up a `NufsAccountAccountBuilder`--it could default to this `Accountbuilder` if it can't find one by name. That would also give you a place to put the expiration setter as opposed to requiring the Account subclass implement a `set_expires_at!`.

Another thing that might clean this up would be to take the parameters you currently have for `build` and add them to the initializer to set instance variables. That way you're not passing them all over the place. Then you could have a class method to keep a similar public API.

```
def self.build(facility, current_user, owner_user, params)
  new(facility, current_user, owner_user, params).build
end
```

app/controllers/facility_accounts_controller.rb
jhanggi: There's global a table of affiliates managed under "global settings", which is what `affiliate_id` is foreign keyed to. When creating a CC/PO, you can either select one of the existing affiliates from the dropdown, or choose "Other" (which has an entry in the affiliates table) and enter a free-text field (`affiliate_other`). If `affiliate_other is not null`, then `affiliate_id = 1` (as that's the id of "Other").

vendor/engines/c2po/app/views/facility_accounts/account_fields/_credit_card_account.html.haml
```diff
@@ -1,4 +1,10 @@
...
+- is_new ||= false
 
-= render :partial => 'facility_accounts/account_fields/affiliate', :locals => { :f => f }
-= render :partial => 'facility_accounts/account_fields/remittance_information', :locals => { :f => f }
\ No newline at end of file
+= f.input :description, required: true, hint: t('facility_accounts.account_fields.label.all_instruct'), hint_html: { class: 'help-inline' }
+= render partial: 'facility_accounts/account_fields/affiliate', locals: { f: f }
+= render partial: 'facility_accounts/account_fields/remittance_information', locals: { f: f }
+
+- if is_new
```
jhanggi: Please fix the ordering of the fields on this form to match the old version.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_credit_card_account.html.haml
```diff
@@ -1,4 +1,10 @@
-= f.input :description, :required => true, :hint => t('facility_accounts.account_fields.label.all_instruct'), :hint_html => { :class => 'help-inline' }
+- is_new ||= false
 
-= render :partial => 'facility_accounts/account_fields/affiliate', :locals => { :f => f }
-= render :partial => 'facility_accounts/account_fields/remittance_information', :locals => { :f => f }
\ No newline at end of file
+= f.input :description, required: true, hint: t('facility_accounts.account_fields.label.all_instruct'), hint_html: { class: 'help-inline' }
+= render partial: 'facility_accounts/account_fields/affiliate', locals: { f: f }
```
jhanggi: Wasn't entirely sure where to put this note: when I load up one of the CC or PO forms, the "Affiliate Name" should be hidden. It should only appear once you've selected "Other". Seems like the JS is hooked up to the change event since changing the dropdown gives me correct behavior, it's ,just not on page load.

vendor/engines/split_accounts/app/services/split_accounts/account_builder_extension.rb
jhanggi: The expiration date of a split account should be the earliest of all the component accounts' expirations.

vendor/engines/split_accounts/app/views/facility_accounts/account_fields/split_accounts/_splits.html.haml
```diff
@@ -0,0 +1,17 @@
+- available_subaccounts = Account.excluding_split_accounts.filter_global_account_types
```
jhanggi: This should exclude expired and suspended accounts. I would almost consider creating a method on `SplitAccount.available_subaccounts` or something like that to wrap this up so it's not in the view.

vendor/engines/split_accounts/app/views/facility_accounts/account_fields/split_accounts/_splits.html.haml
```diff
@@ -0,0 +1,17 @@
...
+= javascript_include_tag "split_account_splits"
+
+%h4 Subaccounts
+
+%div{data: {subaccounts: true}}
+  = f.nested_fields_for :splits, class_name: "SplitAccounts::Split" do |ff|
+    .well
+      = ff.input :subaccount_id, collection: available_subaccounts, wrapper_html: {class: "input-inline"}
```
jhanggi: One worry I have here is that the list will be _very_ long in NU. It might be fine if we just wrap it in Chosen, but we might want to consider an ajax lookup. We can probably leave this as is for the current PR, but we'll probably want to do something at a later point.

vendor/engines/split_accounts/app/views/facility_accounts/account_fields/split_accounts/_splits.html.haml
```diff
@@ -0,0 +1,17 @@
...
+      = ff.input :percent, input_html: {min: 0, max: 100, data: {percent: true}}, wrapper_html: {class: "input-inline"}
+      = ff.input :extra_penny, wrapper_html: {class: "input-inline"}
```
jhanggi: Couple thoughts:

* Would it make sense to default the first subaccount's extra penny field to true?
* Should we have one (or maybe even two) subaccounts exposed when you first land on the page?
* Can the extra penny be a radio button? Or is that too complicated?

jhanggi:
I think I'm done with my initial review. Overall, the refactorings look pretty good and clean up quite a bit.

I think it's fine that the different new forms are on separate pages: definitely cleans up some code, but I'm not sure I'm sold on the tabs for the interface. We could do a dropdown that navigates onchange. I could be swayed, though.

jhanggi:
One additional note: Split Accounts should only be created by global admins, so we'll need to add some permission checks around that.

app/support/order_uncanceler.rb
benlinton: Good call, it is in a rake task.  Hmmm.  I could have the parent task set `Rails.logger = Logger.new(STDOUT)` so it still `puts` to the command line without getting `puts` during spec execution.  Do you have any better ideas?  Thanks!

app/models/account.rb
benlinton: Good idea.  I'll implement Jason's suggestion.

app/models/account.rb
benlinton: Ok, sounds good.

app/models/nufs_account.rb
benlinton: There was a handful of validations taking place in deprecated controller methods.  Such as:
```
account.valid? # populate virtual charstring attributes required by set_expires_at
account.errors.clear

# be verbose with failures. Too many tasks (#29563, #31873) need it
begin
  account.set_expires_at!
  account.errors.add(:base, I18n.t('controllers.facility_accounts.create.expires_at_missing')) unless account.expires_at
rescue AccountNumberFormatError => e
  account.expires_at = Time.zone.now # Prevent expires_at missing message
rescue ValidatorError => e
  account.errors.add(:base, e.message)
end
```

app/views/facility_accounts/account_fields/_nufs_account.html.haml
benlinton: I was hoping to default to false and only use `true` if it's passed in.  I can use `local_assigns` if you like but I think it's more code to achieve the same thing.  Although I'm all for it if you believe it's easier to understand what's going on.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_purchase_order_account.html.haml
benlinton: So all of the different `Account` subclassed forms use a similar order.  I can change it back if we don't want to make waves.

app/services/account_builder.rb
benlinton: I think your approach is better.  Good idea.

vendor/engines/split_accounts/app/views/facility_accounts/account_fields/split_accounts/_splits.html.haml
benlinton: Yeah, I wasn't a fan of putting this in the view either.  Didn't get around to cleaning it up yet.  Good idea.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_credit_card_account.html.haml
benlinton: Ok, I'll fix the JS.

vendor/engines/split_accounts/app/views/facility_accounts/account_fields/split_accounts/_splits.html.haml
benlinton: Ok, chosen or ajax sounds good for a future PR.

vendor/engines/split_accounts/app/views/facility_accounts/account_fields/split_accounts/_splits.html.haml
benlinton: The radio wasn't as simple as I had hoped, so I saved it for a future improvement.  I think all of these suggestions are a good idea.  And I was going to circle back to this once I have to reports done.  Does that work for you?

jhanggi: If radio isn't easy, don't worry about it for now. I'm fine waiting on a few of these improvements until later.

app/models/account.rb
benlinton: Good idea.  I added `AccountConfig` instead since `Account::Config` raised some rails deprecation errors.
```
/.../nucore/nucore-open/app/models/account.rb:47:in `config': Use RbConfig instead of obsolete and deprecated Config.
/.../nucore/nucore-open/app/models/account.rb:47: warning: toplevel constant Config referenced by Account::Config
/.../nucore/nucore-open/app/models/account.rb:47:in `config': Use RbConfig instead of obsolete and deprecated Config.
/.../.rbenv/versions/2.1.2/lib/ruby/2.1.0/rbconfig/obsolete.rb:20:in `method_missing': undefined method `new' for RbConfig:Module (NoMethodError)
	from /.../nucore/nucore-open/app/models/account.rb:47:in `config'
```

app/services/account_builder.rb
benlinton: I started off with instance variables and started passing them all over in the name of making testing easier.  But I'm fine with using the instance variables again.

benlinton:
I've implemented both the new `AccountConfig` as well as subclassing `AccountBuilder` for each account type.  I have a few more things to clean up, but feel free to review those two commits before I finalize this PR.  Thanks!

lib/account_manager.rb
benlinton: `AccountManager` is officially no more.  See `app/models/account_config` for the new account related settings and a fully supported spec.

app/services/account_builder.rb
```diff
@@ -0,0 +1,185 @@
...
+  # `Account` class object, class name string/symbol, or underscored class name
+  # string/symbol.
+  #
+  # Example usage:
+  #   AccountBuilder.for("NufsAccount") => NufsAccountBuilder
+  #   AccountBuilder.for("WithoutBuilderAccount") => AccountBuilder
+  #
+  def self.for(account_type)
+    klass = "#{account_type}_builder".classify
+    Object.const_defined?(klass) ? klass.constantize : self
```
jhanggi: I see the comment about lazily-loaded classes, and that makes sense to me, but is there a specific way to go about avoiding that scenario?

Would it be better to catch a NameError (or at least I think that's the error it would raise)? That would probably handle the lazy-loading issue.
```
  klass = "#{account_type}_builder".classify
  klass.constantize
rescue NameError
  self
```

app/services/account_builder.rb
```diff
@@ -0,0 +1,185 @@
...
+  end
+
+  # Builds and returns a new account subclassed `Account` object.
+  # Used when creating an account.
+  def build
+    @account = assign_params(@account, @params, @params_key)
```
jhanggi: Since these are all instance variables, do they need to passed around? Method will have access to them.

app/services/account_builder.rb
```diff
@@ -0,0 +1,185 @@
...
+    @account = set_affiliate(@account, @params, @params_key)
+    @account = set_updated_by(@account, @current_user)
+    @account
+  end
+
+  protected
+
+  # Returns hash of potentially modified params.
+  # Account can potentially be nil if a `new` form hasn't been submitted yet.
+  def parse_params(account, params, key)
```
jhanggi: Maybe it's because nothing's overriding it yet, but I don't really get what this method is doing. The other two methods this call really don't do anything. I would kind of expect the return value of a method like this to be pretty much equivalent to `params[Account.config.account_type_to_param(@account_type)]`

app/services/account_builder.rb
```diff
@@ -0,0 +1,185 @@
...
+    account
+  end
+
+  # Set the affiliate for an account only if affiliates are supported for the
+  # given account type.  Returns the modified account object.
+  # Silently fails if affiliate id doesn't exist; add validations to the
+  # subclassed `Account` if you'd like to require an affiliate.
+  def set_affiliate(account, params, params_key)
+    return account unless account.class.using_affiliate?
+    id = params.fetch(params_key, {}).fetch(:affiliate_id, nil)
```
jhanggi: Seeing this again, it seems like if it's possible, the `@params` should be just the `params[account_type]`. Passing around the `params_key` is seeming very confusing and possibly doing redundant lookups.

app/services/account_builder.rb
benlinton: Good idea!

app/controllers/facility_accounts_controller.rb
```diff
@@ -70,7 +47,32 @@ def show
...
+  end
+
+  # POST /facilities/:facility_id/accounts
+  def create
+    @owner_user = User.find(params[:owner_user_id])
+    @available_account_types = available_account_types
+    @current_account_type = current_account_type
+
+    @account = AccountBuilder.for(params[:account_type]).new({
+      account_type: params[:account_type],
```
jhanggi: If you are getting the accountbuilder with the factory method, is it possible to avoid also passing it in via the constructor?

app/controllers/facility_accounts_controller.rb
```diff
@@ -201,6 +191,15 @@ def show_statement
 
   private
 
+  def available_account_types
+    Account.config.account_types_for_facility(current_facility)
+  end
+
+  def current_account_type
+    return params[:account_type] if available_account_types.include?(params[:account_type])
```
jhanggi: It might be clearer that you're returning a default if you don't have a valid one if you did:
```ruby
if available_account_types.include?(params[:account_type])
  params[:account_type]
else
  available_account_types.first
end
```

app/models/account.rb
jhanggi: Seems like this might be a good candidate for some refactoring at some point. Doesn't need to be now, though.

app/views/admin/shared/_sidenav_billing.html.haml
```diff
@@ -18,7 +18,7 @@
   %li.nav-header= Journal.model_name.human.pluralize
   = tab t_create_model(Journal), new_facility_journal_path, (sidenav_tab == 'new_journal')
   = tab t_manage_models(Journal), facility_journals_path, (sidenav_tab == 'journals')
-  - if current_facility.single_facility? && AccountManager.using_statements?
+  - if current_facility.single_facility? && Account.config.using_any_statements?
```
jhanggi: `using_any_statements?` seems like somewhat of a clunky name. What about `statements_enabled?` or something like that?

vendor/engines/c2po/app/services/credit_card_account_builder.rb
```diff
@@ -0,0 +1,20 @@
+# Contains overrides for building a `CreditCardAccount` from params.
+# Dynamically called via the `AccountBuilder.for()` factory.
+class CreditCardAccountBuilder < AccountBuilder
+
+  # Override and append `build`.
+  def build
```
jhanggi: What would you think of changing this to an `after_build` method that gets called in the parent. That way it doesn't need to call `super`.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_purchase_order_account.html.haml
jhanggi: This ordering puts the expiration in inconsistent places. Think of the existing ordering this way (not all types will have all the fields):

1. Definition (account number, PO #, Name on credit card--since we don't store a number)
2. Expiration
3. Description
4. Affiliation
5. Bill to

jhanggi: I could see arguments for re-ordering this list, maybe putting description at the top or swapping expiration and description, but I'd like to keep it the same as it was.

vendor/engines/c2po/spec/models/credit_card_account_spec.rb
```diff
@@ -37,7 +37,7 @@
   end
 
   it "should be limited to a single facility" do
-    expect(CreditCardAccount.limited_to_single_facility?).to be true
+    expect(CreditCardAccount.single_facility?).to be true
```
jhanggi: I wasn't going to comment on this initially, because my preference for `be_xxx` isn't that strong, but can we make the credit_card_account_spec and purchase_order_account spec consistent? Either use `single_facility?) to be true` or `.to be_single_facility`.

vendor/engines/split_accounts/app/assets/javascripts/split_account_splits.js.coffee
```diff
@@ -0,0 +1,17 @@
...
+      $inputs = $container.find("[data-percent]").filter(":visible")
+
+      if $inputs.length > 0
+        percent = Math.round((100.0 / $inputs.length) * 100) / 100
+        remainder = Math.round((100.0 - percent * ($inputs.length - 1)) * 100) / 100
+        $inputs.val(percent)
+        $inputs.last().val(remainder)
+
+# Register event listeners when nested_form_fields are added or removed
+$(document).on "fields_added.nested_form_fields", updateSplits
```
jhanggi: I know we're only including the JS on the single view, but what are your thoughts on the specificity of this? Is it possible to attach this only to the specific form (maybe with a `.js--` prefix?

benlinton: As of right now, I only load this code if the related view is loaded; it's not included in the javascript application manifest.  Alternatively, we could add it to the manifest and using the `.js--` prefix (or better yet a html5 `data` attribute) to trigger it, but I wonder since this is so rarely used if it's better to load it on demand?

vendor/engines/c2po/spec/models/credit_card_account_spec.rb
benlinton: Sounds good, this got changed with a search/replace.  But I can improve the code since I'm touching it anways.

vendor/engines/c2po/app/services/credit_card_account_builder.rb
benlinton: Hmmm, sure.  Do you think calling `super` a code smell for some reason or are you looking for enhanced clarity?

app/views/admin/shared/_sidenav_billing.html.haml
benlinton: yeah `statements_enabled` sounds better

app/controllers/facility_accounts_controller.rb
benlinton: Ok

app/controllers/facility_accounts_controller.rb
benlinton: Passing account type to both seemed redundant to me as well.  But multiple account types may use the same builder, so given an account builder class object we shouldn't try to assume an account type.  

We could have the factory return an instance of the account builder with the account type already stored.  And either pass the initialization args to the factory method as well or support setting the additional args with another `new`-like method.  But I felt that the instance factory approach makes it less clear what's going on.  What do you think?

app/services/account_builder.rb
benlinton: These will be overridden.  Should have the commit tonight or tomorrow.

app/services/account_builder.rb
benlinton: To expand on our earlier slack chat...  The instance variables do not need to be passed around.  I've tried both approaches in the past and have started to prefer the more explicit approach.  Although I'm not strongly opinionated on this matter and could be talked into the less-explicit approach.  But a few pros do the explicit approach are:

* As a general rule of thumb, I like to test private methods.  Sometimes for TDD.  But also because if a class's public API breaks, the tests let us know immediately why the API is broken.  
* I've found it's easier/cleaner to unit test a method that accepts args and returns something; rather than setting up instance variables and checking them afterwards.  
* Passing the ivars as arguments is more explicit.  So it's more clear what goes into and comes out of a method.  And helps avoid unexpected/accidental modification of non-returned values.
* As a minor bonus side effect, avoiding instance variables in the private method decouples the methods from the class so it's easier to refactor or paste/copy.
* As far as I understand, we don't take performance hit or do extra garbage collection since ruby is a `pass-reference-by-value` language.  Although I may misunderstand how the memory management works in ruby.

```
Book = Struct.new("Book", :title)
book = Book.new(title: "foo")
book.object_id 
# => 70129693244980

book.title = "bar"
book.object_id 
# => 70129693244980

def change_title(book)
  puts "before change in method: #{book.object_id}"
  book.title = "changed"
  puts "after change in method: #{book.object_id}"
  book
end

book = change_title(book)
# => before change in method: 70129673327740
# => after change in method: 70129673327740

book.object_id
# => 70129673327740
```

I'll be in the office tomorrow, we should talk in person.  And I'll execute whatever you think is best.  :)

benlinton: Typo edited `...do (much) extra garbage collection...`

benlinton: And the `object_id`'s should have all been the same, doh.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_purchase_order_account.html.haml
benlinton: Ok cool, thanks!

app/assets/javascripts/accounts.js
```diff
@@ -1,40 +1,32 @@
...
+  $(".datepicker").datepicker({minDate:+0, maxDate:"+3y", dateFormat: "mm/dd/yy"});
 
+  // ---------------------------------------------------------------------------
   // Autotab
-  $('.account_number_field :input[maxlength]').keyup(function(evt) {
+  // ---------------------------------------------------------------------------
+  $(".account_number_field :input[maxlength]").keyup(function(evt) {
     $this = $(this);
-    // if it's a number key
+    // if it"s a number key
```
jhanggi: Oops find/replace.

app/assets/javascripts/accounts.js
```diff
@@ -1,40 +1,32 @@
...
+  // ---------------------------------------------------------------------------
+  // Date picker
+  // ---------------------------------------------------------------------------
+  $(".datepicker").datepicker({minDate:+0, maxDate:"+3y", dateFormat: "mm/dd/yy"});
 
+  // ---------------------------------------------------------------------------
   // Autotab
-  $('.account_number_field :input[maxlength]').keyup(function(evt) {
+  // ---------------------------------------------------------------------------
+  $(".account_number_field :input[maxlength]").keyup(function(evt) {
```
jhanggi: I know it's not new, but what does `:input` do? Does that also include `select`? Does that matter here because of `maxlength`.

app/controllers/facility_accounts_controller.rb
```diff
@@ -70,7 +47,31 @@ def show
...
+    @account = current_account_type.constantize.new
+  end
+
+  # POST /facilities/:facility_id/accounts
+  def create
+    @owner_user = User.find(params[:owner_user_id])
+    @available_account_types = available_account_types
+    @current_account_type = current_account_type
+
+    @account = AccountBuilder.for(params[:account_type]).new({
```
jhanggi: No need for the curly braces.

app/controllers/facility_accounts_controller.rb
```diff
@@ -70,7 +47,31 @@ def show
...
+      account_type: params[:account_type],
+      facility: current_facility,
+      current_user: current_user,
+      owner_user: @owner_user,
+      params: params,
+    }).build
+
+    if @account.save
+      flash[:notice] = 'Account was successfully created.'
```
jhanggi: I18n

app/controllers/facility_accounts_controller.rb
```diff
@@ -70,7 +47,31 @@ def show
...
+      redirect_to facility_user_accounts_path(current_facility, @account.owner_user)
+    else
+      render action: 'new'
```
jhanggi: Should we have a `flash.now` or do the errors on the form itself suffice? (haven't done a testing yet)

app/controllers/facility_accounts_controller.rb
```diff
@@ -79,13 +80,16 @@ def edit
 
   # PUT /facilities/:facility_id/accounts/:id
   def update
-    class_params = account_class_params
+    account_type = Account.config.account_type_to_param(@account.class)
 
-    if @account.is_a?(AffiliateAccount)
-      class_params[:affiliate_other] = nil if class_params[:affiliate_id] != Affiliate.OTHER.id.to_s
-    end
+    @account = AccountBuilder.for(account_type).new({
```
jhanggi: No need for curly braces

app/models/account.rb
```diff
@@ -26,6 +26,7 @@ def price_groups
   accepts_nested_attributes_for :account_users
 
   scope :active, lambda {{ :conditions => ['expires_at > ? AND suspended_at IS NULL', Time.zone.now] }}
+  scope :global_account_types, -> { where("accounts.type in (:global)", { global: config.global_account_types }) }
```
jhanggi: Curly braces around `global: config.global_account_types` aren't necessary. Matter of personal preference, but I tend to only use named parameters (e.g. `:global` when I have multiple variables or I want to use a single value in multiple places. `where("accounts.type in ?", config.global_account_types)` is a little more succinct.

In this case I might also use `where(accounts: { type: config.global_account_types })`.

Not a strong feeling on it, though.

app/models/order_detail.rb
```diff
@@ -244,15 +244,19 @@ def can_be_viewed_by?(user)
...
-       "products.facility_id = :facility_id
-       AND order_details.state = :state
-       AND problem = :problem
-       AND reviewed_at <= :reviewed_at
-       AND order_details.statement_id IS NULL
-       AND order_details.price_policy_id IS NOT NULL
-       AND accounts.type IN (:accounts)
-       AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)",
-       { :facility_id => facility.id, :state =>'complete', :problem => false, :reviewed_at => Time.zone.now, :accounts => AccountManager::STATEMENT_ACCOUNT_CLASSES }
+      "products.facility_id = :facility_id
```
jhanggi: It might be a good idea to start updating these old-style scopes. Won't call this a show-stopper.

app/models/order_detail.rb
```diff
@@ -244,15 +244,19 @@ def can_be_viewed_by?(user)
...
+        AND order_details.state = :state
+        AND problem = :problem
+        AND reviewed_at <= :reviewed_at
+        AND order_details.statement_id IS NULL
+        AND order_details.price_policy_id IS NOT NULL
+        AND accounts.type IN (:accounts)
+        AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)",
+      :facility_id => facility.id,
```
jhanggi: Should we go ahead and update these to 1.9 hash syntax.

spec/services/nufs_account_builder_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        account_type: "NufsAccount",
+        facility: build_stubbed(:facility),
+        owner_user: build_stubbed(:user),
+        current_user: build_stubbed(:user),
+        params: params,
+      }
+    end
+
+    let(:params) do
+      ActionController::Parameters.new({
```
jhanggi: Curlybraces not necessary.

spec/services/nufs_account_builder_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+        nufs_account: {
+          account_number: "1234",
+          description: "foobar",
+        }
+      })
+    end
+
+    it "sets expired_at" do
+      expect(builder.build.expires_at).not_to be_nil
```
jhanggi: `to be_present` is sometimes a little nicer than a double negative.

jhanggi: And should we do any sort of check that it's setting the expected expires_at?

vendor/engines/c2po/app/services/credit_card_account_builder.rb
```diff
@@ -0,0 +1,42 @@
+# Contains overrides for building a `CreditCardAccount` from params.
+# Dynamically called via the `AccountBuilder.for()` factory.
+class CreditCardAccountBuilder < AccountBuilder
+
+  protected
+
+  # Override strong_params for `build` account.
+  def account_params_for_build
```
jhanggi: Is `for_build` always a superset of `for_update`? Would it be easier to maintain if we did `for_build` as `for_update + [....]`?

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
```diff
@@ -27,5 +29,16 @@ def extra_penny_count
...
     end
 
+    # Stopped using SQL because that didn't seem to work until the built splits
+    # and subaccounts were created.
+    #
+    # SQL version:
+    #   subaccounts.where("expires_at IS NOT NULL").order(expires_at: :asc).first
+    #
+    def earliest_expiring_subaccount
+      subaccounts = splits.map{ |split| split.subaccount }.select(&:expires_at)
```
jhanggi: This confuses me a little. It looks like you'll come out of this as an array of dates. are we not using this method yet?

Can we use the `has_many :subaccounts, through :splits`?
`subaccounts.min_by(&:expires_at)`?

vendor/engines/split_accounts/lib/split_accounts/engine.rb
```diff
@@ -1,22 +1,32 @@
...
-      unless app.root.to_s.match root.to_s
+      unless app.root.to_s.match(root.to_s)
         app.config.paths["db/migrate"] += config.paths["db/migrate"].expanded
       end
     end
 
     # Include factories in main rails app
     initializer "model_core.factories", after: "factory_girl.set_factory_paths" do
-      FactoryGirl.definition_file_paths << File.expand_path('../../../spec/factories', __FILE__) if defined?(FactoryGirl)
+      if defined?(FactoryGirl) && SettingsHelper.feature_on?(:split_accounts)
```
jhanggi: Where were you running into problems with FactoryGirl not being defined. I would expect the `after: "factory_girl.set_factory_paths"` to ensure that FG is always there.

vendor/engines/split_accounts/spec/engine_helper.rb
```diff
@@ -0,0 +1,9 @@
+# Rspec configurations for this engine
+RSpec.configure do |config|
+  
+  # Skip split_accounts tests if split_accounts feature is disabled
+  unless SettingsHelper.feature_on?(:split_accounts)
+    config.filter_run_excluding split_accounts: true 
+  end
+
+end
```
jhanggi: Newline at end of file.

vendor/engines/split_accounts/spec/engine_helper.rb
```diff
@@ -0,0 +1,9 @@
+# Rspec configurations for this engine
+RSpec.configure do |config|
+  
+  # Skip split_accounts tests if split_accounts feature is disabled
+  unless SettingsHelper.feature_on?(:split_accounts)
```
jhanggi: We've tended to use the `feature_setting` metadata to turn on features as they're needed. I don't have a problem not running tests in one of the downstream branches, but I don't like skipping tests in open.

app/assets/javascripts/accounts.js
benlinton: Input selects all form controls including `button`, `textarea`, and `select`.  And this javascript will only target form controls with maxlength present.

app/controllers/facility_accounts_controller.rb
benlinton: The errors on the form suffice.

app/models/order_detail.rb
benlinton: I'll note this and submit a separate PR so this PR isn't doing too many potentially-breaking changes at once.

spec/services/nufs_account_builder_spec.rb
benlinton: I'm not sure since the `ValidatorFactory` sets `expired_at` differently depending on the repository.  I figured I'd let the validator factory handle testing that?

vendor/engines/c2po/app/services/credit_card_account_builder.rb
benlinton: It's not always a subset.

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
benlinton: The `has_many` `through` was not working until I created the `splits` in the database.  I spent some time trying to find a workaround but was unsuccessful.  Is this a rails 3.2 issue or is that expected rails behavior?  At the end of the day, I'd like to use this method on a non-persisted `split_account` (self) with non-persisted `splits` (the through `subaccounts` are persisted).

vendor/engines/split_accounts/lib/split_accounts/engine.rb
benlinton: Copied from here.  https://github.com/thoughtbot/factory_girl_rails/pull/42

vendor/engines/split_accounts/spec/engine_helper.rb
benlinton: I started with the feature_setting metadata but the `@@class` variables and extensions on engine load didn't seem to reset between tests.

benlinton: I wasn't a fan of skipping the tests as well.  I was thinking maybe we could include all split_account code and only use the `split_accounts` feature to update `AccountConfig.new.account_types`.  But then we're loading split accounts code even if we're not using it.

spec/services/nufs_account_builder_spec.rb
jhanggi: That makes sense. I think I'd still prefer a `to be_present` over `not_to be_nil`, though. Not a big deal.

vendor/engines/split_accounts/spec/engine_helper.rb
jhanggi: I'm still a little torn on whether it's redundant to use both the feature flag and gem inclusion to say whether the feature is active. If the gem is loaded, however, I think we should run the tests. And I think we'll need to because circle may still try to run them even if it's not active (we use `vendor/engines/**/spec/**/*_spec.rb` by default in `circle.yml`, but we are restricting that to only one engine in UIC.

jhanggi:
Let's go ahead and merge this. We can clean up remaining things later.

vendor/engines/split_accounts/spec/engine_helper.rb
jhanggi: Your point about the class variables not resetting makes sense. I'm not so worried if the only places you use this metadata are only within the engine.

spec/services/nufs_account_builder_spec.rb
benlinton: That change was made.  :) https://github.com/tablexi/nucore-open/commit/3b55d49db7e0b9254f2504975f870be179c21fed#diff-df5031acd0a4fcde49a8612d191d1505R31

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
jhanggi: I would expect `subaccounts` to work even with unpersisted splits, but I guess it might have to do with how the splits are built. If they're built as `Split.new(parent: parent, subaccount: subaccount)`, I could see how it wouldn't work, but if you do `parent.splits.build(subaccount: subaccount)`, I would expect it to. I'm guessing that it's the nested form gem that's taking care of the building. Maybe if it's doing `parent.splits.build(subaccount_id: id)`, that's not working.

Not using the `has_many :through` is fine, it's just that this is a little hard to read. 

`select(&:expires_at)` confused me a little bit thinking it was doing a kind of map and returning a bunch of dates. All accounts should have an expiration date, so is that check even necessary?

`splits.map(&:subaccount).min_by(&:expires_at)`?

PR #411, opened 2015-12-15
Use SSH for github gems

jhanggi:
NU stopped being able to connect over the git protocol due to firewall
changes. Using SSH is better anyways.


drench:
:+1:

PR #412, opened 2015-12-15
[#98883] Add feature flag for the Billing Administrator role

drench:
This is ~~off by default but should be enabled for UIC once merged~~ ON by default, though currently only UIC is using the role in production.

Most of the changes here happened through spec whack-a-mole, but the root change is in `app/models/user_role.rb` where it checks the flag in the `.global_roles` class method.

### DIFF
--- a/app/validators/user_role_name_validator.rb
+++ b/app/validators/user_role_name_validator.rb
@@ -1,0 +1,9 @@
+class UserRoleNameValidator < ActiveModel::Validator
+
+  def validate(record)
+    unless UserRole.valid_roles.include?(record.role)
+      record.errors[:role] << "is not a valid value"
+    end
+  end
+
+end
--- a/app/validators/user_role_name_validator.rb
+++ b/app/validators/user_role_name_validator.rb
@@ -1,9 +1,0 @@
-class UserRoleNameValidator < ActiveModel::Validator
-
-  def validate(record)
-    unless UserRole.valid_roles.include?(record.role)
-      record.errors[:role] << "is not a valid value"
-    end
-  end
-
-end
--- a/lib/role.rb
+++ b/lib/role.rb
@@ -12,9 +12,11 @@ module Role
     #
     # Creates methods #administrator?, #facility_staff?, etc.
     # Each returns true if #user_roles has the role for any facility.
-    define_method(role.gsub(/\s/, '_').downcase + '?') do
-      roles=user_roles.collect(&:role)
-      roles.include?(role)
+    if role != UserRole::BILLING_ADMINISTRATOR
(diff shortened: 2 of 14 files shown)


app/validators/user_role_name_validator.rb
```diff
@@ -0,0 +1,9 @@
+class UserRoleNameValidator < ActiveModel::Validator
```
jhanggi: I'm curious why you pulled this out into a separate validator class? It like it's doing exactly what it was before. I'm not sure the extra indirection is that valuable.

lib/role.rb
```diff
@@ -50,6 +52,14 @@ module Role
     end
   end
 
+  def billing_administrator?
```
jhanggi: What does separating this out buy us? We've doing a validation in the model. I suppose it would make sure that it always returns false if the feature is off, even if someone manually messes with the database or there's a user who had this role before we implemented the change.

app/validators/user_role_name_validator.rb
drench: I didn't like doing it because it's there purely so specs would pass. There's some detail in the commit comment f283575b535de77056dd9b386d8937326ff713d0 but it came down to `UserRole.valid_roles` being "frozen" at load time, so enabling the `billing_administrator` feature in a context wouldn't necessarily add `billing_administrator` to the list as I'd expect.

jhanggi:
I just heard from Todd. He's actually thinking they may want to leave the role in, and assign the role to Todd. It would likely allow him to handle this case: "A handful of times per year I get the question from business admins who ask: is there any unbilled activity on chart string X. My current workflow is to go to each core individually and search all transactions for that chart string".

I think having the feature flag is a good thing to have, so I think we should leave it; but perhaps default it to true.

app/validators/user_role_name_validator.rb
jhanggi: Would using a proc work?

`validates_inclusion_of :role, in: ->(roles) { valid_roles }, message: "is not a valid value"`

lib/role.rb
drench: If the `billing_administrator` feature is off, the role isn't included in `valid_roles`, so the metaprogramming doesn't create the `billing_administrator?` method. Without it, I was starting to write some awkward conditions in `ability.rb`. It seemed cleaner to go this way and have the flag check in fewer places.

app/validators/user_role_name_validator.rb
drench: Looks like it does (still waiting on Circle); thanks.

drench:
Cleaned up the role validation and set the feature to be on by default.

jhanggi:
:+1:

PR #413, opened 2015-12-16
[#98883] Allow Account Managers to assign account ownership

drench:
Since Account Managers are not considered managers of the "all" facility, they were not given an option to assign the "Owner" role. This change should make "Owner" appear as an option on the form at `/facilities/all/accounts/:account_id/account_users/new?user_id=:user_id` for an Account Manager.

There's a lot of refactoring and cleanup here, but I kept the `manager_of?` check because it felt weird to bring `Ability` checks into a model. Maybe that means `.selectable_user_roles` doesn't belong in the `AccountUser` model at all?

This method is only used in two places that I see, so moving it shouldn't be a big effort, if it's worth taking on.

### DIFF
--- a/app/models/account_user.rb
+++ b/app/models/account_user.rb
@@ -44,16 +44,22 @@ class AccountUser < ActiveRecord::Base
   # Provides an +Array+ of roles that can be assigned
   # to a user. Optionally filters the set by the given
   # arguments
-  # [_user_]
+  # [_granting_user_]
   # The user selecting a role to be applied to
   # another user; the grantor
   # [_facility_]
   # The facility under which the selected role is
   # granted by +user+
-  def self.selectable_user_roles(user=nil, facility=nil)
-    user_roles.reject { |r| r == ACCOUNT_OWNER if user.nil? || facility.nil? || !user.manager_of?(facility) }
+  def self.selectable_user_roles(granting_user = nil, facility = nil)
+    case
+    when granting_user.blank? || facility.blank?
+      user_roles - [ACCOUNT_OWNER]
+    when granting_user.account_manager? || granting_user.manager_of?(facility)
+      user_roles
+    else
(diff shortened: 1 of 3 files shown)


app/models/account_user.rb
```diff
@@ -44,17 +44,23 @@ def self.active
...
   # The user selecting a role to be applied to
   # another user; the grantor
   # [_facility_]
   # The facility under which the selected role is
   # granted by +user+
-  def self.selectable_user_roles(user=nil, facility=nil)
-    user_roles.reject { |r| r == ACCOUNT_OWNER if user.nil? || facility.nil? || !user.manager_of?(facility) }
+  def self.selectable_user_roles(granting_user = nil, facility = nil)
+    case
+    when granting_user.blank? || facility.blank?
```
jhanggi: What's the case where `granting_user` or `facility` would be blank?

jhanggi:
Yeah, `selectable_user_roles` doesn't really feel like it belongs on the model, but I don't really know where it would go. It's almost presenter logic, but I could also see it being useful in the controller to validate the user's allowed to add that role. I don't think that method has seen a lot of churn in a while, so I'm not that concerned with fixing it up too much.

I just have my one question about the case where the params are blank, otherwise this looks good.

app/models/account_user.rb
drench: `app/views/account_users/new.html.haml` calls it without arguments, though it could (maybe should) pass a `granting_user`. Then again if this was a controller helper, it could be using `current_user`/`current_facility`/`current_ability` to figure it out.

jhanggi: Gotcha. I'm not convinced it's worth the effort to clean it up right now, but if you think it'll be easy, go ahead.

Otherwise, feel free to merge. :+1:

PR #414, opened 2015-12-16
[#122474] Restore ability for facility staff to batch update reservations

jhanggi:
I didn’t want to get into the black hole that could be writing the missing specs for this method; I just wanted to get the bug fixed.


drench:
:+1:

PR #415, opened 2015-12-18
[#98883] Remove user switch (order on behalf) link from flash after user creation

drench:
The "order on behalf" link should not appear when in a cross-facility ("all") context.

### DIFF
--- a/vendor/engines/c2po/app/views/users/_new_user.html.haml
+++ b/vendor/engines/c2po/app/views/users/_new_user.html.haml
@@ -1,4 +1,8 @@
 .notice
   %p= "#{t('users.new_user.success')} #{link_to "#{@new_user} (#{@new_user.username})", facility_user_path(current_facility, @new_user)}.".html_safe
   %p= link_to t('users.new_user.add_to_payment_source'), facility_accounts_path(current_facility)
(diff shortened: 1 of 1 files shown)


vendor/engines/c2po/app/views/users/_new_user.html.haml
```diff
@@ -1,4 +1,8 @@
 .notice
   %p= "#{t('users.new_user.success')} #{link_to "#{@new_user} (#{@new_user.username})", facility_user_path(current_facility, @new_user)}.".html_safe
   %p= link_to t('users.new_user.add_to_payment_source'), facility_accounts_path(current_facility)
-  %p= link_to t('users.new_user.order_for'), facility_user_switch_to_path(current_facility, @new_user)
+
+  - if current_facility.single_facility?
```
jhanggi: Would an ability check be better?

jhanggi: Actually, based on my understanding of the ticket, it sounds like this should link should just go away entirely. Users won't have a payment source at this point, but ordering on behalf of them is kind of useless.

drench: Makes sense. I added a commit to remove the link altogether.

jhanggi:
:+1: When you squash, make sure to give it the more accurate description of the second commit.

PR #416, opened 2015-12-21
Fix missing translation error

jhanggi:
While trying to pull the latest open/master into UIC, we were getting
a test failure on a missing translation. I'm not 100% why it wasn't
finding the translation, especially, why it was working in open, but not uic, but this appears to work. There's a circle branch running with the latest from open/master and this change.


drench:
:+1:

PR #417, opened 2015-12-21
[#103993] Set default Payment#processing_fee to $0

jhanggi:
And make it not-null.

I didn’t want to give :amount a default because it’s something that
should always be supplied by the user, whereas processing_fee could
likely not be a feature that’s used.


jhanggi:
Hold off. This changed more in the schema than I wanted it to.

PR #418, opened 2015-12-21
[#103993] Set default Payment#processing_fee to $0

jhanggi:
And make it not-null.

I didn’t want to give :amount a default because it’s something that
should always be supplied by the user, whereas processing_fee could
likely not be a feature that’s used.

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -9,6 +9,7 @@ class Payment < ActiveRecord::Base
   end
 
   validates :source, presence: true, inclusion: { in: valid_sources }
-  validates :account, :amount, presence: true
+  validates :account, :amount, numericality: true, presence: true
+  validates :processing_fee, numericality: true, allow_nil: false
 
 end
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -9,7 +9,8 @@ class Payment < ActiveRecord::Base
   end
 
   validates :source, presence: true, inclusion: { in: valid_sources }
-  validates :account, :amount, numericality: true, presence: true
+  validates :account, presence: true
+  validates :amount, presence: true, numericality: true
   validates :processing_fee, numericality: true, allow_nil: false
 
 end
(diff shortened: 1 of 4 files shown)


benlinton:
Looks good :+1:

app/models/payment.rb
```diff
@@ -9,6 +9,8 @@ def self.valid_sources
   end
 
   validates :source, presence: true, inclusion: { in: valid_sources }
-  validates :account, :amount, presence: true
+  validates :account, presence: true
+  validates :amount, presence: true, numericality: true
+  validates :processing_fee, numericality: true, allow_nil: false
```
benlinton: Shouldn't `:amount` and `:processing_fee` use the same validations?

jhanggi: I was actually going for slightly different validations: `amount` should be not-null, but also not allow zero, (i.e. `present?`). `processing_fee` is allowed to be zero. I did intentionally allow negative values in the case of a refund or something like that.

jhanggi: Actually, I realize that may be incorrect. `0.present? == true`. Let me rethink that a bit.

benlinton: You could use `:numericality`'s `:not_equal_to` zero.

jhanggi: `numericality` doesn't have a `not_equal_to`, but rails 4.X does have an `other_than` (I just checked rails 4.2, might be earlier). Unfortunately, we're still on 3.2, so I used `exclusion: { in: [0] }`, which is a little funky, so I added a #TODO for when we upgrade.

benlinton: Ah yeah, it's `other_than` rather than `not_equal_to`.  Bummer it's not rails 3.2.  Exclusion works for now.  Or `numericality` `greater_than: 0, less_than: 0` might work as well.

benlinton:
Update looks good. :+1:  I don't think it's work the effort to use localizations for the `exclusion` message since it's going to get replaced.

PR #419, opened 2015-12-22
[#123960] All transactions in billing tab were red

jhanggi:
Introduced as part of a refactoring of the OrderDetailPresenter.
`row_class` returns either “reconcile-warning” or “”. “” was evaluating
to true in `needs_reconcile_warning?`.

This removes the `needs_reconcile_warning?` helper method. There could
still be the next step of removing `row_class` helper in favor of the
`Presenter.wrap`, but that was a bigger change than I wanted to make to
get a quick fix.


drench:
:+1:; not sad to see `#needs_reconcile_warning?` go away.

PR #421, opened 2015-12-23
[#120455] Spike split accounts journaling

benlinton:
Spike for review only.

### DIFF
--- a/app/controllers/general_reports_controller.rb
+++ b/app/controllers/general_reports_controller.rb
@@ -41,9 +41,13 @@ class GeneralReportsController < ReportsController
   end
 
   def generate_report_data_csv
-    ExportRawReportMailer.delay.raw_report_email(email_to_address, raw_report)
-    flash[:notice] = I18n.t('controllers.reports.mail_queued', email: email_to_address)
-    redirect_to send("#{action_name}_facility_general_reports_path", current_facility)
+    if Rails.env.development?
+      send_data raw_report.to_csv, type: "text/csv", filename: "report.csv"
+    else
+      ExportRawReportMailer.delay.raw_report_email(email_to_address, raw_report)
+      flash[:notice] = I18n.t('controllers.reports.mail_queued', email: email_to_address)
+      redirect_to send("#{action_name}_facility_general_reports_path", current_facility)
+    end
   end
 
   def raw_report
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -57,9 +57,19 @@ class Account < ActiveRecord::Base
     config.cross_facility?(self.name)
   end
 
-  # Returns true if this account type can assign an affiliate.
+  # Returns true if this account type supports affiliate.
   def self.using_affiliate?
     config.using_affiliate?(self.name)
+  end
+
+  # Returns true if this account type supports statements.
+  def self.using_statements?
+    config.using_statements?(self.name)
+  end  
+
+  # Returns true if this account type supports journal.
+  def self.using_journal?
+    config.using_journal?(self.name)
   end
 
   def self.for_facility(facility)
@@ -176,6 +186,7 @@ class Account < ActiveRecord::Base
     return "The #{self.type_string} has insufficient price groups" unless product.can_purchase?((self.price_groups + user.price_groups).flatten.uniq.collect {|pg| pg.id})
 
     # check chart string account number
+    # TODO: should this include SplitAccount?
     if self.is_a?(NufsAccount)
       accts=product.is_a?(Bundle) ? product.products.collect(&:account) : [ product.account ]
       accts.uniq.each {|acct| return "The #{self.type_string} is not open for the required account" unless self.account_open?(acct) }
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -4,55 +4,14 @@ class Journal < ActiveRecord::Base
   class CreationError < StandardError; end
 
   module Overridable
+
     def create_journal_rows!(order_details)
-      recharge_by_product = {}
-      facility_ids_already_in_journal = Set.new
-      order_detail_ids = []
-      pending_facility_ids = Journal.facility_ids_with_pending_journals
-      row_errors = []
-      recharge_enabled=SettingsHelper.feature_on? :recharge_accounts
-
-      # create rows for each transaction
-      order_details.each do |od|
-        row_errors << "##{od} is already journaled in journal ##{od.journal_id}" if od.journal_id
-        account = od.account
-        od_facility_id = od.order.facility_id
-
-        # unless we've already encountered this facility_id during
-        # this call to create_journal_rows,
-        unless facility_ids_already_in_journal.member? od_facility_id
-
-          # check against facility_ids which actually have pending journals
-          # in the DB
-          if pending_facility_ids.member? od_facility_id
-            raise CreationError.new(I18n.t("activerecord.errors.models.journal.pending_overlap", label: od.to_s, facility: Facility.find(od_facility_id)))
-          end
-          facility_ids_already_in_journal.add(od_facility_id)
-        end
-
-        begin
-          ValidatorFactory.instance(account.account_number, od.product.account).account_is_open!(od.fulfilled_at)
-        rescue ValidatorError => e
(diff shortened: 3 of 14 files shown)


app/controllers/general_reports_controller.rb
```diff
@@ -41,9 +41,13 @@ def email_to_address
   end
 
   def generate_report_data_csv
-    ExportRawReportMailer.delay.raw_report_email(email_to_address, raw_report)
-    flash[:notice] = I18n.t('controllers.reports.mail_queued', email: email_to_address)
-    redirect_to send("#{action_name}_facility_general_reports_path", current_facility)
+    if Rails.env.development?
```
jhanggi: I wonder if there's a better way we can do this. Maybe a feature flag. I don't like testing for environments in the middle of code. Maybe a feature flag? `background_jobs`? I'm open to other suggestions.

app/models/account.rb
```diff
@@ -176,6 +186,7 @@ def validate_against_product(product, user)
     return "The #{self.type_string} has insufficient price groups" unless product.can_purchase?((self.price_groups + user.price_groups).flatten.uniq.collect {|pg| pg.id})
 
     # check chart string account number
+    # TODO: should this include SplitAccount?
```
jhanggi: Possibly, it should check against each component of the split account. Seems like this could be a method call that in `Account` does nothing. In `NufsAccount`, it does this. In `SplitAccount`, it loops through each of the sub-accounts and calls the method on that account.

This might be something to chat with Matt about. I don't remember exactly how "account is open" really works.

app/controllers/general_reports_controller.rb
benlinton: Feature flag is ideal.  I'm planning to add one.

app/models/journal.rb
```diff
@@ -4,55 +4,14 @@ class Journal < ActiveRecord::Base
...
-        recharge_by_product[od.product_id] = recharge_by_product[od.product_id].to_f + od.total if recharge_enabled
-      end
 
-      # create rows for each recharge chart string
-      recharge_by_product.each_pair do |product_id, total|
-        product = Product.find(product_id)
-        JournalRow.create!(journal_row_attributes_from_product_and_total(product, total))
+    def create_journal_rows!(order_details)
+      builder = JournalRowBuilder.new(order_details, self).build
+      if builder.errors.blank?
```
jhanggi: Maybe `builder.valid?`which would just be `errors.any?`.

app/models/journal.rb
```diff
@@ -4,55 +4,14 @@ class Journal < ActiveRecord::Base
...
+        builder.journal_rows.each { |journal_row| JournalRow.create!(journal_row) }
```
jhanggi: I wonder if some of this could be encapsulated inside the builder. `journal_rows` would do `each { |row_attributes| JournalRow.new(row_attributes) }` then `journal_rows` actually returns JournalRows (as opposed to hashes). Then this could do `save!` on them.

jhanggi: Also, maybe I'm misreading the old code, but it looks like it creates all the rows, even if there are errors. Maybe it rolls them back?

benlinton: The old code creates the journal rows and then deletes them there are problems.  For my first pass I was attempting to clean this up by moving it into a service object, but not change too much of the logic all at once.  After I get split accounts working, if time permits this logic could use a major refactoring.

app/models/journal.rb
benlinton: Good idea.  Done.

app/controllers/general_reports_controller.rb
drench: Just curious: is the motivation behind this that testing Rails mail delivery, especially with attachments, kind of sucks? I've heard good things about https://github.com/ryanb/letter_opener but haven't tried it yet.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,153 @@
...
+  def reset
+    @errors = []
+    @journal_rows = []
+    @journaled_facility_ids = Set.new
+    @pending_facility_ids = []
+    @product_recharges = {}
+  end
+
+  # Set an array of facility_ids with pending journals
+  def cache_pending_facility_ids
```
jhanggi: Instead of a `cache` method and the attr_reader, I would memoize this
```ruby
def pending_facility_ids
  @pending_facility_ids ||= Journal.facility_ids_with_pending_journals
end
```

Also, how this is used seems kind of weird (it was weird previously as well). Could we do this check ahead of time? Error out if any order_details have a facility with a pending journal: `(order_details.distinct.pluck(:facility_id) - Journal.facility_ids_with_pending_journals).any?` (or equivalent--I think we don't get `distinct.pluck` in 3.2).

app/controllers/general_reports_controller.rb
benlinton: The letter opener gem looks promising as well.

app/services/journal_row_converter.rb
```diff
@@ -0,0 +1,24 @@
+class JournalRowConverter
+
+  # Returns an array of generated journal_rows given a single order_detail.
+  def self.from_order_detail(order_detail, journal = nil)
+    [{
+      account: order_detail.product.account,
```
jhanggi: Hmm, this is a little weird. I would expect `account: order_detail.account`. I wouldn't change it now, though, as that's what it was before. It actually makes some sense in the context of NU, though. Their "account" is part of the chartstring, which would be something like the numeric code for "Lab Services".

vendor/engines/split_accounts/app/services/split_accounts/journal_row_converter.rb
```diff
@@ -0,0 +1,43 @@
+module SplitAccounts
+
+  # TODO: use an instance rather than a class service object and store journal
+  # and possibly order_detail
+  class JournalRowConverter < ::JournalRowConverter
```
jhanggi: Can you make this not a subclass? The trouble will be when the default converter is not `::JournalRowConverter` (like it probably will be in NU). Maybe it can look up the default converter and rely on it.

app/services/journal_row_converter.rb
benlinton: This didn't make sense to me either.  As well as how we set account within `from_product` (I think we simply use a string?).

app/controllers/general_reports_controller.rb
jhanggi: I think it could be cool try try out the letter_opener em. As long as it works well with delayed_job.

vendor/engines/split_accounts/app/services/split_accounts/journal_row_converter.rb
benlinton: I was thinking nucore-open would use (sudo code):
default: JournalRowConverter,
split_accounts: SplitAccounts::JournalRowConverter < JournalRowConverter

And nucore would use:
default: Nu::JournalRowConverter,
split_accounts:  Nu::SplitAccounts::JournalRowConvert < Nu::JournalRowConverter

app/services/journal_row_converter.rb
jhanggi: In the case of journaling "account" has a different meaning that we use "Account".

Take a look at the NU journal extension. `order_detail.product.account` will be something like `"75430"`, which is an important part of what goes into NU financials. There are some other fields that get populated in NU via `order_detail.account`:

```
activity: order_detail.account.activity,
chart_field1: order_detail.account.chart_field1,
dept: order_detail.account.dept,
...
```

vendor/engines/split_accounts/app/services/split_accounts/journal_row_converter.rb
benlinton: Maybe there is a cleaner approach?

jhanggi: Seems like we'd end up with a lot of duplicate logic between ` SplitAccounts::JournalRowConverter` and `Nu::SplitAccounts::JournalRowConvert`. Maybe I'm not quite understanding what you're planning.

PR #422, opened 2015-12-23
Handle nil param for AccountBuilder.for

benlinton:
`AccountBuilder.for(nil)` should return `AccountBuilder` but was instead returning `Builder`.  Added a guard to squash this bug and provided a spec to verify.


jhanggi:
:+1:

PR #423, opened 2015-12-28
[#123990] Allow operators to switch to users on search

drench:
Previously the "Order For" link would appear in search results to users that could manage the current Facility, which does not necessarily include directors or staff.

This adds explicit switch_to User abilities to roles that are considered Facility operators, and uses this Ability check in the view that displays the "Order For" link in user search results.


drench:
Pulling this in after reviewing with @mlineen in person.

PR #424, opened 2016-01-05
[#123990] Account Managers may not "Order For" a user

drench:
When searching for users under the "Users" tab, the search results should show an "Order For" link if the logged-in user is allowed to `switch_to` a `User`. Account Managers should only encounter this when in the "all" (cross-facility) context, where placing an order wouldn't make sense. This change removes the `switch_to` ability for Account Managers, so they should no longer see "Order For" links.


mlineen:
:+1:

PR #425, opened 2016-01-21
[#121326] Send emails for training requests

jhanggi:
Admins can set a comma-separated list of email addresses who will receive
notification anytime a user submits a training request. If the field is left
blank, no emails will be sent.

This is on a per-instrument basis. We decided against any facility-level or
mass configuration.

### DIFF
--- a/spec/controllers/training_requests_controller_spec.rb
+++ b/spec/controllers/training_requests_controller_spec.rb
@@ -1,0 +1,83 @@
+require "rails_helper"
+# require 'controller_spec_helper'
+
+RSpec.describe TrainingRequestsController do
+  let(:facility) { FactoryGirl.create(:setup_facility) }
+  let(:product) { FactoryGirl.create(:setup_item, facility: facility,
+    requires_approval: true, training_request_contacts: training_request_contacts) }
+  let(:training_request_contacts) { "test@example.com, test2@example.com" }
+  let(:user) { FactoryGirl.create(:user) }
+
+  describe "#new" do
+    def do_request
+      get :new, facility_id: facility.url_name, product_id: product.url_name
+    end
+
+    describe "while not logged in" do
+      it "sends the user to sign in" do
+        do_request
+        expect(response).to redirect_to new_user_session_path
+      end
+    end
+
+    describe "while logged in" do
+      before { sign_in user }
+
+      it "gives access" do
+        do_request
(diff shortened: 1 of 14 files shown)


spec/models/product_spec.rb
```diff
@@ -1,509 +1,557 @@
 require "rails_helper"
 
 RSpec.describe Product do
-  subject(:product) { create(:instrument_requiring_approval) }
+  describe "with everything configured" do
```
jhanggi: Most of the changes in this file are me indenting all the specs inside a new `describe` block because I don't need all the setup in my new specs. Github isn't catching that very well.

spec/models/product_spec.rb
```diff
@@ -1,509 +1,557 @@
...
       end
+    end
+  end
 
-      context 'when it has no access groups' do
-        it 'does not have an access list' do
-          expect(product.has_access_list?).to be false
-        end
-      end
+  describe "#training_request_contacts" do
```
jhanggi: New stuff begins here.

spec/models/product_spec.rb
```diff
@@ -1,509 +1,557 @@
...
+    describe "multiple addresses" do
+      let(:contacts) { "testing@example.com, test2@nucore.com" }
+      it { is_expected.to eq ["testing@example.com", "test2@nucore.com"] }
+    end
 
-      it "is invalid" do
-        expect(new_product).to_not be_valid
-        expect(new_product.errors.messages).to include :url_name
-      end
+    describe "some extra commas in there" do
```
drench: What happens if the input is something that's not an email address at all (no `@` at least)?

drench:
Just one question on input sanitation.

spec/controllers/training_requests_controller_spec.rb
```diff
@@ -0,0 +1,83 @@
+require "rails_helper"
+# require 'controller_spec_helper'
```
drench: Nice to see this not being necessary! Deletable?

spec/models/product_spec.rb
jhanggi: Made an update that validates each of the addresses. I went with a pretty simple regex. We have a more complicated one inside `User`, but I didn't want to go crazy with it.

drench:
:+1:

PR #426, opened 2016-01-22
[#122002] Update spacing/capitalization to "NUcore"

jhanggi:
I'll have a separate PR for UIC where I fix a few merge conflicts and update a couple overrides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# NU Core Open
+# NUcore Open
 
 Open source version of Northwestern University Core Facility Management Software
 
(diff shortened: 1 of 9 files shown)


drench:
For just the text changes, LGTM :+1:, but are there plans to change the `NUCore` namespace in code to match the new branding too?

jhanggi:
I thought about that, but figured it wasn't worth the effort. There's also the fact that the primary namespace of the rails application is `Nucore`, so it's already a little confusing.

PR #428, opened 2016-01-28
[#88724] Error when editing a reconciled OD's reconcile note.

jhanggi:
The `_method` input was being disabled, along with the `utf8` and `authentication_token`
when the other form fields were getting disabled. Because `_method=put` was not
being submitted, the router was treating it as a POST resulting in a route not
found.


drench:
:+1:

PR #429, opened 2016-01-29
[#88724] Ignore rails hidden form fields when deciding when to disable the submit button

jhanggi:
Proper behavior: if any visible/non rails input (utf8, _method) is active, then
the submit button should be enabled. If they are all disabled, then the submit
button should be disabled as well.

After the recent change where we don't disable the hidden utf8, etc fields, the
submit button was always visible.


drench:
:+1:

PR #430, opened 2016-01-29
[#122054] Printable order detail view

jhanggi:
For UIC, they want the ability to print an order detail. I left the default feature flag (which just controls a link) to off because it is specific to them.


drench:
:+1:

PR #431, opened 2016-01-30
Update engine_manager to prevent false positives

benlinton:
The engine_manager used to look at engine names that starts_with
a given string. Which was causing false negatives when I was
trying it out. I updated so it only returns true if you provide
the exact engine name.

### DIFF
--- a/lib/engine_manager.rb
+++ b/lib/engine_manager.rb
@@ -1,17 +1,13 @@
-#
 # Keeper of this app's loaded engines
 class EngineManager
 
-  @@loaded_engines={}
+  def self.loaded_engines
+    Rails.application.railties.engines.map { |e| e.class.name }.to_set
+  end
 
   def self.engine_loaded?(engine_name)
-    normalized=engine_name.to_s.camelize.to_sym
-
-    unless @@loaded_engines.has_key? normalized
-      @@loaded_engines[normalized]=Rails.application.railties.engines.any?{|e| e.class.name.start_with? normalized.to_s }
-    end
-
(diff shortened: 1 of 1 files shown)


jhanggi:
I don't have a problem with this change, but what is the exact case where you were running into an issue?

jhanggi:
@benlinton what's the case where you were running into this?

benlinton:
> @benlinton what's the case where you were running into this?

It noticed it could potentially collide with `nu` and `nucs` while reviewing the code base.  But I did not find a place where it had become a problem yet.  But noticed a quick fix so submitted this PR.

lib/engine_manager.rb
```diff
@@ -1,17 +1,16 @@
-#
 # Keeper of this app's loaded engines
 class EngineManager
 
-  @@loaded_engines={}
+  @@loaded_engines = {}
 
   def self.engine_loaded?(engine_name)
-    normalized=engine_name.to_s.camelize.to_sym
+    class_name = "#{engine_name.to_s.camelize}::Engine"
```
jhanggi: I think going from `start_with?` to a more precise check makes sense, but what if we cleaned up the logic some?

```ruby
# note sure if it's worth memoizing or not
def self.loaded_engines
  Rails.application.railties.engines.map { |e| e.class.name }.to_set
end

def self.engine_loaded?(engine_name)
  class_name = "#{engine_name.to_s.camelize}::Engine"
  loaded_engines.include?(class_name)
end
```

PR #432, opened 2016-01-30
Install letter_opener gem

benlinton:
We'll need to smoke screen test this gem with delayed jobs.  I imagine we shouldn't run into any problems, but let's try it out before merging.  I'll test when time permits.

https://github.com/ryanb/letter_opener

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -15,8 +15,8 @@ Nucore::Application.configure do
 
   # Don't care if the mailer can't send
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.delivery_method = :test
-  config.action_mailer.perform_deliveries = false
+  config.action_mailer.delivery_method = :letter_opener
+  config.action_mailer.perform_deliveries = true
   config.action_mailer.default_url_options = { :host => "localhost", :port => 3000 }
 
   # Print deprecation notices to the Rails logger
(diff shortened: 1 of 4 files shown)


config/environments/development.rb
```diff
@@ -15,7 +15,7 @@
 
   # Don't care if the mailer can't send
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.delivery_method = :test
+  config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_deliveries = false
```
jhanggi: I was playing around with it. It seemed to only work for me if you changed `perform_deliveries = true`. Did you get it working?

jhanggi: @benlinton ^^^

benlinton: I cracked the gem open.  It appears we will have to call `.deliver` for the gem to work.  And I also had to set `perform_deliveries = true` in `config/environments/development.rb`.

jhanggi: It seemed to work for me when I set `perform_deliveries = true` in `development.rb` and if I kicked off delayed_job (`script/delayed_job run`). If it works for you, I'm cool with trying this out for a little while.

benlinton: Firing up delayed job worked for me as well.  Awesome.  I updated `config/environments/development.rb`.  This PR should be ready to go.

PR #433, opened 2016-02-01
Support account numbers with overridden fields

jhanggi:
This fix allows NU and other sites with custom Validators (NU uses fund, dept, program, etc) to support
Strong params. 

It also handles the case of invalid chart strings raising
an error when it tries to set the expiration date (it should end up
swallowing the error).

This should have minimal effect on open, but is necessary in NU.


benlinton:
:+1:

PR #434, opened 2016-02-02
[#120455] Support split accounts journaling

benlinton:
Add support for SplitAccount journal entries.  

Include a new `Converters::ConverterFactory` which is passed an `account` param and returns a converter service object.  For example, the `NufsAccount` will use the default `Converters::OrderDetailToJournalRows` converter, and the `SplitAccounts::SplitAccount` will use a special `SplitAccounts::Converters::OrderDetailToJournalRows` converter.  

Include new `converters` options in `config/settings.yml` so downstream repos can change the default converter to support their specific use case.  Change these settings to cause the `Converters::ConverterFactory` to return a different service object class for each account type.  If an account type is not defined, the factory will use the default.

Move the complex journal row creation and update logic to new `JournalRowBuilder` and `JournalRowUpdater` service objects.  This will allow us to more easily pivot our approach for handling journal row actions.  And it helped with modifying the logic without breaking existing tests.

The `JournaRowUpdater` will update journal row amounts for accounts by default.  Optionally add a `recreate_journal_rows_on_order_detail_update` method to any `Account` model to force the updater to delete and recreate journal rows instead; which is necessary to support split accounts if we do not add a nullable `split_id` column to the `journal_rows` table.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -242,8 +242,8 @@ class OrderDetail < ActiveRecord::Base
   end
 
   scope :need_statement, lambda { |facility| {
-    :joins => [:product, :account],
-    :conditions => [
+    joins: [:product, :account],
+    conditions: [
       "products.facility_id = :facility_id
         AND order_details.state = :state
         AND problem = :problem
@@ -252,23 +252,29 @@ class OrderDetail < ActiveRecord::Base
         AND order_details.price_policy_id IS NOT NULL
         AND accounts.type IN (:account_types)
         AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)",
-      :facility_id => facility.id,
-      :state =>'complete',
-      :problem => false,
-      :reviewed_at => Time.zone.now,
-      :account_types => Account.config.statement_account_types
+      facility_id: facility.id,
+      state: 'complete',
+      problem: false,
+      reviewed_at: Time.zone.now,
+      account_types: Account.config.statement_account_types,
     ]
   }}
 
   scope :need_journal, lambda { {
-    :joins => [:product, :account],
-    :conditions => ['order_details.state = ?
-                     AND problem = ?
-                     AND reviewed_at <= ?
-                     AND accounts.type IN (?)
-                     AND journal_id IS NULL
-                     AND order_details.price_policy_id IS NOT NULL
-                     AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)', 'complete', false, Time.zone.now, Account.config.journal_account_types]
+    joins: [:product, :account],
+    conditions: [
+      "order_details.state = :state
+        AND problem = :problem
+        AND reviewed_at <= :reviewed_at
+        AND accounts.type IN (:account_types)
+        AND journal_id IS NULL
+        AND order_details.price_policy_id IS NOT NULL
+        AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)",
+      state: "complete",
+      problem: false,
+      reviewed_at: Time.zone.now,
+      account_types: Account.config.journal_account_types,
+    ]
   } }
 
   scope :statemented, lambda {|facility| {
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -250,13 +250,13 @@ class OrderDetail < ActiveRecord::Base
         AND reviewed_at <= :reviewed_at
         AND order_details.statement_id IS NULL
         AND order_details.price_policy_id IS NOT NULL
-        AND accounts.type IN (:accounts)
+        AND accounts.type IN (:account_types)
         AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)",
       :facility_id => facility.id,
       :state =>'complete',
       :problem => false,
       :reviewed_at => Time.zone.now,
-      :accounts => Account.config.statement_account_types
+      :account_types => Account.config.statement_account_types
     ]
   }}
 
@@ -265,10 +265,10 @@ class OrderDetail < ActiveRecord::Base
     :conditions => ['order_details.state = ?
                      AND problem = ?
                      AND reviewed_at <= ?
-                     AND accounts.type = ?
+                     AND accounts.type IN (?)
                      AND journal_id IS NULL
                      AND order_details.price_policy_id IS NOT NULL
-                     AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)', 'complete', false, Time.zone.now, 'NufsAccount']
+                     AND (dispute_at IS NULL OR dispute_resolved_at IS NOT NULL)', 'complete', false, Time.zone.now, Account.config.journal_account_types]
   } }
 
   scope :statemented, lambda {|facility| {
@@ -655,7 +655,7 @@ class OrderDetail < ActiveRecord::Base
   end
 
   def update_journal_row_amounts
-    journal_rows.each(&:update_amount)
+    JournalRowUpdater.new(self).update
   end
 
   def assign_estimated_price(second_account=nil, date = Time.zone.now)
@@ -808,7 +808,7 @@ class OrderDetail < ActiveRecord::Base
   end
 
   def self.account_unreconciled(facility, account)
-    if account.is_a?(NufsAccount)
+    if account.class.using_journal?
       joins(:journal)
       .complete_for_facility_and_account(facility, account)
       .where('journals.is_successful' => true)
@@ -942,7 +942,7 @@ class OrderDetail < ActiveRecord::Base
   end
 
   def removable_from_journal?
-    journal.present? && account.is_a?(NufsAccount) && can_reconcile?
+    journal.present? && account.class.using_journal? && can_reconcile?
   end
 
   private
--- a/app/services/converters/order_detail_to_journal_rows.rb
+++ b/app/services/converters/order_detail_to_journal_rows.rb
@@ -1,5 +1,5 @@
 module Converters
-  class OrderDetailToJournalRows
+  class OrderDetailToJournalRowAttributes
 
     attr_accessor :journal, :order_detail, :total
 
--- a/app/services/converters/order_detail_to_journal_rows.rb
+++ b/app/services/converters/order_detail_to_journal_rows.rb
@@ -1,0 +1,23 @@
+module Converters
+  class OrderDetailToJournalRows
+
+    attr_accessor :journal, :order_detail, :total
+
+    def initialize(journal, order_detail, options = {})
+      @journal = journal
+      @order_detail = order_detail
+      @total = options[:total] || order_detail.total
+    end
+
+    def to_a
+      [{
+        account: order_detail.product.account,
+        amount: total,
+        description: order_detail.long_description,
+        order_detail_id: order_detail.id,
+        journal_id: journal.try(:id),
+      }]
+    end
+
+  end
+end
--- a/app/services/converters/order_detail_to_journal_rows.rb
+++ b/app/services/converters/order_detail_to_journal_rows.rb
@@ -9,7 +9,7 @@ module Converters
       @total = options[:total] || order_detail.total
     end
 
-    def to_a
+    def convert
       [{
         account: order_detail.product.account,
         amount: total,
--- a/app/services/converters/product_to_journal_rows.rb
+++ b/app/services/converters/product_to_journal_rows.rb
@@ -1,0 +1,22 @@
+module Converters
+  class ProductToJournalRows
+
+    attr_accessor :journal, :product, :total
+
+    def initialize(journal, product, total)
+      @journal = journal
+      @product = product
+      @total = total
+    end
+
+    def to_a
+      [{
+        account: product.facility_account.revenue_account,
+        amount: total * -1,
+        description: product.to_s,
+        journal_id: journal.try(:id),
+      }]
+    end
+
+  end
+end
--- a/app/services/converters/product_to_journal_rows.rb
+++ b/app/services/converters/product_to_journal_rows.rb
@@ -9,7 +9,7 @@ module Converters
       @total = total
     end
 
-    def to_a
+    def convert
       [{
         account: product.facility_account.revenue_account,
         amount: total * -1,
--- a/app/services/converters/product_to_journal_rows.rb
+++ b/app/services/converters/product_to_journal_rows.rb
@@ -1,5 +1,5 @@
 module Converters
-  class ProductToJournalRows
+  class ProductToJournalRowAttributes
 
     attr_accessor :journal, :product, :total
 
--- a/app/services/journal_row_builder.rb
+++ b/app/services/journal_row_builder.rb
@@ -15,7 +15,6 @@ class JournalRowBuilder
   end
 
   # Reset instance variables both on initialize and build.
-  # Returns self to support chaining.
   def reset
     @errors = []
     @journal_rows = []
@@ -30,15 +29,12 @@ class JournalRowBuilder
   def build
     reset
     order_details.each do |order_detail|
-      validate_order_detail_unjournaled(order_detail)
-      validate_facility_can_journal(order_detail)
-      validate_account(order_detail)
+      validate_order_detail(order_detail)
       new_journal_rows = order_detail_to_journal_rows(order_detail)
       update_product_recharges(order_detail)
       @journal_rows.concat(new_journal_rows)
     end
     add_journal_rows_from_product_recharges
-
     self
   end
 
@@ -46,6 +42,7 @@ class JournalRowBuilder
   # Does not gracefully handle unexpected database failures.
   # Wrapping everything in a transaction was not an option because oracle will
   # only support 1000 transactions per request.
+  # Returns self to support chaining.
   def create
     build
     if valid? && journal_rows.present?
@@ -64,6 +61,13 @@ class JournalRowBuilder
   # Set an array of facility_ids with pending journals
   def pending_facility_ids
     @pending_facility_ids ||= Journal.facility_ids_with_pending_journals
+  end
+
+  # Run all validations on an order detail
+  def validate_order_detail(order_detail)
+    validate_order_detail_unjournaled(order_detail)
+    validate_facility_can_journal(order_detail)
+    validate_account(order_detail)
   end
 
   # Validate the order_detail hasn't already been journaled
--- a/app/services/journal_row_builder.rb
+++ b/app/services/journal_row_builder.rb
@@ -1,0 +1,161 @@
+require "set"
+
+# Ported from Journal model.
+# We should improve upon this logic.
+class JournalRowBuilder
+  include NUCore::Database::ArrayHelper
+
+  attr_reader :order_details, :journal, :journal_rows, :errors, :options,
+              :recharge_enabled, :product_recharges, :journaled_facility_ids
+
+  def initialize(journal, order_details)
+    @order_details = order_details
+    @journal = journal
+    reset
+  end
+
+  # Reset instance variables both on initialize and build.
+  # Returns self to support chaining.
+  def reset
+    @errors = []
+    @journal_rows = []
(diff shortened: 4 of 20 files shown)


app/models/order_detail.rb
```diff
@@ -265,10 +265,10 @@ def can_be_viewed_by?(user)
     :conditions => ['order_details.state = ?
                      AND problem = ?
                      AND reviewed_at <= ?
-                     AND accounts.type = ?
+                     AND accounts.type IN (?)
```
jhanggi: I wouldn't be adverse to converting this scope to a more modern syntax.

app/models/order_detail.rb
```diff
@@ -808,7 +808,7 @@ def can_reconcile_journaled?
   end
 
   def self.account_unreconciled(facility, account)
-    if account.is_a?(NufsAccount)
+    if account.class.using_journal?
```
jhanggi: Would it be worth making `using_journal?` an instance method on `Account` (that may simply delegate to the `class`)?

app/services/converters/order_detail_to_journal_rows.rb
```diff
@@ -0,0 +1,23 @@
+module Converters
+  class OrderDetailToJournalRows
```
jhanggi: I feel like this could have a more descriptive name. This doesn't really imply it's pretty much one-to-one.

app/services/converters/order_detail_to_journal_rows.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    attr_accessor :journal, :order_detail, :total
+
+    def initialize(journal, order_detail, options = {})
```
jhanggi: I need to think a little more as I process how all this works, but my gut reaction is that this class shouldn't need the journal itself. I see we need it for the association, but I wonder if there's a way to do the association outside this class.

app/services/converters/order_detail_to_journal_rows.rb
```diff
@@ -0,0 +1,23 @@
...
+      @journal = journal
+      @order_detail = order_detail
+      @total = options[:total] || order_detail.total
+    end
+
+    def to_a
+      [{
+        account: order_detail.product.account,
```
jhanggi: Is there a specific reason this returns a hash as opposed to a JournalRow? Based on the class name, that is what I would expect.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,161 @@
...
+              :recharge_enabled, :product_recharges, :journaled_facility_ids
+
+  def initialize(journal, order_details)
+    @order_details = order_details
+    @journal = journal
+    reset
+  end
+
+  # Reset instance variables both on initialize and build.
+  # Returns self to support chaining.
```
jhanggi: This comment appears to be lying.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,161 @@
...
+    @recharge_enabled = SettingsHelper.feature_on?(:recharge_accounts)
+  end
+
+  # Idempotent method that builds an array of new journal_rows. Also adds to
+  # errors if any problems occur. Does not save anything to the database.
+  # Returns self to support chaining.
+  def build
+    reset
+    order_details.each do |order_detail|
+      validate_order_detail_unjournaled(order_detail)
```
jhanggi: I wonder if a `valid_order_detail?(order_detail)` that wraps up these three lines would make the loop cleaner.

config/settings.yml
```diff
@@ -12,6 +12,13 @@ financial:
     csv: true
     xml: true
 
+converters:
+  default:
+    product_to_journal_rows: Converters::ProductToJournalRows
+    order_detail_to_journal_rows: Converters::OrderDetailToJournalRows
+  split_accounts/split_account:
+    order_detail_to_journal_rows: SplitAccounts::Converters::OrderDetailToJournalRows
```
jhanggi: What is the advantage of having all of this configurable? Could there be a method on `Account#order_detail_journal_row_converter` method that returns the default `Converters::OrderDetailToJournalRows`. Then `SplitAccount` could override that.

config/settings.yml
```diff
@@ -12,6 +12,13 @@ financial:
     csv: true
     xml: true
 
+converters:
+  default:
+    product_to_journal_rows: Converters::ProductToJournalRows
+    order_detail_to_journal_rows: Converters::OrderDetailToJournalRows
```
jhanggi: Part of me wants a more specific `JournalRowConverters` namespace (or something). `Converters` just seems really general for this use case.

app/services/journal_row_updater.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def recreate_journal_rows
+    if journal_rows.present?
+      order_detail.journal_rows.destroy_all
+      JournalRowBuilder.new(order_detail.journal, [order_detail]).create
+    end
+  end
+
+  def recreate_journal_rows?
+    method = :recreate_journal_rows_on_order_detail_update
```
jhanggi: If we define this method as returning `false` on the base `Account`, would that obviate the need for the `respond_to?` check?

spec/services/journal_row_updater_spec.rb
```diff
@@ -0,0 +1,20 @@
+require "rails_helper"
+
+RSpec.describe JournalRowUpdater, type: :service do
```
jhanggi: Do we have a spec that checks this when there are multiple journal rows for an order detail? I would guess that spec probably belongs in the split account engine.

app/services/converters/order_detail_to_journal_rows.rb
jhanggi: I think it might be helpful to extract an overridable method that returns a hash of non-common fields. I think in the open case, it would just return `account`. Then you could merge that with description, order_detail_id, amount, and journal_id. That would probably make creating the converter class for NU easier, if we can just override that method.

config/settings.yml
jhanggi: I think I see now: for the NU specific overrides. One idea I had was to have a `JournalRowConverters::BaseOrderDetailConverter`, which is always abstract, and then a `DefaultOrderDetailConverter` that inherits and has just the simple implementation of `{ account: order_detail.product.account }` that gets merged in with everything else. That way there's an example of how to build your own.

jhanggi: Another thought: I think I'm now ok with having the default converter be configurable here, but there's a part of me that feels a little weird about having the split_accounts config here. Seems like that should be part of the engine, though I'm not sure how best to handle that.

vendor/engines/split_accounts/app/services/split_accounts/converters/order_detail_to_journal_rows.rb
```diff
@@ -0,0 +1,52 @@
...
+    class OrderDetailToJournalRows
+
+      SplitAmount = Struct.new(:split, :amount)
+
+      attr_accessor :journal, :order_detail, :total, :splits
+
+      def initialize(journal, order_detail, options = {})
+        @journal = journal
+        @order_detail = order_detail
+        @total = options[:total] || order_detail.total
```
jhanggi: Are these options primarily for testing? They feel a little weird to make them overridable in actual work.

app/models/order_detail.rb
benlinton: I would expect to be able to add `delegate :using_journal?, to: :class` in the `Account` object.  But it seems to not work with active support 3.2.22.1.
```
/Users/.../Workspace/txi/nucore/nucore-open/vendor/bundle/gems/activesupport-3.2.22.1/lib/active_support/core_ext/module/delegation.rb:139:in `module_eval': /Users/.../Workspace/txi/nucore/nucore-open/app/models/account.rb:42: syntax error, unexpected '.' (SyntaxError)
            class.__send__(:using_journal?, *ar...
                  ^
/Users/.../Workspace/txi/nucore/nucore-open/app/models/account.rb:44: syntax error, unexpected '.'
            if class.nil?                         ...
                     ^
/Users/.../Workspace/txi/nucore/nucore-open/app/models/account.rb:49: syntax error, unexpected keyword_end, expecting end-of-input
          end                                                 # end
```

app/models/order_detail.rb
benlinton: :+1: Done.

app/services/converters/order_detail_to_journal_rows.rb
benlinton: How about adding some documentation?  Otherwise what name did you have in mind?

benlinton: `order_detail_to_journal_row_attributes`?

app/services/converters/order_detail_to_journal_rows.rb
benlinton: I was attempting to port the existing logic (to adhere the single responsibility principle) without altering the existing logic too much.  Seems like it will be safer to take baby steps.

app/services/journal_row_builder.rb
benlinton: :+1: Fixed.

app/services/journal_row_builder.rb
benlinton: Done.

config/settings.yml
benlinton: I was torn.  Adding the split account settings here is much more clear, and it's non-obtrusive even though the engine is bleeding outside of it's boundaries a bit.  Moving the settings to the engine make me think I'll create more confusion than it's worth.  What do you think we should do?

config/settings.yml
benlinton: I started with 
```
converters:
  for_journal_rows: 
    default:
       order_details_to_journal_rows: "..."
```
But the `for_journal_rows` seemed redundant.

app/services/journal_row_updater.rb
benlinton: Good idea.

vendor/engines/split_accounts/app/services/split_accounts/converters/order_detail_to_journal_rows.rb
benlinton: They help with testing.  But primarily I need to be able to override the `total` from a root converter to a leaf converter.  A converter should be able to pass a new total to another converter anywhere throughout the converter chain.  This way we could go as fas as to support conversions like `Split Account 1 => Split Account 2 => NufsAccount (the repo specific version)`.  Does that make sense?

spec/services/journal_row_updater_spec.rb
benlinton: Working on this.

config/settings.yml
jhanggi: I would agree that's somewhat redundant in that case, but I still feel `Converters` is overly general. I'd maybe go with `JournalRowConverters::FromOrderDetail` and `JournalRowConverters::FromProduct`. Does that make sense?

app/services/converters/order_detail_to_journal_rows.rb
jhanggi: That makes sense for now. We can deal with some of the attribute merging/whatnot for when we get to the NU piece.

I would lean towards either changing the name of the class to something ending with `JournalRowAttributes` if we're not actually going to be returning JournalRows.

app/services/journal_row_updater.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  def update_journal_rows
+    journal_rows.each(&:update_amount)
+  end
+
+  def recreate_journal_rows
+    if journal_rows.present?
+      order_detail.journal_rows.destroy_all
+      JournalRowBuilder.new(order_detail.journal, [order_detail]).create
```
jhanggi: Should we wrap these two lines up in a transaction? If something fails in the builder, it would make me feel better if we roll back the destroy_all.

config/settings.yml
jhanggi: I think you're right that this is the clearest place.

app/services/converters/product_to_journal_rows.rb
```diff
@@ -0,0 +1,22 @@
+module Converters
+  class ProductToJournalRows
+
+    attr_accessor :journal, :product, :total
+
+    def initialize(journal, product, total)
```
jhanggi: Would it make sense to make total an `option` in order to put the interfaces of the two converters more in sync?

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+        account_number: account.account_number_to_s,
+        validation_error: e.message
+      )
+    end
+  end
+
+  # Given an order detail, return an array of one or more new JournalRow
+  # objects.
+  def order_detail_to_journal_rows(order_detail)
+    factory = Converters::ConverterFactory.new(account: order_detail.account)
```
jhanggi: You could combine some of this to clean up the interface:
`Converters::ConverterFactory.for("order_detail_to_journal_rows", account: order_detail.account)` or even (with some slightly different naming):
`klass = JournalRowConverters::Factory.for_order_detail(account: order_detail.account)`

```
module JournalRowConverters
  class Factory
    def self.for_order_detail(options = {})
       new(options).for("order_detail_to_journal_rows")
    end
  ened
end
```

jhanggi: I'm not a huge fan of having to call `new` when using a factory. I tend to prefer class methods, and then the class method can instantiate an instance of itself if needed.

spec/services/converters/order_detail_to_journal_rows_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+  it "initializes journal" do
+    expect(converter.journal).to eq(journal)
+  end
+
+  it "initializes order_detail" do
+    expect(converter.order_detail).to eq(order_detail)
+  end
+
+  describe "#to_a" do
+    let(:returned) { converter.to_a }
```
jhanggi: Not sure if it's worth it, but you could delegate `each` to `to_a`, and then `include Enumerable` and you should be able to call `size` and `first` on the converter itself.

I'm also not sure I'm in love with `to_a` as the primary API. I wouldn't mind leaving it as an alias, but I think something like `#convert` would be clearer.

app/services/converters/product_to_journal_rows.rb
benlinton: The original functionality required `total`.  With this object I think we have to require `total` to support the previous behavior.  Alternatively, with the order detail converter we can infer the `total` if an explicit value is not provided.

app/services/journal_row_builder.rb
benlinton: Ok, sounds good.

spec/services/converters/order_detail_to_journal_rows_spec.rb
benlinton: Ok, I'll update this.

app/models/order_detail.rb
jhanggi: Maybe `to: "self.class"`?

jhanggi: Or we could just create a method that does the delegation for us.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+  # Idempotent method that builds an array of new journal_rows. Also adds to
+  # errors if any problems occur. Does not save anything to the database.
+  # Returns self to support chaining.
+  def build
+    reset
+    order_details.each do |order_detail|
+      validate_order_detail(order_detail)
+      new_journal_rows = order_detail_to_journal_rows(order_detail)
+      update_product_recharges(order_detail)
+      @journal_rows.concat(new_journal_rows)
```
jhanggi: For the sake of keeping related logic together, is it possible to either combine `new_journal_rows = order_detail_to_journal_rows(order_detail)` and `@journal_rows.concat(new_journal_rows)` lines or put them next to each other instead of being separated by the `update_product_recharges(order_detail)` line.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+  def validate_order_detail(order_detail)
+    validate_order_detail_unjournaled(order_detail)
+    validate_facility_can_journal(order_detail)
+    validate_account(order_detail)
+  end
+
+  # Validate the order_detail hasn't already been journaled
+  def validate_order_detail_unjournaled(order_detail)
+    if order_detail.journal_id.present?
+      @errors << "#{order_detail} is already journaled in journal #{order_detail.journal_id}"
```
jhanggi: Since `errors` is an `attr_reader`, you should be able to use it without the `@`.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+  # Run all validations on an order detail
+  def validate_order_detail(order_detail)
+    validate_order_detail_unjournaled(order_detail)
+    validate_facility_can_journal(order_detail)
+    validate_account(order_detail)
+  end
+
+  # Validate the order_detail hasn't already been journaled
+  def validate_order_detail_unjournaled(order_detail)
+    if order_detail.journal_id.present?
```
jhanggi: Would it be worth pulling this into `OrderDetail` as `in_journal?`? We have `in_open_journal?` and `in_closed_journal?` already.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+      @errors << "#{order_detail} is already journaled in journal #{order_detail.journal_id}"
+    end
+  end
+
+  # Validate the order_details's facility doesn't currently have a pending
+  # journal.
+  #
+  # Ported from original journal model logic. Consider replacing the raised
+  # error with a push to `@errors` instead.
+  def validate_facility_can_journal(order_detail)
```
jhanggi: Since this could possibly raise an error, would it be worth adding a bang to the method name? We don't need to refactor now, but I think a push to `errors` would be nice.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+        journaled_facility_ids.add(facility_id)
+      end
+    end
+  end
+
+  # Validate each journal_row account. This is necessary because an order_detail
+  # may generate multiple journal_rows with different accounts, and we need to
+  # validate them all.
+  #
+  # TODO: we may need to add a validator factory for split_accounts. Otherwise
```
jhanggi: I believe it is a safe assumption that the product's account cannot be a split account.

app/services/journal_row_builder.rb
```diff
@@ -0,0 +1,165 @@
...
+    product_recharges.each_pair do |product_id, total|
+      product = Product.find(product_id)
+      new_journal_rows = product_to_journal_rows(product, total)
+      @journal_rows.concat(new_journal_rows)
+    end
+  end
+
+  # Updates the journal_id of each order_detail.
+  def set_journal_for_order_details(journal, order_detail_ids)
+    array_slice(order_detail_ids) do |id_slice|
```
jhanggi: I don't expect this to fail, but I think we should wrap this in a transaction.

app/services/journal_row_updater.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  def recreate_journal_rows?
+    account = order_detail.account
+    account && account.recreate_journal_rows_on_order_detail_update?
```
jhanggi: What's the way that an order detail would be able to get into this section of code without an account?

app/services/converters/order_detail_to_journal_rows.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    attr_accessor :journal, :order_detail, :total
+
+    def initialize(journal, order_detail, options = {})
+      @journal = journal
+      @order_detail = order_detail
+      @total = options[:total] || order_detail.total
+    end
+
+    def convert
```
jhanggi: I suspect for NU we'll need to write a modified version of this converter that handles the additional journal fields there. I'd like to make sure we have a functional PR on NU before we merge this into open/master.

PR #435, opened 2016-02-02
Better handling of validator errors during expiration setting

jhanggi:
Errors in NU such as BlacklistError were getting raised all the way up
to the view as opposed to caught and displayed in a pretty error
message. This changes the behavior so that `set_expires_at` will never
raise its own errors (I’m not clear why it did before).

### DIFF
--- a/app/models/nufs_account.rb
+++ b/app/models/nufs_account.rb
@@ -6,8 +6,13 @@ class NufsAccount < Account
 
   after_find :load_components
 
-  def set_expires_at!
+  def set_expires_at
     self.expires_at = ValidatorFactory.instance(account_number).latest_expiration
+    # If this Validator instance has an error, suppress it. The errors
+    # will be raised later.
+    rescue ValidatorError, AccountNumberFormatError
+      # Prevent the expires_at validation error message from appearing on the front end
+      self.expires_at ||= Time.current
   end
 
   def account_open?(account_num)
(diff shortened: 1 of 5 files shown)


app/models/nufs_account.rb
```diff
@@ -6,8 +6,13 @@ class NufsAccount < Account
...
   after_find :load_components
 
-  def set_expires_at!
+  def set_expires_at
     self.expires_at = ValidatorFactory.instance(account_number).latest_expiration
+    # If this Validator instance has an error, suppress it. The errors
+    # will be raised later.
+    rescue ValidatorError, AccountNumberFormatError
+      # Prevent the expires_at validation error message from appearing on the front end
+      self.expires_at ||= Time.current
```
drench: Is `Time.current` interchangeable with `Time.zone.now`? Asking because we use `Time.zone.now` in this project much more than `Time.current`.

drench:
One minor question.

app/models/nufs_account.rb
jhanggi: Yes, they're pretty much the same. https://github.com/rails/rails/blob/master/activesupport/lib/active_support/core_ext/time/calculations.rb#L36

I'm partial to `Time.current`, but we have lots of older references to `Time.zone.now`.

drench:
:+1:

PR #437, opened 2016-02-04
[#121325] Prevent backdated journals after cutoff

jhanggi:
I did not feature flag this because if you don't add any `JournalCutoffDate`, then the existing behavior remains the same of being able to backdate a journal as far as you want (well, you can't backdate to before the earliest fulfilled_at).

Basically, the point of this is if you set a cutoff for which you can no longer backdate journals into the previous month. For example, if you set a cutoff for March 3, then after March 3 at 4:45, you can no longer create journals with a February date.

I'm not 100% I love the extraction of the journal validation into a validator class. The API is a little awkward because only a single instance of the Validator is created because that's how rails does it. That's why I left that as its own commit.

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -242,8 +242,8 @@ th.currency {
   }
 }
 
-.table {
-  tr.row-warning td {
+.table tr.row-warning {
+   td, th {
     background-color: $errorBackground;
   }
 }
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -242,6 +242,12 @@ th.currency {
   }
 }
 
+.table {
+  tr.row-warning td {
+    background-color: $errorBackground;
+  }
+}
+
 .table .super-th {
   text-align: center;
 }
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -42,7 +42,7 @@ class FacilityJournalsController < ApplicationController
       redirect_to facility_journals_path(current_facility)
     else
       @order_details = OrderDetail.for_facility(current_facility).need_journal
-      set_soonest_journal_date
+      set_earliest_journal_date
       set_pending_journals
 
       # move error messages for pending journal into the flash
@@ -50,7 +50,7 @@ class FacilityJournalsController < ApplicationController
         flash[:error] = @journal.errors.full_messages.join("<br/>").html_safe
       end
 
-      @soonest_journal_date = params[:journal_date] || @soonest_journal_date
+      @earliest_journal_date = params[:journal_date] || @earliest_journal_date
       render :action => :index
     end
   end
@@ -145,21 +145,23 @@ class FacilityJournalsController < ApplicationController
     @pending_journals = @journals.where(:is_successful => nil)
   end
 
-  def set_soonest_journal_date
-    @soonest_journal_date=@order_details.collect{ |od| od.fulfilled_at }.max
-    @soonest_journal_date=Time.zone.now unless @soonest_journal_date
+  def set_earliest_journal_date
+    @earliest_journal_date = [
+      @order_details.collect { |od| od.fulfilled_at }.max,
+      JournalCutoffDate.first_valid_date
+    ].compact.max
   end
 
   def set_default_variables
     @order_details   = @order_details.need_journal
 
     set_pending_journals
-    set_soonest_journal_date
+    set_earliest_journal_date
 
     blocked_facility_ids = Journal.facility_ids_with_pending_journals
     if cross_facility? || !has_pending_journals?
       @order_detail_action = :create
-      @action_date_field = {:journal_date => @soonest_journal_date}
+      @action_date_field = {:journal_date => @earliest_journal_date}
     end
   end
 
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -161,7 +161,7 @@ class FacilityJournalsController < ApplicationController
     blocked_facility_ids = Journal.facility_ids_with_pending_journals
     if cross_facility? || !has_pending_journals?
       @order_detail_action = :create
-      @action_date_field = {:journal_date => @earliest_journal_date}
+      @action_date_field = { journal_date: @earliest_journal_date }
     end
   end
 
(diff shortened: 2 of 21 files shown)


drench:
The validator works for me. :+1:
@benlinton?

jhanggi:
Rebased off of master to prevent merge conflicts. Any thoughts @benlinton ?

app/assets/stylesheets/nucore.scss
```diff
@@ -242,6 +242,12 @@ th.currency {
   }
 }
 
+.table {
+  tr.row-warning td {
```
benlinton: `, tr.row-warning th`  ?

app/controllers/facility_journals_controller.rb
```diff
@@ -145,21 +145,23 @@ def set_pending_journals
...
 
     set_pending_journals
-    set_soonest_journal_date
+    set_earliest_journal_date
 
     blocked_facility_ids = Journal.facility_ids_with_pending_journals
     if cross_facility? || !has_pending_journals?
       @order_detail_action = :create
-      @action_date_field = {:journal_date => @soonest_journal_date}
+      @action_date_field = {:journal_date => @earliest_journal_date}
```
benlinton: Very minor.  Maybe use ruby 2.0 hash syntax.

app/controllers/journal_cutoff_dates_controller.rb
```diff
@@ -0,0 +1,57 @@
...
+
+  private
+
+  def default_next_cutoff
+    max_future_time = [JournalCutoffDate.maximum(:cutoff_date), Time.current].compact.max
+
+    hour, min = Settings.financial.default_journal_cutoff_time.split(":")
+    # in_time_zone is necessary to get beginning_of_month to respect timezones
+    max_future_time.in_time_zone.next_month.beginning_of_month.change(hour: hour, min: min)
+  end
```
benlinton: Does this logic belong in `JournalCutoffDate` ?

app/models/journal.rb
```diff
@@ -110,7 +110,7 @@ def journal_row_attributes_from_product_and_total(product, total)
   validate :must_have_order_details, :on => :create, :if => :order_details_for_creation
   validate :must_not_span_fiscal_years, :on => :create, :if => :has_order_details_for_creation?
   validate :journal_date_cannot_be_before_last_fulfillment, :on => :create, :if => :has_order_details_for_creation?
-
+  validates_with JournalDateMustBeAfterCutoffs, on: :create
```
benlinton: I'm probably just confusing myself.  But why not on `update`?

app/models/journal_cutoff_date.rb
```diff
@@ -0,0 +1,73 @@
+class JournalCutoffDate < ActiveRecord::Base
+  include ActiveModel::ForbiddenAttributesProtection
+
+  validates :cutoff_date, presence: true
+  validate :unique_month_for_cutoff_date, if: :cutoff_date?
+
+  def self.recent_and_upcoming
+    where("cutoff_date > ?", 1.month.ago.beginning_of_month).order(:cutoff_date)
+  end
```
benlinton: Did we land on avoiding scopes?

app/models/journal_cutoff_date.rb
```diff
@@ -0,0 +1,73 @@
...
+  def cutoff_date_time=(hash)
+    formatted = "#{cutoff_date.to_date} #{hash[:hour]}:#{hash[:minute]}#{hash[:ampm]}"
+    self.cutoff_date = Time.zone.parse(formatted)
+  end
+
+  def cutoff_date_time
+    {
+      hour: cutoff_date.strftime("%-l"),
+      minute:  "%02d" % cutoff_date.min,
+      ampm: cutoff_date.strftime("%p")
```
benlinton: `,`

app/models/journal_cutoff_date.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  def past?
+    cutoff_date < Time.current
+  end
+
+  # TODO consider turning this into a reusable module
+  def cutoff_date_time=(hash)
+    formatted = "#{cutoff_date.to_date} #{hash[:hour]}:#{hash[:minute]}#{hash[:ampm]}"
+    self.cutoff_date = Time.zone.parse(formatted)
+  end
```
benlinton: I'm not sure transforming user input belongs in a model.  But it's probably not worth the extra effort to move it somewhere else.

app/views/journal_cutoff_dates/_form.html.haml
```diff
@@ -0,0 +1,8 @@
+= simple_form_for(@journal_cutoff_date) do |f|
+  = f.input :cutoff_date, as: :string, input_html: { class: "datepicker", value: format_usa_date(f.object.cutoff_date) }
+  .time-select= time_select_tag "journal_cutoff_date[cutoff_date_time]", f.object.cutoff_date
+  = f.button :submit, t(".submit"), :class => ['btn', 'btn-primary']
```
benlinton: 2.0 hash syntax.

benlinton:
:+1:

app/models/journal.rb
jhanggi: This validation only needs to happen on journal creation. It's validating that at the time of creation in relation to `Time.current`, it's within the right time window. If we were to do this validation on `update` (e.g. at journal reconciling), we would be outside the window, so the validation would fail.

app/controllers/journal_cutoff_dates_controller.rb
jhanggi: I went back and forth on that and finally settled on it feeling more like it didn't belong there because it doesn't really get used for anything other than filling in a default value in a text box. In the end I didn't have a strong feeling, so I left it here.

app/models/journal_cutoff_date.rb
jhanggi: I've always felt kind of weird of where input parsing (especially into date/times) should happen. On one hand, I want the model to be able to support `assign/update_attributes` with the `params` so that the controller stays normal. On the other hand, yeah, it's a little weird to do it in the model.

I've done things in the past where [I include a module and then call a method from that module](https://github.com/tablexi/rogerebert/blob/develop/lib/formattable_date_fields.rb). something like `datetime_attribute :cutoff_time, :some_other_time_field` that would set up extra methods like this, but I've never been fully satisfied with the results, but I've yet to find a better way. (It especially gets weird if the app is in UTC--NUcore is in Central Time).

app/controllers/journal_cutoff_dates_controller.rb
benlinton: :+1:

benlinton:
@jhanggi All of your comments make sense to me :+1:

PR #438, opened 2016-02-04
[#122054] Add print.css to asset precompile list.

jhanggi:
We'll need to add it to the staging environments in NU/UIC.


drench:
:+1:

PR #439, opened 2016-02-05
[#88683] Maintain report parameters on export raw

jhanggi:
Before, export raw was submitting a normal request so it would reload
the page. This makes the export raw request via ajax so the page does
not do a full reload.

### DIFF
--- a/app/assets/javascripts/app/reports.coffee
+++ b/app/assets/javascripts/app/reports.coffee
@@ -84,15 +84,19 @@ class TabbableReports
 
   init_export_all_handler: ->
     @$emailToAddressField = $('#email_to_address')
-    $('#export-all').click (event) => @export_all_email_confirm(event)
+    $('#export-all')
+      .attr("data-remote", true)
+      .click (event) => @export_all_email_confirm(event)
 
   export_all_email_confirm: (event) ->
     new_to = prompt 'Have the report emailed to this address:', @$emailToAddressField.val()
+
     if new_to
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/app/reports.coffee
```diff
@@ -84,15 +84,19 @@ class TabbableReports
...
   export_all_email_confirm: (event) ->
     new_to = prompt 'Have the report emailed to this address:', @$emailToAddressField.val()
+
     if new_to
       @$emailToAddressField.val(new_to)
       @update_export_urls()
-    else
-      event.preventDefault()
+      # Actual sending handled by remote: true
+      Flash.info("A report is being prepared and will be emailed to #{new_to} when complete")
```
drench: Potential XSS? It looks like `flash.coffee` doesn't try to escape or sanitize the message.

drench:
That's all I've got.

jhanggi:
Good catch on the XSS. For some reason I was thinking Coffeescript _would_ escape the input. Added a fix in flash.coffee.

jhanggi:
I thought about trying to add some validation on the server side to error out on an invalid email, but it didn't seem worth it for this. The only validation I used in the training request feature was that each email had to have an `@` with characters before and after. This is going to be admins/facility directors using these, so I'm not so concerned. We'll probably get redmine errors if someone does try an invalid email.

drench:
:+1:

PR #440, opened 2016-02-08
[#124333] Fix account change from cart

drench:
While this appears to fix the issue in changing the account but keeping it the same (as happens when only one account is available), I have some questions.

The scenario is this:
* Add a product to your cart
* From the `/orders/:order_id` page, click "Change Payment Source".
* Arrive at `/orders/:order_id/choose_account?reset_account=true`
* Choose the same account you used when adding to the cart
* Arrive back to `/orders/:order_id`; note the "You must select a payment method" warning

The `GET  /orders/:order_id/choose_account?reset_account=true` request sets the `account_id` of the `order_details` to `nil`, but retains the `account_id` of the `order`. I'm not sure why we need to `nil` these out at all, but if we do, I would think we should clear the `account_id` on the `order` to match. Ideally I'd get rid of `reset_account` completely; it's a `GET` and is only resetting things partially anyway. But I didn't change this behavior and added specs to document it.

The `POST` on choosing an account is interesting too, in that it would only set `account_id` on the `order`, but not `order_details`. That is, if they've just been reset, they remain `nil`. This patch should set the `account_id`s on `order_details` to match their parent `order`. But then should that even be happening in the controller? While it doesn't happen currently, should changing an `order`'s `account_id` automatically update the accounts for its `order_details`? Are there reasons why we don't want that to happen?

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -160,10 +160,7 @@ class OrdersController < ApplicationController
           begin
             @order.invalidate
             @order.update_attributes!(account_id: account.id)
-            @order.order_details.each do |order_detail|
-              order_detail.update_attributes!(account_id: account.id)
-            end
-          rescue => e
+          rescue ActiveRecord::ActiveRecordError => e
             success = false
             raise ActiveRecord::Rollback
           end
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -186,13 +186,6 @@ class OrdersController < ApplicationController
       end
     end
 
-    if params[:reset_account]
-      @order.order_details.each do |od|
-        od.account = nil
-        od.save!
-      end
-    end
-
     if session[:add_to_cart].blank?
       @product = @order.order_details[0].try(:product)
     else
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -153,11 +153,7 @@ class OrdersController < ApplicationController
   # POST /orders/:id/choose_account
   def choose_account
     if request.post?
-      begin
-        account = Account.find(params[:account_id])
-        raise ActiveRecord::RecordNotFound unless account.can_be_used_by?(@order.user)
-      rescue
-      end
+      account = Account.find(params[:account_id])
       if account
         success = true
         @order.transaction do
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -163,7 +163,10 @@ class OrdersController < ApplicationController
         @order.transaction do
           begin
             @order.invalidate
-            @order.update_attributes!(:account => account)
+            @order.update_attributes!(account_id: account.id)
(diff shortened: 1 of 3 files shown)


jhanggi:
I don't understand why `reset_account=true` is necessary at all. It seems like it should on the GET, it should not nil out any account. 

I would vote for removing the `reset_account` parameter entirely. Then the update to the order details should happen in the `after_save :update_order_detail_accounts, :if => :account_id_changed?` on `Order`.

Does that sounds reasonable to you?

jhanggi:
Also, this portion of the controller is silly:
```ruby
      begin
        account = Account.find(params[:account_id])
        raise ActiveRecord::RecordNotFound unless account.can_be_used_by?(@order.user)
      rescue
      end
```

The `Account#can_be_used_by?` check ends up happening in the OrderDetail validation, so that check is not really necessary. That code in `Account#can_be_used_by?` looks like it duplicates `OrderDetail#account_usable_by_order_owner?` (I think they're the same, but there's some double-negation so it's a little confusing). It might be nice to do some cleanup around there, but not completely necessary.

drench:
> I would vote for removing the `reset_account` parameter entirely. Then the update to the order details should happen in the `after_save :update_order_detail_accounts, :if => :account_id_changed?` on `Order`.

I'm on board. Working on that change now.

drench:
There are several spots in `#choose_account` checking `session[:add_to_cart]`… the only place I'm aware of that sends a user to `#choose_account` is `app/views/orders/show.html.haml`, and I wouldn't expect an `add_to_cart` happening there. Do those `add_to_cart` references make sense to you?

drench:
This is ready for another look; `reset_account` is gone and it relies on `Order` updating its `OrderDetails` with the new `account_id`.

I added specs that should document how `#choose_account` behaves when `session[:add_to_cart]` is set, but as I mentioned above I'm not sure this is something that would actually happen. If we can lose the parts that act on `add_to_cart` it could shrink this method quite a bit.

app/controllers/orders_controller.rb
```diff
@@ -153,17 +153,16 @@ def remove
...
-      end
+      account = Account.find(params[:account_id])
       if account
         success = true
         @order.transaction do
           begin
             @order.invalidate
-            @order.update_attributes!(:account => account)
+            @order.update_attributes!(account_id: account.id)
+            @order.order_details.each do |order_detail|
```
jhanggi: I believe this is redundant since the `after_save` hook on `Order` should take care of this (and it keeps it in the transaction).

drench: Ack: I committed this removal but didn't push. Coming up.

jhanggi:
Maybe adding a spec for if you try to POST with an account the user doesn't have access to since we removed that `begin/rescue` block would be good. Otherwise, :+1:

drench:
Added a test for `POST`ing an invalid `account_id`.

jhanggi:
:+1:

PR #441, opened 2016-02-08
[#102132] Select no payment source on new reservation forms

drench:
This adds a blank to the account drop-down on the new reservation form, and it should be selected by default. If the user selects an account and there's an error on `create`, the account should remain selected.

### DIFF
--- a/app/views/reservations/new.html.haml
+++ b/app/views/reservations/new.html.haml
@@ -25,14 +25,7 @@
 
 = simple_form_for([@order, @order_detail, @reservation]) do |f|
   = f.error_messages
-
-  - unless @order_detail.bundled?
-    - accounts = @order.user.accounts_for_product(@instrument)
-    - selected_account_id = @order.account_id || params[:order_account]
-    - if selected_account_id.blank? && accounts.count == 1
-      - selected_account_id = accounts.first.id
-
-    = render "account_field", accounts: accounts, selected_account_id: selected_account_id
+  = render partial: 'account_field', locals: { f: f } unless @order_detail.bundled?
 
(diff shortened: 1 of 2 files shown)


jhanggi:
Change looks good for the case of multiple available accounts. Here's a question: if a user only has one payment source, should that be selected by default?

drench:
If there's just one option, I suppose I'd want it selected. I'll make the change.

app/views/reservations/new.html.haml
```diff
@@ -25,7 +25,14 @@
 
 = simple_form_for([@order, @order_detail, @reservation]) do |f|
   = f.error_messages
-  = render partial: 'account_field', locals: { f: f } unless @order_detail.bundled?
+
+  - unless @order_detail.bundled?
+    - accounts = @order.user.accounts_for_product(@instrument)
```
drench: Lines 30 through 33 feels like they're on the edge of extraction into a parameter builder of some kind.

jhanggi: I could definitely see it getting extracted into something; not quite sure what. One option would be a helper. Another would be a SimpleForm input. In `account_field`, you could probably just say `f.input :account, as: :account_select` and the logic could be inside an `AccountSelectInput` class. That should give you access to the object that is stored in `@order`.

jhanggi: Even if you were to just leave it in the view, I would lean towards moving it into `_account_field`.

drench:
I moved the account selection logic back to `account_field`.

jhanggi:
:+1:

PR #442, opened 2016-02-10
Fixes for AccountBuilder

jhanggi:
Move load_components call to only be for NufsAccount. This should fix problems
with the call for credit cards and purchase orders.

Correct error messages on account creation

There were extra error messages appearing for invalid accounts such as
“Expiration cannot be blank” and longer text when the GL066 could not
be found. This should revert to the behavior prior to the account
builder refactor

Cherry-pick and rebase of the main-app changes of https://github.com/tablexi/nucore-nu/pull/116


drench:
:+1:

PR #444, opened 2016-02-12
Update RuboCop gem and config

drench:
This upgrades the RuboCop gem to match the version CodeClimate is currently using.

After they upgraded, we began getting errors because our config uses some now-obsolete cops. This change replaces the obsolete config directives with their updated versions.

If this is good, we should apply a similar patch to https://github.com/tablexi/starter.


jhanggi:
:+1:

PR #445, opened 2016-02-12
JavaScript and CoffeeScript updates

drench:
These are a couple updates I wanted to make as I've been digging into a coffeescript-heavy change. It:

* adds the coffeelint gem and the config from the starter project
* upgrades the teaspoon-jasmine gem, allowing us to use newer features like `expect(something).toBeTruthy()` in our specs


jhanggi:
:+1:

PR #446, opened 2016-02-15
[#99279] Abort in-flight XHR when report parameters change

drench:
Since all inputs on the reports form trigger a tab content load when changed, it was easy for a user to fire off a series of Ajax requests, even though the user likely only cared about the result of the one fired last.

It also caused some confusion, because after a successful XHR return, it would update the date start and end input fields, which may have been changed since.

8a8e5ec changes this behavior so if a request is in-flight (that is, triggered by a previous form input change), it should abort that old XHR before making a new one.

adfea82 is cleanup based on CoffeeLint suggestions.

### DIFF
--- a/app/assets/javascripts/app/reports.coffee
+++ b/app/assets/javascripts/app/reports.coffee
@@ -34,10 +34,16 @@ class TabbableReports
         true
 
       beforeLoad: (_, ui) ->
+        @in_flight_xhr.abort() if @in_flight_xhr?
+        @in_flight_xhr = ui.jqXHR
+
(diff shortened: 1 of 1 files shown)


jhanggi:
This is an interesting technique. It seems to work well on the browser side. The server still ends up processing the full request after the browser cancels the xhr. Not sure if there's a way to fix that. Either way, it's behavior on the server side is no worse that it was before. :+1:

jhanggi:
Let's merge this in, I ran across an issue with the export raw that I have a fix for, but it touches some of the code that got linted here.

PR #448, opened 2016-02-18
[#120455] Fix Purchase Order creation after AccountBuilder refactor

jhanggi:
This field uses `formatted_expires_at` to get the correct formatting. It was not
being allowed by strong params. I also discovered a bug in the spec where it
was testing for inverted month/day (May 12 v Dec 5), otherwise we should have
seen this bug.


drench:
:+1:

PR #449, opened 2016-02-22
[#102132] Retain Reservation duration value on error

drench:
If the user changes the default duration time while making a `Reservation`, either directly or by changing the start/end times, the form should retain those values if the form has errors (such as choosing no payment source).

### DIFF
--- a/app/views/reservations/_reservation_fields.html.haml
+++ b/app/views/reservations/_reservation_fields.html.haml
@@ -5,7 +5,9 @@
     .controls
       .string.optional.control-label &nbsp;
       = time_select f, :reserve_start, { minute_step: @instrument.reserve_interval }, disabled: start_disabled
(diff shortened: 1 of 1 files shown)


jhanggi:
I'm not sure if this was an intentional feature or not: If I try to select a time that is not available aka "The reservation spans time that the instrument is unavailable for reservation" as the result of overlapping with an existing reservation or going outside the scheduling rules, the start/end times would get reset to the default.

With this change, if you choose an invalid time, but _do_ select a payment source, then we retain the invalid time. I don't have a problem with it, but this feels like the kind of thing that could have been intentional (the git history does not really help without digging in too deeply: the change from start/end time to start/duration to start/end/duration makes it somewhat hard to understand).

jhanggi:
We're good with the behavior of leaving the user-input info for all cases.
:+1:

PR #450, opened 2016-02-22
Support MySQL 5.7

drench:
Migrations recently started breaking for me because MySQL 5.7+ does not allow `NULL` primary keys, and older versions of Rails create primary key columns `DEFAULT NULL`. See https://github.com/rails/rails/pull/13247 for background.

This workaround should not be necessary after upgrading the application to Rails 4.

### DIFF
--- a/lib/patches/abstract_mysql_adapter.rb
+++ b/lib/patches/abstract_mysql_adapter.rb
@@ -5,6 +5,9 @@
 # Remove this file and its environment.rb reference after upgrading to Rails 4.
 #
 # https://github.com/rails/rails/pull/13247#issuecomment-32425844
+
+raise "Remove #{__FILE__} for Rails 4+" unless Rails.version < "4"
+
 class ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter
   NATIVE_DATABASE_TYPES[:primary_key] = "int(11) auto_increment PRIMARY KEY"
 end
--- a/lib/patches/abstract_mysql_adapter.rb
+++ b/lib/patches/abstract_mysql_adapter.rb
@@ -6,7 +6,7 @@
 #
 # https://github.com/rails/rails/pull/13247#issuecomment-32425844
 
-raise "Remove #{__FILE__} for Rails 4+" unless Rails.version < "4"
+raise "Remove #{__FILE__} for Rails 4+" if Rails.version >= "4"
 
 class ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter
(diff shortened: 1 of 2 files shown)


lib/patches/abstract_mysql_adapter.rb
```diff
@@ -0,0 +1,10 @@
+# TODO:
+# MySQL 5.7+ does not allow NULL primary keys, but older versions of Rails
+# create primary key columns DEFAULT NULL.
+#
+# Remove this file and its environment.rb reference after upgrading to Rails 4.
+#
+# https://github.com/rails/rails/pull/13247#issuecomment-32425844
+class ActiveRecord::ConnectionAdapters::AbstractMysqlAdapter
```
jhanggi: Should this be `ActiveRecord::ConnectionAdapters::Mysql2Adapter` for `mysql2`?

jhanggi:
Will we get any sort of warning when we upgrade, or will we just need to rely on `rake notes`?

drench:
What would you think about having it raise an exception if `Rails.version >= "4"`?

lib/patches/abstract_mysql_adapter.rb
drench: I went with this class since it's specifically overriding this: https://github.com/yahonda/rails/blob/v3.2.19/activerecord/lib/active_record/connection_adapters/abstract_mysql_adapter.rb#L114

jhanggi: Cool. Makes sense. I just noticed that in the conversation on the rails PR.

jhanggi:
I'd be fine with an exception. This should be easy enough to remove as part of the migration. I'd also be OK with a deprecation warning that shows up when we run `rails s` or `rspec`.

lib/patches/abstract_mysql_adapter.rb
```diff
@@ -0,0 +1,13 @@
+# TODO:
+# MySQL 5.7+ does not allow NULL primary keys, but older versions of Rails
+# create primary key columns DEFAULT NULL.
+#
+# Remove this file and its environment.rb reference after upgrading to Rails 4.
+#
+# https://github.com/rails/rails/pull/13247#issuecomment-32425844
+
+raise "Remove #{__FILE__} for Rails 4+" unless Rails.version < "4"
```
jhanggi: `unless Rails.version < "4"` feels a little awkward. Would `if Rails.version >= "4"` be easier to process?

jhanggi:
One minor question, otherwise, :+1:

PR #451, opened 2016-02-24
Rubocop

mlineen:


mlineen:
@jhanggi your wish is my comand

drench:
We're going to need something like this while we're still on Rails 3.2:

```
Rails/ActionFilter:
  EnforcedStyle: filter # TODO: remove after upgrading to Rails 4
```

I don't know if it belongs in `.rubocop_todo.yml` or `.rubocop-txi.yml`

jhanggi:
Looks like Rails/FindEach needs to be undone. It converts `.all.each` to `all.find_each`,  which doesn't work in 3.2 because `all` in 3.2 executes the SQL and turns it into an array.

jhanggi:
@mlineen I think we're at a more stable point in open/master where we can run this and hopefully get it in quickly. Watch out for Rails/ActionFilter and Rails/FindEach as they caused problems the last time. Probably best to just start a new branch and close this one. 

And we should make sure that github and codeclimate are set up to run rubocop on PRs.

PR #452, opened 2016-02-25
Do not load mysql patch for 5.7 in oracle mode

jhanggi:
Related to #450


drench:
Ack, I should have seen that coming. :+1:

benlinton:
:+1:

PR #453, opened 2016-02-26
Ignore order of order details for journal remover spec

benlinton:
Ignore order of order details for journal remover spec.
Cherry pick a commit from https://github.com/tablexi/nucore-nu/pull/121


jhanggi:
:+1:

PR #454, opened 2016-02-26
Move test-unit out of development/test group

jhanggi:
It is not included by default in Ruby 2.2, and the rails console requires it. Without
including it everywhere, we can't run `rails console` on staging or production.
https://github.com/rails/rails/blob/3-2-stable/railties/lib/rails/console/app.rb


drench:
:+1:

PR #455, opened 2016-02-27
[#120455] Nicer display of extra penny boolean

jhanggi:
I’d like to keep the ticket number of the redmine task since future people probably won’t have access to the pivotal queue we’re using. It’s all sub-parts of the same feature.


drench:
:+1:

benlinton:
:+1: Looks good.  Needs conflicts resolved.

PR #456, opened 2016-02-29
[#120455] Split accounts reports

benlinton:
https://www.pivotaltracker.com/story/show/114954883

We have two options to support SplitAccounts for Reporting.

1. Completely rewrite the Reporting logic.
2. Create an "order detail simulator" service object that replaces each instance of an `order_detail` containing a `split_account` with a collection of "simulated" split `order_details`.  Meaning we clone the order detail for each split.  Then mutate the quantity, cost totals, subsidy totals, and associated account for each clone before passing the collection to the Report objects.

This PR opts for the second option.

The first option could eliminate a lot of technical debt, and would be far more performant.  But it would also be a rather time consuming and require risky sweeping changes.  The second option is less performant but it is far less intrusive, and it lends itself fairly well as a stepping stone to rewriting Report logic in the future.

**Note:**  We may want to invest some time into forcing cloned order details to be read-only.  It's not necessary at this time since we're not saving order details within the reporting objects, but it would be a smart safe guard that we could eventually implement.


benlinton:
I believe we will have to convert the `order_detail` `quantity` column from an `integer` to a `decimal` so we can support simulating partial quantities.  @jhanggi do you foresee any problem with this?  If so I could add support for an `overridden_quantity` virtual attribute instead.  Although it would be easiest if we didn't have to deal with the virtual attribute.

vendor/engines/split_accounts/app/services/split_accounts/split_order_detail_simulator.rb
```diff
@@ -0,0 +1,90 @@
+module SplitAccounts
+
+  # TODO: maybe refactor some of this into SplitSimulator
+  # TODO: we can refactor this code to fewer lines and iterate fewer times
+  class SplitOrderDetailSimulator
```
jhanggi: I'm not sure I love the term `Simulator`. I don't really have anything better, though.

vendor/engines/split_accounts/app/services/split_accounts/reports/querier.rb
```diff
@@ -0,0 +1,29 @@
...
+    class Querier < ::Reports::Querier
+
+      def perform
+        super.map do |order_detail|
+          simulate_order_details(order_detail)
+        end.flatten
+      end
+
+      def simulate_order_details(order_detail)
+        if order_detail.account.is_a?(SplitAccounts::SplitAccount)
```
jhanggi: Is it possible to make this check not necessary? Can you call `splits` on a non-split account and have it return `self`? Or a 100% split version of itself? Then the rest of the logic should apply.

jhanggi: My other thought is that if we do keep the `is_a?` check, that it gets pulled into `SplitOrderDetailSimulator`. Then this class becomes pretty trivial and the simulator is what you're actually overriding in Settings, not the Querier, since the query is always the same.

jhanggi:
So overall I'm onboard with where this is heading. I'd definitely like to see some refactoring to `SplitOrderDetailSimulator`, that class is a beast. I'd also be curious to see a separation of the `Querier` logic (I do like splitting that class out of the controller), but then as opposed to overriding it, pass the output to a `Simulator` (or `Transformer` or some better name). The default would just return the input while the split simulator would expand out into the virtual order details.

app/controllers/general_reports_controller.rb
```diff
@@ -139,13 +138,14 @@ def report_data
...
 
-  def report_data_query(stati, date_column)
-    return [] if stati.blank?
-    # default to using fulfilled_at
-    result = OrderDetail.where(:order_status_id => stati)
-                        .for_facility(current_facility)
-                        .action_in_date_range(date_column, @date_start, @date_end)
-                        .includes(:order, :account, :price_policy, :product, :order_status)
+  def report_data_query(order_status_id, date_range_field)
+    Reports::QuerierFactory.instance({
```
jhanggi: Curly braces aren't needed here.

app/services/reports/querier.rb
```diff
@@ -0,0 +1,27 @@
+module Reports
+
+  class Querier
+
+    attr_reader :order_status_id, :current_facility, :date_range_field,
+      :date_range_start, :date_range_end
+
+    def initialize(options = {})
```
jhanggi: Would `options` be worth doing as keyword arguments?

vendor/engines/split_accounts/app/services/split_accounts/split_order_detail_simulator.rb
```diff
@@ -0,0 +1,90 @@
...
+      clone.actual_cost = item.actual_cost
+      clone.actual_subsidy = item.actual_subsidy
+      clone.estimated_cost = item.estimated_cost
+      clone.estimated_subsidy = item.estimated_subsidy
+
+      clone
+    end
+
+    def simulated_splits
+      items = splits.map { |split| simulated_split(split) }
```
jhanggi: As we get deeper into these methods, it becomes very difficult to remember what `items` is. I'm also having trouble wrapping my head around `SimulatedSplit` and how it's different from the simulated order detail.

vendor/engines/split_accounts/app/services/split_accounts/split_order_detail_simulator.rb
```diff
@@ -0,0 +1,90 @@
...
+    end
+
+    def floored_total(items, attr)
+      items.reduce(BigDecimal("0")) { |sum, item| sum + item.send(attr) }
+    end
+
+    def floored_amount(percent, attr)
+      return 0 if percent == 0 || order_detail.send(attr).blank?
+
+      cents = (order_detail.send(attr) * 100) * (BigDecimal(percent) / 100)
```
jhanggi: `BigDecimal#floor` can take an argument for what precision you want. All this math seems overly complicated (and I'm worried about the `to_f / 100` ruining our floating point math.

```ruby
amount = order_detail.send(attr) * BigDecimal(percent) / 100
amount.floor(2)
```

vendor/engines/split_accounts/lib/split_accounts/engine.rb
```diff
@@ -25,7 +27,7 @@ class Engine < Rails::Engine
 
     # Include factories in main rails app
     initializer "model_core.factories", after: "factory_girl.set_factory_paths" do
-      if defined?(FactoryGirl) && SettingsHelper.feature_on?(:split_accounts)
+      if defined?(FactoryGirl)
```
jhanggi: Is this change so that the tests still work if the setting is turned off?

jhanggi:
From testing: not sure if it should be part of this PR or a separate one, but I think we should format the report so that quantity does not have a decimal if it's `.0`. 99% of the time there won't be split charges, so I'd rather it look similar to how it did before.

screen shot 2016-02-29 at 7 10 34 pm

jhanggi:
Is this intended to modify the Export Raw? The export raw is not split out yet. I'm fine with that being another PR, but I saw the pivotal story is finished, but don't see that here.

vendor/engines/split_accounts/lib/split_accounts/engine.rb
benlinton: Yes.

vendor/engines/split_accounts/app/services/split_accounts/split_order_detail_simulator.rb
benlinton: I'm not crazy about `Simulator` either.  I was trying to convey that the resulting order details were `read-only`.

vendor/engines/split_accounts/app/services/split_accounts/reports/querier.rb
benlinton: I originally added `splits` to the `Account` model.  But then decided against it since that means the account split gem is bleeding into the main app.  I like your idea to move it into the `SplitOrderDetailSimulator` object (or whatever we end up calling it.)

jhanggi: We have an `AccountExtension` module where we're adding `parent_split_accounts` and a couple other scopes/associations. We could add `splits` from there. But yeah, we definitely shouldn't put it directly in the Account class.

app/services/reports/querier.rb
benlinton: Yeah.  Good idea.

PR #457, opened 2016-02-29
[#120455] Verify splits have unique subaccounts

benlinton:
Test that you cannot have duplicate accounts in a split

https://www.pivotaltracker.com/story/show/114612951

### DIFF
--- a/vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
+++ b/vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
@@ -6,6 +6,7 @@ module SplitAccounts
 
     validate :valid_percent_total
     validate :one_split_has_extra_penny
+    validate :unique_split_subaccounts
 
     accepts_nested_attributes_for :splits, allow_destroy: true
 
@@ -25,8 +26,21 @@ module SplitAccounts
       end
     end
 
+    def unique_split_subaccounts
+      if duplicate_subaccounts?
+        errors.add(:splits, :duplicate_subaccounts)
+      end
+    end
+
     def extra_penny_count
       splits.select(&:extra_penny?).size
+    end
+
+    def duplicate_subaccounts?
+      splits.map(&:subaccount_id)
+        .group_by{ |e| e }
+        .select{ |k, v| v.size > 1 }
+        .present?
     end
 
     # Stopped using SQL because that didn't seem to work until the built splits
--- a/vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
+++ b/vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
@@ -38,9 +38,8 @@ module SplitAccounts
 
     def duplicate_subaccounts?
       splits.map(&:subaccount_id)
-        .group_by{ |e| e }
-        .select{ |k, v| v.size > 1 }
-        .present?
+        .group_by { |e| e }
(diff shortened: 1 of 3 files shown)


vendor/engines/split_accounts/spec/models/split_account_spec.rb
```diff
@@ -58,8 +61,23 @@
...
-          split_account.splits << build(:split, percent: 50, extra_penny: true, parent_split_account: split_account)
-          split_account.splits << build(:split, percent: 50, extra_penny: true, parent_split_account: split_account)
+          split_account.splits << build(:split, percent: 50, extra_penny: true, parent_split_account: split_account, subaccount: subaccount_1)
+          split_account.splits << build(:split, percent: 50, extra_penny: true, parent_split_account: split_account, subaccount: subaccount_2)
+        end
+      end
+
+      it "is invalid" do
+        expect(split_account).not_to be_valid
+        expect(split_account.errors).to include(:splits)
```
jhanggi: I feel like now that we're adding more validations, we should probably check that it's invalid for the right reason.

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
```diff
@@ -25,10 +26,23 @@ def one_split_has_extra_penny
...
+        errors.add(:splits, :duplicate_subaccounts)
+      end
+    end
+
     def extra_penny_count
       splits.select(&:extra_penny?).size
     end
 
+    def duplicate_subaccounts?
+      splits.map(&:subaccount_id)
```
jhanggi: It's not that bad, but it seems like there should be a cleaner way to write this.

```
subaccounts = splits.map(&:subaccount_id)
subaccounts.length == subaccounts.uniq.length
```
or
```
splits.map(&:subaccount_id)
  .group_by { |e| e }
  .any? { |k, v| v.size > 1 }
```

vendor/engines/split_accounts/spec/models/split_account_spec.rb
benlinton: Done.

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
benlinton: Done.  I believe the second option is supposed to be much more performant.

benlinton:
@jhanggi Ready for a re-review.

vendor/engines/split_accounts/spec/models/split_account_spec.rb
```diff
@@ -30,42 +33,61 @@
...
-          split_account.splits << build(:split, percent: 50, extra_penny: false, parent_split_account: split_account)
+          split_account.splits << build(:split, percent: 50, extra_penny: false, parent_split_account: split_account, subaccount: subaccount_1)
+          split_account.splits << build(:split, percent: 50, extra_penny: false, parent_split_account: split_account, subaccount: subaccount_2)
         end
       end
 
       it "is invalid" do
         expect(split_account).not_to be_valid
-        expect(split_account.errors).to include(:splits)
+        message = I18n.t("activerecord.errors.models.split_accounts/split_account.attributes.splits.only_one_extra_penny")
```
jhanggi: Seems like there should be a cleaner way to do this test, but I haven't seen a better way.

jhanggi:
:+1:

PR #460, opened 2016-03-01
[#120455] Add chosen to split account dropdown

benlinton:
https://www.pivotaltracker.com/story/show/114612023

The subaccount dropdown was unwieldy with a large data set, so we are now using `chosen` for that particular dropdown.

screen shot 2016-03-01 at 1 18 57 pm

**Note:** The vertical alignment of the inputs is slightly off due to loading chosen.  This did not appear to be a trivial fix and will require some investigation.


jhanggi:
We're not particularly consistent elsewhere about sorting lists of accounts, but when we do, we sort by `order(:account_number)` or `order(:account_number, :description)`. I'd be fine if we wanted to do that as a separate PR, though. Otherwise this looks good.

PR #461, opened 2016-03-02
[#120455] Only global admins can manage SplitAccounts

jhanggi:
This prevents non-admins from being able to edit as well.

Pivotal #114612043

### DIFF
--- a/vendor/engines/split_accounts/spec/controllers/facility_accounts_controller_spec.rb
+++ b/vendor/engines/split_accounts/spec/controllers/facility_accounts_controller_spec.rb
@@ -1,0 +1,119 @@
+require "rails_helper"
+require_relative "../split_accounts_spec_helper"
+
+RSpec.describe FacilityAccountsController, :enable_split_accounts do
+  render_views
+
+  let(:facility) { FactoryGirl.create(:setup_facility) }
+  before { sign_in user }
+
+  describe "as an admin" do
+    let(:user) { FactoryGirl.create(:user, :administrator) }
+
+    describe "default new" do
+      before { get :new, facility_id: facility.url_name, owner_user_id: user.id }
+
+      it "sees split account option" do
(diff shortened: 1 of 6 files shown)


jhanggi:
@benlinton Take a look, especially the second commit, which is an attempt at a method of enabling the feature if needed. Let me know your thoughts.

benlinton:
This is what I had in mind.  This is a great start.  Looks good!

benlinton:
Sorry for commenting on a single commit.

vendor/engines/split_accounts/spec/controllers/facility_accounts_controller_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+  describe "as an admin" do
+    let(:user) { FactoryGirl.create(:user, :administrator) }
+
+    describe "default new" do
+      before { get :new, facility_id: facility.url_name, owner_user_id: user.id }
+
+      it "sees split account option" do
+        expect(response.code).to eq("200")
+        expect(response.body).to include("Chart String")
+        expect(response.body).to include("Split Account")
```
benlinton: Should we use `I18n` or is it too much of a pain?

jhanggi: You're probably right. I'll change it before I merge it down.

jhanggi: After looking at it, it's kind of a pain, and also since this is effectively an integration test (we're rendering out the view, just skipping the routing), I think I'd like it to break if we change things.

jhanggi:
I've rebased this off of #467 which has the enabling/disabling code, so only the last commit is really relevant to this feature.

benlinton:
:+1:

PR #462, opened 2016-03-03
[#120455] Split accounts reports export raw

benlinton:
Adds support for split accounts and export raw.

https://www.pivotaltracker.com/story/show/114612217

### DIFF
--- a/app/controllers/general_reports_controller.rb
+++ b/app/controllers/general_reports_controller.rb
@@ -88,7 +88,7 @@ class GeneralReportsController < ReportsController
   end
 
 
-  def init_report_headers(report_on_label)
+  def init_report_headers(report_on_label = nil)
     if !report_data_request?
       @headers=[ report_on_label, 'Quantity', 'Total Cost', 'Percent of Cost' ]
     else
--- a/app/models/reports/export_raw.rb
+++ b/app/models/reports/export_raw.rb
@@ -2,6 +2,8 @@ require 'csv'
 class Reports::ExportRaw
   include DateHelper
 
+  attr_reader :order_status_ids, :facility, :date_range_field
+
   def initialize(arguments)
     [:action_name, :date_end, :date_start, :headers, :facility, :order_status_ids].each do |property|
       if arguments[property].present?
@@ -169,12 +171,14 @@ class Reports::ExportRaw
   end
 
   def report_data_query
-    return [] if @order_status_ids.blank?
-
-    OrderDetail.where(order_status_id: @order_status_ids)
-      .for_facility(@facility)
-      .action_in_date_range(@date_range_field, date_start, date_end)
-      .includes(:account, :order, :order_status, :price_policy, :product, :reservation, :statement)
+    Reports::Querier.new(
+      order_status_id: @order_status_ids,
+      current_facility: @facility,
+      date_range_field: @date_range_field,
+      date_range_start: date_start,
+      date_range_end: date_end,
+      includes: [:reservation, :statement],
+    ).perform
   end
 
   def as_currency(number)
--- a/app/services/reports/querier.rb
+++ b/app/services/reports/querier.rb
@@ -3,7 +3,7 @@ module Reports
   class Querier
 
     attr_reader :order_status_id, :current_facility, :date_range_field,
-      :date_range_start, :date_range_end
+      :date_range_start, :date_range_end, :extra_includes
 
     def initialize(options = {})
       @order_status_id = options[:order_status_id]
@@ -11,6 +11,7 @@ module Reports
       @date_range_field = options[:date_range_field]
       @date_range_start = options[:date_range_start]
       @date_range_end = options[:date_range_end]
+      @extra_includes = options[:includes]
     end
 
     def perform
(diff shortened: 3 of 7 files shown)


app/models/reports/export_raw.rb
```diff
@@ -169,12 +169,14 @@ def csv_body
...
 
   def report_data_query
-    return [] if @order_status_ids.blank?
-
-    OrderDetail.where(order_status_id: @order_status_ids)
-      .for_facility(@facility)
-      .action_in_date_range(@date_range_field, date_start, date_end)
-      .includes(:account, :order, :order_status, :price_policy, :product, :reservation, :statement)
+    Reports::Querier.new(
+      order_status_id: @order_status_id,
```
jhanggi: I think this should be `@order_status_ids`

jhanggi:
The second column for the Order Id is missing the orderdetail part of the ID. Example: should be "187-215", is "187-".

jhanggi:
I made a couple changes that should address what I found while testing.

app/services/reports/querier.rb
```diff
@@ -17,12 +17,20 @@ def perform
       Reports::TransformerFactory.instance(order_details).perform
     end
 
-    def order_details
+    def order_details(options = {})
       return [] if order_status_id.blank?
+
+      includes = default_includes
+      includes.concat(options[:includes]) if options[:includes].present?
```
jhanggi: Is this `includes` coming from the call to this method or the initializer?

benlinton:
Ready for review.

benlinton:
As a random suggestion, I would love to figure out how to reduce/improve the setup required for tests involving orders.  I feel like we investigate using mocks, writing better factories, or decoupling the objects related to orders.

vendor/engines/split_accounts/spec/services/order_detail_splitter_spec.rb
```diff
@@ -13,8 +12,10 @@
     end
   end
 
+  let(:order) { build_stubbed(:order) }
+
   let(:order_detail) do
-    build_stubbed(:order_detail).tap do |order_detail|
+    build_stubbed(:order_detail, order: order).tap do |order_detail|
```
jhanggi: Not saying it needs to be changed now, but is there a reason we're not adding the things inside `tap` just using the standard `build_stubbed` options hash?

vendor/engines/split_accounts/spec/models/reports/export_raw_spec.rb
```diff
@@ -0,0 +1,119 @@
...
+      date_start: 1.day.ago,
+      date_range_field: "ordered_at"
+    }
+  end
+
+  let(:headers) { I18n.t("controllers.general_reports.headers.data") }
+  let(:lines) { report.to_csv.split("\n") }
+  let(:cells) { lines.map{ |line| line.split(",") } }
+  let(:cells_without_headers) { cells[1..-1] }
+  let(:column_values) { cells_without_headers.map {|line| line[column_index]} }
```
jhanggi: Missing space in between `{` and `|`. That can probably wait for us to re-introduce rubocop, though.

app/services/reports/querier.rb
```diff
@@ -17,12 +17,20 @@ def perform
       OrderDetailListTransformerFactory.instance(order_details).perform
     end
 
-    def order_details
+    def order_details(options = {})
```
jhanggi: Do we need this `options` param?

jhanggi: Or it seems like maybe we should be storing `options` from the initializer into `@options`?

app/controllers/general_reports_controller.rb
```diff
@@ -88,7 +88,7 @@ def init_report_params
   end
 
 
-  def init_report_headers(report_on_label)
+  def init_report_headers(report_on_label = nil)
```
jhanggi: Where does this get called without the parameter that it needs a default?

jhanggi:
Just a couple minor comments, but overall looking pretty good.

app/controllers/general_reports_controller.rb
benlinton: It doesn't get called without `report_on_label` yet.  But the `report_on_label` is optional.  I ran into this while experimenting with testing, before finding a better testing approach.

app/services/reports/querier.rb
benlinton: I think we'll need to address this downstream in the `nu` branch.  With that branch I believe we add a few extra `includes`.  It was handled before but rebasing with master may have broke something.

vendor/engines/split_accounts/spec/models/reports/export_raw_spec.rb
benlinton: :+1: Fixed

vendor/engines/split_accounts/spec/services/order_detail_splitter_spec.rb
benlinton: I try to use `stubbed_build` instead of `build` if possible so I don't end up creating database entries for associations (something factory girl likes to do).

app/services/reports/querier.rb
jhanggi: Do you know what branch that is? I don't see where this method is being called publicly anywhere.

vendor/engines/split_accounts/spec/services/order_detail_splitter_spec.rb
jhanggi: I was actually talking about 
```ruby
build_stubbed(:order_detail, order: order, created_by: 1, quantity: 3, estimated_subsidy: BigDecimal("39.99"))
```
instead of the `tap`

benlinton: Oh good point.  Yeah I started with a normal `build` and later changed to `build_stubbed`.

app/services/reports/querier.rb
benlinton: I'll look into this and report back on slack.

jhanggi:
:+1:

PR #463, opened 2016-03-03
[#120455] Ensure split accounts have two or more splits

benlinton:
The split account form should gracefully error out if a user attempts to create a split account with only one 100% split.  We will display an error message that they must have two or more subaccounts.

https://www.pivotaltracker.com/story/show/114612937

### DIFF
--- a/vendor/engines/split_accounts/config/locales/en.yml
+++ b/vendor/engines/split_accounts/config/locales/en.yml
@@ -9,6 +9,7 @@ en:
               percent_total: percent total must equal 100
               only_one_extra_penny: must have one split with extra penny
               duplicate_subaccounts: must have unique subaccounts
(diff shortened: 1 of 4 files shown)


vendor/engines/split_accounts/config/locales/en.yml
```diff
@@ -9,6 +9,7 @@ en:
               percent_total: percent total must equal 100
               only_one_extra_penny: must have one split with extra penny
               duplicate_subaccounts: must have unique subaccounts
+              more_than_one_split: must have two or more
```
jhanggi: This message becomes a little awkward when it displays: "Splits must have two or more". Maybe just add "subaccounts" to the end?

jhanggi:
Other than a possible messaging change, :+1:

PR #464, opened 2016-03-03
[#120455] Prevent 500 when submitting split account without any splits

benlinton:
The `earliest_expiring_subaccount` method in the `split_account` model was expecting valid subaccounts which was resulting in 500 errors when submitting a split account without any subaccounts selected.  This PR fixes the issue.

https://www.pivotaltracker.com/story/show/114804481


jhanggi:
:+1:

PR #465, opened 2016-03-03
[#120455] Ensure split percent is within range

benlinton:
https://www.pivotaltracker.com/story/show/114611681

Users should not be able to create splits with 200% and -100%; that add up to 100% but venture outside of a valid percentage range.

### DIFF
--- a/vendor/engines/split_accounts/app/models/split_accounts/split.rb
+++ b/vendor/engines/split_accounts/app/models/split_accounts/split.rb
@@ -8,6 +8,7 @@ module SplitAccounts
 
     validates :parent_split_account, presence: true
     validates :subaccount, presence: true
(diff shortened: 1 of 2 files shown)


vendor/engines/split_accounts/app/models/split_accounts/split.rb
```diff
@@ -8,6 +8,7 @@ class Split < ActiveRecord::Base
 
     validates :parent_split_account, presence: true
     validates :subaccount, presence: true
+    validates :percent, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }
```
jhanggi: I'd really like to see this be `greater_than: 0` and `less_than: 100`.

jhanggi:
:+1: We've decided to push preventing 100%/0% splits until later as it causes testing headaches if it's invalid.

PR #467, opened 2016-03-04
Split account spec fixes

jhanggi:
Many of the SplitAccount specs haven't been running because the feature is still enabled. This enables the feature for those that need them, and then disables it after running the tests.

The second commit fixes a test that was failing due to a change, but the test that would have caught it wasn't running.

Keep this as two separate commits.


benlinton:
:+1:

PR #468, opened 2016-03-04
[#121871] Make print button respect feature setting

jhanggi:
This accidentally got committed with the check for the feature flag commented
out.


drench:
:+1:

PR #469, opened 2016-03-05
[#120455] Fix for pricing updates when on a failed journal

jhanggi:
If you were to try to update the pricing of an order detail that was
part of a failed journal and on a split account, it would fail because
trying to update the failed journal rows would fail because the order
details’ journal_ids get set to nil when a journal gets marked as
failed.

Pivotal #115078649


benlinton:
:+1:

spec/services/journal_row_updater_spec.rb
```diff
@@ -17,4 +17,35 @@
     end
   end
 
+  describe "when the order detail is on a failed journal" do
+    let(:user) { FactoryGirl.create(:user) }
+    let(:facility) { FactoryGirl.create(:setup_facility) }
+    let(:account) { FactoryGirl.create(:setup_account, owner: user) }
+    # Original costs: actual: 20, subsidy: 10
```
benlinton: I wonder if instead of adding this comment, we should explicitly setup these values in the test?  Unless it's a lot of hassle.

spec/services/journal_row_updater_spec.rb
```diff
@@ -17,4 +17,35 @@
...
+        order_detail.update_attributes(actual_cost: 20.37)
+      end
+
+      it "updates the order detail" do
+        expect(order_detail.reload.actual_cost).to eq(20.37)
+        expect(order_detail.total).to eq(10.37)
+      end
+
+      it "updates the journal row" do
+        expect(JournalRow.where(order_detail_id: order_detail.id).first.amount).to eq(10.37)
```
benlinton: `find_by()` instead o `where().first`?

spec/services/journal_row_updater_spec.rb
jhanggi: It's a pain. No easy way to inject it into `place_and_complete_item_order`. That's why I ended up and just did the comment.

spec/services/journal_row_updater_spec.rb
jhanggi: Good call, except 3.2 doesn't have that `find_by`, but we can use `find_by_order_detail_id()`

PR #471, opened 2016-03-08
[#120455] Hide insignificant decimals on reports

jhanggi:
Only show decimals in quantities where needed on reports. Don’t show if
it’s .0 (which will be the majority of time since split accounts will
be the special case).

Pivotal #115081495

### DIFF
--- a/app/views/general_reports/_report_rows.html.haml
+++ b/app/views/general_reports/_report_rows.html.haml
@@ -2,11 +2,19 @@
   - @rows.each do |row|
     %tr
       - row.each_with_index do |col, i|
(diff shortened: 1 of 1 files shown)


app/views/general_reports/_report_rows.html.haml
```diff
@@ -2,11 +2,19 @@
   - @rows.each do |row|
     %tr
       - row.each_with_index do |col, i|
-        %td{:class => i == 0 ? '' : 'currency'}= i == 2 ? number_to_currency(col) : i == 3 ? number_to_percentage(col, :precision => 1) : col
+        %td{class: i == 0 ? '' : 'currency'}
+          - if i == 2
```
drench: I'd probably lean toward a `case`/`when` structure instead but it's not a showstopper. This is so much clearer than what was there.

drench:
:+1:

PR #472, opened 2016-03-09
[#120455] Refactor split journal creation to share code with reports

jhanggi:


vendor/engines/split_accounts/app/services/split_accounts/order_detail_list_transformer.rb
```diff
@@ -0,0 +1,25 @@
...
+  class OrderDetailListTransformer
+
+    attr_reader :order_details
+
+    def initialize(order_details = [])
+      @order_details = order_details
+    end
+
+    def perform
+      order_details.map do |order_detail|
```
jhanggi: I'll put this in a separate PR once this is in since it will cause conflicts: I was doing some performance testing. Testing against 1000 OrderDetails all being on a SplitAccounts, using the following using `each_with_object` instead of the `map`/`flatten` had a 30% speed increase (~7.2 seconds down to ~4.7). For reference, testing against 1000 OrderDetails on a normal account, there is negligible performance impact of using this Transformer.

```ruby
      def perform
        order_details.each_with_object([]) do |order_detail, results|
          # We will need to refactor the general_reports_controller_spec in
          # order to remove the `try` methods below.
          if order_detail.account.try(:splits).try(:present?)
            results.concat SplitAccounts::OrderDetailSplitter.new(order_detail).build_split_order_details
          else
            results << order_detail
          end
        end
      end
```

jhanggi: I also tried `order_details.flat_map` and while that's an improvement over the `map`/`flatten`, it still tended to be slower than the `each_with_object`, running around ~5.2 seconds.

app/services/journal_row_builder.rb
```diff
@@ -31,17 +30,29 @@ def reset
...
   def build
-    reset
-    order_details.each do |order_detail|
-      validate_order_detail(order_detail)
-      new_journal_rows = order_detail_to_journal_rows(order_detail)
-      update_product_recharges(order_detail)
-      @journal_rows.concat(new_journal_rows)
+    build_order_detail_journal_rows do |virtual_order_detail|
+      validate_order_detail(virtual_order_detail)
     end
```
benlinton: :+1:

db/migrate/20160309224521_add_account_id_to_journal_rows.rb
```diff
@@ -0,0 +1,6 @@
+class AddAccountIdToJournalRows < ActiveRecord::Migration
+  def change
+    add_column :journal_rows, :account_id, :integer
+    add_index :journal_rows, :account_id
+  end
+end
```
benlinton: Should we set `account_id` for pre-existing `journal_rows`?

vendor/engines/split_accounts/spec/services/journal_row_builder_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    it "comes out with 8 rows" do
+      expect(builder.build.journal_rows.length).to eq(8)
+    end
+
+    it "splits the first order detail into three rows" do
+      rows = builder.build.journal_rows.select { |row| row.order_detail_id == order_details.first.id }
+
+      expect(rows.map(&:amount)).to eq([6.66, 6.68, 6.66])
```
benlinton: Maybe use `contain_exactly(6.66, 6.68, 6.66)` since I don't think we want to test order?

jhanggi: Actually, I do want to test order. It's the middle one with `extra_penny: true`.

jhanggi:
@benlinton any other thoughts? I'd like to get this merged in.

benlinton:
:+1:

PR #473, opened 2016-03-09
Do not crash if no earliest journal date

jhanggi:
If there are no order details and there are no JournalCutoffDates
configured, the Journal#new view would crash because
`@earliest_journal_date` is `nil`.

### DIFF
--- a/app/views/facility_journals/new.html.haml
+++ b/app/views/facility_journals/new.html.haml
@@ -1,19 +1,20 @@
-= content_for :head_content do
-  :javascript
-    $(function(){
-      var today = "#{l(Time.zone.now.to_date, format: :usa)}";
+- if @earliest_journal_date
+  = content_for :head_content do
+    :javascript
+      $(function(){
+        var today = "#{l(Time.zone.now.to_date, format: :usa)}";
 
-      $("#journal_date").val(today).datepicker({
-        "minDate": "#{l(@earliest_journal_date, format: :usa)}",
-        "maxDate": today
+        $("#journal_date").val(today).datepicker({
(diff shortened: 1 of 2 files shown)


drench:
:+1:

app/views/facility_journals/new.html.haml
```diff
@@ -1,19 +1,20 @@
...
+- if @earliest_journal_date
+  = content_for :head_content do
+    :javascript
+      $(function(){
+        var today = "#{l(Time.zone.now.to_date, format: :usa)}";
 
-      $("#journal_date").val(today).datepicker({
-        "minDate": "#{l(@earliest_journal_date, format: :usa)}",
-        "maxDate": today
+        $("#journal_date").val(today).datepicker({
```
benlinton: `#js--journalDate`?

app/views/facility_journals/new.html.haml
```diff
@@ -1,19 +1,20 @@
...
+          "maxDate": today
+        });
       });
-    });
 
-    $(function(){
-      $('#journals_create_form').submit(function(e) {
-        $(e.target).find(':submit').attr('disabled', 'true');
+      $(function(){
+        $('#journals_create_form').submit(function(e) {
```
benlinton: `#js--journalsCreateForm`?

benlinton:
Looks good.  May not be worth changing HTML IDs at this point.

jhanggi:
Yeah, I don't think it's worth changing the IDs right now. NUcore is an old project, way before the ideas of prefixing `js--`, so that would be a larger undertaking.

PR #475, opened 2016-03-10
Improve performance of Account General Report

jhanggi:
This has negligible performance implications in open, but in NU where
`to_s` does a query to pmu_department, this saves quite a bit of time
when many OrderDetails use the same account. Testing against 1000 order
details on the same account, it cuts the load time from ~3.2 seconds
down to ~2.5 due to the savings of N+1 queries.

This really only works because the report view itself will call `to_s`
implicitly when outputting the account as the first column, but it does
improve performance, so I think it’s worth it.

I'll let the rubocop changes later deal with the spacing and 1.9 syntax on the spec.

### DIFF
--- a/app/controllers/general_reports_controller.rb
+++ b/app/controllers/general_reports_controller.rb
@@ -7,7 +7,7 @@ class GeneralReportsController < ReportsController
 
 
   def account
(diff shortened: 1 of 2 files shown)


drench:
:+1:

benlinton:
:+1:

PR #477, opened 2016-03-11
[#124787] Feature flag user-based price groups

drench:
This adds a `user_based_price_groups` feature flag that is on by default, but would be off for UIC.

It became an issue because UIC, which does not have user-based price groups, still had a "Users" tab when admining price groups. There already was [an ability check to hide the tab](https://github.com/tablexi/nucore-open/blob/2d1ce5bb18ab3570c84198e66b8a145d62266202/app/views/admin/shared/_tabnav_price_group.html.haml#L3-L4), but `Ability` was not disallowing it.

Note this would be the first time we have feature flag checks directly inside `Ability`.

### DIFF
--- a/app/controllers/price_groups_controller.rb
+++ b/app/controllers/price_groups_controller.rb
@@ -28,7 +28,7 @@ class PriceGroupsController < ApplicationController
 
   # GET /facilities/:facility_id/price_groups/:id/users
   def users
-    unless @price_group_ability.can?(:read, UserPriceGroupMember)
+    unless SettingsHelper.feature_on?(:user_based_price_groups) && @price_group_ability.can?(:read, UserPriceGroupMember)
       raise ActiveRecord::RecordNotFound
     end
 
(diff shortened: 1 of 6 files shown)


jhanggi:
Hmm. I'm not completely opposed to using a feature flag in Ability, but I'd almost rather keep it at the view/controller level. How difficult would this be to fix at that layer?

Also, just a thought, but would it be worth it to add the feature flag check into `User#price_groups` so that if it's off the method returns `[]`. Then we could possibly get rid of the `UserExtension` in UIC.

drench:
Refactored to remove the feature checks from `Ability`. The behavior ought to be the same, with a smaller diff from `master`.

app/controllers/price_groups_controller.rb
```diff
@@ -28,7 +28,7 @@ def show
 
   # GET /facilities/:facility_id/price_groups/:id/users
   def users
-    unless @price_group_ability.can?(:read, UserPriceGroupMember)
+    unless SettingsHelper.feature_on?(:user_based_price_groups) && @price_group_ability.can?(:read, UserPriceGroupMember)
```
jhanggi: Do we have any specs that hit this controller action that will now fail if the feature is off?

jhanggi:
I think I like this better. If the answer to my question is "no", :+1:

app/controllers/price_groups_controller.rb
drench: No; really the [specs for this controller](https://github.com/tablexi/nucore-open/blob/cc49546d164be98308dade4c441ed38b85609f64/spec/controllers/price_groups_controller_spec.rb) could use some attention.

drench: I take that back; there are a couple failures with the flag off. PR coming soon.

PR #478, opened 2016-03-11
Test the user_based_price_groups flag in the PriceGroupsController specs

drench:
There's a lot of spec cleanup here but the change to handle the feature flag is in 41479f3. The specs should pass no matter how `user_based_price_groups_on` is set in `config/settings.yml`.


jhanggi:
:+1:

PR #479, opened 2016-03-11
[#124958] Update payment source link logic

drench:
When listing payment sources, if the user can manage the account, this should generate a link to the `facility_account_path`, as opposed to the account's orders index as before.

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -6,6 +6,14 @@ module AccountsHelper
     end
   end
 
+  def payment_source_link_or_text(account)
+    if current_ability.can?(:manage, account)
+      link_to account, facility_account_path(current_facility, account)
+    else
+      account.to_s
+    end
+  end
+
   private
 
   def available_accounts_array
--- a/app/views/shared/_accounts_table.html.haml
+++ b/app/views/shared/_accounts_table.html.haml
@@ -6,11 +6,6 @@
   %tbody
     - @accounts.each do |account|
       %tr
-        %td
-          - if current_ability.can?(:manage, account)
-            = link_to account, facility_account_path(current_facility, account)
-          - else
-            = account
-
+        %td= payment_source_link_or_text(account)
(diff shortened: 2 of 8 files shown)


app/views/shared/_accounts_table.html.haml
```diff
@@ -7,16 +7,10 @@
...
-          - case
-          - when !current_ability.can?(:manage, account)
-            = account
-
-          - when current_facility.cross_facility?
+          - if current_ability.can?(:manage, account)
             = link_to account, facility_account_path(current_facility, account)
-
           - else
-            = link_to account,
```
jhanggi: Is this page still accessible somehow? Should it be?

drench: You're right, with this change I don't see any way for a user to navigate  to `facility_account_orders_path`.

What do you think about adding an "orders" tab to `app/views/admin/shared/_tabnav_payment_method.html.haml`? This is the horizontal nav when you're at `/facilities/:facility_id/accounts/:account_id`.

jhanggi: Was it accessible before the account administrator role was added? I feel like it went to a statement view. I think adding the tab sounds like a reasonable plan.

drench:
Updated to move the account order listing into a tab:
screen shot 2016-03-16 at 9 58 39 am

app/helpers/accounts_helper.rb
```diff
@@ -6,6 +6,14 @@ def account_input(form)
     end
   end
 
+  def payment_source_link_or_text(account)
+    if current_ability.can?(:manage, account)
```
jhanggi: I'm wondering if we should be checking on something a little less restrictive like `:show`. While testing this locally, I have some split accounts, which cannot be managed by an account manager. I'm thinking they should still be able to see it, though; just not manage it. Let me do a little more testing.

jhanggi:
:+1: We'll deal with split account permissions separately.

PR #480, opened 2016-03-11
[#125280] Clean up journal creation instructions

drench:
The existing text has some awkward phrasing and typos.


jhanggi:
:+1: Can you make an update in UIC as well, since it overrides this text. You can either add it to or replace https://github.com/tablexi/nucore-uic/pull/49

PR #482, opened 2016-03-11
[#120455] Improve performance of OrderDetail split transformer

jhanggi:
Testing against NU/Oracle with 1000 OrderDetails all being on a
SplitAccounts, using each_with_object instead of the map/flatten had a
30% speed increase (~7.2 seconds down to ~4.7).

I also tried order_details.flat_map and while that's an improvement
over the map/flatten, it still tended to be slower than the
each_with_object, running around ~5.2 seconds.


drench:
:+1:

benlinton:
:+1:

PR #484, opened 2016-03-14
Add codeclimate_circle_ci_coverage gem

rdunlop:


spec/spec_helper.rb
```diff
@@ -1,3 +1,11 @@
+# run coverage when on CI
+if ENV['CI']
+  require 'simplecov'
+  SimpleCov.start 'rails' do
+    add_filter '/spec/'
```
jhanggi: Should this also run over `vendor/engines/**/spec/`? Or does this filter already hit it?

rdunlop: I don't know.

benlinton:
Nice addition!

`CI=true bundle exec rspec`
`open ./coverage/index.html`

We're at ~34% coverage.

jhanggi:
I got 84% when I ran it.

Looks like it is filtering the engines spec directories, so I think we're good there. I think we should probably `.gitignore` the coverage directory, but other than that :+1:

PR #485, opened 2016-03-15
[#114611969] Default to having two splits on #new

jhanggi:
I considered `params[:action] == “new”` as another option, but think I
like this better. If you remove all the sub accounts and try to create,
this rebuilds them, and since the account is blank on them, it’s
invalid. The messaging does get a little funky in that it shows “Splits
subaccount may not be blank” and “Splits must have unique subaccounts”,
but maybe that can be addressed in #114804299. I’m not sure where you’re at with that, @benlinton.

Also addresses #114611991.


benlinton:
:+1:

PR #486, opened 2016-03-15
[#120455] Suspend a parent account if any subaccount is suspended

benlinton:
When suspending a subaccount, any parent split accounts will inherit the farthest out `suspended_at` timestamp from its subaccounts.  If all subaccounts become unsuspended for a suspended split account, the split account does **NOT** un-suspend.

When expiring a subaccount, any parent split accounts will inherit the most recent `expires_at` timestamp from its subaccounts.  If all subaccounts become unexpired for an expired split account, the split account will become unexpired as well.

https://www.pivotaltracker.com/story/show/114611945
https://www.pivotaltracker.com/story/show/114611905

This approach does have some potential pitfalls.  If the database fails or network fails during the `account_extension.rb` `after_save` callback (which would update the parent split account), then the database will could end up with inconsistent data.  

To prevent inconsistent data, we could override the `expired_at?`, `suspsended_at?`, `active`, etc. methods on the SplitAccount model instead.  This approach would be safer in terms of data integrity, but it would also be more difficult to maintain as the `Account` model changes in the future.

### DIFF
--- a/vendor/engines/split_accounts/app/models/split_accounts/account_extension.rb
+++ b/vendor/engines/split_accounts/app/models/split_accounts/account_extension.rb
@@ -8,6 +8,27 @@ module SplitAccounts
       has_many :parent_split_accounts, through: :parent_splits
 
       scope :excluding_split_accounts, -> { where("accounts.type != ?", "SplitAccounts::SplitAccount") }
+
+      after_save :update_suspended_at_for_parent_split_accounts
+      after_save :update_expires_at_for_parent_split_accounts
+    end
+
+    def update_suspended_at_for_parent_split_accounts
+      if parent_split_accounts.any? && suspended_at_changed?
+        parent_split_accounts.each do |split_account|
+          split_account.set_suspended_at_from_subaccounts
+          split_account.save
+        end
+      end
+    end
+
+    def update_expires_at_for_parent_split_accounts
+      if parent_split_accounts.any? && expires_at_changed?
+        parent_split_accounts.each do |split_account|
+          split_account.set_expires_at_from_subaccounts
+          split_account.save
+        end
+      end
     end
 
     # Allows all account types to appropriately respond to a method that is
--- a/vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
+++ b/vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
@@ -11,6 +11,9 @@ module SplitAccounts
 
     accepts_nested_attributes_for :splits, allow_destroy: true
 
+    before_save :set_expires_at_from_subaccounts
+    before_save :set_suspended_at_from_subaccounts
+
     def valid_percent_total
       if percent_total != 100
         errors.add(:splits, :percent_total)
@@ -49,15 +52,32 @@ module SplitAccounts
       end
     end
 
-    # Stopped using SQL because that didn't seem to work until the built splits
-    # and subaccounts were created. `subaccounts` has_many :through seems to not
-    # work here, either
-    #
-    # SQL version:
-    #   subaccounts.where("expires_at IS NOT NULL").order(expires_at: :asc).first
-    #
+    # Stopped using SQL because that didn't work until the built splits
+    # and subaccounts were persisted. And `subaccounts` has_many :through seems
+    # to not work here either.
     def earliest_expiring_subaccount
-      splits.map(&:subaccount).compact.min_by(&:expires_at)
+      subaccounts = splits.map(&:subaccount).compact.select(&:expires_at?)
+      subaccounts.min_by(&:expires_at)
+    end
+
+    def earliest_suspended_subaccount
+      subaccounts = splits.map(&:subaccount).compact.select(&:suspended_at?)
+      subaccounts.min_by(&:suspended_at)
+    end
+
+    # Updates the parent account expires_at with earliest subaccount expires_at.
+    # Gracefully handles whenever all subaccounts have expires_at set to nil
+    # (even though techincally expires_at should always be set).
+    def set_expires_at_from_subaccounts
+      self.expires_at = earliest_expiring_subaccount.try(:expires_at)
+    end
+
+    # Suspend the parent account if any subaccounts are suspended.
+    # Do not automatically unsuspend the parent account if all subaccounts
+    # become unsuspended.
+    def set_suspended_at_from_subaccounts
+      subaccount = earliest_suspended_subaccount
+      self.suspended_at = subaccount.suspended_at if subaccount
     end
 
     def recreate_journal_rows_on_order_detail_update?
--- a/vendor/engines/split_accounts/spec/models/split_account_spec.rb
+++ b/vendor/engines/split_accounts/spec/models/split_account_spec.rb
@@ -146,4 +146,141 @@ RSpec.describe SplitAccounts::SplitAccount, :enable_split_accounts, type: :model
     end
   end
 
+  describe "#suspended?" do
+    context "when subaccounts and parent account are not suspended" do
+      let(:split_account) do
+        build(:split_account, without_splits: true).tap do |split_account|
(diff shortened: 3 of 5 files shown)


vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
```diff
@@ -36,10 +39,24 @@ def extra_penny_count
...
       subaccounts.sort_by(&:expires_at).first
     end
 
+    def outmost_suspended_subaccount
+      subaccounts = splits.map(&:subaccount).select(&:suspended_at)
+      subaccounts.sort_by(&:suspended_at).last
+    end
+
+    def set_expires_at_from_subaccounts
+      self.expires_at = earliest_expiring_subaccount.expires_at
```
rdunlop: if all subaccounts have un-expired, won't `earliest_expiring_subaccount` be nil?

vendor/engines/split_accounts/spec/models/split_account_spec.rb
```diff
@@ -111,4 +111,107 @@
...
+
+      context "when subaccount unexpires" do
+        before do
+          split_account
+          expired_subaccount.expires_at = Time.current + 1.year
+          expired_subaccount.save(validate: false)
+        end
+
+        it "unexpires the parent split_account" do
+          expect(split_account.reload.expired?).to eq(false)
```
rdunlop: could this be `expect(split_account.reload).not_to be_expired` to be more meaningful?

vendor/engines/split_accounts/spec/models/split_account_spec.rb
```diff
@@ -111,4 +111,107 @@
...
+      let(:split_account) do
+        build(:split_account, without_splits: true).tap do |split_account|
+          split_account.splits << build(:split, percent: 50, extra_penny: true, parent_split_account: split_account, subaccount: expired_subaccount)
+          split_account.splits << build(:split, percent: 50, extra_penny: false, parent_split_account: split_account, subaccount: create(:setup_account))
+          split_account.save
+        end
+      end
+
+      it "expires the parent split_account" do
+        expect(split_account.expired?).to eq(true)
```
rdunlop: can this `.be_expired`?

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
benlinton: Technically expires_at should always be set.  But I like the idea of adding a guard clause to make this more robust for potential future enhancements or mocked tests.

vendor/engines/split_accounts/spec/models/split_account_spec.rb
benlinton: Yeah, good idea.

vendor/engines/split_accounts/spec/models/split_account_spec.rb
benlinton: :+1:

vendor/engines/split_accounts/app/models/split_accounts/account_extension.rb
```diff
@@ -8,6 +8,27 @@ module AccountExtension
...
+          split_account.set_suspended_at_from_subaccounts
+          split_account.save
+        end
+      end
+    end
+
+    def update_expires_at_for_parent_split_accounts
+      if parent_split_accounts.any? && expires_at_changed?
+        parent_split_accounts.each do |split_account|
+          split_account.set_expires_at_from_subaccounts
```
jhanggi: Does this end up running `set_expires_at_from_subaccounts` twice since it's also a `before_save` on the SplitAccount?

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
```diff
@@ -36,10 +39,30 @@ def extra_penny_count
     #   subaccounts.where("expires_at IS NOT NULL").order(expires_at: :asc).first
     #
     def earliest_expiring_subaccount
-      subaccounts = splits.map{ |split| split.subaccount }.select(&:expires_at)
+      subaccounts = splits.map(&:subaccount).select(&:expires_at)
```
jhanggi: Something always bothers me about the `.select(&:attribute)` construct. Even though I know better, I always initially read it as `map` or `pluck` where the result is the values, when actually it's selecting all the items that have that attribute present.

I'd almost rather see something more like:
`splits.map { |s| s.subaccount.expires_at }.compact.sort.first`

Or even adding `expires_at` and `suspended_at` as delegates on `Split`.

In the expires_at case, the `compact` should be unnecessary as all `Accounts` should have an expiration due to the `validates_presence_of`.

jhanggi: Or actually, master has `splits.map(&:subaccount).compact.min_by(&:expires_at)` since subaccount could be `nil` on part of `create`.

That wouldn't necessarily work for `suspended_at` since `suspended_at` could definitely be `nil`.

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
```diff
@@ -36,10 +39,30 @@ def extra_penny_count
...
       subaccounts.sort_by(&:expires_at).first
     end
 
+    def outmost_suspended_subaccount
+      subaccounts = splits.map(&:subaccount).select(&:suspended_at)
+      subaccounts.sort_by(&:suspended_at).last
```
jhanggi: `suspended_at` will always be in the past. What's the reason you chose to use the maximum `suspended_at`?

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
```diff
@@ -36,10 +39,30 @@ def extra_penny_count
...
+    end
+
+    # Updates the parent account expires_at with earliest subaccount expires_at.
+    # Gracefully handles whenever all subaccounts have expires_at set to nil
+    # (even though techincally expires_at should always be set).
+    def set_expires_at_from_subaccounts
+      self.expires_at = earliest_expiring_subaccount.try(:expires_at)
+    end
+
+    # Suspend the parent account if any subaccounts are suspsended.
```
jhanggi: Typo on "suspsended". In the next line as well.

vendor/engines/split_accounts/app/models/split_accounts/account_extension.rb
benlinton: `set_expires_at_from_subaccounts` could potentially be ran twice

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
benlinton: If multiple child accounts are suspended, I was thinking we should display when the parent account was initially suspended at.

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
benlinton: Ok.

vendor/engines/split_accounts/app/models/split_accounts/split_account.rb
jhanggi: That makes sense. However, wouldn't that be the earliest `suspended_at`? Another option might be to `||=` suspended_at so it only gets updated if it wasn't already suspended.

vendor/engines/split_accounts/app/models/split_accounts/account_extension.rb
jhanggi: It's probably not an issue, but is there a way we can avoid running it twice?

benlinton: I'll look into this and submit another PR if I figure something out.

jhanggi: I'm cool with that happening later.

PR #487, opened 2016-03-15
[#125280] Remove reference to NU Financials from journal instructions

drench:
Forks may customize this by editing `config/locales/override/en.yml`.


jhanggi:
:+1:

PR #488, opened 2016-03-15
Move tabnav_payment_method I18n under admin.shared

drench:
This should correct the missing translations in the tabs for `/facilities/:facility_id/accounts/:account_id`.


drench:
Obsolete as I've included this in #479.

PR #492, opened 2016-03-17
Update libv8, ruby-ole, spring gems

drench:
When updating gem versions in https://github.com/tablexi/nucore-nu/pull/128, there were 3 gems where NU was ahead. This bumps the versions of those gems.


jhanggi:
:+1:

PR #493, opened 2016-03-17
Add custom Code Climate configuration

drench:
Code Climate is not currently looking at `/vendor/engines`, but should for this project.
This also should make Code Climate's coffelinter use `.coffeelint.json` (added in #445).

https://www.pivotaltracker.com/story/show/115736563

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,30 @@
+engines:
+  brakeman:
+    enabled: true
+  bundler-audit:
+    enabled: true
+  csslint:
+    enabled: true
+  coffeelint:
+    enabled: true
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -0,0 +1,30 @@
...
+  coffeelint:
+    enabled: true
+    config: .coffeelint.json
+  duplication:
+    enabled: true
+    config:
+      languages:
+      - ruby
+      - javascript
+  rubocop:
```
jhanggi: I'm scared of rubocop until we re-do Matt's #451. But it's probably fine for now.

jhanggi:
:+1:

PR #494, opened 2016-03-18
[#120455] Refactor Export Raw and show percentage of split charges

jhanggi:
This is extracted from https://github.com/tablexi/nucore-nu/pull/130

This will definitely be squashed when we merge, but I think the history and the thought process as revealed through the commit history is worth saving.


lib/hash_helper.rb
```diff
@@ -0,0 +1,13 @@
...
+  # additions = { e: 5, f: 6 }
+  # insert_into_hash_after(original_hash, :c, additions)
+  # => { a: 1, b: 2, c: 3, e: 5, f: 6, d: 4 }
+  def insert_into_hash_after(original_hash, after_column, additions)
+    original_hash.each_with_object({}) do |(k, v), new_hash|
+      new_hash[k] = v
+      new_hash.merge!(additions) if k == after_column
+    end
+  end
+end
```
benlinton: Good thinking, extracting this out.  :+1:

vendor/engines/split_accounts/config/locales/en.yml
```diff
@@ -20,3 +20,8 @@ en:
       split_accounts/split_account:
         one: Split Account
         other: Split Accounts
+
+  controllers:
+    general_reports:
+      headers:
+        split_percentage: "Split Percentage"
```
benlinton: Very minor.  If we want to remain consistent with the rest of the `split_accounts` gem lingo, we may want to use `split_percent` instead of `split_percentage`?

benlinton:
Looks great!

vendor/engines/split_accounts/config/locales/en.yml
jhanggi: That's a good call. I'll change it before I merge.

jhanggi:
@benlinton I found an issue where non-split rows were breaking because they don't get the `split` method from the decorator. Take a look at the last commit. Penultimate commit is the `split_percent` change.

vendor/engines/split_accounts/app/models/split_accounts/reports/export_raw_transformer.rb
```diff
@@ -12,7 +12,9 @@ def transform(original_hash)
       private
 
       def split_percent(order_detail)
-        number_with_precision(order_detail.split.percent, strip_insignificant_zeros: true) + "%" if order_detail.split
+        if order_detail.respond_to?(:split) && order_detail.split
+          number_with_precision(order_detail.split.percent, strip_insignificant_zeros: true) + "%"
+        end
       end
     end
   end
```
benlinton: This looks good.  Or we could extend order_detail to have a `def split; []; end`  Hmm, I'm not sure I'm crazy about that idea.

jhanggi: I don't think I'm a fan either. The other option would be to decorate everything that goes through the transformer, even if it's just a single item. I think this is simpler.

jhanggi:
Closing in favor of #495

PR #497, opened 2016-03-22
[#120455] Improve new split account view render

jhanggi:
The available_subaccounts list gets very big, and the
`available_subaccounts` query gets run for every nested sub account.
Since we’re starting with 2, it doubles the load. This especially gets
bad if you try to submit 4 or more sub accounts with a validation
error. This cuts it down so we only do a single query.


benlinton:
:+1:

PR #502, opened 2016-03-23
[#120455] Move split percent column to end

jhanggi:
Per Aaron: “we have distributed Excel templates that auto-sort "Export
Raw" reports, and we are not in a position to edit and redistribute
these templates at the moment.”

Pivotal #116148175

### DIFF
--- a/vendor/engines/split_accounts/app/models/split_accounts/reports/export_raw_transformer.rb
+++ b/vendor/engines/split_accounts/app/models/split_accounts/reports/export_raw_transformer.rb
@@ -1,10 +1,12 @@
+require "hash_helper"
 module SplitAccounts
   module Reports
     class ExportRawTransformer
+      include HashHelper
       include ActionView::Helpers::NumberHelper
 
       def transform(original_hash)
-        original_hash.merge(split_percent: method(:split_percent))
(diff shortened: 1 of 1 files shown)


vendor/engines/split_accounts/app/models/split_accounts/reports/export_raw_transformer.rb
```diff
@@ -1,12 +1,10 @@
...
 module SplitAccounts
   module Reports
     class ExportRawTransformer
-      include HashHelper
       include ActionView::Helpers::NumberHelper
 
       def transform(original_hash)
-        insert_into_hash_after(original_hash, :actual_total, split_percent: method(:split_percent))
+        original_hash.merge(split_percent: method(:split_percent))
       end
```
benlinton: I wonder if we should change the `:actual_total` in `insert_into_hash_after(original_hash, :actual_total,...)` to the last column rather than remove `HashHelper` and merge to wherever the end happens to be?

jhanggi: Yeah, that's probably a good idea to keep consistency if we add new columns later.

jhanggi:
Updated.

benlinton:
:+1:

PR #503, opened 2016-03-23
Add CSS and CoffeeScript linter configs

jhanggi:
CodeClimate was generating these files, so I was seeing a warning
inside codeclimate. I downloaded this patch from that warning.


drench:
It seems like CodeClimate is ignoring our coffeelint config filename override. 

We already had `.coffeelint.json` at the root of the project, taken straight from https://github.com/tablexi/starter/blob/master/.coffeelint.json, and had set CodeClimate to use `.coffeelint.json` in its config here: https://github.com/tablexi/nucore-open/blob/master/.codeclimate.yml#L8-L10

We could try `mv .coffeelint.json coffeelint.json` and delete `config: .coffeelint.json` line from `.codeclimate.yml` and see if that makes a difference.

drench:
I went ahead with the coffeelint file move and would like to see if CodeClimate likes it. I'm :+1: on the changes.

PR #504, opened 2016-03-23
Exclude spec directories from codeclimate duplication checks

jhanggi:
duplication in specs is not really a sign of a problem.


drench:
:+1:

PR #505, opened 2016-03-23
[#120455] Do not double count reservations in utilization reports

jhanggi:
This also fixes a bug where the actual hours report was using the
reservation hours. It got changed, by what appears to be an accident in
978ac8a


drench:
:+1:

benlinton:
:+1:

PR #506, opened 2016-03-23
[#120455] Privileged users should be able to edit, but not create split accounts

jhanggi:
And refactor how we extend Abilitiy.


vendor/engines/split_accounts/spec/controllers/facility_accounts_controller_spec.rb
```diff
@@ -66,8 +39,61 @@
...
+  describe "as an admin" do
+    let(:user) { FactoryGirl.create(:user, :administrator) }
+
+    include_examples "allows editing"
+    include_examples "allows suspending"
+
+    describe "default new" do
+      before { get :new, facility_id: facility.url_name, owner_user_id: user.id }
+
+      it "sees split account option" do
```
drench: `:aggregate_failures`?

vendor/engines/split_accounts/spec/controllers/facility_accounts_controller_spec.rb
```diff
@@ -66,8 +39,61 @@
...
+        expect(response.code).to eq("200")
+        expect(response.body).to include("Chart String")
+        expect(response.body).to include("Split Account")
```
drench: This feels like a candidate for `.and` chaining.

lib/ability_extension_manager.rb
```diff
@@ -0,0 +1,19 @@
+class AbilityExtensionManager
```
benlinton: Can we add a short comment at the top?  About how the extensions build on each other and order matters?

vendor/engines/split_accounts/spec/controllers/facility_accounts_controller_spec.rb
jhanggi: I played with it, and it got super awkward for the examples with `not_to`, and I'd rather be a little verbose but remain consistent within the file. Using `:aggregate_failures` should accomplish the same upside, though, so I went ahead and added that.

jhanggi:
Updated.

lib/ability_extension_manager.rb
```diff
@@ -1,3 +1,17 @@
+# This class is used by the Ability class in order for you to customize the
+# ability behavior inside of an engines. Extensions should take the ability as
```
benlinton: Minor typo `an engines`

benlinton:
:+1:

drench:
:+1:

PR #509, opened 2016-03-24
Update Capfile

jhanggi:
Found this while pulling the rubocop changes into NU. This is the
capfle for capistrano 3, which we are now using in our projects.

In NU and UIC, there’s a
require "rollbar/capistrano3"
require "eye/patch/capistrano3"

that should stay there. But open doesn’t have those dependencies


drench:
:+1:

PR #510, opened 2016-03-24
[#120455] Maintain first suspended_at in split accounts

jhanggi:
If a sub account gets suspended and unsuspended and re-suspended, the
parent should use maintain the first suspension date.

If you set the suspended_at of a sub account to some time in the past
before the parent’s set suspended_at, it won’t update. The logic to
make this update was much uglier and harder to understand without
significant commenting, so I decided not to worry about it because the
act of `suspend!` always happens “now” in the context of the app.


benlinton:
:-1:

benlinton:
Oops, didn't mean to add the thumbs down. :+1:

PR #512, opened 2016-03-24
Apply remainder to reservation fields

jhanggi:


vendor/engines/split_accounts/app/services/split_accounts/remainder_applier.rb
```diff
@@ -0,0 +1,47 @@
...
+    def object_to_receive_remainder
+      split_objects[remainder_index]
+    end
+
+    def remainder_index
+      @remainder_index ||= splits.find_index(&:extra_penny?)
+    end
+
+    def floored_total(attribute)
+      split_objects.reduce(BigDecimal(0)) { |sum, item| sum + item.public_send(attribute) }
```
drench: I'm not completely on-board with `Style/SingleLineBlockParams` imposing variable names for one-line `reduce`. But the line is long enough to become a 3-line `do`/`end` block, which should clear up the Code Climate :x:.

drench:
LTGM

vendor/engines/split_accounts/app/services/split_accounts/remainder_applier.rb
jhanggi: I was just planning to ignore that X. But if it's only for one-liners, then splitting to three lines makes sense. I'll make that update.

vendor/engines/split_accounts/app/services/split_accounts/attribute_splitter.rb
```diff
@@ -14,6 +14,10 @@ def split(original_object, split_object, split)
       end
     end
 
+    def to_ary
```
benlinton: `ary`?  Maybe something with whole words?

vendor/engines/split_accounts/app/services/split_accounts/remainder_applier.rb
```diff
@@ -0,0 +1,49 @@
+module SplitAccounts
+
+  class RemainderApplier
+
+    attr_reader :original, :split_objects, :splits
+
+    # split_objects and splits must have the same order
+    def initialize(original, split_objects, splits)
+      raise "There should be the same number of split objects as splits" if splits.size != split_objects.size
```
benlinton: `raise ArguementError, "..."`?

vendor/engines/split_accounts/app/services/split_accounts/attribute_splitter.rb
jhanggi: Fair question. `to_ary` is actually a standard ruby method. `Array(xxx)` will call `to_ary` under the hood if it supports it. `nil` implements it as `[]` and `Array` implements it as `self`.

vendor/engines/split_accounts/app/services/split_accounts/remainder_applier.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def floored_total(attribute)
+      split_objects.reduce(BigDecimal(0)) do |sum, item|
+        sum + item.public_send(attribute)
+      end
+    end
+
+  end
+
+end
```
benlinton: Nice job.

vendor/engines/split_accounts/app/services/split_accounts/attribute_splitter.rb
jhanggi: It's a complicated question. I wanted to be able to use `Array`, which now researching might also try `to_a`, which I think I would prefer. I'll test it out.

vendor/engines/split_accounts/spec/services/order_detail_splitter_spec.rb
```diff
@@ -1,77 +1,134 @@
 require "rails_helper"
 
 RSpec.describe SplitAccounts::OrderDetailSplitter, type: :service do
-  let(:subaccount_1) { build_stubbed(:nufs_account) }
-  let(:subaccount_2) { build_stubbed(:nufs_account) }
+  describe "item/service orders" do
```
benlinton: Can we be more explicit here?  What is an `item`?

Something like `describe "product and service orders" do` or ...?

jhanggi: An "item" is an `Item`. Item is a subclass of product. Not quite sure how to be more specific. Although, technically I'm only testing an item order, but items and services behave identically in this case.

benlinton:
:+1: Very minor comments.

vendor/engines/split_accounts/app/services/split_accounts/remainder_applier.rb
jhanggi: Good call.

vendor/engines/split_accounts/spec/services/order_detail_splitter_spec.rb
benlinton: I think it's fair to say using a more explicit name isn't going to work.  :smile:

PR #513, opened 2016-03-24
[#120455] Hide insignificant zeros on instrument util reports

jhanggi:
If there is 1 reservation on a split account, it is showing 1.0 in
quantity columns. We don’t need to show this, although, we should show
the decimals if they are there. We will continue showing 1 decimal
precision on the hours columns.

### DIFF
--- a/app/models/reports/instrument_day_report.rb
+++ b/app/models/reports/instrument_day_report.rb
@@ -44,6 +44,7 @@ class Reports::InstrumentDayReport
   end
 
   class DayValue
+
     include ActionView::Helpers::NumberHelper
 
     def initialize(reservation)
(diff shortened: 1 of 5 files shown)


benlinton:
Is it a problem that we're changing the type in the model?  Should we be using a decorator pattern?  Or is it not worth the extra effort for this particular case?

benlinton:
Other than one question.  :+1:

jhanggi:
It's a totally fair question. These report classes are weird, they're somewhere between models and presenters. Possibly service objects? There's already other formatting going on, which is why I wasn't that concerned about it. `to_hours` rounds, `format_percent` converts to a string. You could call it [broken windows](https://en.wikipedia.org/wiki/Broken_windows_theory), but I don't think it's worth the effort it would take to clean it up right now.

benlinton:
:+1:

PR #515, opened 2016-03-25
[#115285123] Send email notification on order, if configured by facility

drench:
This adds `order_notification_recipient` to the `Facility` model.

If given a value, this address will receive order notifications. These are generated the same time as receipts, and use the same templates, with minor differences: the `Subject` line is different, and only receipts should say "Thank you for your order".

A major part of this is a refactor of `OrdersController#purchase`. I think it's an improvement though we could definitely go further if want to put more time into it. I left this method's specs mostly as-is since its coverage was near-complete.


app/models/facility.rb
```diff
@@ -43,6 +43,10 @@ def accounts
   validates_format_of    :abbreviation, with: /^[a-zA-Z\d\-\.\s]+$/, message: "may include letters, numbers, hyphens, spaces, or periods only"
   validates_format_of    :journal_mask, with: /^C\d{2}$/, message: "must be in the format C##"
 
+  validates :order_notification_recipient,
+    format: { with: /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,6}$/i },
```
drench: This is lifted from a validation in the `User` model. Do we want to extract a validator, or use something like https://github.com/balexand/email_validator ?

app/controllers/orders_controller.rb
```diff
@@ -238,71 +238,38 @@ def update
...
+        redirect_to receipt_order_path(@order)
       end
-    rescue => e
-      flash[:error] = I18n.t("orders.purchase.error")
-      flash[:error] += " #{e.message}" if e.message
-      puts e.message
-      @order.reload.invalidate!
-      redirect_to(order_path(@order)) && return
     end
+  rescue NUCore::QuantityUpdateChangeException => e
```
jhanggi: Why do we need both this error handling _and_ `order_purchaser.quantities_changed?`?

app/models/facility.rb
jhanggi: I'm not sure I want to pull in a dependency for this, but I do think standardizing our email validation is a good idea. For the training request notification feature, I did minimal validation https://github.com/tablexi/nucore-open/blob/master/app/models/product.rb#L31

For public sites, I definitely lean towards lenient email validation to allow that random person with `☃@snowman.com`, but in NUcore's case, I'd be fine with continuing with this regex.

app/services/order_detail_update_param_hash_extractor.rb
```diff
@@ -0,0 +1,17 @@
+class OrderDetailUpdateParamHashExtractor
+  attr_reader :params
+
+  def initialize(params)
+    @params = params
+  end
+
+  def updates_as_hash
```
jhanggi: This method is really difficult to reason about, but I'm not sure how to make it clearer. I think a comment explaining what's it's doing would help. Also, the `updates` part of the method name implies it's going to be actually doing something.  I think maybe something like `extract` or even just `to_h` might be clearer.

app/services/order_detail_updater.rb
```diff
@@ -0,0 +1,34 @@
...
+    @quantities_changed
+  end
+
+  def update
+    update_result = order.update_details(params)
+    @quantities_changed = (@initial_quantities != quantities)
+    update_result
+  end
+
+  def update!
```
jhanggi: Do we really need this? I'd love to get rid of the exception handling if we can help it.

app/services/order_detail_updater.rb
```diff
@@ -0,0 +1,34 @@
...
+    @params = params
+    @initial_quantities = quantities
+  end
+
+  def quantities_changed?
+    @quantities_changed
+  end
+
+  def update
+    update_result = order.update_details(params)
```
jhanggi: Would it make sense to pull this method out of the model into this class?

app/services/order_purchaser.rb
```diff
@@ -0,0 +1,81 @@
...
+    @acting_as = acting_as
+    @order = order
+    @order_in_past = order_in_past
+    @params = params
+    @user = user
+  end
+
+  def purchase!
+    order.transaction do
+      return if order_detail_updater.update! && quantities_changed?
```
jhanggi: If the quantities have changed, won't that raise an error (which would rollback the transaction)? And the `quantities_changed?` would either not get called or return `false`?

app/services/order_purchaser.rb
```diff
@@ -0,0 +1,81 @@
+class OrderPurchaser
+  attr_reader :order, :params, :user
+
+  delegate :order_notification_recipient, to: :order
+  delegate :quantities_changed?, to: :order_detail_updater
+
+  def initialize(acting_as:, order:, order_in_past:, params:, user:)
```
jhanggi: For the sake of this parameter, would `can_order_in_past` be a little clearer?

app/controllers/orders_controller.rb
```diff
@@ -238,71 +238,38 @@ def update
...
+    logger.debug "errors #{@order.errors.full_messages}"
+    flash[:error] = @order.errors.full_messages.join("<br/>").html_safe
+    render :show
+  rescue => e
+    flash[:error] = I18n.t("orders.purchase.error")
+    flash[:error] += " #{e.message}" if e.message
+    puts e.message
```
jhanggi: Do we need this `puts`?

jhanggi:
I think this is a good start for the refactoring. I think we can spend a little more time and push it a little bit further. I'd love to see if there's more we could get out of the controller, and I'd love to be able to get rid of some of the exception handling.

Also, there are a few rubocop issues that codeclimate found.

app/services/order_purchaser.rb
drench: The name `QuantityUpdateChangeException` was misleading because it was wrong. A quantity update should update `order_detail` but not move the `order` to `purchased`, but not raise an exception. I renamed the exception (which I'd still like to phase out).

jhanggi:
Let me know once this is ready for another look.

jhanggi:
Looks like there are merge conflicts.

app/services/order_detail_update_param_hash_extractor.rb
```diff
@@ -0,0 +1,22 @@
+class OrderDetailUpdateParamHashExtractor
+
+  attr_reader :params
+
+  def initialize(params)
+    @params = params
+  end
+
+  # This extracts and transforms a hash (like controller params) into the form
+  # of a hash that Order#update_details expects.
```
jhanggi: Could you include an example?

app/controllers/orders_controller.rb
```diff
@@ -336,36 +302,61 @@ def index
...
+
+  def ordering_on_behalf_with_date_params?
+    params[:order_date].present? && params[:order_time].present? && acting_as?
+  end
+
+  def single_reservation?
+    @order.order_details.size == 1 &&
+      @order.order_details.first.product.is_a?(Instrument) &&
+      !@order.order_details.first.bundled? &&
+      !acting_as?
```
jhanggi: It feels like the `acting_as?` check doesn't belong in this method. Would it make sense to pull the rest of the method into `Order`?

app/models/facility.rb
```diff
@@ -43,6 +43,10 @@ def accounts
   validates_format_of    :abbreviation, with: /^[a-zA-Z\d\-\.\s]+$/, message: "may include letters, numbers, hyphens, spaces, or periods only"
   validates_format_of    :journal_mask, with: /^C\d{2}$/, message: "must be in the format C##"
 
+  validates :order_notification_recipient,
+            format: { with: /^[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,6}$/i },
```
jhanggi: Can we share this regex with the user model?

config/locales/en.yml
```diff
@@ -852,6 +852,9 @@ en:
       is_active: "Is Active?"
       contact: "The following contact information will appear on statements and should indicate where payment should be sent"
       show_instrument_availability: Show Instrument Availability Status?
+      order_notification: |
+        The email address that will receive order notifications
```
jhanggi: I think this text could be a little clearer. Without knowing, it's unclear what a "order notification" would be. "This email address will be notified whenever an order is placed by a user" or something similar. The next line is fine.

app/services/order_purchaser.rb
```diff
@@ -0,0 +1,76 @@
...
+  def order_update_params
+    @order_update_params ||= OrderDetailUpdateParamHashExtractor.new(params).to_h
+  end
+
+  def quantities
+    order.order_details.order("order_details.id").pluck(:quantity)
+  end
+
+  def send_notification?
+    order_notification_recipient.present?
```
jhanggi: We should probably not be sending the notification when placing an order on behalf of.

drench: I thought we would? We're suppressing receipts when `acting_as?`, but this is meant to have facility-level order notifications still go out either way.

Is the method name is too vague, or do we want it to suppress notifications?

jhanggi: When we wrote this up for them: "This email would not be sent if placed as a user proxy." If someone from the facility is ordering, there's not really a need for the email. The primary use-case from DC is low-volume resources who aren't checking NUcore every day.

.gitignore
```diff
@@ -30,3 +30,6 @@ config/settings/*.local.yml
 config/environments/*.local.yml
 
 public/images/bg-search-form.pxm
+
+# Ignore RSpec's persistence file:
```
jhanggi: FYI, I'm going to cherry-pick this commit into master.

app/services/order_purchaser.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def order_update_params
+    @order_update_params ||= OrderDetailUpdateParamHashExtractor.new(params).to_h
+  end
+
+  def quantities
+    order.order_details.order("order_details.id").pluck(:quantity)
+  end
+
+  def send_notification?
```
jhanggi: Perhaps `send_facility_notification?` or `send_notification_to_facility?` just to make it a little more specific.

app/validators/email_format_validator.rb
```diff
@@ -0,0 +1,16 @@
+class EmailFormatValidator < ActiveModel::EachValidator
+
+  EMAIL_PATTERN = /\A[a-z0-9._%+-]+@[a-z0-9.-]+\.[a-z]{2,6}\z/i
```
jhanggi: I still wonder if we need to be this strict with email validation, but I like that we have it encapsulated within a validator now.

jhanggi:
:+1:

PR #516, opened 2016-03-25
[#115962549] Remove "-nu" from the logo filename

drench:
Note that the alt text here remains "Northwestern Logo" since nucore-open will still "ship" with that logo. ~~We can override that in the other forks.~~ This text is already overridden in the other forks.


jhanggi:
:+1:

PR #517, opened 2016-03-25
Remove 2010 check when generating copyright dates

drench:
This is something I noticed when updating the footer for #516. Is there any reason to keep that 2010 check?

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -83,7 +83,7 @@ en:
 
   shared:
     footer:
-      copyright_html: "&copy; Copyright 2010%{to_date} Northwestern University"
+      copyright_html: "&copy; Copyright 2010&ndash;%{to_date} Northwestern University"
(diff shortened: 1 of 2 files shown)


config/locales/en.yml
```diff
@@ -83,7 +83,7 @@ en:
 
   shared:
     footer:
-      copyright_html: "&copy; Copyright 2010%{to_date} Northwestern University"
+      copyright_html: "&copy; Copyright 2010&ndash;%{to_date} Northwestern University"
```
jhanggi: I was wondering if we should we remove the reference to Northwestern University here? But they started nucore, and we can easily override this in the forks.

jhanggi:
I support removing that check.

jhanggi:
:+1: I think we're probably fine leaving NU as the default in open, but we should override it in the other branches (it's already overridden in UIC)

PR #519, opened 2016-03-26
[#115962443] Remove NU-specific language from locales

drench:
Forks that require site-specific language should make overrides in `config/locales/override/en.yml`.


jhanggi:
:+1: Do we want to pull the "Unreconciled Total" change up into open?

drench:
5136c57 should shorten the header to "Unreconciled Total" now. NU will need a patch to remove its override (since it would be the same now).

PR #521, opened 2016-03-29
[#120455] Show error when trying to reactivate split with suspended child

jhanggi:
I tried to use a validation, but ended up not just overriding the `unsuspend` method because the ordering gets complicated between the validation and `:set_suspended_at_from_subaccounts`.

### DIFF
--- a/spec/models/account_spec.rb
+++ b/spec/models/account_spec.rb
@@ -147,10 +147,10 @@ RSpec.describe Account do
   it "should set suspend_at on suspend and unsuspend" do
     @owner   = FactoryGirl.create(:user)
     @account = FactoryGirl.create(:nufs_account, account_users_attributes: account_users_attributes_hash(user: @owner))
-    assert_nothing_raised { @account.suspend! }
-    expect(@account.suspended_at).not_to eq(nil)
-    assert_nothing_raised { @account.unsuspend! }
-    expect(@account.suspended_at).to eq(nil)
+    expect(@account.suspend).to be_truthy
+    expect(@account.suspended_at).to be_present
+    expect(@account.unsuspend).to be_truthy
+    expect(@account.suspended_at).to be_blank
   end
 
   context "account users" do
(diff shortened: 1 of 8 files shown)


spec/models/account_spec.rb
```diff
@@ -147,10 +147,10 @@
   it "should set suspend_at on suspend and unsuspend" do
     @owner   = FactoryGirl.create(:user)
     @account = FactoryGirl.create(:nufs_account, account_users_attributes: account_users_attributes_hash(user: @owner))
-    assert_nothing_raised { @account.suspend! }
-    expect(@account.suspended_at).not_to eq(nil)
-    assert_nothing_raised { @account.unsuspend! }
-    expect(@account.suspended_at).to eq(nil)
+    expect(@account.suspend).to be_truthy
```
benlinton: ` expect(@account.suspend).to eq(true)` to be more explicit?

spec/models/account_spec.rb
```diff
@@ -147,10 +147,10 @@
...
+    expect(@account.suspended_at).to be_present
+    expect(@account.unsuspend).to be_truthy
+    expect(@account.suspended_at).to be_blank
```
benlinton: `be_nil`

spec/models/account_spec.rb
```diff
@@ -147,10 +147,10 @@
   it "should set suspend_at on suspend and unsuspend" do
     @owner   = FactoryGirl.create(:user)
     @account = FactoryGirl.create(:nufs_account, account_users_attributes: account_users_attributes_hash(user: @owner))
-    assert_nothing_raised { @account.suspend! }
-    expect(@account.suspended_at).not_to eq(nil)
-    assert_nothing_raised { @account.unsuspend! }
-    expect(@account.suspended_at).to eq(nil)
+    expect(@account.suspend).to be_truthy
+    expect(@account.suspended_at).to be_present
+    expect(@account.unsuspend).to be_truthy
```
benlinton: `eq(true)`

spec/models/account_spec.rb
jhanggi: I don't have a strong feeling either way. I feel like @drench has pushed me this way before.

benlinton:
Minor comments.  Looks good :+1:

spec/models/account_spec.rb
benlinton: `eq(true)` is more likely to catch a bug if one pops up, but overall it's not a big deal.  :+1:

PR #523, opened 2016-03-30
[#125520] Transaction export as CSV fails with 1000+ orders

jhanggi:
This also does some refactoring using a `where_ids_in` method I built using the fact that `WHERE IN (?, ?, ) OR WHERE IN (?, ?, )` works fine as long as each has less than 1000 items.

### DIFF
--- a/app/views/shared/transactions/_table_inside.html.haml
+++ b/app/views/shared/transactions/_table_inside.html.haml
@@ -19,7 +19,7 @@
       %th.currency= OrderDetail.human_attribute_name("cost")
       %th= OrderDetail.human_attribute_name("order_status")
   %tbody
-    - query_in_slices(order_details).each do |order_detail|
+    - order_details.each do |order_detail|
       %tr{ class: row_class(order_detail) }
         - if @order_detail_action
           %td= check_box_tag "order_detail_ids[]", order_detail.id, false, {:class => 'toggle'}
--- a/spec/lib/bulk_email_helper_spec.rb
+++ b/spec/lib/bulk_email_helper_spec.rb
@@ -216,21 +216,22 @@ RSpec.describe BulkEmailHelper do
     end
   end
 
+  # SLOW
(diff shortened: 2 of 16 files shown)


app/views/shared/transactions/_table_inside.html.haml
```diff
@@ -19,7 +19,7 @@
       %th.currency= OrderDetail.human_attribute_name("cost")
       %th= OrderDetail.human_attribute_name("order_status")
   %tbody
-    - query_in_slices(order_details).each do |order_detail|
+    - order_details.each do |order_detail|
```
jhanggi: This ended up no longer being necessary since the driver has fixed its issues with eager loading association to avoid more than 1000 items.

spec/lib/bulk_email_helper_spec.rb
```diff
@@ -219,23 +219,18 @@ class BulkEmailTest
...
-  #     puts 'doing query'
-  #     users = @controller.do_search(@params)
-  #     puts "type: #{users.class}"
-  #     users.size.should == 1001
-  #   end
-  # end
+  context "being ready for Oracle" do
+    before :each do
+      1001.times do
+        user = FactoryGirl.create(:user)
```
drench: I had to see if doing a `create_list(:user, 1001)` and iterating through that would be any faster, but it barely made a difference for me (from 45 seconds down to 44).

spec/lib/bulk_email_helper_spec.rb
```diff
@@ -219,23 +219,18 @@ class BulkEmailTest
...
+        od = place_product_order(user, @facility, @product, @account)
+      end
+      expect(OrderDetail.all.size).to eq(1001)
+    end
+    it "should return 1001 users" do
```
drench: If this were going to hang around, there's some more minor cleanup that should happen ("should", `@` variables, spacing).

drench:
:+1:

PR #525, opened 2016-03-31
[#125502] Hide deleted AccountUsers

jhanggi:
In the order detail modal, the payment source dropdown was showing
duplicate payment sources if the user had deleted AccountUsers.

This now makes the association always exclude `deleted_at`. The
association from the User side was already ready doing this


drench:
:+1:

PR #530, opened 2016-04-04
[#125414] Update Export Raw email subject and filename

jhanggi:
The export raw actually has nothing to do with the current action
(product, account_owner, price_group, etc). This also removes the need
for that parameter.


drench:
:+1:

PR #531, opened 2016-04-04
[#125047] Remove "Add Payment Source" button from user

jhanggi:
Facility directors were using this button thinking they were going to
add the user as a purchaser and instead adding them as the owner.

Also of note, the feature flag check was probably referencing the wrong
flag.


drench:
:+1:

PR #532, opened 2016-04-07
[#115962443] Remove "NU" from payment option text

drench:
This would be overridden in the forks that use `c2po`.


jhanggi:
:+1:

jhanggi:
Extracting "Chart String" into a more generic term is a task for another time since it's much bigger and might not be something we need to worry too much about, but something to think about.

jhanggi:
Went ahead and merged this down and pulling into DC so I can accept the story.

PR #533, opened 2016-04-08
[#124331] Move new user flow back into the main app

jhanggi:
This is related to some refactorings that I'm doing for UIC around excluding deactivated users from search results.

In https://pm.tablexi.com/issues/29581

Chris: 

> I had to move Jason's most recent changes down to the c2po engine. The new_facility_account_path helper usage in users/_new_user.html.haml was breaking the UIC builds because it doesn't exist for them.

https://github.com/tablexi/nucore-open/commit/ca733fd27aef586ed888f03bcc324725a3075eb5

However, we've since changed the flow, so the path helper is no longer used.


drench:
:+1:

PR #534, opened 2016-04-09
[#124331] Support deactivated users

jhanggi:
* Display "(DEACTIVATED) next to a user's name in search results
* Exclude deactivated users from bulk email list
* Refactor search result partials
* Prevent ordering on behalf of a suspended user

This includes #533.

There are still some things I might do at a later point, but I wanted to go ahead and get this out there.

* Change the I18ns in the search screens to default to not having "NetID" in them
* Make sure we're including (DEACTIVATED) elsewhere in the app
* Prevent being able to add a deactivated user to a payment source or price group (although this might be fine as is). Also prevent modifying their access lists.


app/models/user.rb
```diff
@@ -157,8 +157,20 @@ def full_name
     end
   end
 
+  def last_first_name
+    "#{last_name}, #{first_name}" + deactivated_string
+  end
+
   def to_s
-    full_name
+    full_name + deactivated_string
```
jhanggi: There are still a lot of places where we use `full_name`. I'll probably want to clean that up at some point.

config/locales/en.yml
```diff
@@ -767,6 +767,11 @@ en:
         search_term: "Search by name, NetID, or username"
 
   search:
+    results_table:
+      order_for: Order For
+      name: Name
+      username: Username # or NetID
```
jhanggi: This is mainly as a reminder that I'll want to override this in NU.

app/services/bulk_email_searcher.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    @order_details = order_details.ordered_or_reserved_in_range(start_date, end_date)
+  end
+
+  def find_order_details_for_roles(roles)
+    find_order_details.joins(account: :account_users).where(account_users: { user_role: roles })
+  end
+
+  def start_date
+    parse_usa_date(search_fields[:start_date].to_s.to_s.tr("-", "/")) if search_fields[:start_date]
```
drench: It looks like this has been around for a long time (ba825dd93b8958fd6aab209b7f3226b70f82fede) but do you know why this needs `.to_s` twice?

app/views/search/_account_account_user_link.html.haml
```diff
@@ -1 +1 @@
-= link_to "#{user.last_name}, #{user.first_name}", new_account_account_user_path(@account, :user_id => user.id)
\ No newline at end of file
+= link_to user.last_first_name, new_account_account_user_path(@account, :user_id => user.id)
```
drench: I'd un-hashrocket this.

spec/lib/ability_spec.rb
```diff
@@ -32,8 +32,19 @@
     end
   end
 
+  shared_examples_for "it allows switch_to on active, but not deactivated users" do
+    let(:stub_controller) { UsersController.new }
+    let(:active_user) { User.new }
```
drench: Any reason to favor this over `FactoryGirl.build`ing?

app/views/search/_results_table.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  %tbody
+    - users.each do |user|
+      %tr
+        - if order_for
+          %td.centered
+            - if current_ability.can?(:switch_to, user)
+              = link_to t(".order_for"),
+                facility_user_switch_to_path(current_facility, user)
+        %td
+          = render "search/#{params[:search_type] || "manage_user"}_link", user: user
```
drench: Code Climate pointed this out, though given the prefix and suffix I can't think of a way it would be exploitable. It would still be nice to avoid it.

drench:
A few minor comments/questions but otherwise LGTM :+1:

app/services/bulk_email_searcher.rb
jhanggi: I'm going to guess that was an accident. I'll update it.

app/views/search/_account_account_user_link.html.haml
jhanggi: Thanks for catching this. I was doing a bunch of find/replace for these `_link` files. Guess I missed the hash style here.

jhanggi:
Updated. This could use another look.

drench:
:+1:

PR #535, opened 2016-04-10
Updates from DC

drench:
These are cherry-picks from https://github.com/tablexi/nucore-dartmouth/pull/15 that would make senses to be in all forks.

One upgrades the `spreadsheet` gem (786f916) and another suppresses duplication checks (ecaf785) in engine files that tend to have similarities that we don't really mind.


jhanggi:
:+1:

PR #536, opened 2016-04-12
[#86777] Disable statement search specs if statements are disabled

jhanggi:
This should address the test failures on uic/develop.


drench:
:+1:

PR #539, opened 2016-04-13
[#47665] Add specs for reservation cancelation

jhanggi:
This will prepare us for being able to update the account on order details.


drench:
:+1: (for the changes since the commits I made)

PR #540, opened 2016-04-14
[#47665] Use strong params for OrderDetails

jhanggi:


spec/controllers/order_management/order_details_controller_spec.rb
```diff
@@ -493,6 +493,25 @@
...
     end
 
+    describe "changing quantity" do
+      before do
+        @params[:order_detail] = { quantity: 2 }
+      end
+
+      it "updates the quanity" do
+        expect { do_request }.to change { order_detail.reload.quantity }.to(2)
+
```
drench: Super minor, but I thought RuboCop would have flagged this newline. It does if I add a 2nd one.

drench:
Looks like the only CC issue is going beyond our artificial `Metrics/ClassLength` by 1. :+1:

PR #541, opened 2016-04-14
[#125686] Remove JS flash removal after 10 seconds

jhanggi:
While testing some issues around the credit card provider for NU, it
felt that hiding the flash after 10 seconds could obscure the errors.
This will leave the flash message in place until the page is refreshed.
If a new flash is added, the old one is still removed, so we don’t have
to worry about flash messages piling up.


drench:
:+1:

PR #543, opened 2016-04-15
[#47665] Purchaser/Owner/BA can change change order detail account

jhanggi:
I’ve implemented this so that a user may change the account up to the end of the review period (unless it gets disputed, then the feature continues to work until the dispute gets resolved). I've also set it up so the user can only select non-expired, non-suspended accounts. I agree that it could be problematic if users are given the ability to assign it to an expired account. We could implement logic that takes fulfillment date into account, but that logic will get messy, and I'm not sure that it's worth the effort and additional complexity. If a user need to move the order to an expired account, they can always contact the facility directly.

The original purchaser, account owners and business admins have this interface. Account Purchasers don't have access (they currently can't even see other users purchases).

I’m awaiting final confirmation of these decisions from NU, but the code is still reviewable (especially some of the controller refactoring).

### DIFF
--- a/app/assets/stylesheets/app/typography.scss
+++ b/app/assets/stylesheets/app/typography.scss
@@ -40,6 +40,10 @@ label {
   margin-bottom: 0;
 }
 
+.normal-weight {
+  font-weight: normal;
+}
+
 // Custom headers
 .header-tight {
   line-height: 1.1;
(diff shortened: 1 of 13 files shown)


drench:
LGTM, and the CC issue doesn't seem legitimate.

PR #544, opened 2016-04-16
[#122397] Prevent multiple reservations on the same order detail

jhanggi:
It is possible for two reservations to be created for the same user if
they double click the “Create” button in quick succession and the
transactions overlap. This only happens on a multi-threaded server like
unicorn.

This also cleans up duplicates and adds a unique constraint on
`reservations.order_detail_id`. All duplicates I found in practice are
exact duplicates.

Also related to NU #124969, UIC #122397.

### DIFF
--- a/app/views/reservations/new.html.haml
+++ b/app/views/reservations/new.html.haml
@@ -40,7 +40,7 @@
           = t('reservations.new.notify')
 
   %ul.inline
-    %li= f.submit 'Create', class: 'btn', id: 'reservation_submit'
+    %li= f.submit 'Create', class: 'btn', id: 'reservation_submit', data: { disable_with: 'Create' }
     - url_after_action = facility_path(@instrument.facility)
     - if @order_detail.bundled?
       %li= link_to 'Cancel', url_after_action
--- a/db/migrate/20160415231343_clean_duplicate_reservations.rb
+++ b/db/migrate/20160415231343_clean_duplicate_reservations.rb
@@ -1,0 +1,27 @@
+class CleanDuplicateReservations < ActiveRecord::Migration
+  class Reservation < ActiveRecord::Base
+  end
+
+  def up
+    order_details_with_multiple_reservations = Reservation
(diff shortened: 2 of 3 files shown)


app/views/reservations/new.html.haml
```diff
@@ -40,7 +40,7 @@
           = t('reservations.new.notify')
 
   %ul.inline
-    %li= f.submit 'Create', class: 'btn', id: 'reservation_submit'
+    %li= f.submit 'Create', class: 'btn', id: 'reservation_submit', data: { disable_with: 'Create' }
```
jhanggi: I wanted to keep the change small here, so I didn't go all rubocop on this file.

db/migrate/20160415231343_clean_duplicate_reservations.rb
```diff
@@ -0,0 +1,27 @@
...
+      .where("order_detail_id is not null")
+      .group(:order_detail_id)
+      .having("count(*) > 1")
+      .pluck(:order_detail_id)
+
+    order_details_with_multiple_reservations.each do |order_detail_id|
+      Reservation.where(order_detail_id: order_detail_id).last.destroy
+    end
+    if NUCore::Database.oracle?
+      remove_index :reservations, :order_detail_id
```
jhanggi: I realize this is awkward. MySQL ties foreign keys very tightly to an index. So I can't remove this index without also removing the foreign key.  Oracle keeps the foreign keys and indexes completely separate so I can actually replace the index.

db/schema.rb
```diff
@@ -175,9 +175,9 @@
   create_table "journal_rows", :force => true do |t|
     t.integer "journal_id",                                                   :null => false
     t.integer "order_detail_id"
+    t.string  "account",         :limit => 5
```
jhanggi: I ran db:migrate from a 100% clean database and it moved this field's position.

drench:
:+1:

PR #547, opened 2016-04-21
Clean up recharge chart string partials and I18n

drench:
This is extracted from https://github.com/tablexi/nucore-dartmouth/pull/22.

It does not address the odd taborder issue we've noticed when filling out a chart string (that's where halfway through, tabbing may bounce your cursor to the search input).


jhanggi:
👍  I've played around a little with the taborder. In chrome, if I tab after the last of the account number fields, focus moves to the "NUcore" header. If I just remove the `taborder` attribute, everything seems to work again. But I've only tested that in chrome.

PR #548, opened 2016-04-21
[#47665] Fix strong params for OrderDetail popup

jhanggi:
Our specs were using `reserve_start_at` as the parameter instead of
what really gets submitted: `reserve_start_date, reserve_start_hour,
reserve_start_min, reserve_start_meridian` so the specs were passing,
but would error when used in the interface.


drench:
:+1:

PR #549, opened 2016-04-21
[#124331] Fix broken search result

jhanggi:
The user approval search was broken because of a syntax error. This was
broken by the refactoring in #534 for supporting deactivated users.

While writing tests, I also discovered that this causes a problem where
facility admins cannot order for after a search. I never noticed it
because I was always testing as a global admin.


drench:
:+1:

PR #550, opened 2016-04-22
[#47665] Fix edit account permissions based on state

jhanggi:
#update was already using logic to determine whether you could modify
the order. This adds the logic to edit and the “Change” link.


drench:
👍

PR #551, opened 2016-04-22
[#47665] Do not duplicate current account when changing source

jhanggi:


spec/factories/accounts.rb
```diff
@@ -7,7 +7,7 @@
     # Every account must have an account_user "owner" in order for the account
     # to be valid in rails. And foreign key constraints require that each
     # account_user has an account inserted before the account_user is inserted.
-    callback(:after_build, :before_create) do |account, evaluator|
```
jhanggi: This callback was actually being run twice (once `after_build` and again `before_create`) resulting in duplicate `AccountUser`s.

drench:
:+1:

PR #552, opened 2016-04-22
[#47665] Customer can change account until journaled or statement

jhanggi:
Per NU’s request, a customer should be allowed to change the payment
source up until the time the order detail is journaled or statemented.

### DIFF
--- a/app/controllers/order_details_controller.rb
+++ b/app/controllers/order_details_controller.rb
@@ -121,7 +121,7 @@ class OrderDetailsController < ApplicationController
   end
 
   def order_editable?
-    @order_detail.customer_editable?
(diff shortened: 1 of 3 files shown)


app/controllers/order_details_controller.rb
```diff
@@ -121,7 +121,7 @@ def prevent_edit_based_on_state
   end
 
   def order_editable?
-    @order_detail.customer_editable?
```
drench: I think the `OrderDetail#customer_editable?` method can go away now. I haven't found it being used anywhere else.

drench:
Just one note on a possible dead method to clean up.

PR #553, opened 2016-04-22
[#125794] Hide dispute box from Account Purchaser 

jhanggi:
The user would receive a 403 when they tried to submit the dispute
reason. This now hides the box entirely from those users.


drench:
`app/views/order_details/show.html.haml` retains some outdated style that may be worth cleaning up but otherwise :+1:

PR #554, opened 2016-04-23
[#125700] Refactor Reconciliation Actions

jhanggi:
This refactors the reconciliation actions to set up being able to add a new statementable/reconcilable account type for UIC. It gets rid of the C2PO FacilityAccountControllerExtenstion and pulls the reconciliation functionality into a new FacilityAccountsReconciliationController. It then sets up routes based on the `Account.config.statement_account_types` that all route to the same controller action.

If I did my job right, there should be no difference on the user side (including URLs) from what was there before.


app/controllers/facility_accounts_reconciliation_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+      .account_unreconciled(current_facility, selected_account)
+      .order(%w(
+               order_details.account_id
+               order_details.statement_id
+               order_details.order_id
+               order_details.id))
+      .paginate(page: params[:page])
+  end
+
+  def update_account
```
jhanggi: This method is pulled directly out of C2PO::FacilityAccountsControllerExtension. It could use some refactoring, but that seemed like more than I wanted to take on for this refactoring.

vendor/engines/c2po/spec/controllers/facility_accounts_reconciliation_controller_spec.rb
```diff
@@ -0,0 +1,207 @@
...
+    account.account_users_attributes = [{ user_id: @purchaser.id, user_role: AccountUser::ACCOUNT_OWNER, created_by: @admin.id }]
+    assert account.save
+    statement = FactoryGirl.create(:statement, facility_id: @authable.id, created_by: @admin.id, account: account)
+    @order_detail.to_complete!
+    @order_detail.update_attributes(account: account, fulfilled_at: 1.day.ago, actual_cost: 10, actual_subsidy: 2)
+    @order_detail.statement = statement
+    @order_detail.save
+  end
+
+  def make_order_statemented
```
drench: Dead code?

drench:
:+1:

PR #557, opened 2016-04-26
Add :with_completed_order trait to the Journal factory

drench:
This is extracted from https://github.com/tablexi/nucore-dartmouth/pull/24

It's a fairly complicated trait but it should eliminate a lot of setup from specs themselves. I could see using it in `spec/models/journal_spec.rb`.

### DIFF
--- a/spec/factories/journals.rb
+++ b/spec/factories/journals.rb
@@ -3,5 +3,47 @@ FactoryGirl.define do
     is_successful true
     created_by 1
     journal_date { Time.zone.now }
+
+    trait :with_completed_order do
+      facility
+
+      transient do
+        quantities [1]
+
+        facility_account do
+          facility
+            .facility_accounts
+            .create(FactoryGirl.attributes_for(:facility_account))
+        end
+
+        order { FactoryGirl.create(:purchased_order, product: product) }
+
+        product do
+          FactoryGirl.create(:setup_item, facility: facility, facility_account: facility_account)
+        end
+
+      end
+
+      after(:build) do |_journal, evaluator|
+        order_details = evaluator.order.order_details
+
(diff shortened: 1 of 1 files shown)


spec/factories/journals.rb
```diff
@@ -3,5 +3,47 @@
...
+
+        facility_account do
+          facility
+            .facility_accounts
+            .create(FactoryGirl.attributes_for(:facility_account))
+        end
+
+        order { FactoryGirl.create(:purchased_order, product: product) }
+
+        product do
```
jhanggi: Does `association :product, factory: :setup_item, facility: facility, facility_account: facility_account` work?

Similar for the `order` association. Seems like it might be a little cleaner.

jhanggi: And would better handling `build`/`build_stubbed`.

spec/factories/journals.rb
```diff
@@ -3,5 +3,47 @@
...
+          FactoryGirl.create(:setup_item, facility: facility, facility_account: facility_account)
+        end
+
+      end
+
+      after(:build) do |_journal, evaluator|
```
jhanggi: Is there a `before(:create)` hook? It seems like it should be better to not persist stuff during `build`.

jhanggi:
Do we want to pull this in or close it out?

drench:
I'd like to keep the non-engine parts of the app as close as we can across forks, though it would be a little odd to bring in a factory that this app isn't actually using yet. I lean toward bringing it in.

BTW @rdunlop had some good suggestions for improvements, like to pull much of the `after(:build)` (which doesn't directly touch the `journal`) into the product factory.

drench:
Closing, but I'd like to revive this in some form next time we work on `Journal`-related specs.

PR #558, opened 2016-04-27
[#125700] Hide reconciliation links from cross-facility

jhanggi:
Reconciliation doesn’t work in the cross-facility context, so we should
hide it from the menu (similar to statements). We have a separate story
for enabling cross-facility statements (#125702), so we can look into
enabling this when we play that story.


drench:
:+1:

PR #559, opened 2016-04-27
Statement spec should pass for newer pdf versions

jhanggi:
UIC uses a PNG with transparency, which is only supported by PDF 1.4+.
Prawn is apparently smart enough to use the lowest compatibility level
based on the features that get used.

https://acrobatusers.com/tutorials/understanding-pdf-compatibility-levels


jhanggi:
This should address the test failures in https://github.com/tablexi/nucore-uic/pull/70

drench:
:+1:

PR #560, opened 2016-04-27
[#118500913] Ability to disable creation of certain account types

jhanggi:
This will be used to disable the creation of NufsAccounts in branches where we are using other custom account types. Disabling just the interface for creation instead of completely is much simpler than fixing all of the tests that would be broken.


drench:
:+1:

PR #562, opened 2016-04-28
Revenue account validations

drench:
This pulls in the parts of https://github.com/tablexi/nucore-nu/pull/151 that are not specific to NU, to generalize the `revenue_account` validations to allow integers. Number ranges and format validations, if any, would be up to the forks to implement.


jhanggi:
👍

PR #564, opened 2016-04-28
[#118597103] Move default revenue and product accounts to Settings

jhanggi:
There's one spec in the NU branch that uses `COMMON_ACCOUNT`, so we'll need to change that there.


drench:
:+1:

PR #565, opened 2016-04-28
[#115962501] Remove instances of "facility" from views

jhanggi:
* Refactor products#index to share a partial

With this, there should be no remaining references to “Facility” or “Facilities” coded into the views. They should only remain in the locales.


drench:
:+1:

PR #566, opened 2016-04-29
I18n "Facility" from Controllers and Views

jhanggi:
This is a work in progress. It continues from #565.

There are still instances of "facility"/"facilities" that I haven't updated yet.

I've been somewhat inconsistent with whether or not I lint the views (quotes/hash syntax). I've been doing some, but sometimes it just feels like it's more work than it's worth. Similarly, when I decide whether to extract things out into their own en.xxx.yml file: I do it when it makes sense and makes things nicer without being too much work.

I'd just like some feedback on the how it's looking so far. And this will definitely be a cleaner diff if we can merge in #565.


config/locales/en.yml
```diff
@@ -29,8 +29,15 @@ en:
       calendar: "%a, %d %b %Y %H:%M:%S"
       usa: "%m/%d/%Y %l:%M %p"
 
+  Facility: "!activerecord.models.facility.one!"
+  Facilities: "!activerecord.models.facility.other!"
+  facility_downcase: facility
```
jhanggi: Here's the meat of the technique. If I need capital facility, I can do something easy like `Manage !Facility!`.
For when it happens in longer sentences, I chose to go with `blah blah !facilities_downcase! blah blah`. It doesn't read quite as nicely, but I chose to go with this as opposed to needing to pass in `Facility.model_name.human` everywhere we need it. I couldn't use plain `facilities` because that's a later key in the yaml with children.

config/locales/override/en.yml
```diff
@@ -2,6 +2,13 @@
 en:
   # app_name: NUcore
   # institution_name: Northwestern
+  activerecord:
```
jhanggi: I committed this accidentally. This is where we would do the override for DC.

config/locales/en.yml
```diff
@@ -105,6 +86,9 @@ en:
 
   shared:
     cancel: Cancel
+    edit: Edit
+    save: Save
+    create: Create
```
drench: Nice additions

app/views/admin/products/_details_fields.haml
```diff
@@ -0,0 +1,24 @@
+= f.input :name, hint: text(".hints.name", field: f.object.class.name.downcase)
+
+- url_name_hint = begin send("facility_#{f.object.class.name.downcase}_url", current_facility, f.object) rescue send("facility_#{f.object.class.name.downcase}_url", current_facility, "url-name") end
```
drench: Agreed on not taking on all view cleanup right now, but I'd make an exception for this because it's got a `rescue` in it.

drench:
It's looking good to me. I'm still trying to think if there's an alternative to `!facilities_downcase!` that might read better, but I've got nothing yet.

jhanggi:
This should now be everything in the controllers and views. There are still some references to facilities remaining in en.models.yml. Those will require a little more thought on how to clean them up since the I don't believe model errors and attributes can use `text_helpers` easily. There are some cases where it can be made to use text_helpers since it's using a direct call to `I18n.t`, but we can't hook into presence validations and whatnot.

jhanggi:
I'm going to now work on verify the overrides work correctly in NU/UIC.

jhanggi:
Made a couple more changes here. The important one is the load_path ordering. `views/en.facilities.yml` was taking precedence over `en.override.yml`, which I discovered with the `facilities.index.welcome` translation.

I'll have an NU PR shortly. UIC will probably happen tomorrow.

drench:
This has been working well for me in tests. The only places where I could see "Facility" leak to the front end is where we display hardcoded user roles like "Facility Staff" or "Facility Director" but that feels like a separate PR, if that's something we need to take on. :+1:

PR #567, opened 2016-04-29
[#116373301] Add Projects index and new/create

drench:
This adds a new facility admin tab for viewing lists of Projects and and creating new ones.

I initially had the tab display at the end of the admin's tab list, but committed e80bc0b to place it second to last. I think it works better there.

### DIFF
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -19,7 +19,7 @@ module Projects
       @project = current_facility.projects.new(project_params)
       if @project.save
         flash[:notice] =
-         I18n.t("controllers.projects.projects.create.success", project_name: @project.name)
+          I18n.t("controllers.projects.projects.create.success", project_name: @project.name)
         redirect_to facility_projects_path(current_facility)
       else
         render action: :new
@@ -33,4 +33,5 @@ module Projects
     end
 
   end
+
 end
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -2,6 +2,9 @@ module Projects
 
   class ProjectsController < ApplicationController
 
+    admin_tab :all
+    before_filter { @active_tab = "admin_projects" }
+
     load_and_authorize_resource
 
     def index
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -1,0 +1,26 @@
+module Projects
+
+  class ProjectsController < ApplicationController
+
+    load_and_authorize_resource
+
+    def index
+      @projects = current_facility.projects
+    end
+
+    def new
+      @project = Projects::Project.new(facility: current_facility)
+    end
+
+    def create
+      @project = current_facility.projects.new(params[:projects_project])
+      if @project.save
+        flash[:notice] =
+         I18n.t("controllers.projects.projects.create.success", project_name: @project.name)
+        redirect_to facility_projects_path(current_facility)
+      else
+        render action: :new
+      end
+    end
+  end
+end
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -13,7 +13,7 @@ module Projects
     end
 
     def create
-      @project = current_facility.projects.new(params[:projects_project])
+      @project = current_facility.projects.new(project_params)
       if @project.save
         flash[:notice] =
          I18n.t("controllers.projects.projects.create.success", project_name: @project.name)
@@ -22,5 +22,12 @@ module Projects
         render action: :new
       end
     end
+
+    private
+
+    def project_params
+      params.require(:projects_project).permit("description", "name")
+    end
+
   end
 end
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -8,7 +8,7 @@ module Projects
     load_and_authorize_resource
 
     def index
-      @projects = current_facility.projects
+      @projects = current_facility.projects.paginate(page: params[:page])
     end
 
     def new
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -12,7 +12,6 @@ module Projects
     end
 
     def new
-
     end
 
     def create
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -5,14 +5,14 @@ module Projects
     admin_tab :all
     before_filter { @active_tab = "admin_projects" }
 
-    load_and_authorize_resource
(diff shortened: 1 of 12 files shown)


vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def new
+      @project = Projects::Project.new(facility: current_facility)
+    end
+
+    def create
+      @project = current_facility.projects.new(project_params)
+      if @project.save
+        flash[:notice] =
+          I18n.t("controllers.projects.projects.create.success", project_name: @project.name)
```
jhanggi: Does I18n sanitize the interpolation options? Probably not a big deal here, though; just something to watch out for.

vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    before_filter { @active_tab = "admin_projects" }
+
+    load_and_authorize_resource
+
+    def index
+      @projects = current_facility.projects.paginate(page: params[:page])
+    end
+
+    def new
+      @project = Projects::Project.new(facility: current_facility)
```
jhanggi: I believe CanCanCan sets this object up for you already.

vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  class ProjectsController < ApplicationController
+
+    admin_tab :all
+    before_filter { @active_tab = "admin_projects" }
+
+    load_and_authorize_resource
+
+    def index
+      @projects = current_facility.projects.paginate(page: params[:page])
```
jhanggi: If you used `load_and_authorize_resource through: :current_facility`, then you could say `@projects = @projects.paginate(...)`. Cancan has some nice loading behavior when you configure it well (most of nucore is not set up to use it well).

vendor/engines/projects/app/views/projects/projects/index.html.haml
```diff
@@ -0,0 +1,17 @@
...
+  = t(".head")
+
+- if current_ability.can?(:create, Projects::Project)
+  %p= link_to t(".add"), new_facility_project_path, class: "btn-add"
+
+- if @projects.any?
+  %table.table.table-striped.table-hover
+    %thead
+      %tr
+        %th= Projects::Project.human_attribute_name(:name).titlecase
```
jhanggi: Is the `titlecase` necessary?

vendor/engines/projects/app/views/projects/projects/index.html.haml
```diff
@@ -0,0 +1,17 @@
+= content_for :h1 do
+  = t(".head")
+
+- if current_ability.can?(:create, Projects::Project)
+  %p= link_to t(".add"), new_facility_project_path, class: "btn-add"
```
jhanggi: I've started setting up some generic helper localse for these over in my other branch. We don't need to do anything right now, but we could revisit these once we get text_helpers in place.

vendor/engines/projects/lib/projects/engine.rb
```diff
@@ -2,8 +2,13 @@ module Projects
 
   class Engine < Rails::Engine
 
+    def self.enable!
```
jhanggi: `enable!` is never actually getting called. If I log in as a non-admin, I don't see the menu item. Admins see it because of `can :manage, :all`

vendor/engines/projects/spec/controllers/projects/projects_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
+require "rails_helper"
+require_relative "../../projects_spec_helper"
+
+RSpec.describe Projects::ProjectsController, type: :controller do
```
jhanggi: I don't know how you feel about it, but I don't feel strongly about needing to namespace the engine specs in the filenames. I.e. I would have no problem if this were `vendor/engines/projects/spec/controllers/projects_controller_spec.rb`. The `app` directory needs it for autoloading to work properly, but specs don't.

vendor/engines/projects/spec/controllers/projects/projects_controller_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+          do_request
+        end
+
+        it "shows the new view", :aggregate_failures do
+          expect(response.code).to eq("200")
+          expect(assigns(:project)).to be_kind_of(Projects::Project).and be_new_record
+        end
+      end
+
+      %i(facility_administrator facility_director senior_staff staff).each do |role|
```
jhanggi: `facility_operators` from `controller_spec_helpers` has this list. I'm not a huge fan of most of those helper methods, but the role lists are useful so we don't need to maintain them in multiple places.

vendor/engines/projects/spec/projects_spec_helper.rb
```diff
@@ -0,0 +1,3 @@
+RSpec.configure do |config|
+  config.before(:all) { Projects::Engine.enable! }
```
jhanggi: Oh, here's where it gets enabled in specs. Since we're using the existence of the gem in the gemfile we should always `enable!` it. The big reason we used `enable!`/`disable!` in split accounts was because we were using a separate feature flag.

vendor/engines/projects/app/controllers/projects/projects_controller.rb
drench: i18n doesn't sanitize, but the flash value is when displayed.

vendor/engines/projects/app/views/projects/projects/index.html.haml
drench: Maybe? I think the chance of this translation changing to something that's more than one word is low, but if it did become 2 or more words, would we want it `titlecase`d?

There are other places where we have header labels with more than one word, and that's what we're doing. One example:
https://github.com/tablexi/nucore-open/blob/master/app/views/facility_journals/index.html.haml#L54-L55

vendor/engines/projects/spec/controllers/projects/projects_controller_spec.rb
drench: I like having the spec structure reflect the files they relate to. Honestly the thing I like most is that using this convention, https://github.com/tpope/vim-rails picks up on the two files being related.

vendor/engines/projects/app/controllers/projects/projects_controller.rb
jhanggi: Cool, just wanted to verify this was getting sanitized somehow.

drench:

vendor/engines/projects/app/views/projects/projects/new.html.haml
```diff
@@ -0,0 +1,12 @@
+= render "shared/headers/ckeditor"
+
+= content_for :h1 do
+  = t(".head")
+
+= simple_form_for @project, url: facility_projects_path(current_facility), method: :post do |f|
+  .form-inputs
+    = f.input :name, hint: t(".name")
```
jhanggi: Not a big deal, but I like to use a `hints` prefix, so this would be `.hints.name`. You could also put the hints in the (`simple_form.hints` space)[https://github.com/plataformatec/simple_form#i18n] and have it automatically picked up. Currently we have a few things in the `simple_form.yml` file using that technique, but I've become less of a fan using that file because it separates the hints from other translations related to the model. But it might be fine to use within the engine file. I don't have a strong preference.

jhanggi:
Just one minor thought on translation locations, but otherwise 👍

PR #568, opened 2016-04-30
[#125075] Manually set reconciliation date when reconciling

jhanggi:
This handles both reconciling CC/PO and "succeeded, with errors" journals. It also sets the reconciled_at to the current time for "succeeded, no errors".

There is still an open question around what we want to do with the historical manually-reconciled details (those that are not part of a statement or PR). Currently I have them being set to `reviewed_at`, but that may change.

### DIFF
--- a/app/assets/javascripts/app/datepicker-data.coffee
+++ b/app/assets/javascripts/app/datepicker-data.coffee
@@ -1,0 +1,9 @@
+$ ->
+  # This will set up a datepicker based on data attributes
+  # Make sure you call `to_s` or `iso8601` on any dates you set in the views to
+  # ensure the proper format.
+  picker = $(".datepicker__data")
+  picker.datepicker(
+    minDate: new Date(picker.data("min-date")) # will be unbounded if not provided
+    maxDate: new Date(picker.data("max-date")) # will be unbounded if not provided
+  )
--- a/app/assets/javascripts/app/datepicker-data.coffee
+++ b/app/assets/javascripts/app/datepicker-data.coffee
@@ -2,8 +2,8 @@ $ ->
   # This will set up a datepicker based on data attributes
   # Make sure you call `to_s` or `iso8601` on any dates you set in the views to
   # ensure the proper format.
-  picker = $(".datepicker__data")
-  picker.datepicker(
-    minDate: new Date(picker.data("min-date")) # will be unbounded if not provided
-    maxDate: new Date(picker.data("max-date")) # will be unbounded if not provided
+  $picker = $(".datepicker__data")
+  $picker.datepicker(
+    minDate: new Date($picker.data("min-date")) # will be unbounded if not provided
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/app/datepicker-data.coffee
```diff
@@ -0,0 +1,9 @@
+$ ->
+  # This will set up a datepicker based on data attributes
+  # Make sure you call `to_s` or `iso8601` on any dates you set in the views to
+  # ensure the proper format.
+  picker = $(".datepicker__data")
```
drench: `$picker` since it's a jQuery object?

app/views/facility_accounts_reconciliation/_table.html.haml
```diff
@@ -22,5 +21,4 @@
               .order-detail-extra.order-detail-note= order_detail.note
 
         %td= show_actual_total(order_detail)
-        %td= text_field_tag "order_detail[#{order_detail.id}][notes]", begin params[:order_detail][order_detail.id.to_s][:notes] rescue "" end
-  = render partial: "action_row"
+        %td= text_field_tag "order_detail[#{order_detail.id}][reconciled_note]", begin params[:order_detail][order_detail.id.to_s][:reconciled_note] rescue "" end
```
drench: Another pre-existing `rescue` in a view! There aren't many left. Same note as https://github.com/tablexi/nucore-open/pull/566#discussion_r61766376

spec/controllers/facility_accounts_reconciliation_controller_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+          expect(flash[:error]).to include("cannot be in the future")
+        end
+      end
+
+      describe "with a reconciliation date before the statement" do
+        let(:reconciled_at) { 10.days.ago }
+
+        it "does not reconcile the order" do
+          expect { perform }.not_to change { order_detail.reload.state }
+          expect(order_detail.state).to eq("complete")
```
drench: Could you combine these like:
```ruby
expect { perform }.not_to change { order_detail.reload.state }.from("complete")
```

app/views/facility_accounts_reconciliation/_table.html.haml
jhanggi: Yeah, I originally just didn't want to try thinking through why this was there. It appears to be a funky way around doing `params.try(:[], :order_detail).try(:[], order_detail.id.to_s).try(:[], :reconciled_note)` or if were were on ruby 2.3 `params.dig(:order_detail,...)`.

It turns out the only place we're using a direct render is in the `index` view where these parameters won't be there, so we can simply remove the second `text_field_tag` argument.

jhanggi:
Thanks for the feedback. Updated.

drench:
:+1:

PR #569, opened 2016-05-02
[#116373789] Add Project show view

drench:
This adds a bare-bones show view for a `Project`, and adds links to these from project index listings.


jhanggi:
👍

PR #570, opened 2016-05-02
Allow HTML in Project descriptions to display as HTML

drench:
This is a loose end from #569: we're allowing HTML in `Project#description` but the "show" view was escaping it. This marks the string as "safe", though it does mean trusting that a privileged user (facility operator) wouldn't do anything malicious.

### DIFF
--- a/vendor/engines/projects/app/views/projects/projects/show.html.haml
+++ b/vendor/engines/projects/app/views/projects/projects/show.html.haml
@@ -2,4 +2,4 @@
   = @project.facility
 
 %h2= @project.name
-%p= @project.description
+%p= @project.description.html_safe
--- a/vendor/engines/projects/app/views/projects/projects/show.html.haml
+++ b/vendor/engines/projects/app/views/projects/projects/show.html.haml
@@ -2,4 +2,4 @@
   = @project.facility
 
 %h2= @project.name
-%p= @project.description.html_safe
+= sanitize @project.description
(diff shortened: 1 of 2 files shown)


vendor/engines/projects/app/views/projects/projects/show.html.haml
```diff
@@ -2,4 +2,4 @@
   = @project.facility
 
 %h2= @project.name
-%p= @project.description
+%p= @project.description.html_safe
```
jhanggi: Since `description` will already have the `<p>` tag, do we want it here?

Also, I'm wondering if using the `sanitize` helper method might be useful (probably apply it elsewhere as well). After a quick test, we would need to add `style` to the list of allowed attributes.

I think we could do that in an initializer
`ActionView::Base.sanitized_allowed_attributes << "style"`

http://apidock.com/rails/v3.2.13/ActionView/Helpers/SanitizeHelper/sanitize

What do you think?

drench: I like it. There are definitely other places in this app that could use it too.

drench:
Updated with `sanitize`.

jhanggi:
I don't want to deal with this now, but at some point when we start rolling this out, we either need to disable tables in the wysiwyg or add the necessary tags and attributes (the attributes is where the list gets a little nasty).

👍

PR #571, opened 2016-05-03
[#116373897] Add edit and update actions for Projects

drench:
This adds a basic interface for editing projects by adding an edit button to the project show view.

It also includes a few `active` flag to the `Project` model. A checkbox appears on the edit form to toggle `active` but the flag has no impact on anything else yet.

### DIFF
--- a/vendor/engines/projects/app/views/projects/projects/edit.html.haml
+++ b/vendor/engines/projects/app/views/projects/projects/edit.html.haml
@@ -3,10 +3,12 @@
 = content_for :h1 do
   = @project.facility
 
+%h2= t(".update")
+
 = simple_form_for @project, url: facility_project_path(@project.facility, @project), method: :put do |f|
   .form-inputs
     = f.input :name, hint: t(".name")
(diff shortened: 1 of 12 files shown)


vendor/engines/projects/app/views/projects/projects/edit.html.haml
```diff
@@ -0,0 +1,15 @@
+= render "shared/headers/ckeditor"
+
+= content_for :h1 do
+  = @project.facility
+
+%h2= t(".update")
+
+= simple_form_for @project, url: facility_project_path(@project.facility, @project), method: :put do |f|
```
jhanggi: can we share the form with `new`?

drench: It's close; `edit` has the `:active` checkbox and `new` doesn't. I lean toward keeping them separate for now.

jhanggi: Ok, that's fine.

jhanggi:
👍

PR #572, opened 2016-05-04
Only reset locales in specs where necessary

jhanggi:
Doing the `TextHelpers::RSpec.reset_spec_translations` on every single spec significantly increased the testing time on CI, especially for UIC, which went from around 18 minutes to 30-40. I'm not sure why it didn't have the same level of impact on open and NU. This drops the time back down to 17-18 minutes on UIC.


drench:
:+1:

PR #573, opened 2016-05-04
[#115962679] Use `app_name` so the name is more easily overridable 

jhanggi:
The definition of `app_name` should be the only place "NUcore" appears in the locales. Once this is in, we should be able to clean up a lot of the overrides in UIC.


drench:
:+1:
(if we could get `app_name` to update the README too…)

PR #574, opened 2016-05-04
[#116374821] Assign a project to an OrderDetail

drench:
This injects Project selection into the order detail edit screen (modal, typically).

Highlights:

* The `order_details` table gets another attribute: `project_id`. There are related changes to the order detail updater for the Projects engine to make changes to permitted params.
* There are validations to check that the project:
  - belongs to the same facility as the order/order detail
  - is active, unless the project was already associated (presumably when it was active)
* There's some cleanup of the edit view partial
* The project selection is injected through a view hook
* There's a change to how it activates `chosen`, in that if the `<select>` is optional, you should be able to "x" out a previous selection.

### DIFF
--- a/vendor/engines/projects/app/models/projects/order_detail_extension.rb
+++ b/vendor/engines/projects/app/models/projects/order_detail_extension.rb
@@ -33,6 +33,7 @@ module Projects
         errors.add(:project_id, :project_is_inactive)
       end
     end
+
   end
 
 end
--- a/vendor/engines/projects/app/models/projects/order_detail_extension.rb
+++ b/vendor/engines/projects/app/models/projects/order_detail_extension.rb
@@ -1,0 +1,26 @@
+module Projects
+
+  module OrderDetailExtension
+
+    extend ActiveSupport::Concern
+
+    included do
+      belongs_to :project,
+                 class_name: "Projects::Project",
+                 foreign_key: :project_id,
+                 inverse_of: :order_details
+
+      validate :project_facility_matches?
+    end
+
+    private
+
+    def project_facility_matches?
+      if project_id.present? && project.facility != facility
+        errors.add(:project_id, :project_facility_mismatch)
+      end
+    end
+
+  end
+
+end
--- a/vendor/engines/projects/app/models/projects/order_detail_extension.rb
+++ b/vendor/engines/projects/app/models/projects/order_detail_extension.rb
@@ -13,11 +13,11 @@ module Projects
       validate :project_facility_must_match
       validate :project_must_be_active, if: :project_id_changed?
 
-      delegate :projects, to: :facility, allow_nil: true
+      delegate :projects, to: :facility, prefix: true
     end
 
     def selectable_projects
-      (projects.active + [project]).uniq.compact
+      (facility_projects.active + [project]).uniq.compact
     end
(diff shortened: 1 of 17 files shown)


vendor/engines/projects/app/models/projects/project.rb
```diff
@@ -5,12 +5,15 @@ class Project < ActiveRecord::Base
...
 
     belongs_to :facility, foreign_key: :facility_id
+    has_many :order_details, inverse_of: :project
 
     validates :facility_id, presence: true
     validates :name,
               presence: true,
               uniqueness: { case_sensitive: false, scope: :facility_id }
 
+    scope :active, conditions: { active: true }
```
jhanggi: Does 3.2 allow `scope :active, -> { where(active: true) }`? I feel like we should be using a more modern syntax.

vendor/engines/projects/spec/services/projects/order_details/param_updater_extension_spec.rb
```diff
@@ -0,0 +1,10 @@
+require "rails_helper"
+
+RSpec.describe Projects::OrderDetails::ParamUpdaterExtension do
```
jhanggi: I feel like a controller spec that makes sure that the project_id gets saved would be nice to have.

vendor/engines/projects/app/views/projects/shared/_select_project.html.haml
```diff
@@ -0,0 +1,9 @@
+.row
+  .span10
+    = f.input :project_id,
+      collection: order_detail.selectable_projects,
+      input_html: { class: "js--chosen", data: { placeholder: t(".placeholder") } },
+      selected: order_detail.project_id,
+      include_blank: true
+
+= javascript_include_tag "activate_chosen" if modal
```
jhanggi: Could we do something like define a method in the JS for this, call it on load of that file to get the normal behavior, and trigger that same method call here rather than loading another file. I would be surprised if chosen didn't already handle having `chosen` called on it multiple times. Could we not deal with the `modal` check?

drench:
Updated. I also switched the existing controller `#update` specs to `PUT`s instead of `POST`s.

jhanggi:
👍

vendor/engines/projects/app/models/projects/order_detail_extension.rb
```diff
@@ -0,0 +1,39 @@
...
+    included do
+      belongs_to :project,
+                 class_name: "Projects::Project",
+                 foreign_key: :project_id,
+                 inverse_of: :order_details
+
+      validate :project_facility_must_match
+      validate :project_must_be_active, if: :project_id_changed?
+
+      delegate :projects, to: :facility, allow_nil: true
```
drench: Just noticed: this probably would be better with a `prefix: true`. Updating.

jhanggi: Actually, what is this being used for?

jhanggi: Oh, I found it.

drench:
One minor update on a delegation.

jhanggi:
👍

PR #575, opened 2016-05-06
Ensure tmp exists at app root, and use it

drench:
This is partially from https://github.com/tablexi/nucore-dartmouth/pull/38. We need `tmp` to exist at the app root, and it does, though it seems it's only because gems like `sprockets` and `letter_opener` will create it if it's not there.

This also reworks the only code I'm aware of that was relying on `Dir.tmpdir`, in journal creation. With this in place, some of the overrides in the DC fork may become redundant.


jhanggi:
👍

PR #576, opened 2016-05-06
[#116374821] Present project selection if there are selectable projects

drench:
This is a loose end from #574: if there are no projects available to select, it shouldn't show a selection widget.


jhanggi:
👍

PR #577, opened 2016-05-06
[#119031765] New email interceptor that allows a whitelist

jhanggi:
As opposed to capturing everything and sending it to a list, this sets up a
whitelist of people who can receive emails. If an email is sent to someone not
on the list, it is captured and redirected to a list of developers.

This should require no configuration changes unless we want to add people to the whitelist that shouldn’t receive the intercepted emails.

### DIFF
--- a/spec/lib/staging_mail_interceptor_spec.rb
+++ b/spec/lib/staging_mail_interceptor_spec.rb
@@ -24,6 +24,14 @@ RSpec.describe StagingMailInterceptor do
         end
       end
 
+      describe "case insensitivity" do
+        let(:to) { ["ALLOWED@example.org"] }
+
+        it "lets the message through" do
+          expect(message.to).to eq(to)
+        end
+      end
+
       describe "multiple addresses" do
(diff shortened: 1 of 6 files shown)


spec/lib/staging_mail_interceptor_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+
+    before do
+      allow(interceptor).to receive(:whitelist) { whitelist }
+      allow(interceptor).to receive(:send_to_addresses) { send_to }
+      interceptor.process
+    end
+
+    describe "when the email is sent to someone on the list" do
+      describe "single address" do
+        let(:to) { ["allowed@example.org"] }
```
drench: If I change this to `allowed@EXAMPLE.ORG` the spec fails. Should it be case-insensitive?

jhanggi: Updated to make the whitelist case-insensitive.

drench:
:+1:

PR #578, opened 2016-05-07
[#115962501] More support of "Facility" overriding

jhanggi:
This takes care of what #566 did not. After this, there should be no more references to "facility" or "facilities" directly in any strings. They should all go through `facility_downcase` or `model_name.human`.

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -88,7 +88,7 @@ class FacilitiesController < ApplicationController
     @facility = Facility.new(facility_params)
 
     if @facility.save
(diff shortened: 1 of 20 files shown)


jhanggi:
I have all the specs fixed up on this one now. There is a duplication warning from code climate that I don't want to deal with here. I think it's somewhat addressed in the WIP #371, which is probably pretty out of date at this point.

drench:
I'm aware of one more (my fault) in `vendor/engines/projects/config/locales/en.yml`, though it's a message that nobody should ever see unless they're messing with form values. :+1: otherwise.

PR #579, opened 2016-05-10
[#125075] Manual reconciliation date updates

jhanggi:
These are based on feedback from our last discussion with NU about the feature.

One big thing we decided was to leave historical non-journaled reconciliation dates blank. They'd rather have no data than incorrect data. Journaled orders' reconciled_at will always be set to the journal_date.

Also, they gave us the go ahead to change the export raw template so I moved the split percentage to a spot where it makes a little more logical grouping.

The migrations haven't gone out on NU & UIC, and there is no important data on DC, so I felt OK changing the migration in place.

### DIFF
--- a/app/views/facility_journals/show.html.haml
+++ b/app/views/facility_journals/show.html.haml
@@ -37,10 +37,12 @@
 - else
   = form_tag facility_journal_reconcile_path(current_facility, @journal), method: :post do
     - journal_is_submittable = @journal.submittable?
-    - if journal_is_submittable
-      = render "facility_accounts_reconciliation/action_row", date: true, unreconciled_order_details: @order_details, starting: :all
     %table.table.table-striped.table-hover
       %thead
+        - if journal_is_submittable
+          %tr.borderless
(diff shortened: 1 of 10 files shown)


app/views/facility_journals/show.html.haml
```diff
@@ -37,10 +37,12 @@
 - else
   = form_tag facility_journal_reconcile_path(current_facility, @journal), method: :post do
     - journal_is_submittable = @journal.submittable?
-    - if journal_is_submittable
-      = render "facility_accounts_reconciliation/action_row", date: true, unreconciled_order_details: @order_details, starting: :all
     %table.table.table-striped.table-hover
       %thead
+        - if journal_is_submittable
+          %tr.borderless
+            %th{colspan: 3}= link_to "Select All", "JavaScript:void(0);", class: "select_all"
```
drench: "Select All" seems like a good candidate for a shared translation.

drench:
:+1: with one note on i18n.

app/views/facility_journals/show.html.haml
jhanggi: Good call. I'll do a separate PR for that.

PR #581, opened 2016-05-10
[#119019273] Use "facilities" locale in routes

jhanggi:
There’s cleanup that could be done to the routes (why all the reports
routes are like that at the bottom instead of inside the `resources
:facilities`, I don’t know), but I don’t want to tackle that here.


drench:
Amazed at how small this diff is. It breaks if `facilities_downcase` has whitespace (like "`shared resource`") but we could consider snake_case if that ever happens. :+1:

PR #582, opened 2016-05-10
[#116374999] Project filter on transaction search

jhanggi:
There is definitely more refactoring I would love to do to TransactionSearch to get rid of a lot of the metaprogramming and get it more wrapped up in tighter objects rather than making very large controllers, but this felt like a good stopping point.

First commit is a refactor; second is actually adding the project search.


drench:
Filtering by project is working for me. Really happy to see refactoring begin on that code. :+1:

PR #583, opened 2016-05-10
[#116372857] Add Project when ordering-on-behalf

drench:
I wanted to put this out in case @jhanggi had time this evening to review, though there are two outstanding issues I am working on cleaning up that don't directly affect how this feature functions:

* The placement of the project selection widget on the cart screen needs improvement
* Adding `inverse_of` to the `Order` and `OrderDetail` models broke specs for `Accessories::ChildUpdater`


app/views/orders/_form.html.haml
```diff
@@ -0,0 +1,52 @@
...
+    %ul.inline{style: "float: left"}
+      %li= submit_tag t("shared.update"), class: "btn"
+      - if order.validated?
+        %li
+          = f.submit t("shared.purchase"),
+            class: %w(btn btn-primary),
+            confirm: t(".confirm")
+
+      %li
+        = render_view_hook "acting_as",
```
jhanggi: This might be related to the cleaning up the styling, but this is displaying even if you aren't "ordering as"

jhanggi:
Not sure how much we want to worry about it, but if I change a quantity and the project and click purchase, I am redirected back to the cart with "Quantities have changed. Please review updated prices then click "Purchase"" and the project is lost.

app/models/order_detail.rb
```diff
@@ -38,7 +38,7 @@ def set_problem_order
   belongs_to :price_policy
   belongs_to :statement, inverse_of: :order_details
   belongs_to :journal
-  belongs_to :order
+  belongs_to :order, inverse_of: :order_details
```
drench: We put the reverse in the `Order` model yesterday, but that caused the `Accessories::ChildUpdater` specs to fail. I'd still like that to happen. but I've been looking at it too long and need to step away.

drench:
I'm not 100% happy with the positioning of the project selection widgets when purchasing, but this is ready for a review.

app/controllers/orders_controller.rb
```diff
@@ -225,16 +233,18 @@ def update_or_purchase
   def update
     load_statuses
     params[:order_datetime] = build_order_date if acting_as?
+
     @order.transaction do
+      @order.assign_attributes(order_params)
```
jhanggi: I'm not sure if there's something we can do easily to avoid the code climate warning about attribute sanitization. I wonder if it's because there's a return case where it's `nil` as opposed to a `ActionController::Parameters`.

app/controllers/reservations_controller.rb
```diff
@@ -94,7 +94,8 @@ def list
   # POST /orders/1/order_details/1/reservations
   def create
     raise ActiveRecord::RecordNotFound unless @reservation.nil?
-    @reservation = @order_detail.build_reservation(reservation_create_params)
+    @reservation = @order_detail.build_reservation
+    @reservation.assign_attributes(reservation_create_params)
```
jhanggi: This was already a sanitization issue before. We can deal with this when we do the rails upgrade.

app/models/order_detail.rb
jhanggi: So leaving the `inverse_of: :order` on `Order` caused the spec failures, so you're leaving it off for now? Or did you figure something out?

jhanggi:
I'm thinking for reservations, the place to put the dropdown would be up by the note. I don't love the positioning in the cart where the "More options" pushes the purchase button down, but I also don't hate it, nor do I have a better idea for it.

app/models/order_detail.rb
drench: That's where it's at now. It's not obvious to me why the `inverse_of: :order` breaks child order_detail updates. I was hoping it was something in the way the specs are set up, but it doesn't seem to be. They're either not happening or getting reverted.

jhanggi: That's what I figured, just wanted to check. I'm fine with not spending any more time on it.

drench:
I added a commit to move it where it's below the account selection:
new reservation project selection

I considered putting it next to "note" but changed my mind since we're planning to expand that into a textarea very soon. What do you think?

jhanggi:
Works for me. 👍

PR #585, opened 2016-05-10
Support overridden "facility" in spec

jhanggi:
This was causing a spec failure in DC where "facility" is "resource".


drench:
:+1:

PR #586, opened 2016-05-11
Ensure tests work in timezones other than Central

jhanggi:
https://github.com/rails/rails/commit/39374aa925a7d670b039c0c0c9aa9f4aef19466b

I pulled in some spec changes after seeing failing tests when DC ran in Eastern time. I set this to run in Pacific to make sure it works that way as well. We'll revert that change before merging.


drench:
:+1: with the TZ revert

PR #587, opened 2016-05-11
[#118125127] Add Projects to export raw

jhanggi:
Also, add a default to the export raw column headers so we don’t need
so many extra locales.

I'm not set on this column location within the report, but we could easily change it.


drench:
:+1:

PR #588, opened 2016-05-11
Fix i18n route specs 119019273

jhanggi:
I tried to only use the I18n version in routing specs and elsewhere
switch to the url helper methods. I also kept them using string routes
in presenters because switching to the helpers ends up just asserting
exactly what the methods do.

DC’s latest pull.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -28,7 +28,7 @@ en:
   Facility: "!activerecord.models.facility.one!"
   Facilities: "!activerecord.models.facility.other!"
   facility_downcase: facility
-  facilities_downcase: resources
+  facilities_downcase: facilities
 
   admin:
     shared:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -28,7 +28,7 @@ en:
   Facility: "!activerecord.models.facility.one!"
   Facilities: "!activerecord.models.facility.other!"
   facility_downcase: facility
-  facilities_downcase: facilities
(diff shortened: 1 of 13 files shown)


drench:
I think we'll want to update https://github.com/tablexi/nucore-open/blob/07adb56012ecd0dbb92ccfb122d783442aeb8ac4/spec/controllers/users_controller_spec.rb#L191-L194 too; these would always be non-routable if facilities was renamed.

jhanggi:
I fixed the user controller routes in the subsequent commit after I saw the tests fail.

drench:
Right, that's for the "feature enabled" specs. For the "feature disabled" at https://github.com/tablexi/nucore-open/blob/07adb56012ecd0dbb92ccfb122d783442aeb8ac4/spec/controllers/users_controller_spec.rb#L191-L194 -- those use hard-coded `/facilities/`. If you were to change the translation to `resources`, these would never be routable, whether the `:create_users` feature was on or off, right?

config/locales/en.yml
```diff
@@ -28,7 +28,7 @@ en:
   Facility: "!activerecord.models.facility.one!"
   Facilities: "!activerecord.models.facility.other!"
   facility_downcase: facility
-  facilities_downcase: facilities
```
drench: Revert after CI?

jhanggi:
Gotcha on the user specs. And yes, I'll revert the locale before merge.

drench:
:+1:

PR #589, opened 2016-05-11
[#116374977] List order_details on the Project show view

drench:


vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -14,6 +14,10 @@ def index
     def new
     end
 
+    def backend?
+      true
```
drench: Every action the `ProjectsController` handles currently should be considered backend, giving us the modal order detail popup.

jhanggi:
👍

PR #590, opened 2016-05-12
[#116374999] Fix Project Filter on Transactions view

jhanggi:
If the page (all transaction, send notifications, etc) only has one
unique project, then that one project was being shown in the box and
the box was disabled. The logic was relying on the fact that all of the
other select fields were non-nullable.


drench:
:+1:

PR #591, opened 2016-05-12
[#117175831] Projects Report Tabs

jhanggi:
The diff is a little confusing because I moved some files around. Here are the major points of the refactor:

* Moved report controllers to `app/controllers/reports`
* Moved `lib/instrument_reporter.rb` into `app/controllers/concerns/reports`
* Extracted the export raw into its own controller (this allowed getting rid of a lot of ugly conditions)
* All the reports now go through a single action on each controller making the routes nicer
* The reports are now a hash of the action => a lambda (it was actually using blocks before, so this transition wasn't too difficult)

There's definitely more refactoring that could be done, but my goal was to refactor enough to make adding the Projects reports easy.


drench:
Adding reports this way looks nice. :+1:

PR #592, opened 2016-05-12
[#124331] Bulk Email returns duplicates on authorized user search

jhanggi:
If a user is authorized on multiple products, they will be returned as
multiple rows in the results.


drench:
:+1:

PR #593, opened 2016-05-12
[#116374361] Project batch assignment

drench:
The biggest change here is the extraction and (light) refactor of `OrderDetail.batch_update`. It still needs refactoring and I18n, but is at least now in its own service object with specs.

The project select widget does not use `chosen` to be consistent with the other `select` widgets on the page.


app/services/order_detail_batch_updater.rb
```diff
@@ -0,0 +1,144 @@
...
+    @msg_hash = {}
+  end
+
+  def update!
+    unless order_detail_ids.present?
+      msg_hash[:error] = "No #{msg_type} selected"
+      return msg_hash
+    end
+
+    if order_details.any? { |od| !(od.state.include?("inprocess") || od.state.include?("new")) }
```
jhanggi: This line feels especially awkward. I wouldn't mind cleaning it up now.

app/services/order_detail_batch_updater.rb
```diff
@@ -0,0 +1,144 @@
...
+    order_details.each do |order_detail|
+      order_detail.public_send("#{attribute}=", value)
+    end
+  end
+
+  def update_all_order_status_ids
+    order_details.each do |order_detail|
+      # cancel reservation order details
+      if order_status.id == OrderStatus.canceled.first.id && order_detail.reservation
+        unless order_detail.cancel_reservation(user, order_status, true)
```
jhanggi: I don't know if it's worth doing now, but `OrderDetail#update_order_status!` is essentially the same as `change_status!` but will handle the reservation cancelation.

app/services/order_detail_batch_updater.rb
```diff
@@ -0,0 +1,144 @@
...
+  def update_all(attribute, value)
+    order_details.each do |order_detail|
+      order_detail.public_send("#{attribute}=", value)
+    end
+  end
+
+  def update_all_order_status_ids
+    order_details.each do |order_detail|
+      # cancel reservation order details
+      if order_status.id == OrderStatus.canceled.first.id && order_detail.reservation
```
jhanggi: I'm actually curious if this logic is wrong. If there is a sub-status of Canceled, this seems like it might not cancel the reservation.

vendor/engines/projects/app/helpers/projects/application_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  module ApplicationHelper
+
+    def project_id_assignment_options(projects)
+      blank_options + options_from_collection_for_select(projects, :id, :name)
+    end
+
+    private
+
+    def blank_options
```
jhanggi: It might be a good idea to give this a more specific name. Private methods in modules might get weird, especially with these helpers.

```ruby
module A
  private
  def priv
    "a"
  end
end

module B
  private
  def priv
    "b"
  end
end

class Testing
  include A
  include B
  def test
    priv
  end
end

> Testing.new.test
=> "b"
```

vendor/engines/projects/lib/projects/engine.rb
```diff
@@ -4,10 +4,12 @@ class Engine < Rails::Engine
 
     config.to_prepare do
       ::AbilityExtensionManager.extensions << "Projects::AbilityExtension"
+      ::ApplicationHelper.send :include, Projects::ApplicationHelper
```
jhanggi: I actually thought this would get loaded automatically if it's in `engines/xxx/app/helpers`.

app/views/shared/_order_detail_action_form.html.haml
```diff
@@ -1,13 +1,18 @@
...
-  %select{ :name => "order_status_id", :class => 'sync_select', :id => nil }
-    = options_for_select([['Status...', nil]])
-    - OrderStatus.non_protected_statuses(current_facility).each do |os|
-      = options_for_select([[os.name_with_level, os.id]])
+  %select{ name: "order_status_id", class: "sync_select", id: nil }
+    = options_for_select([["Status...", nil]])
+    - OrderStatus.non_protected_statuses(current_facility).each do |order_status|
+      = options_for_select([[order_status.name_with_level, order_status.id]])
+
+= render_view_hook "before_bulk_action_submit_button", facility: current_facility
```
jhanggi: I feel like this might be better over the "Product" column. Right now it's pushing the submit button off the table.

screen shot 2016-05-12 at 5 57 24 pm

jhanggi:
Not directly tied to this story, but we should probably think about how we want to display projects on some of the order detail listings. When you assign to a project in these views, there's no obvious change. You need to drill into the modal.

drench:
Updated to cover the major issues.

jhanggi:
Orders looks fine. Reservations New/In Process I think needs a colspan tweak.

Other than that 👍

jhanggi:
👍

PR #594, opened 2016-05-13
[#115962501] Do not show save success strings on Users#show

jhanggi:
I added this to make testing easier for #578 so I didn't have to go through the
full user creation process while testing the strings, but forgot to take it out
when I was done.


drench:
:+1:

PR #595, opened 2016-05-13
[#116375533] Split the Projects index into active and inactive tabs

drench:
This is similar to what we do for Products, down to using the same `archived` query string parameter.

One minor update to `#update` is that after a successful save, it bounces the user to the appropriate index where the just-edited project will be. It doesn't attempt to find the right `page` value though.

### DIFF
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -47,11 +47,15 @@ module Projects
       params.require(:projects_project).permit("active", "description", "name")
     end
 
+    def project_index_params
+      @project.active? ? {} : { archived: "true" }
+    end
+
     def save_project
       if @project.save
         flash[:notice] =
           I18n.t("controllers.projects.projects.#{action_name}.success", project_name: @project.name)
-        redirect_to facility_projects_path(@project.facility)
+        redirect_to facility_projects_path(@project.facility, project_index_params)
       end
     end
 
--- a/vendor/engines/projects/app/controllers/projects/projects_controller.rb
+++ b/vendor/engines/projects/app/controllers/projects/projects_controller.rb
@@ -9,8 +9,13 @@ module Projects
 
     def index
       @all_projects = @projects
-      @projects = showing_inactive? ? @projects.inactive : @projects.active
-      @projects = @projects.paginate(page: params[:page])
+      @projects = @all_projects.active.paginate(page: params[:page])
+    end
+
+    def inactive
+      @all_projects = @projects
+      @projects = @all_projects.inactive.paginate(page: params[:page])
+      render action: :index
     end
 
     def new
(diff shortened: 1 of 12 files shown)


jhanggi:
I guess this is mostly fine, but I'd almost rather split the active and inactive into separate actions. I think it would make a lot of things much cleaner. What do you think?

vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -34,20 +36,34 @@ def update
...
       if @project.save
         flash[:notice] =
           I18n.t("controllers.projects.projects.#{action_name}.success", project_name: @project.name)
-        redirect_to facility_projects_path(@project.facility)
+        redirect_to facility_projects_path(@project.facility, project_index_params)
       end
     end
 
+    def filter_projects_for_index
+      @projects = showing_archived? ? @projects.inactive : @projects.active
```
jhanggi: All this logic feels kind of convoluted. What about something like

```ruby
def index
  @active_projects = Project.active
  @inactive_project = Project.inactive
  @projects = showing_archived? @inactive_projects : @active_projects
end
```

I also feel the `@(in)active_project_count` instance variables don't add much. If we already have the actual relations, we could just call `.count` on it.

vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -34,20 +36,34 @@ def update
...
+    def project_index_params
+      @project.active? ? {} : { archived: "true" }
+    end
+
     def save_project
       if @project.save
         flash[:notice] =
           I18n.t("controllers.projects.projects.#{action_name}.success", project_name: @project.name)
-        redirect_to facility_projects_path(@project.facility)
+        redirect_to facility_projects_path(@project.facility, project_index_params)
```
jhanggi: What if we redirected to the show view? That's what I believe we do for Products. I feel like this might be adding unnecessary complexity.

drench:
> I'd almost rather split the active and inactive into separate actions
What did you have in mind for routes? We could go with `/projects/active` / `/projects/archived` with `/projects` redirecting to `/projects/active`.

jhanggi:
I think `/projects` and `/projects/inactive` should be fine. I don't think we need the extra `projects/active` and redirect.

drench:
Updated, aside from the switch to `/projects/inactive`. I have a WIP branch that will take more time.

jhanggi:
Other than the routes/actions, 👍

vendor/engines/projects/config/routes.rb
```diff
@@ -1,5 +1,9 @@
 Rails.application.routes.draw do
   resources :facilities, only: [] do
-    resources :projects, controller: "projects/projects", except: [:destroy]
+    resources :projects, controller: "projects/projects", except: [:destroy] do
+      collection do
```
drench: BTW this was the key to everything: putting `inactive` inside a `collection` block made `load_and_authorize_resource` do the right thing.

jhanggi: Ah, cool. You can also write this as `get "inactive", on: :collection` as a one-liner.

jhanggi:
👍

PR #597, opened 2016-05-14
[#117175831] Missing argument on reports routing error

jhanggi:
RoutingError requires a message. The message should not get displayed
to the user, so it’s just for developers.


drench:
:+1:

PR #598, opened 2016-05-14
Vendor javascripts and remove unused assets

jhanggi:
This moves several of the javascript libraries we use into the `vendor/assets` folder, which allows us to exclude them from CodeClimate checks. It also removes some unused images that appear to be leftover from the pre-bootstrap days.


drench:
I didn't run into any 404s but we ought to see those as routing errors in Rollbar, if it turns out there are any. :+1:

jhanggi:
We're currently ignoring routing errors in rollbar because they generally produce a lot of noise. I'd be comfortable turning them back on temporarily.

My process for moving the JS was to find what pages it was being included on and making sure those pages still worked.

PR #599, opened 2016-05-14
Support Ruby 2.3

jhanggi:
I'm not saying we should pull this in just now: I think we should deal with this as part of the Rails 4 upgrade, but I did want to leave this here for posterity and reference.

Here's where I was able to find the solution to the test failures: https://github.com/rails/rails/issues/25010


jhanggi:
Closing as we don't need to deal with this now that we're on Rails 4+.

PR #600, opened 2016-05-17
Remove vestigial gems

jhanggi:
These gems have been present since the initial git commit. However, it does not appear that that are being used. I found these while doing an initial audit of our gems for a Rails 4 upgrade.

We have no references to `PDF::Reader` from https://github.com/yob/pdf-reader I found no references in the initial git commit.

fast-aes's README states "This gem is a relic from 5 years ago, when libraries such as OpenSSL did not work correctly with Ruby. Use in new projects is strongly discouraged. The core Ruby OpenSSL library is faster and more secure." https://github.com/nateware/fast-aes
It appears that it was used way back to encrypt credit card numbers, which we no longer do. The change was introduced in 9d75a5725dce8dd54ede0fadc9bba03e3a1a8c59, but there is little explanation.


drench:
:+1:

PR #601, opened 2016-05-17
Fix OrderStatus table on migration from scratch

jhanggi:
When running `rake db:migrate db:seed` from scratch, the migrations would
result in improperly ordered OrderStatuses.

Adding the Settings.yml's order statuses based on the `order_details.status_change_hooks`
we listen to as observers happens before db:seed, as is the reconciled status. This
will now add them as part of `db:seed`.

I'd like to see if we can remove the `ActiveRecord::Observer` as part of the
Rails 4 upgrade since that is moved out into a separate gem, but that may not
be worth the effort.

The ultimate goal of wanting to be able to run `db:migrate` from scratch are for better support of Oracle and MySQL being able to live in harmony since they end up with slightly different `schema.rb`.


db/seeds.rb
```diff
@@ -14,6 +14,10 @@
 
 Affiliate.find_or_create_by_name("Other")
 
+# TODO revisit in Rails 4. If you run `rake db:reset db:create db:migrate db:seed` as one step,
```
drench: TIL `rake notes` doesn't look for `TODO` comments under `db/`.

db/seeds.rb
```diff
@@ -22,3 +26,8 @@
   price_group.display_order = index + 1
   price_group.save(validate: false)
 end
+
+new_status = OrderStatus.new_os.first
+OrderDetailObserver.status_change_hooks.keys.each do |status|
+  OrderStatus.find_or_create_by_name_and_facility_id(name: status.to_s.titleize, facility_id: nil, parent: new_status)
```
drench: `TODO` convert from a dynamic method for Rails 4?

drench:
:+1: with one comment suggestion, though I'm sure we'll run into it when we look at `OrderDetailObserver` during the upgrade.

PR #603, opened 2016-05-18
[#116375253] Add Projects to global search

drench:
This changes `GlobalSearch` to make it more easily pluggable.

`GlobalSearchController.searcher_classes` is an `Array` of "searcher" classes that need to implement three methods:

* `sanitize_search_string`, which accepts a search string (the query) and transforms it as necessary (such as removing excess whitespace). This method should be private.
* `execute_search_query`, which should return an `Array` of objects found. The sanitized query string is available as `query`. The result will be memoized, so it's not necessary to memoize this method. This method should also be private.
* `template`, which returns the partial name that can render results, and must be public.

This adapts `OrderSearcher` and `StatementSearcher` to this interface, and injects `ProjectSearcher` via the `Projects` engine when enabled.

If the user is searching from the context of a single facility, project name searches will be scoped to that facility. In practice that means there will be at most 1 project returned in that scenario.
When searching in "all" facility or no facility context, project name searches are global and may return multiple results.

Project name search should be case-insensitive.


app/services/global_search/common.rb
```diff
@@ -0,0 +1,22 @@
+module GlobalSearch
+
+  module Common
```
drench: I'm open for suggestions for a better module name.

vendor/engines/projects/app/services/projects/global_search/project_searcher.rb
```diff
@@ -0,0 +1,37 @@
...
+      def query_object
+        if facility.try(:single_facility?)
+          facility.projects
+        else
+          Projects::Project
+        end
+      end
+
+      def execute_search_query
+        query_object.where("lower(name) = ?", query).select do |project|
```
drench: This is to make the query on Oracle (and Postgres, someday) case-insensitive. MySQL already is.

jhanggi: I agree it should be case-insensitive, but it should probably support partial matches as well.

vendor/engines/projects/app/controllers/projects/global_search_controller_extension.rb
```diff
@@ -0,0 +1,13 @@
+module Projects
+
+  module GlobalSearchControllerExtension
+
+    extend ActiveSupport::Concern
+
+    included do
+      searcher_classes.unshift(Projects::GlobalSearch::ProjectSearcher)
```
jhanggi: What's your reasoning for doing an `unshift` instead of `<<`? I'm inclined towards orders always being top priority, but could possible be swayed.

jhanggi:
My general feeling is that the required API for searchers is now a bit too complicated. I feel like they should only be _required_ to implement a single `search` method. If they want to clean the string they can, although I might put the basic `query.to_s.strip` into the `Common` class since that is standard across all the classes.

I could also see pulling the `restrict` methods up into `Common` as well, so `search` returns everything that matches and then the public `results` applies memoization and filtering.

app/services/global_search/order_searcher.rb
```diff
@@ -3,12 +3,15 @@ module GlobalSearch
   class OrderSearcher
 
     include NUCore::Database::CaseSensitivityHelper
+    include Common
 
-    def initialize(user = nil)
-      @user = user
+    def template
```
jhanggi: Part of me wonders if this method actually belongs in the class or if it should be the responsibility of the controller (i.e. we go back to using a hash instead of an array). It just feels weird to have what is essentially a view concern inside a service object..

app/services/global_search/common.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    extend ActiveSupport::Concern
+
+    included do
+      attr_reader :facility, :query, :user
+    end
+
+    def initialize(user = nil, facility = nil, query = nil)
```
jhanggi: Hmm, it feels a little strange to me to have an initializer in a concern. This is the kind of thing where I feel like inheritance might be better. If you named the class `GlobalSearch::Base`, it would work.

vendor/engines/projects/app/controllers/projects/global_search_controller_extension.rb
drench: It's arbitrary. It seemed like the case of a search query matching a project name **and** a statement or order number would be rare and ordering wouldn't matter.

But then if this turns back into a `Hash`, injections like this will always be appends.

app/services/global_search/common.rb
drench: Good call on inheritance. I've been in a concerns mindset with most of the recent work and didn't give that possibility a lot of consideration.

app/services/global_search/statement_searcher.rb
```diff
@@ -1,35 +1,31 @@
...
+      "statements"
     end
 
     private
 
-    def restrict(statement)
-      return unless statement
+    def search
+      return [] unless Account.config.statements_enabled? && query.present?
+      Array(Statement.find_by_invoice_number(query.sub(/\A#/, "")))
```
jhanggi: Can we keep the comment about why the `sub` is there?

vendor/engines/projects/app/services/projects/global_search/project_searcher.rb
```diff
@@ -0,0 +1,32 @@
...
+          facility.projects
+        else
+          Projects::Project
+        end
+      end
+
+      def search
+        return [] if query.blank?
+        query_object.where("lower(name) LIKE ?", "%#{query.downcase}%").select do |project|
+          Ability.new(user, project).can?(:show, project)
```
jhanggi: Isn't `Base` doing the restriction? Can we just return the query?

vendor/engines/projects/app/services/projects/global_search/project_searcher.rb
```diff
@@ -0,0 +1,32 @@
...
+      def query_object
+        if facility.try(:single_facility?)
+          facility.projects
+        else
+          Projects::Project
+        end
+      end
+
+      def search
+        return [] if query.blank?
```
jhanggi: Would it be worth pulling the `query.present?`/`blank?` up into `Base`?

jhanggi:
Nothing to major. I think we're just about there.

vendor/engines/projects/app/services/projects/global_search/project_searcher.rb
drench: Accidentally left in. Removing.

drench:
Updated

jhanggi:
👍

PR #604, opened 2016-05-18
[#116374977] Enable pagination of order_details for Project "show" views

drench:
The shared partial this uses is already set up for pagination, but it was not calling `paginate` on the collection of order details. This change should fix that.

The `order_details` are not ordered by `ordered_at`. Should they be?

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -94,6 +94,7 @@ class OrderDetail < ActiveRecord::Base
   ## TODO validate order status is global or a member of the product's facility
   ## TODO validate which fields can be edited for which states
 
+  scope :by_ordered_at, -> { joins(:order).order("orders.ordered_at") }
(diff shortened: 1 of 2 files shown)


jhanggi:
Yeah, I think ordering by `ordered_at` is a good idea. 👍  on what's here so far.

drench:
Updated to sort by `ordered_at`, descending.

app/models/order_detail.rb
```diff
@@ -94,6 +94,7 @@ def journal_or_statement_date
   ## TODO validate order status is global or a member of the product's facility
   ## TODO validate which fields can be edited for which states
 
+  scope :by_ordered_at, -> { joins(:order).order("orders.ordered_at DESC") }
```
jhanggi: I'm amazed we didn't have this scope before.

jhanggi:
👍

PR #606, opened 2016-05-18
[#119749091] Disable chosen dropdown for reconciled orders

jhanggi:
The form gets disabled by JS, so this makes sure that any chosen
selects get updated from their `select`’s attributes.


drench:
:+1:

PR #608, opened 2016-05-18
[#126252] Could not upload files when adding to existing order

jhanggi:
This was broken by #543

### DIFF
--- a/spec/controllers/order_details_controller_spec.rb
+++ b/spec/controllers/order_details_controller_spec.rb
@@ -346,4 +346,33 @@ RSpec.describe OrderDetailsController do
       end
     end
   end
+
+  describe "#order_file" do
+    let(:product) { create(:setup_service, :with_order_form) }
+    let(:order_detail) { order.order_details.first }
+    let(:facility) { order.facility }
+
+    let(:params) { { order_id: order.id, id: order_detail.id } }
+    before { sign_in user }
+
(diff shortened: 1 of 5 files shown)


spec/controllers/order_details_controller_spec.rb
```diff
@@ -346,4 +346,33 @@ def perform_update
...
+      it "has access" do
+        get :order_file, params
+        expect(response).to be_success
+      end
+    end
+
+    describe "adding to an existing order" do
+      let(:order) { create(:purchased_order, product: product) }
+      let(:user) { create(:user, :staff, facility: facility) }
+      let(:merge_order) { create(:setup_order, product: product, account: order.account, user: order.user, merge_with_order_id: order.id) }
```
drench: More of a `TODO`, but merge order setup is complex enough to become a `trait` of `:setup_order`.

drench:
:+1:, with the `:trait` suggestion if it's worth taking on now.

spec/controllers/order_details_controller_spec.rb
jhanggi: Updated. I went with making `merge_order` a full factory. I feel like `create(:merge_order)` is easier to understand than `create(:setup_order, :merge_order)`

drench:
:+1:

PR #609, opened 2016-05-18
[#119754369] Project dropdown had project if sister detail did

jhanggi:
If an order detail did not have a project, but the first detail in the
same order does, that project was appearing in the dropdown.

This also now renders on purchase with quantity changes rather than
redirects. This makes sure the project and the “more options” fields
don’t get lost.


drench:
:+1:

PR #613, opened 2016-05-21
[#119937533] Add total sum to project#show

jhanggi:


drench:
:+1: though I'd like to note a `TODO` on cleaning up that show partial at some point (I18n, hashrockets, maybe a presenter, etc.)

PR #616, opened 2016-05-23
[rails4] Fix affiliate spec

jhanggi:
This is kind of a weird one. We’re testing some affiliate code in the
base builder, but NufsAccount don’t support affiliates, so they’re
excluded from the `account_params_for_build` and get filtered out by
strong params. CreditCardAccountBuilder & PurchaseOrderAccountBuilder
include affiliate_id and affiliate_other.

Turns out the `using_affiliate? && account_params.key?(:affiliate_id)`
is somewhat redundant, but I like keeping it in there, but this keeps
the data clean.


drench:
Thanks for digging into this one. :+1:

PR #617, opened 2016-05-24
Use chosen-rails

jhanggi:
jQuery in Rails 4 breaks the old version of chosen we were using. This uses the gem version of chosen-rails. I ended up cherry-picking this from a branch I made off of `rails4` because it's not really dependent on rails 4.

It also fixes a minor bugs (un-reported):

* When choosing "Authorized Users" in bulk email, narrow the product list to only restricted products. This had been broken for a long time.


drench:
Nice to see all that code go away. :+1:

jhanggi:
@drench I'm pulling this into master. We definitely need this is the `rails4` branch as no JS functionality works on that branch right now because of an error within the old Chosen version.

PR #618, opened 2016-05-24
[rails4] Clean up deprecations

jhanggi:
`rails server` will now boot without deprecations, and this takes care of a deprecation on how we test if engines are loaded.


drench:
👍

PR #619, opened 2016-05-24
[rails4] Kill `remove_ugly_params` filter

jhanggi:
This was a “nice” feature I added when I first created the transaction
search which strips the hidden field parts a GET search to make URL query strings
look just a little nicer.

In Rails 4, the `redirect_to params` raises an "Cannot redirect to a parameter hash!"
error, which seems like a reasonable security feature. It seems the easiest way
to fix the error is to kill the feature altogether. I don’t think anyone will notice.


drench:
:+1:

PR #620, opened 2016-05-24
[#125690] Add facility to statement subject interpolations

jhanggi:
This is needed by UIC in one of their locale overrides.


drench:
:+1:

PR #622, opened 2016-05-24
[Rails4] AASM upgrade to 4.10

jhanggi:
This merges #621 into the Rails 4 branch *This also includes a few other updates from master*

The syntax is the obvious change, but it's pretty straightforward.

The bigger issue is that prior to AASM 3.0, `AASM::InvalidTransition` would only be raised if the transition did not exist. In 3.0+, guard failures also raise the error. So for example

```ruby
event :validate_order do
  transitions to: :validated, from: [:new, :validated], guard: :cart_valid?
end

# order.state == "purchased"
order.validate_order! # Raises an error in both versions

# order.cart_valid? == false, order.state == "new"
order.validate_order! # Returns false in 2.X, raises in 3.0+
```

`validate_order!` is the big place where this problematic since it gets called several places with no `rescue`. I chose to keep the change minimal by defining my own `validate_order!` method that calls the event/transition, but swallows the error.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -18,7 +18,7 @@ gem "devise_ldap_authenticatable", "~> 0.8.5"
 gem "cancancan",        "1.10"
 
 ## models
-gem "aasm",             "2.2.0"
+gem "aasm",             "~> 4.10.1"
 gem "paperclip",        "~> 4.2.0"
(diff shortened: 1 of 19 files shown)


drench:
This pulls in a bit more than `aasm` but the rails4 branch needs it. :+1:

PR #623, opened 2016-05-24
[rails4] Clean up deprecation notices from simple_form

jhanggi:
These methods now accept a `wrappers_options` argument.

https://github.com/plataformatec/simple_form/pull/997


drench:
:+1:

PR #624, opened 2016-05-25
[#116375253] Restrict Projects by Facility abilities

drench:
When determining permissions for the `show` view of a `Project`, it uses the project's `facility` as its base resource. When restricting global searches, it should use the same.

This means that of the 3 current global search types, only one, `OrderSearcher`, would be using the `restrict` method from the base class. It still felt general enough to belong there so I left it.

### DIFF
--- a/vendor/engines/projects/app/services/projects/global_search/project_searcher.rb
+++ b/vendor/engines/projects/app/services/projects/global_search/project_searcher.rb
@@ -18,6 +18,12 @@ module Projects
         end
       end
 
+      def restrict(projects)
+        projects.select do |project|
+          Ability.new(user, project.facility).can?(:show, project)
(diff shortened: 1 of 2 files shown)


jhanggi:
I added specs around this. My ideal change would be to make the `Ability` use `Project` as the resource, similar to how  `OrderDetail`s work. That's how I feel CanCan _should_ be used, but that's a bigger issue and would probably have larger repercussions than I'd like to deal with right now. If you're good with my spec, feel free to merge. 👍 

If in the future we add more searchers, I'd like to see us wrap up the `@searchers` into a cleaner interface. Testing with `assigns[:searchers].find { |s| s.template == "projects" }.results` feels really ugly.

PR #625, opened 2016-05-25
Initial Sanger Sequencing Setup

jhanggi:
* Set up the engine
* Landing page for coming from the cart
* Add the `/submissions/new` URL as an "Online Order Form" for a service
* "Sanger" navigation tab and landing page for facility staff
* Pass quantity from cart to external service [#119570551]
* Bring in Capybara and use for integration tests of the engine

This is pretty much just the bootstrap, so it doesn't include things like permissions or serious I18n since a lot of that will change shortly. I think we'll also want a per-facility flag for enabling it.

One thing I know about: while in the cart, if you change the quantity but don't click "Update" and click out to the order form, it passes the original quantity. We'll need a little JS to fix that.

We'll want to make sure to disable this engine downstream until we're ready.


drench:
LGTM. Happy to see capybara added to the project. :+1:

PR #626, opened 2016-05-25
[#125075] Use correct format for min-date on reconciliation date widget

jhanggi:
Chrome can parse `Date#to_s` correctly, but Firefox does not. They both handle iso8601.


drench:
:+1:

PR #627, opened 2016-05-26
Upgrade jquery-ui

jhanggi:
jquery-ui-sass-rails is no longer maintained. Switch to using
jquery-ui-rails instead.

jquery-ui-rails uses sprockets `require` syntax, and SASS is not able
to do the dependent requires when called with `@import` so we need to
explicitly require each individual component that we need.

I went through the whole list of components, and I believe these are
the only ones we need.

This also pulls the latest from master which includes the chosen update because without it, JS doesn't work at all.


drench:
Pulling this in as the date parsing CI failures are unrelated to this change and we're working on them separately.

PR #628, opened 2016-05-26
Date parsing updates

drench:
I extracted this from the `rails4` branch for potential discussion. We don't necessarily need to merge it to `master` separately but it should work if we do.

The issue was in replacing the deprecated `to_time_in_current_zone` call in `date_helper` with `in_time_zone` as the deprecation notice suggests:

```
[1] pry(main)> DateTime.strptime("5/26/2016", "%m/%d/%Y").to_time_in_current_zone
=> Thu, 26 May 2016 00:00:00 CDT -05:00
[2] pry(main)> DateTime.strptime("5/26/2016", "%m/%d/%Y").in_time_zone
=> Wed, 25 May 2016 19:00:00 CDT -05:00
```

When importing dates, we want the time to be beginning-of-day in the current zone, so I went with this instead:

```
[3] pry(main)> DateTime.strptime("5/26/2016", "%m/%d/%Y").to_date.beginning_of_day
=> Thu, 26 May 2016 00:00:00 CDT -05:00
```

This works with Rails 3.2 and 4.0, with no deprecation notices.

### DIFF
--- a/spec/helpers/date_helper_spec.rb
+++ b/spec/helpers/date_helper_spec.rb
@@ -14,8 +14,7 @@ RSpec.describe DateHelper do
     context "with valid dates" do
       %w(1/1/2014 3/31/2014 10/2/2014 12/31/2014).each do |date_string|
         it "considers '#{date_string}' a valid date" do
-          expect(parse_usa_import_date(date_string))
-            .to eq DateTime.strptime(date_string, "%m/%d/%Y").to_time_in_current_zone
+          expect(parse_usa_import_date(date_string)).to be_present
         end
       end
 
--- a/spec/helpers/date_helper_spec.rb
+++ b/spec/helpers/date_helper_spec.rb
@@ -18,9 +18,14 @@ RSpec.describe DateHelper do
         end
       end
 
-      it "reads dates in USA-style MM/DD/YYYY format" do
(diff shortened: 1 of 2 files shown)


spec/helpers/date_helper_spec.rb
```diff
@@ -14,14 +14,18 @@
     context "with valid dates" do
       %w(1/1/2014 3/31/2014 10/2/2014 12/31/2014).each do |date_string|
         it "considers '#{date_string}' a valid date" do
-          expect(parse_usa_import_date(date_string))
-            .to eq DateTime.strptime(date_string, "%m/%d/%Y").to_time_in_current_zone
+          expect(parse_usa_import_date(date_string)).to be_present
```
drench: The old test was essentially doing the same parsing as the method it was testing. I've added tests below to check that the day, month, and years are what we expect.

jhanggi:
I'm fine with the change. It might be nice to have a test that's something like:

```ruby
Time.use_zone("Eastern Time (US & Canada)") do
  expect(parse_usa_import_date("05/01/2014").to_s).to eq("Sun, 01 May 2016 00:00:00 EDT -05:00")
end
```
(We probably have this check, but we definitely want to make sure this gets parsed as May 1, not Jan 5).

drench:
I've added the Eastern time spec to #647.

PR #629, opened 2016-05-26
[#119570591] Pre-populate Sanger Customer Sample IDs

jhanggi:
This extends from #625, which is where most of the commits come from. Only the last two are specific to this PR.

### DIFF
--- a/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/sample.rb
+++ b/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/sample.rb
@@ -4,6 +4,20 @@ module SangerSequencing
 
     self.table_name = "sanger_sequencing_samples"
     belongs_to :submission
+    # customer_sample_id is based off of the ID, which we don't have until
+    # after the initial persistence.
+    after_create { customer_sample_id && save }
+
+    def customer_sample_id
+      self[:customer_sample_id] ||= default_customer_sample_id
+    end
+
+    private
+
+    def default_customer_sample_id
+      # last four digits of the id as a zero-padded string
+      id.to_s.last(4).rjust(4, "0")
+    end
 
   end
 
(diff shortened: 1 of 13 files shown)


jhanggi:
Rebased off of master now that #625 is in.

vendor/engines/sanger_sequencing/app/models/sanger_sequencing/sample.rb
```diff
@@ -4,6 +4,20 @@ class Sample < ActiveRecord::Base
...
+
+    def customer_sample_id
+      self[:customer_sample_id] ||= default_customer_sample_id
+    end
+
+    private
+
+    def default_customer_sample_id
+      # last four digits of the id as a zero-padded string
+      id.to_s.last(4).rjust(4, "0")
```
drench: I don't know if it's much more clear but I would have gone with `format("%04d", id).last(4)`, presuming `id` won't never be `nil` here.

vendor/engines/sanger_sequencing/spec/features/purchasing_spec.rb
```diff
@@ -1,24 +1,41 @@
...
+      click_link "Complete Online Order Form"
+    end
+
+    it "sets up the right number of text boxes" do
+      expect(page).to have_css(customer_id_selector, count: 5)
+    end
+
+    it "has prefilled values in the text boxes with unique four digit numbers" do
+      values = page.all(customer_id_selector).map(&:value)
+      expect(values).to all(match(/\d{4}/))
```
drench: `match(/\A\d{4}\z/)` to be sure they're all four, and only four digits?

drench:
A couple comments but nothing big.

vendor/engines/sanger_sequencing/app/models/sanger_sequencing/sample.rb
jhanggi: Yeah, I like that. I always forget about `format`. I'm fine erroring if `id` is nil, as that would represent an actual bug.

jhanggi:
Feedback addressed.

drench:
:+1:

PR #630, opened 2016-05-27
Enable features in specs where necessary

jhanggi:
These are testing features that can be disabled with a feature flag which removes the routes.

I haven't found any documentation around this, but it seems like Rails 3 controller specs didn't check the routes, it just called methods directly on the controller. Now Rails 4 does check. This should fix some of the tests in UIC.


drench:
:+1:

PR #631, opened 2016-05-27
Upgrade to Rails 4.0

drench:
This appears 'clean', meaning all specs pass and there are no deprecation warnings, with one exception: we've marked the "does not allow overriding the account_type" test in `FacilityAccountsReconciliationController` as `pending` because it appears to only be a problem in the spec. We will revisit that with Rails 4.1.


app/models/order_status.rb
```diff
@@ -15,11 +15,12 @@ class OrderStatus < ActiveRecord::Base
...
-  scope :new_os, conditions: { name: "New" }, limit: 1
-  scope :inprocess, conditions: { name: "In Process" }, limit: 1
-  scope :canceled, conditions: { name: "Canceled" }, limit: 1
-  scope :complete, conditions: { name: "Complete" }, limit: 1
-  scope :reconciled, conditions: { name: "Reconciled" }, limit: 1
+  scope :for_facility, -> (facility) { where(facility_id: [nil, facility.id]).order(:lft) }
+  scope :new_os, -> { where(name: "New").limit(1) }
+  scope :inprocess, -> { where(name: "In Process").limit(1) }
+  scope :canceled, -> { where(name: "Canceled").limit(1) }
+  scope :complete, -> { where(name: "Complete").limit(1) }
```
jhanggi: One day I want to kill these scopes and replace them with the `_status`/`first` versions below.

app/models/price_policy.rb
```diff
@@ -53,13 +53,25 @@ def self.for_price_groups(price_groups)
   end
 
   def self.current_date(product)
-    ipp = product.price_policies.find(:first, conditions: [dateize("start_date", " <= ? AND ") + dateize("expire_date", " > ?"), Time.zone.now, Time.zone.now], order: "start_date DESC")
```
jhanggi: Are there any usages of `dateize` left? Can we get rid of it?

app/models/product.rb
```diff
@@ -38,10 +38,10 @@ class Product < ActiveRecord::Base
     end
   end
 
-  scope :active,             conditions: { is_archived: false, is_hidden: false }
-  scope :active_plus_hidden, conditions: { is_archived: false }
-  scope :archived,           conditions: { is_archived: true }
-  scope :not_archived,       conditions: { is_archived: false }
+  scope :active, -> { where(is_archived: false, is_hidden: false) }
+  scope :active_plus_hidden, -> { where(is_archived: false) } # TODO: is this scope named correctly?
```
jhanggi: There's probably a better name for it, but `active` and `hidden` are two separate concepts. I think `active` may actually be the incorrectly named one.

app/models/reservation.rb
```diff
@@ -95,9 +95,11 @@ def self.in_range(start_time, end_time)
...
     # If this is a named scope differences emerge between Oracle & MySQL on #reserve_end_at querying.
     # Eliminate by letting Rails filter by #reserve_end_at
-    reservations = find(:all, conditions: "reservations.canceled_at IS NULL AND (orders.state = 'purchased' OR orders.state IS NULL)", order: "reserve_end_at asc", joins: ["LEFT JOIN order_details ON order_details.id = reservations.order_detail_id", "LEFT JOIN orders ON orders.id = order_details.order_id"])
-    reservations.delete_if { |r| r.reserve_end_at < t }
-    reservations
+    joins("LEFT JOIN order_details ON order_details.id = reservations.order_detail_id")
+      .joins("LEFT JOIN orders ON orders.id = order_details.order_id")
+      .where(canceled_at: nil, "orders.state" => [nil, "purchased"])
+      .order(reserve_end_at: :asc)
+      .delete_if { |reservation| reservation.reserve_end_at < t }
```
jhanggi: I wonder if we could replace this with `where("reserve_end_at > ?", t)`.

config/environments/production.rb
```diff
@@ -5,6 +5,12 @@
   # Code is not reloaded between requests
   config.cache_classes = true
 
+  # Eager load code on boot. This eager loads most of Rails and
+  # your application in memory, allowing both threaded web servers
+  # and those relying on copy on write to perform better.
+  # Rake tasks automatically ignore this option for performance.
+  config.eager_load = true
```
jhanggi: Let's make sure to remember to add this to `staging` in the downstream forks.

config/routes.rb
```diff
@@ -338,20 +346,20 @@
...
   get   "/#{I18n.t("facilities_downcase")}/:facility_id/:product/:product_id/files/product_survey",                           to: 'file_uploads#product_survey',        as: "product_survey"
   post  "/#{I18n.t("facilities_downcase")}/:facility_id/:product/:product_id/files/create_product_survey",                    to: 'file_uploads#create_product_survey', as: "create_product_survey"
   put   "/#{I18n.t("facilities_downcase")}/:facility_id/services/:service_id/surveys/:external_service_passer_id/activate",   to: 'surveys#activate',                 as: "activate_survey"
   put   "/#{I18n.t("facilities_downcase")}/:facility_id/services/:service_id/surveys/:external_service_passer_id/deactivate", to: 'surveys#deactivate',               as: "deactivate_survey"
-  match "/#{I18n.t("facilities_downcase")}/:facility_id/services/:service_id/surveys/:external_service_id/complete",          to: 'surveys#complete',                 as: "complete_survey", via: [:get, :post]
+  get "/#{I18n.t("facilities_downcase")}/:facility_id/services/:service_id/surveys/:external_service_id/complete",          to: 'surveys#complete',                 as: "complete_survey", via: [:get, :post]
```
jhanggi: This one still has the `via` option.

config/routes.rb
```diff
@@ -295,20 +298,25 @@
...
-      match "purchase",       via: [:get, :put]
-      match "choose_account", via: [:get, :post]
-      match "update_or_purchase", via: [:get, :put]
-      get   "receipt"
-      put   "clear"
+      get "add"
+      put "add"
+      get "purchase"
+      put "purchase"
+      get "choose_account"
```
jhanggi: Are you allowed to do `match "choose_account", via: [:get, :post]`?

spec/controllers/bundle_products_controller_spec.rb
```diff
@@ -27,7 +27,7 @@
     it_should_require_login
 
     it_should_allow_all facility_operators do
-      expect(assigns(:bundle_products)).to be_kind_of Array
+      expect(assigns(:bundle_products)).to all be_kind_of(BundleProduct)
```
jhanggi: I like this change.

jhanggi:
Looks good. A couple comments, but not really much that needs to change now. 👍

app/models/price_policy.rb
drench: I don't see any. I'm removing it.

app/models/reservation.rb
drench: I'd like to. I didn't dig up a lot in history about why it's necessary I've left it alone for now.

jhanggi: Yeah, don't need to deal with it now. Just wanted to note it.

PR #632, opened 2016-05-31
DO NOT MERGE: Reduce spec to one failing case

drench:
I'm opening this PR to debug, not merge.

I rewrote this to isolate one case: transitioning from `new` to `inprocess` when an `OrderDetail` has an accessory (child `OrderDetail`). The existing tests did some stubbing, but this version attempts to create real objects persisted to disk. From the web app, these transitions seem to work as expected, but this spec fails in Rails 3.2 and Rails 4.1.

### DIFF
--- a/spec/services/accessories_child_updater_spec.rb
+++ b/spec/services/accessories_child_updater_spec.rb
@@ -1,77 +1,29 @@
 require "rails_helper"
 
 RSpec.describe Accessories::ChildUpdater do
-  let(:order) { Order.new }
-  let(:order_detail) do
-    build_stubbed :order_detail,
-                  order: order,
-                  child_order_details: [],
-                  attributes: {
-                    order_id: order.id,
-                  }
+  let!(:child_order_detail) do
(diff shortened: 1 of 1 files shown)


jhanggi:
The problem is that the child_order_details don't have an initial status in the tests. I'm not sure what the best way to fix this is.

The options include:
* Change the `Accessories::ChildUpdater#status_changed?` to something like `@order_detail.order_status_id_changed? && (@order_detail.order_status_id_was == od.order_status_id || od.order_status_id.blank?)`.
* Add `order_status { OrderStatus.new_status }` to the order detail factory (I don't like this because it could have larger performance impact).
* In the `before`, say `child_order_detail.set_default_status!`
* Add `order_status: OrderStatus.new_status` to `child_order_detail` attributes on creation

I think the last one is my preference as it has the smallest footprint.

drench:
Setting the initial order status did it. I've adapted this change to #638 and there's no pressing need to merge it into `master` so I'm closing this.

PR #633, opened 2016-05-31
[#119570693] Sanger sequencing edit form

jhanggi:
This is the basic edit functionality. This does not have the JS add/remove functionality yet.


jhanggi:
Rebased now that #629 is in. I've since made changes to what's happening here to deal with add/remove, but I think this keeps the changeset small.

drench:
:+1:

PR #635, opened 2016-06-01
Upgrade to Rails 4.2.6

jhanggi:


jhanggi:
We should run a rubocop over everything again at some point, maybe once we're done with the 4.2 upgrade.

PR #636, opened 2016-06-02
[#126377] Fix slow page load on products index

jhanggi:
In #565, we refactored to share Products#index views. As part of this,
we got rid of the need for the specific `@instruments`, `@services`,
etc in the views. CanCan was still loading these variables, but without
a facility restriction.

There was some vestigial code in the instruments controller that loaded
from when we used to include a lot more information like upcoming
reservations and relay status. See
d4579e6b67419ff51a052a82bb5c93ed0626ead5. I believe this was removed
when we added the timeline/daily view.

This was now loading upcoming reservation status for _every_ instrument
in NUcore, not just the ones for the facility.

### DIFF
--- a/app/controllers/instruments_controller.rb
+++ b/app/controllers/instruments_controller.rb
@@ -12,14 +12,6 @@ class InstrumentsController < ProductsCommonController
 
   skip_before_filter :init_product, only: [:instrument_statuses]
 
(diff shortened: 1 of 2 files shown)


drench:
:+1: when Circle finishes

PR #637, opened 2016-06-02
[#119570633] Sanger sequencing Add/Remove samples via JS

jhanggi:
This adds Poltergeist for Capybara/JS testing.

One behavior that might differ than we expect, and the way it behaved in the previous PR: Say you have a form with 10 samples. Leave number 3, 9, and 10 blank. Click submit. 3 will be highlighted as "can't be blank", but 9 & 10 will be gone. That's because we mark them for destruction, which allows them to be invalid since they'll go away on save. `nested_fields_for` hides anything marked for destruction. I tried several things, but couldn't come up with a solution that allowed that to work in conjunction with pulling down new IDs via ajax. I eventually gave up as it didn't seem worth the effort for a minor and likely rarely seen UI enhancement.


drench:
Works for me. I had a UI problem with the "Add" button on this form but it's not worth addressing given the overhaul in #642 being out there. :+1:

PR #638, opened 2016-06-02
Upgrade to Rails 4.1.15

drench:
Highlights:
* Product sorting by name with AR instead of `Array#sort`. There's still a `<=>` method in the `Product` model though it may not need to exist anymore.
* Some code relies on methods to return `Array` but some of these became AR relations. This converts them to `Array` in those spots to keep the changes minimal, but these are refactor opportunities.
* Some tests started failing because changes made through `update_attributes` were no longer automatically bubbling up to their existing objects; calling `reload` fixed them.
* Dynamic `find_or_create_by_*` is deprecated and so converted.


spec/models/reservation_spec.rb
```diff
@@ -529,6 +529,7 @@
         before do
           ScheduleRule.destroy_all
           @instrument.update_attributes(requires_approval: true)
+          @instrument.reload
```
jhanggi: I think this might be better off as `@instrument.schedule_rules.reload` so it's a little more explicit about why we need to do it (and move it up a line). It's not actually the `update_attributes` that's causing the problem, it's that we're destroying ScheduleRules and  not telling `@instrument` about it and it already has the association loaded. Not sure why this was working before (maybe update_attributes used to do a `reload` implicitly?)

spec/models/reservation_spec.rb
```diff
@@ -921,6 +922,7 @@
       before :each do
         @rule.destroy
         @instrument.update_attribute :reserve_interval, 15
+        @instrument.reload
```
jhanggi: Same with the rest of these, although, it could also be better to say `@instrument.schedule_rules.destroy_all` instead of `@rule.destroy`.

jhanggi:
Looks good. I just had a comment about the `reload`s in the reservation_spec.

PR #639, opened 2016-06-02
[#119570693] Disable quantity editing when survey controls quantity

jhanggi:
If a quantity is returned by the outside application, then disable the quantity box. This will also affect IMSERC, but this appears to be the correct behavior there as well.

### DIFF
--- a/spec/models/external_service/survey_response_spec.rb
+++ b/spec/models/external_service/survey_response_spec.rb
@@ -83,6 +83,24 @@ RSpec.describe SurveyResponse do
       receiver = survey_response.save!
       expect(receiver.receiver.quantity).to eq quantity
     end
+
+    it "locks the quantity for order if there is a quantity param" do
+      params[:quantity] = order_detail.quantity
+      receiver = survey_response.save!
+      expect(order_detail.reload).to be_quantity_locked_by_survey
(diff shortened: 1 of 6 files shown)


spec/models/external_service/survey_response_spec.rb
```diff
@@ -83,6 +83,24 @@
       receiver = survey_response.save!
       expect(receiver.receiver.quantity).to eq quantity
     end
+
+    it "locks the quantity for order if there is a quantity param" do
+      params[:quantity] = order_detail.quantity
+      receiver = survey_response.save!
+      expect(order_detail.reload).to be_quantity_locked_by_survey
```
drench: Minor, but I'd probably have written these as `expect { survey_response.save! }.to change…` just to be sure the bit flipped. Also, is there any reason to assign `receiver`?

jhanggi: Not really any reason to assign `receiver`. I think that's just a holdover from copying and pasting some setup. I'll change that as well as to using the `expect {}.to change`. I like that better as well.

jhanggi:
Updated.

drench:
:+1: after CI finishes

PR #640, opened 2016-06-02
[#119570551] Pass quantity to Sanger even without clicking "Update"

jhanggi:
Only the last commit is relevant here, the rest of the commits are from #637, primarily for the poltergeist integration.

### DIFF
--- a/app/assets/javascripts/app/cart.coffee
+++ b/app/assets/javascripts/app/cart.coffee
@@ -1,3 +1,9 @@
+class window.CartQuantityReplacer
+  constructor: (@originalPath, @newQuantity) ->
+
+  toString: ->
+    @originalPath.replace(/\bquantity=\d+(&|$)/, "quantity=#{@newQuantity}$1")
+
 $ ->
   class Cart
     constructor: (cartSelector) ->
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/app/cart.coffee
```diff
@@ -0,0 +1,13 @@
...
+  class Cart
+    constructor: (cartSelector) ->
+      @$cart = $(cartSelector)
+      @$cart.find("[data-quantity-field]").each @setupQuantityListeners
+
+    setupQuantityListeners: (i, link) ->
+      trackedField = $(link).data("quantity-field")
+      $("##{trackedField}").change ->
+        newQuantity = $(@).val()
+        link.href = link.href.replace(/quantity=\d+&/, "quantity=#{newQuantity}&")
```
drench: Wondering if it would be a bit safer to do something like:

```javascript
link.href.replace(/\bquantity=\d+(&|$)/, "quantity=#{newQuantity}$1");
```

That should be able to handle `quantity` being the last param, as well as avoid accidentally replacing the value of another param that happens to end with "quantity".

drench:
:+1: aside from the regexp suggestion

jhanggi:
That's a good suggestion. I'm working on writing a couple jasmine tests for the regex replacement.

jhanggi:
Updated and added a couple jasmine specs.

drench:
:+1:

PR #641, opened 2016-06-02
[#119570737] A user can only edit their own submissions

jhanggi:
Extends from #637. Only the last commit is relevant.

I decided to abandon the `SangerSequencing::OrderDetail` wrapper concept. I was originally going for super-loose coupling with the NUcore app, but given the low likelihood of ever wanting to extract this into its own stand-alone app, I gave up on that.

This also uses a different technique for Ability, where we have a custom ability for the engine instead of extending the base Ability. That might need to change when I get over into the admin side.


drench:
It would be great if the `Ability` technique could survive. :+1:

PR #642, opened 2016-06-02
Improve Sanger Sequencing form design

jhanggi:
Extends from #641. Only the last commit is relevant.

* Tightens up the form
* Adds breadcrumbs
* Maintains the My Orders, My Reservations tabs while in the form

### DIFF
--- a/app/assets/stylesheets/app/tables.scss
+++ b/app/assets/stylesheets/app/tables.scss
@@ -1,0 +1,8 @@
+.table-tight {
+  width: inherit;
+
+  input,
+  .control-group {
+    margin-bottom: 0;
+  }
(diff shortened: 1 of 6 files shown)


drench:
The one issue I have: when clicking "Add", the "Remove" link for the newly-inserted row is right under my mouse cursor. It's easy to accidentally click "Remove" on it.

jhanggi:
I moved the "Add" button over to the left side. Now your mouse will be over the newly created text input, which I think is a decent place to be. I initially thought that put it too close to the "Submit" button, but I think it's fine. How does that work for you?

drench:
Works for me :+1:

PR #643, opened 2016-06-03
[#119570781] Add note to cart form for users

jhanggi:
* Refactor cart form
* Show note on order receipt
* Clean up unnecessary method on InstrumentsController (ProductsCommon could handle it)
* Move note field into same cart row. This enhances the visual connection between
  the note and the order detail. It was getting a little confusing when you have
  bundles where one item has a note field, and another doesn't.

PR #644, opened 2016-06-03
Upgrade to Rails 4.2.6

drench:


app/views/orders/_cart_row.html.haml
```diff
@@ -1,6 +1,6 @@
 %tr
   - if first_of_group
-    - row_span = @order.order_details.find(:all, :conditions => {:group_id => order_detail.group_id}).length
+    - row_span = @order.order_details.where(group_id: order_detail.group_id).count
```
drench: …with the understanding that this partial gets an overhaul in #643

db/migrate/20160602153918_increase_decimal_precision.rb
```diff
@@ -0,0 +1,11 @@
+class IncreaseDecimalPrecision < ActiveRecord::Migration
+  def up
+    change_column :price_policies, :overage_rate, :decimal, precision: 12, scale: 4
+    change_column :price_policies, :reservation_rate, :decimal, precision: 12, scale: 4
```
drench: `usage_rate` is already at 12/4, and we had specs with rates like 0.002777777777777783333333333, which translate to zero if the field is 10/2.

app/models/payment.rb
```diff
@@ -9,7 +9,7 @@ def self.valid_sources
     @@valid_sources ||= [:check]
   end
 
-  validates :source, presence: true, inclusion: { in: valid_sources }
+  validates :source, presence: true, inclusion: { in: valid_sources.map(&:to_s) }
   validates :account, presence: true
   # TODO: use `numericality: { other_than: 0 }` in newer rails
```
jhanggi: Can we update this TODO?

config/application.rb
```diff
@@ -74,6 +74,8 @@ class Application < Rails::Application
       g.test_framework :rspec
     end
 
+    # TODO: this appears to be necessary to suppress deprecations from fog:
+    config.active_record.raise_in_transactional_callbacks = true
```
jhanggi: This is actually something that should be added as this is opt-in new behavior that will become the default later. See "3.3 Error handling in transaction callbacks" in the upgrade guide. http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html

db/migrate/20160602153918_increase_decimal_precision.rb
jhanggi: Cool. This increase in precision makes perfect. Seems like it should have been this before.

db/schema.rb
```diff
@@ -9,714 +9,696 @@
...
+ActiveRecord::Schema.define(version: 20160602153918) do
 
-  create_table "account_users", :force => true do |t|
-    t.integer  "account_id",               :null => false
-    t.integer  "user_id",                  :null => false
-    t.string   "user_role",  :limit => 50, :null => false
-    t.datetime "created_at",               :null => false
-    t.integer  "created_by",               :null => false
+  create_table "account_users", force: :cascade do |t|
+    t.integer  "account_id", limit: 4,  null: false
```
jhanggi: These additions of `limit` happen just from a straight schema:dump? Just curious.

spec/models/instrument_price_policy/matrix_spec.rb
```diff
@@ -62,24 +62,24 @@
       describe "overage" do
         include_context "overage"
         it "calculates cost and subsidy" do
-          expect(subject[:cost].round 4).to eq overage[:cost]
-          expect(subject[:subsidy].round 4).to eq overage[:subsidy]
+          expect(subject[:cost].round 2).to eq overage[:cost]
```
jhanggi: Do we need these `round`s anymore?

jhanggi:
Are emails that use DelayedJob's `Notifier.delay.order_receipt` style still sent? Should we update those to `Notifier.order_receipt.deliver_later`?

spec/models/instrument_price_policy/matrix_spec.rb
drench: Rounding to 4, it would produce results like `expected: 20, got: 19.998`.

db/schema.rb
drench: Yes. `bundle exec rake db:schema:dump` gives me the same.

drench:
> Are emails that use DelayedJob's Notifier.delay.order_receipt style still sent?

Delayed job emails look like they're getting queued up the same as before, though so far I've only tested delivery via letter opener.

jhanggi:
I'd like to see us switch to `deliver_later`, but I'm comfortable putting that in a later PR. Remove the #todo on `raise_in_transactional_callbacks`, then I think we're good there.

PR #645, opened 2016-06-03
Fix Sanger migration for Oracle

jhanggi:
Oracle doesn't support ON UPDATE CASCADE. We don't really need it for MySQL either since our primary keys will never change, so it's easier to just get rid of it.

This has not been released yet, so it shouldn't cause any problems, although devs
should roll back and re-run it to make sure their schema.rb is in sync.


drench:
:+1:

PR #648, opened 2016-06-06
[#120813297] Enable Sanger tab per-facility

jhanggi:
I chose to not add a way to do this in the UI, since the vast majority
of facilities will not need it. We can turn it on for facilities as
needed.


jhanggi:
Rebased off of latest master.

drench:
:+1:

PR #650, opened 2016-06-07
Fix Rails 4.2 Upgrade Issues

jhanggi:
This fixes a couple issues, primarily around POST/PUTs changing to GETs in the ajax searches.

It also begins to add some fixtures, which I would like to continue to use in feature specs. I'll have a few more fixtures as I track down the reservation creation stack overflow.

### DIFF
--- a/spec/fixtures/order_statuses.rb
+++ b/spec/fixtures/order_statuses.rb
@@ -1,0 +1,2 @@
+new_order_status:
+  name: New
--- a/app/assets/javascripts/_common.js
+++ b/app/assets/javascripts/_common.js
@@ -15,7 +15,7 @@ $(document).ready(function() {
     submit.val("Please Wait...");
     submit.attr("disabled", true);
     jQuery.ajax({
-      type: "post",
+      type: "get",
       data: form.serialize(),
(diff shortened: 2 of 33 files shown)


spec/fixtures/order_statuses.rb
```diff
@@ -0,0 +1,2 @@
+new_order_status:
+  name: New
```
drench: Code Climate flagged this. Guessing it was meant to be `.yml` but is it being used?

drench:
One question on a fixture but otherwise LGTM. Looking forward to more feature specs in this project.

For the other 2 RuboCop/CC complaints in `app/controllers/account_price_group_members_controller.rb` I'd run an autocorrect.

spec/fixtures/order_statuses.rb
jhanggi: Shouldn't be used. Removed it and we should see green specs.

drench:
:+1: with green CI.

jhanggi:
I also had a mistake in the feature filename. Didn't end in `_spec.rb`, so it wouldn't have run... dangerous.

PR #652, opened 2016-06-07
[#126567] Fix infinite loop while ordering reservation on behalf of

jhanggi:
The callbacks when ordering on behalf of (specifically with a note) was
causing an infinite loop in Rails 4. We weren’t catching it before
because we were never submitting a note in the tests.


drench:
I'm not really concerned with the unescaped warnings from CC. :+1:

PR #653, opened 2016-06-07
[#126572] Upgrade to qtip2

jhanggi:
Solves the `Cannot read property 'msie' of undefined` which then causes
the calendar not to load when “Show user information” is enabled for an
instrument


drench:
:+1: though we should configure csslint to ignore everything under `vendor/assets/stylesheets/`

PR #654, opened 2016-06-08
Fix timezone spec

jhanggi:
The `Time.use_zone` needs to be within the `it` block in order to work. The test
was failing in DC where the app is in Eastern time.


drench:
:+1:

PR #655, opened 2016-06-08
Use locale for "Facility" when testing adding staff

jhanggi:


drench:
The code climate complaint feels trivial. :+1:

PR #656, opened 2016-06-08
Remove fixtures

jhanggi:
Fixtures were causing lots of problems with test ordering. They don't play well with the `before(:all)` in rails_helper where we `find_or_create` order statuses and price groups and other tests rely on an otherwise empty database. I tried various permutations of DatabaseCleaner, but it's messy and doesn't seem worth the effort right now.


spec/features/purchasing_a_reservation_on_behalf_of_spec.rb
```diff
@@ -1,17 +1,18 @@
...
-  fixtures :all
 
-  let(:facility) { facilities(:facility) }
-  let(:instrument) { products(:reservation_only_instrument) }
-  let(:facility_admin) { users(:facility_admin) }
-  let(:user) { users(:normal_user) }
+  let!(:instrument) { FactoryGirl.create(:setup_instrument) }
+  let!(:facility) { instrument.facility }
+  let!(:account) { FactoryGirl.create(:nufs_account, :with_account_owner, owner: user) }
+  let!(:price_policy) { FactoryGirl.create(:instrument_price_policy, price_group: PriceGroup.base.first, product: instrument) }
```
jhanggi: The price group might need to change for DC, or likely, we'll need to add the user or account explicitly into a price group (UIC doesn't have user-based price groups).

drench:
:+1: though I'd like to catch up with you on the recent test suite updates soon.

PR #657, opened 2016-06-08
Spec updates via the DC fork

drench:
These are cherry-picked from https://github.com/tablexi/nucore-dartmouth/pull/50

* bd6183f and 880b56f set up `AccountPriceGroupMember`s in feature specs.
* 7db8273 replaces an unreliable regular expression with Nokogiri to test for HTML attribute values


jhanggi:
👍

PR #658, opened 2016-06-08
[#119570851] Sanger admin index view

jhanggi:
Also a basic submission show view. This will likely get cleaned up as we get further along.

### DIFF
--- a/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/ability.rb
+++ b/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/ability.rb
@@ -4,8 +4,14 @@ module SangerSequencing
 
     include CanCan::Ability
 
-    def initialize(user)
+    def initialize(user, facility = nil)
+      return unless user
+
       can [:show, :create, :update, :fetch_ids], Submission, user: user
+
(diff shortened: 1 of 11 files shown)


vendor/engines/sanger_sequencing/app/models/sanger_sequencing/ability.rb
```diff
@@ -4,8 +4,14 @@ class Ability
 
     include CanCan::Ability
 
-    def initialize(user)
+    def initialize(user, facility = nil)
+      return unless user
+
       can [:show, :create, :update, :fetch_ids], Submission, user: user
+
+      if facility && user.operator_of?(facility)
```
drench: Do you want to check `facility.sanger_sequencing_enabled?` as well?

jhanggi: That's a good point. I'm not sure if the ability is the right place to do that. Part of that thought is if it's not enabled, I feel 404 is more appropriate than 403. I'm thinking of doing it in a `before_action`. I'll add that in another PR.

drench:
:+1:

PR #660, opened 2016-06-09
Raise 404 if the service does not have sanger enabled

jhanggi:


vendor/engines/sanger_sequencing/app/controllers/sanger_sequencing/base_controller.rb
```diff
@@ -0,0 +1,15 @@
+module SangerSequencing
+
+  class BaseController < ApplicationController
+
+    before_action :assert_sanger_enabled_for_facility
+
+    private
+
+    def assert_sanger_enabled_for_facility
+      raise ActionController::RoutingError, "Sanger not enabled for this facility" unless current_facility.sanger_sequencing_enabled?
```
drench: `current_facility` might be `nil`. Implementing some kind of a `NullFacility` object is on my wish list.

jhanggi: Yeah, the NullFacility would be nice, and I was looking into it, but it still seems like a lot of effort right now (the big thing that's weird to me is the `facility_id = params[:facility_id] || params[:id]` and then not 404ing if it doesn't find the `url_name`. Ideally, we would have a NullFacility if there is no facility_id requested, but 404 if the facility_id/url_name is not found.

I adjusted this for now so it does a `try`.

drench:
:+1:

PR #661, opened 2016-06-09
Replace better_errors with web-console

drench:


Gemfile
```diff
@@ -62,11 +62,10 @@ gem "split_accounts", "~> 0.0.1", path: "vendor/engines/split_accounts"
 gem "synaccess_connect", "0.2.2", github: "tablexi/synaccess"
 
 group :development do
-  gem "better_errors"
-  gem "binding_of_caller"
```
jhanggi: Does `pry` need binding of caller?

drench: I used starter's `Gemfile` as a guideline, and it doesn't include `binding_of_caller`. It looks like only a couple TXI projects include both `web-console` and `binding_of_caller` in their `Gemfile`.

`pry` works for me without it, though I don't know if I'm hitting all its features.

jhanggi: If pry works, then, 👍

jhanggi:
👍

PR #662, opened 2016-06-10
[#126615] Restore User access list

drench:
This removes a stray `scope` missed during the Rails upgrade. It also adds some tests for `UsersController#access_list`, which relies on `Product.group_by_type`.


jhanggi:
👍  assuming green build.

PR #663, opened 2016-06-13
[#126619] Set relay "switch" param to checked value

drench:
This sets the `switch` param on relay Ajax requests to "on" if the relay checkbox (switch) is checked, and "off" if it is not.

Before Rails 4 and related jQuery upgrades, this needed to be the opposite; the `:checked` value of the checkbox did not appear to be updated immediately on click, so this used the opposite value.


mlineen:
👍

PR #664, opened 2016-06-21
[#126687] AM/PM and Date changes did not trigger pricing update

jhanggi:
See https://github.com/tablexi/nucore-open/pull/155 for where this
broke. We were trying to fix a 500 error when the ajax was triggering
with an empty quantity. However, this then ignored the call for AM/PM
and date changes.

This now just makes sure there is at least something in the field. If
you do something silly like type in a non-numeric into quantity, the
quantity and pricing changes get thrown away by the backend (see the
change in PR #155).

### DIFF
--- a/app/assets/javascripts/app/manage_orders.coffee
+++ b/app/assets/javascripts/app/manage_orders.coffee
@@ -45,7 +45,7 @@ class OrderDetailManagement
   initPriceUpdating: ->
     self = this
     @$element.find('[name^="order_detail[reservation]"]:not([name$=_display]),[name="order_detail[quantity]"],[name="order_detail[account_id]"]').bind "change keyup", (evt) ->
-      self.updatePricing(evt) if this.value.match(/^\d+$/)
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/app/manage_orders.coffee
```diff
@@ -45,7 +45,7 @@ class OrderDetailManagement
   initPriceUpdating: ->
     self = this
     @$element.find('[name^="order_detail[reservation]"]:not([name$=_display]),[name="order_detail[quantity]"],[name="order_detail[account_id]"]').bind "change keyup", (evt) ->
-      self.updatePricing(evt) if this.value.match(/^\d+$/)
+      self.updatePricing(evt) if this.value.match(/.+/)
```
mlineen: would length > 0 be better here?

jhanggi: Good call. Definitely clearer. I'll make that change.

PR #665, opened 2016-06-22
[#126669] Rails4 Fix password editing and reset

jhanggi:
We messed up the routing a little bit, but I discovered that even before the upgrade there were some controller actions we were never accessing, so I removed the actions and those routes. We're using the devise actions. Also added capybara specs and removed the controller specs.


mlineen:
👍

PR #669, opened 2016-06-27
[#120869561] Add colors to sanger well plates

jhanggi:
This also adds the ability to remove a submission from the well plate, but not as an on-hover button. There's some extra logic in the second commit so that once a submission has been added to the well plate, even if it is removed and added back in, it gets the same color.

Still need:

* Multiple well-plates within a batch
* On-hover link for removing a submission from the well plate grid
* ~~Handle when adding more submissions than we have colors~~

PR #670, opened 2016-06-28
[#120869561] Sanger sequencing - support multiple plates

jhanggi:
This extends from #669 and adds some refactoring including moving the color array into JS. This allowed me to get rid of the magic number that needed to be know between CSS and JS.

It also changes how the rendering of a well plate works so we only "render" whenever we add or remove a submission vs what it was doing before of rendering on every cell when called with `sampleAtCell`. Once we got into multiple plates, I was starting to see some UI slowness. This is much snappier.

### DIFF
--- a/vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/vue_well_plate_app.coffee
+++ b/vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/vue_well_plate_app.coffee
@@ -19,19 +19,9 @@ window.vue_sanger_sequencing_well_plate_app = {
         submission.id == submissionId
       )[0]
 
-    colorForCell: (cell) ->
-      @colorForSubmissionId(@sampleAtCell(cell.name).submission_id())
-
     colorForSubmissionId: (submissionId) ->
-      # 18 is a magic number coming from the number of colors we have defined in
-      # our CSS classes
-      index = (@submissionIndex(submissionId) % 18) + 1
-      "sangerSequencing--colorCoded__color#{index}"
-
-    submissionIndex: (submissionId) ->
-      @builder.allSubmissions.map((submission) ->
-        submission.id
-      ).indexOf(submissionId)
+      @colorBuilder ||= new SangerSequencing.WellPlateColors(@builder)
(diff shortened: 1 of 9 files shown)


drench:
It's working well for me, and https://github.com/tablexi/nucore-open/compare/16cb68c...e368ad2 LGTM aside from the minor code climate newline issue.

vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/vue_well_plate_app.coffee
```diff
@@ -1,30 +1,32 @@
...
 
     findSubmission: (submissionId) ->
       @submissions.filter((submission) =>
         submission.id == submissionId
       )[0]
 
-    hasNotBeenAdded: (submissionId) ->
-      !@builder.hasBeenAdded(@findSubmission(submissionId))
+    styleForSubmissionId: (submissionId) ->
+      @colorBuilder ||= new SangerSequencing.WellPlateColors(@builder)
```
theotherzach: It might make sense to explicitly assign `@colorBuilder` in the `ready` hook so that this method becomes somewhat simpler.

vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/vue_well_plate_app.coffee
```diff
@@ -1,30 +1,32 @@
 window.vue_sanger_sequencing_well_plate_app = {
   props: ["submissions"]
   data: ->
-    plate: SangerSequencing.WellPlateBuilder.rows()
     builder: new SangerSequencing.WellPlateBuilder
 
   ready: ->
     vueBus.$on "submission-added", @addSubmission
```
theotherzach: Sorry I missed this in the last PR, but I think this might be dead code?

theotherzach:
👍

vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/vue_well_plate_app.coffee
jhanggi: Good call. I like that.

PR #671, opened 2016-06-28
[#120869561] Temporarily disable "Save Batch" for Sanger well plates

jhanggi:
The functionality doesn't work yet and currently errors. This will avoid errors
while we're building out the next feature.


drench:
:+1:

PR #676, opened 2016-06-29
Don't exclude vendor from RSpec's stack traces

jhanggi:
This is why our stack traces in engine test failures are so unhelpful.
All we were seeing was the `rails_helper.rb` line, but nothing within
the engine.


drench:
Thank you for this. I kind of agree with code climate, but it's only one escaped `/`. :+1:

drench:
:+1:

PR #677, opened 2016-06-30
[#126388] Allow an instrument to be marked offline/down

drench:
This adds `OfflineReservation`s and a CRUD UI temporarily restricted to global administrators.

The STI of `Reservation` isn't as clean as I'd like. I'd like to remove the type checks in the validations if possible, and not have to rely on the `untyped` scope to filter offline reservations from the others.

I've marked this as WIP as there's still an annoying CI failure in `spec/features/managing_instrument_offline_spec.rb` I'm trying to solve.


jhanggi:
I haven't actually run any of the code, but I'm a little confused by the `untyped` scope. If we're using STI, then the `type` field should never be `nil`. We should probably have plain `Reservation`, `OfflineReservation`, and `AdminReservation`. Or perhaps we create a `UserReservation` class for hanging specific validations off of, and there are no instances of actual `Reservation`s.

app/controllers/instruments_controller.rb
```diff
@@ -81,7 +81,13 @@ def show
 
   # GET /facilities/:facility_id/instruments/:instrument_id/schedule
   def schedule
-    @admin_reservations = @instrument.schedule.reservations.where("reserve_end_at > ? AND order_detail_id IS NULL", Time.zone.now).order("reserve_start_at ASC")
+    @admin_reservations =
+      @instrument
+        .schedule
+        .reservations
+        .admin
```
jhanggi: I definitely like using the scopes here.

app/controllers/offline_reservations_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+      flash[:notice] = text("bring_online.success")
+    else
+      flash[:error] = text("bring_online.error")
+    end
+
+    redirect_to facility_instrument_schedule_path
+  end
+
+  def edit
+    @reservation = @instrument.offline_reservations.find(params[:id])
```
jhanggi: Maybe pull this into a `before_action` to share with `update`?

app/models/reservation.rb
```diff
@@ -165,22 +169,26 @@ def admin?
...
-
-    Reservation.joins(:order_detail).where(where, Time.zone.now, reserve_start_at, product_id).first.nil?
+
+    Reservation
+      .not_started
+      .where("reserve_start_at > :now", now: Time.current)
+      .where("reserve_start_at < :reserve_start_at", reserve_start_at: reserve_start_at)
+      .where(product_id: product_id)
+      .joins(:order_detail)
+      .where("order_detail_id IS NULL OR order_details.state IN ('new', 'inprocess')")
```
jhanggi: Seems like this could be split into `admin` and some other `new_or_in_process` scope (I think we have that already).

jhanggi:
Not sure if this ends up outside of the scope of this project, but it would be nice to split the admin reservation create/edit out of `FacilitiesReservationController`, which has a weird mix of admin editing and normal reservation editing (this will be related to the ticket from Aaron that just came in, so let's not do this just yet, but something to think about).

app/models/reservation.rb
drench: The existing `new_or_inprocess` scope unfortunately comes with baggage. It looks like it's only used in 3 spots (outside specs) so it might not be too painful to clean up.

app/models/reservation.rb
```diff
@@ -118,9 +120,15 @@ def self.relay_in_progress
     where("actual_start_at IS NOT NULL AND actual_end_at IS NULL")
   end
 
+  scope :not_offline, -> { where(type: nil) }
```
jhanggi: Do we still have `nil` types? I thought the base reservation would be saved as `Reservation`

app/models/reservation.rb
```diff
@@ -44,19 +44,21 @@ class Reservation < ActiveRecord::Base
 
   def self.active
     not_canceled
+      .not_offline
       .where(orders: { state: ["purchased", nil] })
       .joins_order
   end
 
+  scope :admin, -> { where(order_detail_id: nil) }
```
jhanggi: Can admin instead be `where(type: "AdminReservation")`?

app/support/products/scheduling_support.rb
```diff
@@ -98,6 +98,14 @@ def available_schedule_rules(user)
     end
   end
 
+  def online!
+    offline_reservations.current.update_all(reserve_end_at: Time.current)
```
jhanggi: Should we set actuals as well?

app/support/reservations/date_support.rb
```diff
@@ -11,7 +11,7 @@ module Reservations::DateSupport
                 :actual_end_date, :actual_end_hour, :actual_end_min, :actual_end_meridian,
                 :actual_duration_mins
 
-    before_validation :set_all_split_times
+    before_validation :set_all_split_times, if: :end_at_required?
```
jhanggi: I get that we need it in certain cases, but could we name it something else? I'd be fine being an alias, but this feels confusing as to why it's here. I'd be ok if it's just an alias.

app/models/reservation.rb
drench: They unfortunately aren't. The base `Reservation` gets a type of `nil` by default:
```
[1] pry(main)> Instrument.first.reservations.create.type
  Instrument Load (0.4ms)  SELECT  `products`.* FROM `products` WHERE `products`.`type` IN ('Instrument')  ORDER BY `products`.`id` ASC LIMIT 1
   (0.1ms)  SAVEPOINT active_record_1
  Product Load (0.4ms)  SELECT  `products`.* FROM `products` WHERE `products`.`id` = 3 LIMIT 1
   (0.1ms)  RELEASE SAVEPOINT active_record_1
=> nil
```

app/views/offline_reservations/new.html.haml
```diff
@@ -0,0 +1,19 @@
...
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_product", sidenav_tab: "instruments"
+
+= content_for :tabnav do
+  = render "admin/shared/tabnav_product", secondary_tab: "reservations"
+
+%h2= @instrument
+
+= simple_form_for([@instrument, @reservation], url: facility_instrument_offline_reservations_path) do |f|
```
jhanggi: Seems like there's a lot that could be shared between new and edit. I think the only difference is the text on the submit button.

spec/features/managing_instrument_offline_spec.rb
```diff
@@ -0,0 +1,79 @@
...
+  describe "marking an instrument offline" do
+    before(:each) do
+      visit facility_instrument_schedule_path(facility, instrument)
+      click_link "Mark Instrument Offline"
+      fill_in "offline_reservation[admin_note]", with: "The instrument is broken"
+      click_button "Take instrument offline"
+    end
+
+    it "marks the instrument offline" do
+      expect(page).to have_content("The instrument has been marked offline")
```
jhanggi: Was this where the CI failure was happening? I'd like to confirm that it actually got marked offline. I'm wondering if it's just a timing issue.

app/views/instruments/schedule.html.haml
```diff
@@ -1,40 +1,84 @@
...
+%p
+  = link_to t(".add_admin_reservation"),
+    new_facility_instrument_reservation_path,
+    class: "btn-add"
+
+  - if current_user.administrator? # TODO: restrict to global admins until the offline feature is complete
+    - if @instrument.online?
+      = link_to t(".mark_instrument_offline"),
+        new_facility_instrument_offline_reservation_path(current_facility, @instrument),
+        class: "btn-add"
```
jhanggi: I feel like this button should use a different styling. They kind of blur together. It also might be nice to separate them with some space. Possibly float this one over to the right?

spec/features/managing_instrument_offline_spec.rb
drench: What get me is the page content test would fail if `instrument.online?` was `true`. So it's timing or maybe threading? If I put a `sleep 1` before the `instrument.reload.online?` test and it that worked on CI. But I never got it to fail outside CI.

app/models/reservation.rb
jhanggi: OK, makes sense. I didn't realize that's how it worked. One idea I found on stackoverflow is to set the default value in the DB to "Reservation", then it would always be there. We don't need to do it, just an idea.

app/models/reservation.rb
jhanggi: Is this not checking if the reservation is purchased? That seems to be what the extra baggage on the scope is. (Although I could see it being tricky to get that scope combined with the `OR`).

app/views/instruments/schedule.html.haml
drench: I know what you mean. Here's `float: right;` what do you think?

screen shot 2016-07-01 at 7 45 19 pm

I also considered a red button, but this is one step removed from actually taking it offline so it didn't feel right.

app/support/products/scheduling_support.rb
drench: Maybe? Reserve times should equal actuals if we do, though at this point nothing is looking at offline actuals. Will it matter with reporting?

app/controllers/facility_reservations_controller.rb
```diff
@@ -149,8 +149,8 @@ def update_admin
     raise ActiveRecord::RecordNotFound unless @reservation.order_detail_id.nil?
     set_windows
 
-    @reservation.assign_times_from_params(params[:reservation])
-    @reservation.admin_note = params[:reservation][:admin_note]
+    @reservation.assign_times_from_params(params[:admin_reservation])
```
jhanggi: While testing, I'm seeing this coming in as `params[:reservation]`.

app/views/instruments/schedule.html.haml
```diff
@@ -1,40 +1,84 @@
...
+        method: :put,
+        class: "btn-add"
 
 - if @admin_reservations.empty?
-  %p.notice There are no upcoming administrative reservations.
+  %p.notice= t(".no_upcoming_admin_reservations")
+
 - else
+  - unless @instrument.online?
+    %p.notice= t(".instrument_is_offline")
```
jhanggi: This should be something more severe than the light blue. I think `.alert` is the red one.

app/controllers/facility_reservations_controller.rb
jhanggi: Sorry, looking at the wrong action. I am unable to create an admin reservation. Looks like `create` isn't using `assign_times_from_params`.

app/controllers/offline_reservations_controller.rb
```diff
@@ -0,0 +1,64 @@
...
+  before_action :load_reservation, only: %i(edit update)
+
+  load_and_authorize_resource class: Reservation
+
+  def new
+    @reservation = @instrument.offline_reservations.new
+  end
+
+  def create
+    @reservation = @instrument.offline_reservations.new(params[:offline_reservation])
```
jhanggi: Even if `Reservation` isn't forcing strong params, can we use them here?

jhanggi:
If I have an admin reservation, I'm not seeing it on the timeline view. Not sure if it's related, but if I create an Admin reservation, its `type` is not being set to `AdminReservation`.

app/controllers/facility_reservations_controller.rb
drench: The admin reservation create action was relying on the old `set_all_split_times` callback, which it can't anymore. Updating.

app/controllers/facility_reservations_controller.rb
```diff
@@ -121,8 +121,10 @@ def new
 
   # POST /facilities/:facility_id/instruments/:instrument_id/reservations
   def create
-    @instrument   = current_facility.instruments.find_by_url_name!(params[:instrument_id])
-    @reservation  = @instrument.reservations.new(params[:reservation])
+    @instrument =
+      current_facility.instruments.find_by!(url_name: params[:instrument_id])
+    @reservation =
+      @instrument.admin_reservations.new(params[:reservation].merge(split_times: true))
```
jhanggi: Can we use the `assign_times_from_params`? I think that's what we've been using in most places in the application. The `split_times: true` is mostly in the specs.

spec/controllers/facility_reservations_controller_spec.rb
```diff
@@ -101,54 +101,65 @@
     it_should_allow_operators_only :redirect
   end
 
-  context '#create' do
-    before :each do
+  describe 'POST #create' do
```
jhanggi: Code climate is complaining about the single quotes.

app/controllers/facility_reservations_controller.rb
```diff
@@ -123,8 +123,8 @@ def new
   def create
     @instrument =
       current_facility.instruments.find_by!(url_name: params[:instrument_id])
-    @reservation =
-      @instrument.admin_reservations.new(params[:reservation].merge(split_times: true))
+    @reservation = @instrument.admin_reservations.new
+    @reservation.assign_times_from_params(params[:reservation])
```
jhanggi: This ends up throwing out the note field. In normal `Reservations#create` we do both `assign_attributes` and `assign_times_from_params`.

drench: Good catch; the specs weren't checking for `admin_note` (they do now)

app/controllers/facility_reservations_controller.rb
```diff
@@ -121,8 +121,11 @@ def new
 
   # POST /facilities/:facility_id/instruments/:instrument_id/reservations
   def create
-    @instrument   = current_facility.instruments.find_by_url_name!(params[:instrument_id])
-    @reservation  = @instrument.reservations.new(params[:reservation])
+    @instrument =
+      current_facility.instruments.find_by!(url_name: params[:instrument_id])
+    @reservation = @instrument.admin_reservations.new
+    @reservation.assign_attributes(params[:reservation])
```
jhanggi: You could do it in two lines without the `assign_attributes`. Not a big deal though.

```
@reservation  = @instrument.reservations.new(params[:reservation])
@reservation.assign_times_from_params(params[:reservation])
```

jhanggi:
One note that's not a big deal. Otherwise, 👍

app/controllers/facility_reservations_controller.rb
drench: That was my first thought, but I went with the pattern we're using in `ReservationsController#create` (half expecting CodeClimate to complain about duplication).

PR #678, opened 2016-06-30
[#120869939] Download Sanger Sequencing template file

jhanggi:
This modifies the batch creation so that reserved wells in unused rows don't get persisted: they appear as if they're blank in the well-plate export. But they still show up as "reserved" when creating the well plate.

It also adds a `ActionController::Renderer` for CSV to the root app so we can do `render csv: ObjectRespondingToToCSV.new, filename: "somefilename"`. There are a few places elsewhere in the app I'd like to swap this in in place of CSVHelper, but that seemed outside the scope of this PR.

There are a few fields in the CSV that I'm not really sure if they're supposed to be dynamic or not. I figure we can address those when we meet with Chris next week.

### DIFF
--- a/vendor/engines/sanger_sequencing/spec/presenters/sanger_sequencing/well_plate_presenter_spec.rb
+++ b/vendor/engines/sanger_sequencing/spec/presenters/sanger_sequencing/well_plate_presenter_spec.rb
@@ -24,18 +24,18 @@ RSpec.describe SangerSequencing::WellPlatePresenter do
     let(:sample_rows) { presenter.sample_rows }
     let(:expected_results_group) { a_kind_of(String) }
     let(:expected_instrument_protocol) { a_kind_of(String) }
-    let(:expected_analysis_protocal) { a_kind_of(String) }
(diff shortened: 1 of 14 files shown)


vendor/engines/sanger_sequencing/spec/presenters/sanger_sequencing/well_plate_presenter_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+        "H01" => samples[6].id,
+      }
+    end
+
+    let(:well_plate) { SangerSequencing::WellPlate.new(mapping, samples: samples) }
+    let(:presenter) { described_class.new(well_plate) }
+    let(:sample_rows) { presenter.sample_rows }
+    let(:expected_results_group) { a_kind_of(String) }
+    let(:expected_instrument_protocol) { a_kind_of(String) }
+    let(:expected_analysis_protocal) { a_kind_of(String) }
```
drench: Typo?

drench:
It seems to work well and I'm looking forward to applying the csv renderer elsewhere. :+1: after the typo fix.

PR #680, opened 2016-06-30
[#126814] Instrument link should take you to schedule, not details

jhanggi:
This was broken in https://github.com/tablexi/nucore-open/pull/565 when
we went to share the index view between the different product types.


drench:
:+1: though I wouldn't mind seeing the `if/else` pulled into a helper for clarity if you think it's worth doing.

PR #681, opened 2016-06-30
Fix asset precompile error in dev

jhanggi:
Was not causing any problems in stage/prod when trying to upload results files.

This also removes some IE-specific stylesheets that are not referenced
anywhere.


drench:
:+1:

PR #683, opened 2016-07-01
Replace old-style finders in the demo seed

drench:
I needed to do a complete database reset and ran into some issues in the demo seed. This should allow the demo seed to work from a completely empty database.


jhanggi:
👍  I like the use of `find_or_create_by` with the block.

PR #684, opened 2016-07-01
[#120869885] Sanger Sequencing batch show view

jhanggi:
Extends from #678

### DIFF
--- a/vendor/engines/sanger_sequencing/app/views/sanger_sequencing/admin/batches/new.html.haml
+++ b/vendor/engines/sanger_sequencing/app/views/sanger_sequencing/admin/batches/new.html.haml
@@ -5,7 +5,7 @@
   = javascript_include_tag "sanger_sequencing/well_plate"
   = stylesheet_link_tag "sanger_sequencing/application"
 
-%vue-sanger-sequencing-well-plate-app(inline-template){ ":submissions" => @submissions.to_json(include: :samples) }
+%vue-sanger-sequencing-well-plate-editor-app(inline-template){ ":submissions" => @submissions.to_json(include: :samples) }
 
   = simple_form_for :batch, url: facility_sanger_sequencing_admin_batches_path do |f|
     = hidden_field_tag "batch[submission_ids]", "{{submissionIds()}}"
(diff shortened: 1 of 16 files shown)


vendor/engines/sanger_sequencing/app/views/sanger_sequencing/admin/batches/new.html.haml
```diff
@@ -5,49 +5,17 @@
   = javascript_include_tag "sanger_sequencing/well_plate"
   = stylesheet_link_tag "sanger_sequencing/application"
 
-%vue-sanger-sequencing-well-plate-app(inline-template){ ":submissions" => @submissions.to_json(include: :samples) }
+%vue-sanger-sequencing-well-plate-editor-app(inline-template){ ":submissions" => @submissions.to_json(include: :samples) }
```
drench: This reminded me to give a :+1: to @theotherzach's https://github.com/haml/haml/issues/876

drench:
One odd UI thing I ran into was in the submission modal on the view page. It popped up (seemingly) blank for me, but it turned out it was just scrolled all the way to the right. I got the content when all the way left, but it wasn't obvious I had to do that. This was with Firefox.

With Chrome, the content inside the modal is still wide, but it always opens already scrolled left so it doesn't seem to be a problem there.

vendor/engines/sanger_sequencing/spec/presenters/sanger_sequencing/well_plate_presenter_spec.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
+
+RSpec.describe SangerSequencing::WellPlatePresenter do
+  let(:samples) { FactoryGirl.build_stubbed_list(:sanger_sequencing_sample, 10) }
+
+  describe "sample_rows" do
+    let(:mapping) do
```
theotherzach: If you pull this out into its own factory then you're able to use that same data structure in your frontend tests too. Might not be important enough to warrant the extraction, but thought I would put it out there.

jhanggi:
I was able to fix the modal issue. There was an extra bootstrap class that we didn't need.

drench:
:+1:

PR #685, opened 2016-07-05
[#126870] Do not error on blank product url names

jhanggi:
If you submit the product form without a url_name, the form would error
while trying to build the hint.

```No route matches {:action=>"show", :controller=>"services",
:facility_id=>"nugene", :id=>""} missing required keys: [:id]
app/helpers/products_helper.rb:44:in `product_url_hint'
```


drench:
:+1:

PR #686, opened 2016-07-05
[#123935071] Sanger sequencing results file upload

jhanggi:
This uses FineUploader for handling file uploads. There's not a supported rails gem (FineUploader only recently went back to being open-source), so I wrapped it up in an engine. If it all works well, we could extract the engine into a gem.


mlineen:
👍

vendor/engines/sanger_sequencing/spec/services/sanger_sequencing/sample_result_file_saver_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+
+RSpec.describe SangerSequencing::SampleResultFileSaver do
+  let(:user) { FactoryGirl.create(:user) }
+  let(:facility) { FactoryGirl.create(:facility) }
+  let(:product) { FactoryGirl.build(:service, facility: facility).tap { |s| s.save(validate: false) } }
```
drench: I'd prefer having more of this setup into factory traits (though really I wish test setup didn't have to be this elaborate).

drench:
Belated :+1:

PR #687, opened 2016-07-06
Sanger sequencing: link to batch from submission index

jhanggi:
There was already a number from before I added the show views, so this just links
it up.


mlineen:
👍

PR #688, opened 2016-07-06
[#123935331] My Files Tab

jhanggi:
The “My Files” tab contains both results files and uploaded order forms. It should only contain files attached to purchased orders (i.e. doesn’t not include uploaded files still in the cart).

The next step will be an extension point so that we can link to sanger sequencing forms.


mlineen:
👍

PR #690, opened 2016-07-07
Sanger Polish

jhanggi:
Adds some style and additional information to the sanger views

PR #691, opened 2016-07-08
[#119570751] Sanger sample listing on receipt

jhanggi:
Includes the commits from #690 

This adds the sample listing to both the displayed order receipt and the one that is emailed out. I added a couple additional links to the facility and order. At some point, we really should do some styling to the email receipt.

### DIFF
--- a/vendor/engines/sanger_sequencing/app/presenters/sanger_sequencing/receipt_submission_presenter.rb
+++ b/vendor/engines/sanger_sequencing/app/presenters/sanger_sequencing/receipt_submission_presenter.rb
@@ -11,7 +11,7 @@ module SangerSequencing
     end
 
     def to_s
-      return unless submission
+      return "" unless submission
       render("sanger_sequencing/submissions/samples_table", submission: submission, caption: caption)
         .safe_concat(render(body: print_me_text))
     end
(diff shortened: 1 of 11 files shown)


vendor/engines/sanger_sequencing/app/presenters/sanger_sequencing/receipt_submission_presenter.rb
```diff
@@ -0,0 +1,39 @@
...
+    attr_reader :view_context, :order_detail
+    delegate :render, :link_to, to: :view_context
+
+    def initialize(view_context, order_detail)
+      @view_context = view_context
+      @order_detail = order_detail
+    end
+
+    def to_s
+      return unless submission
```
drench: Nitpick: I'd expect `#to_s` to always return a `String`, even if it's `""`.

drench:
Works for me. :+1: with one nitpick.

vendor/engines/sanger_sequencing/app/presenters/sanger_sequencing/receipt_submission_presenter.rb
jhanggi: Good call. I can agree with that.

PR #692, opened 2016-07-08
Fix 95-96 samples would not show second plate

jhanggi:
In the plate count, we weren’t accounting for the two reserved cells,
so if you had 95 or 96 samples added to a plate, the next plate was not
appearing.


drench:
:+1:

PR #693, opened 2016-07-11
Rework OrderStatus factory to eliminate helper

drench:
The factory change allows `FactoryGirl.create(:setup_instrument)` to work from the Rails console without having to load helper methods.


jhanggi:
👍

PR #695, opened 2016-07-13
[#126391] Flag ongoing reservations as problem orders when taking an instrument offline

drench:


app/controllers/offline_reservations_controller.rb
```diff
@@ -52,6 +53,12 @@ def update
 
   private
 
+  def flag_ongoing_reservations_as_problem
+    OrderDetail
+      .where(id: @instrument.reservations.ongoing.pluck(:order_detail_id))
+      .update_all(problem: true)
```
jhanggi: The `problem` field is a field that gets updated as part of a `before_save` callback. We should not be setting it directly. Looking at that code, it looks like we need to mark the order as complete. Since it'll be missing the `actual_end_at`, it should get itself into problem status. I think this is what we do when we auto_expire reservations.

app/controllers/offline_reservations_controller.rb
```diff
@@ -8,6 +8,7 @@ class OfflineReservationsController < ApplicationController
   before_action :init_current_facility
   before_action :load_instrument
   before_action :load_reservation, only: %i(edit update)
+  after_action :flag_ongoing_reservations_as_problem, only: %i(create)
```
jhanggi: Won't this get called on an unsuccessful `create`? I don't think we'd want that to happen.

app/models/reservation.rb
```diff
@@ -80,6 +80,10 @@ def self.not_this_reservation(reservation)
     end
   end
 
+  scope :ongoing, -> {
+    where("actual_start_at <= ?", Time.current).where(actual_end_at: nil)
```
jhanggi: We have a `not_ended` scope that handles the `where(actual_end_at: nil)`.

jhanggi: The `relay_in_progress` would probably work as well. It checks that `actual_start_at IS NOT NULL`. In theory, you won't have an actual start at in the future.

app/controllers/offline_reservations_controller.rb
jhanggi: I don't love doing this in an `after_action`. I would lean towards moving it into the success block of the action or possibly an after create callback on the OfflineReservation.

PR #696, opened 2016-07-14
[#126394] Add offline warnings and banners

drench:
~~This includes #677; I will rebase once that's merged.~~


app/views/instruments/public_schedule.html.haml
```diff
@@ -23,6 +23,9 @@
 
 %h2 #{@instrument} #{t('.title')}
 
+- unless @instrument.online?
```
jhanggi: Would it be worth adding an `Instrument#offline?` method that's just the inverse of `online?` so we can make these `if`s?

app/views/shared/_reservations_table.html.haml
```diff
@@ -24,6 +24,10 @@
             = link_to "#{od.product.facility.abbreviation} / #{product_name}".html_safe, facility_instrument_path(od.product.facility, od.product)
             - if od.note.present?
               .order-detail-extra.order-detail-note= "Note: #{od.note}"
+            - if od.product.respond_to?(:online?)
```
jhanggi: What would you think of an `online?` method on `Product` that always responds to `true`?

config/locales/en.yml
```diff
@@ -1055,6 +1053,13 @@ en:
         max_res: "Maximum Reservation Minutes"
         cancel_hours: "Minimum Cancelation Hours"
         cancel_mins: "Automatic Cancelation Minutes"
+    offline:
+      note: Instrument is down
+      notice: |
+        This instrument is currently down.
+        Contact facility for more information.
+        It will be canceled for $0 if the instrument is still down at the time
```
jhanggi: What would you think of "It will be canceled at no charge if..."? I'm thinking that might read a little nicer from the customer's perspective.

app/views/shared/_reservations_table.html.haml
```diff
@@ -24,6 +24,10 @@
...
+              - unless od.product.online?
+                = tooltip_icon "icon-warning-sign",
```
jhanggi: I'm thinking we should make this red to make it stand out more. What do you think?

app/views/shared/timeline/_instrument.html.haml
```diff
@@ -1,17 +1,26 @@
...
+    %h4
+      - if public_timeline?
+        = link_to instrument,
+          facility_instrument_path(current_facility, instrument)
+      - else
+        = link_to instrument,
+          facility_instrument_schedule_path(current_facility, instrument)
+
+      - unless instrument.online?
+        = t(".offline")
```
jhanggi: I'm wondering if we should make this more obvious and red as well?

config/locales/en.yml
jhanggi: The "It" is a little unclear. Perhaps: "This instrument is currently down. Contact facility for more information. Reservations may still be placed, but they will be canceled at no charge if the instrument is still down at the time of scheduled reservation."

drench: I initially had it as "This instrument is currently offline. If it is not back online before your scheduled time, you will receive a notice and your reservation will be canceled" but the wording in 2a07e1ce295cd8c67c65cda39b4f6e029a7130fd came from the ticket. I'll rewrite.

app/views/shared/timeline/_instrument.html.haml
jhanggi: I like that.

app/support/products/scheduling_support.rb
```diff
@@ -94,12 +94,16 @@ def available_schedule_rules(user)
...
+  end
+
+  def offline_via_schedule_share?
+    schedule.products.any?(&:offline?)
   end
 
-  def online?
-    offline_reservations.current.none?
+  def online!
+    offline_reservations.current.update_all(reserve_end_at: Time.current)
```
jhanggi: I don't think there are any downsides to not running validations in this case, are there?

app/views/reservations/new.html.haml
```diff
@@ -21,8 +21,8 @@
 
 .wysiwyg= @instrument.description
 
-- unless @instrument.online?
-  %p.alert.alert-danger= t(".offline_notice")
+- if @instrument.offline? || @instrument.offline_via_schedule_share?
```
jhanggi: Is the `offline?` check redundant? This instrument is included in `schedule.products`, so it's status should get checked by `offline_via_schedule_share?`. 

Also, that name seems a little awkward. `offline_via_shared_schedule`?

app/support/products/scheduling_support.rb
```diff
@@ -94,12 +94,16 @@ def available_schedule_rules(user)
     end
   end
 
-  def online!
-    offline_reservations.current.update_all(reserve_end_at: Time.current)
+  def offline?
+    offline_reservations.current.any?
+  end
+
+  def offline_via_schedule_share?
```
jhanggi: Is there a way we can hide this so the only public method is `offline?` and that checks the shared schedules as well. 

Thinking about it, I'm becoming less convinced that the shared schedule behavior we're implementing is correct, and would like to make it easy to change later.

app/views/instruments/public_schedule.html.haml
```diff
@@ -23,6 +23,9 @@
 
 %h2 #{@instrument} #{t('.title')}
 
+- if @instrument.offline?
```
jhanggi: This should probably use the shared schedule check.

app/support/products/scheduling_support.rb
drench: Though I wouldn't expect validation to fail, even if it did, I think I'd want this method to finish with no active `OfflineReservation`s afterward.

app/views/instruments/schedule.html.haml
```diff
@@ -40,7 +40,7 @@
   %p.notice= t(".no_upcoming_admin_reservations")
 
 - else
-  - unless @instrument.online?
+  - if @instrument.offline?
```
jhanggi: I could see this one going either way, but I lean towards having the icon for each of the items in the shared calendar.

app/support/products/scheduling_support.rb
drench: I wanted a distinction between being offline and being offline by proxy, particularly on the admin side.

I agree the behavior is awkward. If we need to back out and rethink, the commits are isolated.

app/views/reservations/new.html.haml
drench: If we stick with this concept, I'm open to a new name. Most of the alternatives I could think of were longer.

app/views/instruments/schedule.html.haml
jhanggi: I think I might have been thinking this was a different view. It could be a different method, but there should probably be some warning here if it's offline via the shared schedule.

jhanggi:
Do we need a warning on the admin New/In Process Reservations tab? It's probably fine without it, just wondering your thoughts.

drench:
I could see a :warning: in the new/inprocess list being useful. After drilling down into how that's generated currently through the `order_detail_cell` partial and `OrderDetailPresenter#description_as_html` it seems more complicated than I'd like.

jhanggi:
Could it go right in `order_detail_cell` right after `order_detail_description(od)`? Doesn't seem like it needs to go inside that.

jhanggi:
I think we should probably go ahead and merge this in so we don't have all the PRs cherry-picking and whatnot. We can address the shared schedule concerns tomorrow.

Only remaining comment on this one is I believe `@instrument.offline? || @instrument.offline_via_schedule_share?` is redundant. I think we just need the `offline_via_schedule_share?` check.

app/views/reservations/new.html.haml
drench: It's true checking `#offline?` and `#offline_via_schedule_share?` is redundant. I left it for documentation value. Without `#offline?` it gave me the impression that this only has to do with instruments that have shared schedules. A better name for `#offline_via_shared_schedule?` might fix that.

app/views/shared/_order_detail_cell.html.haml
```diff
@@ -6,6 +6,10 @@
     - if show_reservation && od.reservation
       %br
       %em= od.reservation
+
+    - if od.product.offline?
```
drench: The unfortunate thing is this only reflects current offline status. If I'm looking at the "All" tab of "My Reservations", it might show the warning sign on a reservation completed at a time when the instrument was online, which could be misleading.

jhanggi: That's a good point. I could definitely see it being confusing. While it might be nice for the "New/In Process" tab, it's probably not worth the confusion of having it in both places and the complexity of only showing it in one.

drench: I'll take it out. I could see a method like `Reservation#overlaps_downtime?` (or something with a better name) being useful here if we want to revive the warning sign here.

drench: I commented in the wrong file; the offline warning sign I meant is in `app/views/shared/_reservations_table.html.haml`.

app/helpers/products_helper.rb
```diff
@@ -58,7 +58,7 @@ def public_calendar_options(product)
   end
 
   def public_calendar_availability_options(product)
-    if product.available?
+    if product.available? && product.online?
       { class: ["icon-calendar", "available"],
         title: t("instruments.public_schedule.available") }
```
jhanggi: This needs to be `text` since it uses `!facility_downcase!`

app/views/shared/_order_detail_cell.html.haml
jhanggi: I was thinking we were talking about the reservations tab while managing a facility. Under "My Reservations" we definitely need some sort of warning/icon there. In this view, if it's hard to only show it in one of the views, then I think I would lean towards the potential confusion if that means the user will see a notice where they need it.

jhanggi:
It doesn't necessarily need to be in this PR, but in the admin reservation table on the page where you toggle the status, can we have the offline reservation render as a range, something more like "Thu, 07/14/2016 2:32 PM –" so it better implies that it's ongoing?

spec/models/offline_reservation_spec.rb
```diff
@@ -16,4 +16,31 @@
...
+      it "renders as a range with no known end" do
+        expect(subject.to_s).to eq "Fri, 07/01/2016 12:00 PM -"
+      end
+    end
+
+    context "when the offline event has ended" do
+      let(:reserve_end_at) { reserve_start_at + 1.month }
+
+      it "renders as a range with an ending" do
+        pending "TODO: Reservation#range_to_s is not handling this correctly"
```
drench: `#range_to_s` only checks that the start and end `#day`s match, so it considers July 1 and Aug 1 the same. Something to fix, though it doesn't affect display for this PR, which should only display offline events that are ongoing (that is, no end yet).

jhanggi:
:+1:

PR #697, opened 2016-07-14
[#126390] Zero-cost cancellation of reservations for offline instruments

drench:
This adds `InstrumentOfflineReservationCanceler`, a service object to:
* find reservations about to start for instruments that are offline
* cancel these reservations without cost, and
* mail a notification to the user about the cancellation

It also adds a `POST` endpoint to run the canceler under a new `/admin/services` namespace, with authentication/authorization methods TBD. Basic auth with credentials in `/config`?

Caveat: the email content is placeholder text; real content TBD in a future story

~~This includes #677; will rebase after that's merged.~~


app/mailers/notifier.rb
```diff
@@ -67,6 +67,15 @@ def statement(args)
     send_nucore_mail args[:user].email, text("views.notifier.statement.subject", facility: @facility)
   end
 
+  def offline_cancellation_notification(reservation)
```
drench: I'm having second thoughts about not extracting this into its own mailer class.

jhanggi: Haven't looked any closer than this comment, but I would lean towards breaking out into a separate mailer class. I'd like to break up this generic Notifier mailer at some point.

jhanggi: Also curious, how does this get serialized now in 4.2? Is it using global_id?

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
...
+      reservation.transaction do
+        cancel_reservation(reservation)
+        Notifier.offline_cancellation_notification(reservation)
+      end
+    end
+  end
+
+  private
+
+  def admin
```
jhanggi: Could this be called something like `admin_user` or `admin_dummy_user` so it's a little clearer how it's being used?

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
+class InstrumentOfflineReservationCanceler
+
+  def cancel!
+    reservations_to_cancel.each do |reservation|
+      reservation.transaction do
+        cancel_reservation(reservation)
+        Notifier.offline_cancellation_notification(reservation)
```
jhanggi: Does this need a `deliver_later`?

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def admin
+    # OrderDetail#cancel_reservation needs an object that responds to #id
+    @admin ||= OpenStruct.new(id: 0)
+  end
+
+  def cancel_reservation(reservation)
+    reservation
+      .order_detail
+      .cancel_reservation(admin, OrderStatus.canceled.first, true, false)
```
jhanggi: I was thinking it might be nice to move the canceled_reason into this method call, but that would make the signature really gross without doing some extra refactoring elsewhere.

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
...
+    reservation
+      .update_attribute(:canceled_reason, "The instrument was offline")
+  end
+
+  def reservations_to_cancel
+    Reservation
+      .not_offline
```
jhanggi: Can we come up with a better name for this? This also excludes `AdminReservation`. Perhaps `user` or `user_reservation`? I don't love those, but I'd like to have something a little clearer.

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
...
+      .where(product_id: offline_instrument_ids)
+      .not_canceled
+      .not_ended
+      .where("reserve_start_at <= ?", 1.minute.from_now)
```
jhanggi: What's the `1.minute.from_now`? Why not current time?

app/services/instrument_offline_reservation_canceler.rb
drench: Agreed. `OrderDetail#cancel_reservation` needs replacement or refactoring but this didn't feel like the time to jump into it.

spec/services/instrument_offline_reservation_canceler_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    context "when the instrument is offline" do
+      let!(:instrument) { FactoryGirl.create(:setup_instrument, :offline) }
+
+      it_behaves_like "the instrument is offline"
+    end
+
+    context "when the instrument is online" do
+      let!(:instrument) { FactoryGirl.create(:setup_instrument, schedule: nil) }
+
+      context "but shares a schedule with an instrument that is offline" do
```
jhanggi: I see we settled on that in the ticket. If that's the case, does all the bannering apply to shared schedules as well?

jhanggi:
It can be a separate PR, but we should figure out how to actually ping the heartbeat URL.

app/services/instrument_offline_reservation_canceler.rb
drench: `user` sounds all right to me until a `UserReservation` model can happen.

spec/services/instrument_offline_reservation_canceler_spec.rb
drench: Probably not. I'll look at what might need to change in `Instrument#online?`/`#offline?` to make that happen when updating #696.

app/mailers/notifier.rb
drench: Glad you mentioned it; it appears it's just serializing it as yaml. I'll work that out with the extraction into its own mailer class.

app/controllers/admin/services_controller.rb
```diff
@@ -0,0 +1,12 @@
+module Admin
+
+  class ServicesController < ApplicationController
+
+    def cancel_reservations_for_offline_instruments
+      InstrumentOfflineReservationCanceler.new.cancel!
+      render nothing: true
```
jhanggi: I'm assuming we just want the 500, so we don't need to worry about the error case, correct?

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
+class InstrumentOfflineReservationCanceler
+
+  def cancel!
+    reservations_to_cancel.each do |reservation|
+      reservation.transaction do
+        cancel_reservation(reservation)
+        OfflineCancellationMailer.delay.send_notification(reservation)
```
jhanggi: Looks like the method is expecting the `id`. Can we use `OfflineCancellationMailer.send_notification(reservation).deliver_later` and start moving towards the ActiveJob syntax? I think if we do that, it'll use global_id automatically.

app/controllers/admin/services_controller.rb
drench: Right, my assumption is the client is a cron job running `curl` so a standard 500 seemed reasonable for now.

jhanggi:
Can we add a mailer preview for this one? Also, can we add the cron to `schedule.rb`?

config/schedule.rb
```diff
@@ -1,3 +1,6 @@
+rails = File.join Dir.getwd, "config", "environment.rb"
+require rails
```
drench: The Rails environment doesn't appear to be loaded when building the crontab so I added this. Really I just needed `url_helpers` so I'm unsure if this is overkill.

config/environments/development.rb
```diff
@@ -17,7 +17,8 @@
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_deliveries = true
-  config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
+  Rails.application.routes.default_url_options =
+    config.action_mailer.default_url_options = { host: "localhost", port: 3000 }
```
drench: We'll need to watch this when merging into the other forks, and update the production environments. It looks like url helpers are defaulting to `example.com`.

config/schedule.rb
jhanggi: Is there a way we could point directly to the path? Then we don't need the rails helper. It should all be fine on `localhost`, the only issue I could see is the port number.

drench: If we hit `http://localhost/` we get an https redirect though `https://localhost/` will give us a certificate error (which we could ignore/override). And it would mean hardcoding the path in this file. Workable?

We could possibly hit the unicorn unix domain socket, though that would be fork-specific, and we would need to upgrade the version of `curl` we have on the production servers to get that feature (curl 7.40.0+).

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -0,0 +1,44 @@
...
+      .not_ended
+      .where("reserve_start_at <= ?", Time.current)
+  end
+
+  def offline_instrument_ids
+    Instrument.where(schedule_id: offline_schedule_ids)
+  end
+
+  def offline_schedule_ids
+    OfflineReservation.current.joins(:product).pluck(:schedule_id)
```
jhanggi: Looks like there's still a shared schedule check lurking around.

config/schedule.rb
```diff
@@ -13,6 +16,10 @@
   rake "order_details:auto_logout"
 end
 
+every 1.minute, roles: [:db] do
+  command "curl #{Rails.application.routes.url_helpers.admin_services_cancel_reservations_for_offline_instruments_url}"
```
jhanggi: Does this have output on success? Otherwise we'll be getting cron emails every minute.

spec/mailers/previews/mailer_preview.rb
```diff
@@ -0,0 +1,7 @@
+class MailerPreview < ActionMailer::Preview
```
jhanggi: Can we call this `OfflineCancellationMailer`?

config/schedule.rb
drench: It has no output on success. Though I just noticed I missed making `curl` do a `POST` instead of a `GET` in the crontab.

app/services/instrument_offline_reservation_canceler.rb
drench: Good catch. There's a vestige in #705 to clean up as well.

app/models/reservation.rb
```diff
@@ -121,7 +121,15 @@ def self.relay_in_progress
     where("actual_start_at IS NOT NULL AND actual_end_at IS NULL")
   end
 
-  scope :not_offline, -> { where(type: nil) }
+  def self.upcoming_offline(start_at_limit)
```
drench: This is borrowed from #705. Both PRs need it, and as it turns out they needed different `start_at_limit` values so it made sense to do away with the default limit.

app/mailers/offline_cancellation_mailer.rb
```diff
@@ -0,0 +1,20 @@
...
+    @reservation = reservation
+    @instrument = reservation.product
+    mail(to: reservation.user.email, subject: subject)
+  end
+
+  private
+
+  def subject
+    text(
+      "views.notifier.offline_cancellation_notification.subject",
```
jhanggi: I'm not convinced this locale should live in`notifier`. If we pull the text out of the templates, then it probably makes sense to give it its own file. I'm still not sure what an optimal file organization is for locales, but I think I'm liking the one file per feature (so it could include views, controller, mailers, and possibly models in the same file).

jhanggi:
Just one minor concern with locales, otherwise 👍

app/mailers/offline_cancellation_mailer.rb
```diff
@@ -0,0 +1,20 @@
...
+    @reservation = reservation
+    @instrument = reservation.product
+    mail(to: reservation.user.email, subject: subject)
+  end
+
+  private
+
+  def subject
+    text(
+      "offline_cancellation_mailer.send_notification.subject",
```
drench: If we didn't need the interpolation, this method wouldn't need to exist, but I thought there would be potential to put more in the subject (facility name, downtime category, etc.).

config/locales/mailers/offline_cancellation_mailer.en.yml
```diff
@@ -0,0 +1,4 @@
+en:
+  offline_cancellation_mailer:
+    send_notification:
+      subject: Your reservation for %{instrument} has been canceled
```
drench: Borrowed this translation pattern from GS

jhanggi:
👍

config/locales/mailers/offline_cancellation_mailer.en.yml
jhanggi: I think that's fine for now. I'm honestly not that crazy about the GS structure. I think you end up with too many deeply nested and tiny files. I like what we've started to do here where we keep controller and view locales in the same file. But this is a special mailer, so I'm not sure where the best place for it is.

PR #698, opened 2016-07-15
[#123935371] Download all results files for an order detail as a zip

jhanggi:
This also does some refactoring to move all of the file actions specific to an order detail out of the order detail controller.
I've also cut us down to just using the fog-aws gem, which significantly decreases the number of dependencies.

One thing I ran into that I can't seem to solve (but won't likely be an issue): if I'm using S3 and one of the files is not found, it causes an exception (expected behavior). But when I shut down the server, I get `zlib(finalizer): the stream was freed prematurely.` so it appears some stream is being left open. I've tried `ensure` and `.close` all over the place, but can't figure it out.


drench:
There are probably more opportunities to refactor in the parts extracted from `OrderDetailsController` but just getting it out is nice to see. :+1:

PR #699, opened 2016-07-15
[#126391] Flag ongoing reservations as problem orders when taking an instrument offline

drench:
This replaces #695.

### DIFF
--- a/app/controllers/offline_reservations_controller.rb
+++ b/app/controllers/offline_reservations_controller.rb
@@ -8,7 +8,6 @@ class OfflineReservationsController < ApplicationController
   before_action :init_current_facility
   before_action :load_instrument
   before_action :load_reservation, only: %i(edit update)
-  after_action :flag_ongoing_reservations_as_problem, only: %i(create)
 
   load_and_authorize_resource class: Reservation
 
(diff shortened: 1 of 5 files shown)


app/models/order_detail.rb
```diff
@@ -465,6 +465,10 @@ def complete!
     change_status!(OrderStatus.complete_status)
   end
 
+  def force_complete!
```
drench: I considered `#to_problem!` though I thought this name was more direct. This is used to get ongoing reservations flagged as "problem" though it's done indirectly.

app/models/order_detail.rb
```diff
@@ -465,6 +465,10 @@ def complete!
...
+    update(state: "complete", order_status: OrderStatus.complete.first)
```
jhanggi: We have `OrderStatus.complete_status` which wraps up the `complete.first`.

app/controllers/offline_reservations_controller.rb
```diff
@@ -52,6 +53,12 @@ def update
 
   private
 
+  def flag_ongoing_reservations_as_problem
+    OrderDetail
+      .where(id: @instrument.reservations.ongoing.pluck(:order_detail_id))
```
jhanggi: I'm not sure a callback is the right answer, but this seems a little too knowledgable about order details for a controller. Not sure I have a better solution, though.

jhanggi:
👍

PR #702, opened 2016-07-15
[#126425461] Restrict access to my files

jhanggi:
Prevent a 500 error when trying to access the page while not logged in.
Also prevent access while acting as.


drench:
:+1:

PR #703, opened 2016-07-15
Make types consistent when creating AdminReservations

drench:
This should correct the issue where correcting a form after a failed `create` would raise an exception.

To reproduce:
* As an administrator, visit `/facilities/:id/instruments/:id/reservations/new`
* Specify a zero-length reservation
* Submit the form
* Get an error screen
* Fill in a valid reservation length
* Submit the form


jhanggi:
I went ahead and merged this because I'm going to be merging master down into DC and wanted to get this bug fix included.

PR #704, opened 2016-07-15
[#125568281] Send notification on results file upload

jhanggi:
I decided to debounce email sending based on the order detail so that I could include the order detail link in the email. If we went with a per-facility or global per-user then it wouldn't be a good idea to include the OD link; only the My Files link would be good. 

I also debounced it at 24 hours. That could pretty easily be extracted into a setting if we want.

This is feature-flagged on My Files right now (mainly because I wanted to include the my files links), but that could change as well.

### DIFF
--- a/app/models/stored_file.rb
+++ b/app/models/stored_file.rb
@@ -11,6 +11,7 @@ class StoredFile < ActiveRecord::Base
   validates_inclusion_of  :file_type, in: %w(info template template_result sample_result import_error import_upload)
 
   delegate :user, to: :order_detail
+  delegate :sample_result?, :template_result?, to: :type_inquirer
 
   scope :import_error, -> { where(file_type: "import_error") }
   scope :import_upload, -> { where(file_type: "import_upload") }
@@ -31,6 +32,12 @@ class StoredFile < ActiveRecord::Base
   def swf_uploaded_data=(data)
     data.content_type = MIME::Types.type_for(data.original_filename).first.to_s
     self.file = data
+  end
+
+  private
+
+  def type_inquirer
+    ActiveSupport::StringInquirer.new(file_type)
   end
 
 end
--- a/app/models/stored_file.rb
+++ b/app/models/stored_file.rb
@@ -10,6 +10,8 @@ class StoredFile < ActiveRecord::Base
   validates_presence_of   :order_detail_id, if: ->(o) { o.file_type == "template_result" || o.file_type == "sample_result" }
   validates_inclusion_of  :file_type, in: %w(info template template_result sample_result import_error import_upload)
 
+  delegate :user, to: :order_detail
+
   scope :import_error, -> { where(file_type: "import_error") }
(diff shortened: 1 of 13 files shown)


vendor/engines/sanger_sequencing/app/services/sanger_sequencing/sample_result_file_saver.rb
```diff
@@ -23,7 +23,9 @@ def save
                                    file_type: "sample_result", created_by: current_user.id,
                                    order_detail: sample.submission.order_detail,
                                    product_id: sample.submission.order_detail.product_id)
-      stored_file.save
+      success = stored_file.save
+      trigger_notification(stored_file) if success
+      success
```
drench: Might not be worth persuing but I'm wondering if these 3 lines could be reduced to `trigger_notification(stored_file) if stored_file.save`
If we can count on `trigger_notification` succeeding and returning something truthy…

app/models/stored_file.rb
```diff
@@ -31,4 +34,10 @@ def swf_uploaded_data=(data)
     self.file = data
   end
 
+  private
+
+  def type_inquirer
+    ActiveSupport::StringInquirer.new(file_type)
```
drench: Nice, I never think to use `StringInquirer`

app/views/results_file_notifier_mailer/file_uploaded.html.haml
```diff
@@ -0,0 +1,5 @@
+= html("html", order_detail: @order_detail, facility: @order_detail.facility,
+  order_detail_path: order_order_detail_url(@order_detail.order, @order_detail),
+  my_files_path: my_files_url)
+
+
```
drench: Some excess whitespace

config/locales/views/en.results_file_notifier.yml
```diff
@@ -0,0 +1,15 @@
...
+  views:
+    results_file_notifier_mailer:
+      file_uploaded:
+        html: |
+          Results file(s) for your !app_name! order #[%{order_detail}](%{order_detail_path})
+          have been uploaded by %{facility}.
+
+          You may download them from [My Orders](%{order_detail_path}) or [My Files](%{my_files_path})
+
+
```
drench: Excess whitespace

vendor/engines/sanger_sequencing/app/services/sanger_sequencing/sample_result_file_saver.rb
jhanggi: Yeah, I thought about that, but I didn't want to need to rely on the return value of `trigger_notification`, or rather, didn't want to rely on the return value of `ResultsFileNotifier.new(stored_file).notify`, which returns nil for the case where the feature is turned off or it doesn't actually send an email because it's already sent one recently.

drench:
:+1:

PR #705, opened 2016-07-16
[#126392] Add daily offline notices

drench:
This sets up a cron job to run at 12:30am daily to notify users with reservations for "down" instruments for that day.

Mail previews should be available in the browser at `/rails/mailers/mailer/upcoming_offline_reservation`.

There's a bit of redundancy that's cherry-picked from related unmerged PRs (the `not_offline` scope rename and the `:offline` trait in the `:setup_instrument` factory).


app/services/upcoming_offline_reservation_notifier.rb
```diff
@@ -0,0 +1,28 @@
...
+  def offline_instrument_ids
+    Instrument.where(schedule_id: offline_schedule_ids)
+  end
+
+  def offline_schedule_ids
+    OfflineReservation.current.joins(:product).pluck(:schedule_id)
+  end
+
+  def upcoming_offline_reservations
+    Reservation
```
jhanggi: I feel like I've seen a very similar chain of scopes somewhere else. Would it be worth rolling them up in `Reservation` into a new scope?

app/views/upcoming_offline_reservation_mailer/generate_mail.html.haml
```diff
@@ -0,0 +1,10 @@
+%p
+  The instrument you have reserved for today at
```
jhanggi: Can we I18n this?

app/views/upcoming_offline_reservation_mailer/generate_mail.html.haml
```diff
@@ -0,0 +1,10 @@
...
+  #{l(@reservation.reserve_start_at, format: :timeonly)},
+  #{@reservation.product}, is down.
+
+  If the instrument is still down at your scheduled time, your reservation will
+  be canceled at no cost.
+
+%p
+  Please contact the #{t("facility_downcase")} for more information.
```
jhanggi: Can we include a link to the facility? And possibly a link to the order detail/reservation?

config/schedule.rb
```diff
@@ -17,6 +17,10 @@
   rake "order_details:remove_merge_orders"
 end
 
+every :day, at: "12:30am", roles: [:db] do
```
jhanggi: Let's make sure to double check when we release this that the time zone is taken into account.

app/views/upcoming_offline_reservation_mailer/generate_mail.html.haml
```diff
@@ -0,0 +1,10 @@
+%p
```
jhanggi: `generate_mail` is a little ambiguous. Is there a more precise name we can come up with?

app/services/upcoming_offline_reservation_notifier.rb
drench: Yes, there is nearly identical code in #697. I like the scope idea.

config/locales/en.yml
```diff
@@ -1131,6 +1131,31 @@ en:
...
       subject: "%{facility} - User Requested Training"
 
+  upcoming_offline_reservation_mailer:
+    subject: Regarding your upcoming reservation for %{instrument}
+    send_offline_instrument_warning:
+      body:
+        html: |
+          <p>
+            %{product}, the instrument you
+            <a href="%{reservation_url}">have reserved for %{reservation_time}</a>,
```
jhanggi: If we use the `html` text helper method, we can use markdown in the locale. No need for the `<p>` tags and cleaner links.

app/models/reservation.rb
```diff
@@ -121,7 +121,15 @@ def self.relay_in_progress
     where("actual_start_at IS NOT NULL AND actual_end_at IS NULL")
   end
 
-  scope :not_offline, -> { where(type: nil) }
+  def self.upcoming_offline(start_at_limit = 1.day.from_now)
```
jhanggi: Where else will this be used? I almost think that the default should be now and the service object should be in charge of deciding how far out to look.

app/views/upcoming_offline_reservation_mailer/send_offline_instrument_warning.html.haml
```diff
@@ -0,0 +1,6 @@
+%p
+  = html "upcoming_offline_reservation_mailer.send_offline_instrument_warning.body.html",
```
jhanggi: Can we use the shorthand? That would mean putting this under `views.upcoming_offline_reservation_mailer.send_offline_instrument_warning`.

spec/mailers/previews/mailer_preview.rb
```diff
@@ -0,0 +1,8 @@
+class MailerPreview < ActionMailer::Preview
```
jhanggi: Can we call this `UpcomingOfflineReservationMailer`?

app/models/reservation.rb
drench: It's only used in `UpcomingOfflineReservationNotifer` so far. I can remove the default limit from here.

jhanggi:
👍

PR #706, opened 2016-07-18
[#126389] Prevent reservation start when the instrument is offline

drench:
This should prevent "Create & Start" as well as starting reservations for an instrument that is offline. It includes cherry-picks from #696 to bring in the `Product#online?` and `#offline?` methods.

### DIFF
--- a/app/support/reservations/relay_support.rb
+++ b/app/support/reservations/relay_support.rb
@@ -3,6 +3,7 @@ module Reservations::RelaySupport
   def can_switch_instrument_on?(check_off = true)
     return false if canceled?
     return false unless product.relay # is relay controlled
+    return false if product.offline?
     return false if check_off && can_switch_instrument_off?(false) # mutually exclusive
     return false unless actual_start_at.nil?   # already turned on
     return false unless actual_end_at.nil?     # already turned off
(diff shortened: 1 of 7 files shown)


app/views/reservations/_js_variables.html.haml
```diff
@@ -13,3 +13,4 @@
   var reserveInterval = #{@instrument.reserve_interval};
   var reserveMinimum  = #{@instrument.min_reserve_mins || 0};
   var reserveMaximum  = #{@instrument.max_reserve_mins || 0};
+  var instrumentOnline = #{@instrument.online? || false};
```
jhanggi: What is the `|| false` for? Won't `online?` always return a boolean?

app/support/reservations/relay_support.rb
```diff
@@ -3,6 +3,7 @@ module Reservations::RelaySupport
   def can_switch_instrument_on?(check_off = true)
     return false if canceled?
     return false unless product.relay # is relay controlled
+    return false if product.offline?
```
jhanggi: Should this be checking the shared schedule? I know we haven't necessarily settled that question yet, but other PRs have been assuming we do share the schedule.

drench: I hesitated bringing that in since would specifically need the `#offline_via_schedule_share` method, still isolated in #696 as I type this.

spec/models/instrument_spec.rb
```diff
@@ -818,20 +818,22 @@
     end
   end
 
-  describe "#online?" do
+  describe "#offline? and #online?" do
     before { instrument.save! }
 
-    context "when an offline reservation does not exist" do
+    context "when an offline reservation does not exist", :aggregate_failures do
```
jhanggi: Does this need the `aggregate_failures` since they're two separate tests?

jhanggi:
One minor comment in the specs, otherwise, 👍

PR #707, opened 2016-07-18
[#123935141] Replace old upload widget with fineuploader

jhanggi:
Also allows us to get rid of a middleware that caused problems with better_errors/webconsole breaking at a weird point on exceptions.


drench:
I ran into that middleware issue yesterday and am very happy to see it fixed. :+1:

PR #708, opened 2016-07-18
[#121327715] Setting to allow skipping netid lookup

jhanggi:
If the school doesn’t use a netid or other directory, then there’s no
need to go through that option when creating a user. This will now just
skip you straight to “Add External User”.

I didn’t think it was worth doing any additional disabling of routes
since it’s still fine if those routes work, this is just much more
efficient.

I’ll have another PR for DC that incorporates this change and changes some locales to make it less confusing for them.


drench:
:+1:

PR #709, opened 2016-07-21
[#123935371] Handle duplicate filenames in zip download

jhanggi:
If there are duplicate filenames being downloaded, the mac unarchiver
will break with “Error 1 - Operation not permitted”. If you use the
command line `unzip`, it will expand it, but will be missing a file.

This appends a -N to the end of the filename before the extension. I’m
planning another PR that will prevent duplicate filenames at upload
time.


drench:
:+1:

PR #711, opened 2016-07-21
[#126393] Support offline reservation categories

drench:
`OfflineReservation::CATEGORIES` is a hardcoded list but the idea is to eventually turn this into something more flexible (and admin-editable).

The categories for offline reservations appear in the CRUD interface as well as the admin reservation list, but nowhere else yet. They will appear in reports.

### DIFF
--- a/app/helpers/facility_reservations_helper.rb
+++ b/app/helpers/facility_reservations_helper.rb
@@ -1,5 +1,11 @@
 module FacilityReservationsHelper
 
+  def offline_category_collection
+    OfflineReservation::CATEGORIES.map do |c|
+      [I18n.t("offline_reservations.categories.#{c}"), c]
+    end
+  end
+
   def reservation_links(reservation)
     links = []
     if reservation.admin?
--- a/app/helpers/facility_reservations_helper.rb
+++ b/app/helpers/facility_reservations_helper.rb
@@ -6,12 +6,8 @@ module FacilityReservationsHelper
     end
   end
 
-  def reservation_category_label(reservation_category)
-    if reservation_category.present?
-      I18n.t("offline_reservations.categories.#{reservation_category}")
-    else
-      ""
-    end
+  def reservation_category_label(category)
+    t(category.presence, scope: "offline_reservations.categories", default: "")
   end
 
   def reservation_links(reservation)
--- a/app/helpers/facility_reservations_helper.rb
+++ b/app/helpers/facility_reservations_helper.rb
@@ -1,8 +1,16 @@
 module FacilityReservationsHelper
 
   def offline_category_collection
-    OfflineReservation::CATEGORIES.map do |c|
-      [I18n.t("offline_reservations.categories.#{c}"), c]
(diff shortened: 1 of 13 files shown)


app/support/reservations/validations.rb
```diff
@@ -21,10 +21,22 @@ module Reservations::Validations
...
     validate :starts_before_ends
     validate :duration_is_interval
+    validate :category_is_valid
   end
 
   # Validation Methods
 
+  def category_is_valid
+    return if category.blank?
+    if self.class.const_defined?(:CATEGORIES)
```
jhanggi: I'm suspecting that the reason for this complexity is so that we can add categories to admin reservations as well. Is that correct?

jhanggi: I'm not sure this is necessarily better, but what about something like (not fully tested)

```ruby
validates :category, inclusion: { in: -> (r) { r.class::CATEGORIES }, allow_blank: true },
      if: -> (r) { r.class.const_defined?(:CATEGORIES) }
 validates :category, absence: true, unless: -> (r) { r.class.const_defined?(:CATEGORIES) }
```

app/support/reservations/validations.rb
```diff
@@ -21,10 +21,22 @@ module Reservations::Validations
...
+      unless self.class::CATEGORIES.include?(category)
+        errors.add(:category, "must be a defined category")
```
jhanggi: If we don't use my other suggestion, can we use `errors.add(:category, :invalid)` or whatever the `inclusion` validator uses so it grabs from the locale?

app/support/reservations/validations.rb
jhanggi: And should we be allowing blank category on offline? I think it's a fine assumption for now, but I could see that changing.

app/helpers/facility_reservations_helper.rb
```diff
@@ -1,5 +1,19 @@
...
 
+  def offline_category_collection
+    OfflineReservation::CATEGORIES.map do |category|
+      [reservation_category_label(category), category]
+    end
+  end
+
+  def reservation_category_label(reservation_category)
+    if reservation_category.present?
+      I18n.t("offline_reservations.categories.#{reservation_category}")
```
jhanggi: I was hoping we could do something like `t(reservation_category, scope: "offline_reservations.categories", default: "")` to get rid of the conditional, but you still need some value going into the first argument.

Do we need the `I18n.` namespace?

app/support/reservations/validations.rb
drench: > I'm suspecting that the reason for this complexity is so that we can add categories to admin reservations as well. Is that correct?

Yes, I wanted to leave the door open for admin categories. This also has to do with how we don't have STI completely for `Reservation` yet. If the model currently known as `Reservation` (that is, `type: nil`) could truly became `UserReservation`, I think it would help reduce the complexity of the validations. Or at least let us split the validations out into the 3 different types where we could.

jhanggi:
👍

PR #712, opened 2016-07-21
[#127148] Do not show not authorized message when adding to an order

jhanggi:
As a facility admin, if you add a reservation to an existing order, the
“The user you are ordering for is not on the authorized list for this
instrument” message was checking the acting user (i.e. the facility
admin’s access) instead of the order’s user.


drench:
:+1:

PR #713, opened 2016-07-22
Fix migrations for Rails 4.2

jhanggi:
So we can do a full rake db:migrate from scratch.

This adds a few `null: false` to timestamp creations because that was the default in older rails versions. This will keep the schema diff minimal.

The big difference is in index names. I’m not sure there’s a good way to go about fixing that.


drench:
Worked for me with MySQL though Oracle will be the test. It would be nice to get the index names consistent but it's something I suppose we can deal with when merging into the other forks if we want. :+1:

PR #714, opened 2016-07-23
[#126916383] Show product name on submission lists

jhanggi:
This will allow the resource to identify if they need to do any special
handing such as always use full plate submissions as the only
submission in the batch.


drench:
:+1:

PR #716, opened 2016-07-26
Explicitly set RAILS_ENV in schedule.rb

jhanggi:
Without this, when rails loaded, it did not have an environment set, so
it was trying to load up the `development` database.


drench:
:+1:

PR #717, opened 2016-07-26
Prevent race conditions when debouncing email sending

jhanggi:
In a multithreaded environment, a second find_or_initialize_by might do the SELECT
before the save in the other thread has happened, causing a duplicate key violation
in the database when the second one tries to save. This locks the table so the second
doesn't do the select until the first one is done.

This seemed to be a little clearer than rescuing from ActiveRecord::RecordNotUnique
and RecordInvalid errors and throwing them away, although those might be slightly
more performant. In my local tests, the lock seemed just fine, even when uploading
10+ files at once.


drench:
:+1:

PR #718, opened 2016-07-26
[#127184] Suppress curl progress meter in cron

drench:
This should prevent `curl` from producing progress output for the rolling-cancel `cron` job.


jhanggi:
👍

PR #719, opened 2016-07-26
[#115963269] Refactor footer for override

jhanggi:
DC as a much wider logo. This will handle it much better, including breakpoints.

I'm planning to use a PNG for DC. It feels a little weird to put the filename in the locales, but I settled on there so that it’s right next to the `logo_alt` translation and will be
clean to override.


drench:
:+1:

PR #720, opened 2016-07-26
Attempt 2: Prevent race conditions when debouncing email sending

jhanggi:
The major change here is that we will simply bail in the case of
getting a record not unique error. We also don’t need to lock the table
in this case.

In the case where the uniqueness validation becomes invalid in the race
condition, then `update_attributes` will fail, so we won’t yield and
bail instead.

I actually tested this code on the staging server, and looks to be working better than the previous code.

I spent about an hour trying to write a spec for this and was unable to come up with anything that worked.


drench:
:+1:

PR #721, opened 2016-07-27
[#127212] Cross-facility journal view inaccessible in rails 4.2

jhanggi:
```
Mysql2::Error: Unknown column 'orders.facility_id' in 'where clause':
SELECT  journals.* FROM `journals` WHERE (orders.facility_id IN (1, 2,
3) AND `journals`.`is_successful` IS NULL  ORDER BY journals.created_at
DESC LIMIT 10 OFFSET 0
```


mlineen:
👍

PR #722, opened 2016-07-28
[#127220] Could not create price groups for facility

jhanggi:
Users were receiving “Facility cannot be blank”. Appears to be an issue
stemming from removing `has_many :price_groups, finder_sql: …` into a
regular method.

https://github.com/tablexi/nucore-open/pull/631/commits/f0eb065598f716acbb06fc14cf94cad39191bb9d

Looks like we handled this for `OrderStatus`es in
https://github.com/tablexi/nucore-open/pull/631/commits/39c2c4f773484aced489d2146ea6822bd0642234

I chose to do one success feature spec at the lowest possible
permissions level and one failure spec at the highest possible level,
and left the rest of the levels to be handled by the Ability spec. I
would probably have only done the success feature spec, but there’s
some weird ability stuff in PriceGroup controller where new/create use
Facility as the ability’s resource, while the member actions on an
existing PriceGroup use the price group itself as the resource.

### DIFF
--- a/spec/features/admin/price_group_management_spec.rb
+++ b/spec/features/admin/price_group_management_spec.rb
@@ -1,0 +1,40 @@
+require "rails_helper"
+
+RSpec.describe "Managing Price Groups", :aggregate_failures do
+
+  let(:facility) { FactoryGirl.create(:facility) }
+
+  describe "create" do
+    describe "as a facility admin" do
+      let(:user) { FactoryGirl.create(:user, :facility_director, facility: facility) }
+
+      before do
+        login_as user
+        visit facility_price_groups_path(facility)
+        click_link "Add Price Group"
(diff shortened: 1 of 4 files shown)


spec/features/admin/price_group_management_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        visit new_facility_price_group_path(facility)
+      end
+
+      it "does not allow access" do
+        expect(page.status_code).to eq(403)
+      end
+    end
+  end
+
+  # TODO: Move tests out of price_groups_controller_spce.rb
```
drench: Typo

drench:
:+1: (the comment typo is about the most minor nitpick possible, and the Code Climate issue isn't really new)

PR #724, opened 2016-07-28
[#126395] Add admin and offline reservation (downtime) reporting

drench:
This adds a third kind of instrument report (the existing two being handled in `Reports::InstrumentReportsController` and ` Reports::InstrumentDayReportsController`).

The reports for offline and admin reservations didn't fit into the existing types, so this adds `Reports::InstrumentUnavailableReportsController`. The data comes out of `Reports::InstrumentUnavailableReport` though I decided to test it through the controller since this still needs to implement the interface and set the instance variables `Reports::ReportsController` and its templates expect.


app/models/reports/instrument_unavailable_report.rb
```diff
@@ -0,0 +1,73 @@
+module Reports
+
+  class InstrumentUnavailableReport
+
+    QUANTITY_INDEX = 3
+    HOURS_INDEX = 4
```
jhanggi: Can we share these constants between the controller and here?

app/models/reports/instrument_unavailable_report.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def total_quantity
+      rows.map { |row| row[QUANTITY_INDEX] }.sum
+    end
+
+    private
+
+    def raw_values
+      query.each_with_object({}) do |row, report|
```
jhanggi: This report seems like it might be more complicated than it needs to be.

Using `group_by` and `sum` might simplify things a bit:

```ruby
grouped = query.group_by { |res| [res.product.name, res.type, res.category] }
rows = grouped.each do |key, reservations|
   key + [reservations.count, format_seconds(reservations.sum_by { |r| duration_in_sections(r) })]
end
```

app/models/reports/instrument_unavailable_report.rb
```diff
@@ -0,0 +1,73 @@
...
+        key = "#{row.product_id}/#{row.type}/#{row.category}"
+        report[key] ||= row_to_h(row)
+        report[key][:quantity] += 1
+        report[key][:seconds] += row_duration_in_seconds(row)
+      end.values
+    end
+
+    def row_duration_in_seconds(row)
+      reserve_end_at = row.reserve_end_at || end_time
```
jhanggi: This is a fairly hard to grok. I'm not sure how to improve it and still accomplish the same results. Maybe adding some comments might help.

app/views/reports/instrument_unavailable_reports/_report_data_rows.csv.haml
```diff
@@ -0,0 +1,4 @@
+- @report_data.each do |res|
+  - reserved_hours = to_hours(res.duration_mins)
+  - actual_hours = to_hours(res.actual_duration_mins)
+  - csv << (@report_on.call(res) + [ 1, reserved_hours, reserved_hours / @totals[0], actual_hours, actual_hours / @totals[1] ] + report_attribute_values(res, res.product))
```
jhanggi: Where does this get used? This looks like it's maybe adding percentages? I don't see them anywhere including the downloaded export. (I don't think percentages are worth adding right now)

app/views/reports/instrument_unavailable_reports/_report_rows.html.haml
```diff
@@ -0,0 +1,17 @@
+%tbody
```
jhanggi: Is this duplicated with any of the other reports?

app/views/reports/instrument_unavailable_reports/_report_tabs.html.haml
```diff
@@ -0,0 +1,13 @@
+%ul
```
jhanggi: Seems like this duplicates `reports/instrument_reports/report_tabs`. Instrument day reports just do a render on that partial.

app/views/reports/instrument_unavailable_reports/report_table.html.haml
```diff
@@ -0,0 +1,16 @@
+.updated_values{style: "display: none"}
```
jhanggi: This looks to duplicate reports/report_table.html.haml Can we share the partial?

If we're not going to share some of the partials, then we don't really need to abide by all the instrument reporter APIs and could probably make the reports much cleaner. I could go either way.

drench: The partials are *very* close, but they are different in how they decide which columns are numeric (and get the `.currency` css class). The existing `report_table` partial makes assumptions about column indexes, where this one explicitly defines the numeric columns.

I'd rather use something like the `@numeric_columns` check for the existing reports too. Add a `TODO` or dive in now?

app/views/reports/instrument_unavailable_reports/_report_rows.html.haml
drench: Partially duplicated. This is another place that uses `@numeric_columns` to determine when to apply the `.currency` class.

app/views/reports/instrument_unavailable_reports/report_table.html.haml
jhanggi: A TODO is probably fine for now.

jhanggi:
👍

PR #725, opened 2016-07-28
Update offline cancellation mail content

drench:
This brings the wording more in line with the upcoming offline reservation mail and offline warnings from the UI.

`/app/views/offline_cancellation_mailer/send_notification.*` and `/app/views/upcoming_offline_reservation_mailer/send_offline_instrument_warning.*` are identical now. Do we want to combine these? I could see them diverging.


app/views/offline_cancellation_mailer/send_notification.html.haml
```diff
@@ -1,4 +1,6 @@
--# TODO: content needs fleshing out
-
-Because #{@instrument} is unavailable, your reservation for
-#{@reservation.reserve_start_at} has been canceled.
+%p
+  = html "body.html",
+    facility_url: facility_url(@reservation.facility),
+    product: @reservation.product,
+    reservation_time: I18n.l(@reservation.reserve_start_at, format: :timeonly),
+    reservation_url: order_order_detail_reservation_url(@reservation.order, @reservation.order_detail, @reservation)
```
jhanggi: I'm thinking the order detail link might be a little more useful. It includes the time as well as the cost and other information. The reservation view doesn't have a link back to the order detail (which it probably should at some point)

jhanggi:
I could possibly see combining the two mailers into one `InstrumentOfflineMailer` with two methods. If we did that, then combining the views might make some sense. It would also make the previews listing a little more organized.

jhanggi:
I think we had discussed including the facility contact email address in this.

jhanggi:
Looks pretty good. Do we want to include the facility name somewhere in there?

config/locales/views/en.offline_cancellation_mailer.yml
```diff
@@ -0,0 +1,21 @@
...
+    offline_cancellation_mailer:
+      send_notification:
+        subject: Your reservation for %{product} has been canceled
+        body:
+          html: |
+            %{product}, the instrument you
+            [have reserved for %{reservation_time}](%{order_detail_url}), is down.
+            Your reservation has been canceled at no cost.
+
+            Please contact [%{facility_name}](%{facility_url})
```
drench: It may be safe to change this to "Please contact the `facility_name`" given the values I see in NU's database but can we count on that?

jhanggi: I think leaving out "the" is fine. We can change it if they say anything.

jhanggi:
👍

PR #727, opened 2016-07-29
Restrict Reservation.upcoming_offline

drench:
This scope should only include reservations whose status is either `new` or `inprocess` and should never include problem orders.


app/models/reservation.rb
```diff
@@ -120,6 +120,9 @@ def self.upcoming_offline(start_at_limit)
       .where(product_id: OfflineReservation.current.pluck(:product_id))
       .not_canceled
       .not_ended
+      .joins(:order_detail)
+      .where("order_details.state" => %w(new inprocess))
+      .where("order_details.problem" => false)
```
drench: Is it possible (outside changing the database directly) to have an `OrderDetail` be a problem order and also be in `new` or `inprocess`? Just wondering if it's necessary.

jhanggi: I forget exactly how it works, but I think the all the jobs move the order to completed. I would double check that.

I _think_ you can do `where(order_details: { state: %w(new inprocess), problem: false })`. Also, does this scope ignore un-purchased reservations sitting in carts? We could probably use `merge(OrderDetail.new_or_inprocess)` which includes that check as well.

jhanggi: Were you able to confirm how the job that sends reservations to the problem queue works? And does it ignore cart reservations?

app/models/reservation.rb
```diff
@@ -63,6 +63,7 @@ def self.joins_order
   scope :not_canceled, -> { where(canceled_at: nil) }
   scope :not_started, -> { where(actual_start_at: nil) }
   scope :not_ended, -> { where(actual_end_at: nil) }
+  scope :not_in_cart, -> { joins(:order).where.not("orders.ordered_at" => nil) }
```
jhanggi: Theoretically, we shouldn't need this additional check because `OrderDetail#state` _should_ be null until it's purchased, so the `state: %w(new inprocess)` _should_ be good enough. However, there's a weird case with merge orders that we recently discovered in #731 where merge orders that need attention (need a reservation or missing an order form) have `ordered_at` set and have the `state` set to `new`. In that PR, we get rid of the `ordered_at` setting, but the state piece seemed to complicated to deal with. 

We can probably ignore this edge case because if there's no reservation, we don't need to worry about this hitting it.

There's no harm in doing a check on purchased, so if you do think we want to keep it, I would recommend using the `purchased` state of order, not `ordered_at IS NULL`. (there's already a `purchased` scope on `OrderDetail` that does the `joins(:order).merge(Order.purchased)`. I'd also recommend sticking to `purchased` as the name of the scope if we add a new scope here for consistency.

jhanggi:
👍

PR #729, opened 2016-07-29
[#127295589] Add hook point in shared/head

jhanggi:
This will be used by DC for adding icon link tags. If we end up adding icons to other schools,
then I could see pulling the partial up and using a feature flag.


drench:
:+1:

PR #730, opened 2016-08-01
Support for asynchronous file processing

jhanggi:
This is going to be used by DC for their account uploads (it takes a minute or two to run). It will only allow one file to be in process at a time.

This also makes sure we're using the delayed job adapter for ActiveJob. Turns out that any of the emails we were sending using `deliver_later` were going out synchronously because we didn't turn this on.

I chose to have the job not raise an error on failure because it will continue to retry and we don't really want that in the case of bad input. I'm not currently exposing the error message as it's a backtrace. We'll see if errors become problematic, and if so then we can deal with it then.

I could see order imports being a good candidate to use this as well.

### DIFF
--- a/app/models/concerns/async_file_processing.rb
+++ b/app/models/concerns/async_file_processing.rb
@@ -1,0 +1,60 @@
+# Include this module in one of your models to support asynchronous file processing
+# via ActiveJob. It will only allow one instance of the model to be in process at
+# a time.
+#
+# Requirements:
+# * Model must have `status`, `processed_at`, and `error_message` columns, and
+#   should have "processing" as the default for the `status` column
+#   `add_column :status, default: "processing", null: false
+# * Model must implement `process_file!` (see `FileProcessingJob`)
+#   * `process_file!` should raise an error if anything goes wrong
+#
+# Example:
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -90,15 +90,16 @@ en:
 
   shared:
     cancel: Cancel
+    create: Create
```
drench: Yes: alphabetization in locales!

app/models/concerns/async_file_processing.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  delegate :processing?, :successful?, :failed?, to: :status_inquirer
+
+  def enqueue
+    return unless save
+    FileProcessingJob.perform_later(self)
+  end
+
+  def another_in_queue?
+    self.class.where(status: :processing).any?
```
drench: Consistency nitpick: symbol here where they're strings elsewhere?

drench:
One nitpick; :+1:

app/models/concerns/async_file_processing.rb
jhanggi: Good call. I'll change that.

PR #731, opened 2016-08-02
[#130559] Exclude non-purchased "add-to" orders from export raw

jhanggi:
Prevent order details that were “added to” another order by the
facility staff, but were not complete from appearing in export raw.
These would be instrument orders that don’t have a reservation or
service orders where the order form has not been uploaded.

This takes care of it in two ways, either one would work on its own:
* Remove the `ordered_at` from the merge order. Now the querier won’t
find an ordered_at to search. Once the merge order
is ready, the order details are moved onto the main order and then the order is destroyed. So that `ordered_at` that got set when we created the merge order goes away.
* Force the query to only return purchased orders. The merge order is in `new` state.

The only thing I’d like to figure out is how to have it keep a null
order status until the purchase is complete (this is how carts work).
This seemed more tricky than it was worth in dealing with the
OrderDetailObserver.

### DIFF
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -131,7 +131,6 @@ class FacilityOrdersController < ApplicationController
       account_id: @order.account_id,
       user_id: @order.user_id,
       created_by: current_user.id,
-      ordered_at: Time.zone.now,
     )
   end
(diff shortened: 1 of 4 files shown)


drench:
:+1:

PR #732, opened 2016-08-02
Remove sanger sample ON UPDATE foreign key

jhanggi:
Oracle doesn’t support ON UPDATE, and since it’s just the
auto-increment primary key, it’ll never change, so we don’t need to
worry about ON UPDATE.


drench:
:+1:

PR #733, opened 2016-08-02
[#127182] Use different feature flag for notification on results file upload

jhanggi:
Originally it used `my_files`, but NU will keep `my_files` on and turn these
notifications off. If we can find a good way to add a checkbox or somehow make
the notifications optional, then NU might want it.


drench:
:+1:

PR #735, opened 2016-08-04
Sanger Sequencing: Don't delete blanks from end of sample list

jhanggi:
Now, if you have blank customer sample ids at the end of the form, they no longer
get deleted. Instead, they'll get an error for "may not be blank".

This will help make things simpler in NU where the user might not be using the
default IDs (the core sample id is irrelevant to them).

This also adds a default ordering to Sample. MySQL returns ordered by id if no
ORDER BY is specified. Oracle does not do this, so the samples were getting out
of order there.


jhanggi:
Found a problem with this. I'll reopen once I have a fix.

jhanggi:
Weird. The problem I was having does not seem to be an issue here. I think I need to figure it out in the acgt form.

drench:
Oracle! :+1:

PR #736, opened 2016-08-04
Upgrade sass-rails to 5.0.6 to prevent deprecation warnings

jhanggi:
```
DEPRECATION WARNING: Sprockets method `register_engine` is deprecated.
Please register a mime type using `register_mime_type` then
use `register_compressor` or `register_transformer`.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -110,7 +110,7 @@ group :test do
 end
 
 group :assets do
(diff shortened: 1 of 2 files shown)


drench:
Tested against nucore-nu and the warnings are gone. :+1:

PR #737, opened 2016-08-05
Format instrument downtime hours to 2 decimal places in reports

drench:
This should format offline times in hours to two decimal places in all cases.


jhanggi:
👍

PR #738, opened 2016-08-05
Move jasmine files under sanger_sequencing namespace

jhanggi:
When using Teaspoon in the browser, it lumps all the engines’ specs
together and shows them alphabetically. This makes sure that it knows
to group them and we can see which engine they belong to.

I also got rid of an extraneous comment in the teaspoon config


drench:
:+1:

PR #739, opened 2016-08-05
Fix specs when My Files is disabled

jhanggi:
The results_file_notifications uses my_files_url in the mailer, so if notifications
are enabled, My Files needs to be as well. This fixes a broken spec when my files
was off.

It might be good at some point to address this so the mailer only includes the
my files link if it's enabled, but that's for another time.


drench:
:+1:

PR #740, opened 2016-08-08
[#127246] Add extension point for sanger submission params

jhanggi:
I’ll be using this in NU for adding a well plate fill order parameter.


drench:
:+1:

PR #742, opened 2016-08-12
Do not display sanger submission form errors in flash message

jhanggi:
If there are relevant validation errors, they will be displayed by simple form
near the appropriate field. In open, the only validation is the customer_sample_id,
but NU will have a few more.


drench:
:+1:

PR #743, opened 2016-08-15
[WIP] [#130628] Fragment Analysis

jhanggi:
This adds a new table that associates a Product with a "group", which in this case will only be "fragment", or nothing. The batches#new splits out the submissions based on which group the product is in. This will allow us to have multiple products in a group if we need to allow separate pricing. Fragment Analysis does not include control wells.

Once the batch is submitted it now has a group associated as well so it can be split into two separate views. 

The piece that is not done is a different plate file. I wanted to get some eyes and feedback on this before I moved forward to that piece.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20160802202924) do
+ActiveRecord::Schema.define(version: 20160812230426) do
 
   create_table "account_users", force: :cascade do |t|
     t.integer  "account_id", limit: 4,  null: false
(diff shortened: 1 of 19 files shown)


db/schema.rb
```diff
@@ -528,15 +528,26 @@
...
+    t.string   "group",           limit: 255
   end
 
   add_index "sanger_sequencing_batches", ["created_by_id"], name: "index_sanger_sequencing_batches_on_created_by_id", using: :btree
   add_index "sanger_sequencing_batches", ["facility_id"], name: "index_sanger_sequencing_batches_on_facility_id", using: :btree
+  add_index "sanger_sequencing_batches", ["group"], name: "index_sanger_sequencing_batches_on_group", using: :btree
 
-  create_table "sanger_sequencing_samples", force: :cascade do |t|
-    t.integer  "submission_id",      limit: 4,   null: false
+  create_table "sanger_sequencing_product_groups", force: :cascade do |t|
```
jhanggi: I'm thinking about using this for NU/ACGT and replacing the more specific `products.acgt_service_type`. Thoughts?

drench: I like how it could mean eliminating an engine-specific column to a "core" table like `products`.

vendor/engines/sanger_sequencing/README.md
```diff
@@ -29,3 +29,12 @@ Facility.find_by(url_name: "facility-name").update_attributes(sanger_sequencing_
...
 * Click "Activate" to turn it on
+
+### Enable Fragment Analysis Well Plate Creation
+
+* Create a Fragment Analysis Product
+* Add a "fragment" group under `sanger_sequencing_product_group` that maps to
+  the new product
+
+A product can only be part of a single group. If they are not part of a group,
+they fall back to a "default" group, which is the standard Sanger Sequencing.
```
drench: Singular "it" ought to be safe since it's a product.

drench:
Once I understood to add a row to `sanger_sequencing_product_groups` with group "fragment" for my `Service` it worked well for me. I'm all for using this table for ACGT service types, and the code issues all look Rubocop auto-correctable.

jhanggi:
Updated. Cool if we merge this and the plate file happens as a separate PR?

drench:
Looks good to merge for me :+1:

PR #744, opened 2016-08-16
[#126392] Handle empty facility email in offline notifications

drench:
This updates the offline notifications to handle the case where `Facility#email` is `nil`.

When `email` has a value, the emails should read "Please contact _Facility_ for more information at _address@example.com_" as before. If the `email` field is unset, it should read "Please contact _Facility_ for more information."

This change also consolidates common argument hashes into a helper.


jhanggi:
👍

PR #745, opened 2016-08-17
[#126394] Add warning icon to offline instruments in product listings

drench:
This adds the same red warning icon (with tooltip)  to the product listing page for offline instruments as seen on the daily view:

downtime-warning-icon


jhanggi:
👍

PR #746, opened 2016-08-17
{#126389] Show reservation "begin now" link only if the instrument is online

drench:
Commit cba973e should address the specific issue @nutodd brought up: if the instrument is down, the user should not see a "Begin Now" link.

Since `Reservation#can_move?` was only being used to determine if "Begin Now" should appear or not, this changes the method name in 8a54658 to `#startable_now?`, which sounded more appropriate to how we use it (it reads a bit less awkwardly in tests too).

### DIFF
--- a/app/presenters/reservation_user_action_presenter.rb
+++ b/app/presenters/reservation_user_action_presenter.rb
@@ -3,7 +3,7 @@ class ReservationUserActionPresenter
   attr_accessor :reservation, :controller
   delegate :order_detail, :order,
            :can_switch_instrument?, :can_switch_instrument_on?, :can_switch_instrument_off?,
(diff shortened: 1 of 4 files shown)


jhanggi:
I agree `can_move?` doesn't really reveal what it's for. I wonder if we can come up with something still better. I've always hated the term "Begin Now". It's always felt more like "Move forward" or something like that. Especially as it may move it forward to an hour from now if the instrument is booked for the next hour.

But other than that name, this looks good.

drench:
For the text, the best I've got so far is "Reschedule Earlier" though that might imply heading back to the scheduling screen to enter a new start time.

For the method name, maybe `startable_sooner?` or `advanceable?` would do.

jhanggi:
I think the name is fine for now. 👍

PR #747, opened 2016-08-17
[#130619] Do not send pending approval email on order import

jhanggi:
Also, do a little refactoring to the order state machine to remove a predicate
that had side-effects.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -39,7 +39,7 @@ class Order < ActiveRecord::Base
   aasm column: :state do
     state :new, initial: true
     state :validated
-    state :purchased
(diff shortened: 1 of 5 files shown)


app/models/order.rb
```diff
@@ -105,11 +109,8 @@ def clear_cart?
     save
   end
 
-  # set the ordered time and send emails
-  def place_order?
-    # set the ordered_at date
+  def set_ordered_at
```
drench: It's great to see a `?`-suffix method with side-effects go away (we have more).

drench:
:+1:

PR #748, opened 2016-08-18
[#128102345] Move bulk mail into an engine

drench:
This engine-izes the existing bulk mail feature. It should not look or function any differently than it had before.

Some notes:

This adds an `AbilityExtension` but it's a stub waiting for changes down the line. The new specs for the `send_bulk_emails` ability are here to document existing behavior.

The controller specs are virtually unchanged. While they could use an update, I plan to do that in a separate PR. I wanted to be sure the old specs remained green here.


Gemfile
```diff
@@ -57,6 +57,7 @@ gem "lograge"
 gem "logstash-event"
 
 ## custom
+gem "bulk_email", "~> 0.0.1", path: "vendor/engines/bulk_email"
```
jhanggi: Theoretically since we're just pointing to a path, we don't really need the version here. We could conceivably remove all of the version strings until they get extracted into an action gem.

app/views/admin/shared/_sidenav_users.html.haml
```diff
@@ -1,7 +1,11 @@
 %ul.sidebar-nav
   - if current_ability.can?(:administer, User)
     = tab t("activerecord.models.user", count: 2), facility_users_path, (sidenav_tab == "users")
-  - if current_ability.can?(:send_bulk_emails, current_facility)
-    = tab t("bulk_email.title"), facility_bulk_email_path, (sidenav_tab == "bulk_email")
+
+  = render_view_hook "after_facility_users",
+    ability: current_ability,
```
jhanggi: Won't the included partial have access to `current_ability` and `current_facility` helpers? Do we need to pass them in?

jhanggi:
👍

PR #749, opened 2016-08-18
Fix Oracle 1000 where in error on reconcilable accounts

jhanggi:
This is https://github.com/tablexi/nucore-nu/pull/188 pulled into open.


drench:
:+1:

PR #750, opened 2016-08-18
Upgrade RSpec to 3.5

drench:
This bumps RSpec to 3.5.1, and also includes a few version bumps where NU and DC were ahead of nucore-open: thin, oracle, web-console, poltergeist, parser.


jhanggi:
👍 with green CI.

PR #751, opened 2016-08-18
[#130862] Facilty Admin/Directors did not have access to error reports

jhanggi:
Also fixes an issue where facility staff/senior staff could access the
import if they had the URL. It was only protected by not showing the
tab.

### DIFF
--- a/app/views/admin/shared/_tabnav_order.html.haml
+++ b/app/views/admin/shared/_tabnav_order.html.haml
@@ -4,5 +4,5 @@
     = display_tab "Problem Orders", show_problems_facility_orders_path, action: :show_problems
   - if SettingsHelper.has_review_period? && current_ability.can?(:disputed, Order)
     = display_tab "Disputed", disputed_facility_orders_path, action: :disputed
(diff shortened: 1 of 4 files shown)


app/views/admin/shared/_tabnav_order.html.haml
```diff
@@ -4,5 +4,5 @@
     = display_tab "Problem Orders", show_problems_facility_orders_path, action: :show_problems
   - if SettingsHelper.has_review_period? && current_ability.can?(:disputed, Order)
     = display_tab "Disputed", disputed_facility_orders_path, action: :disputed
-  - if current_user.manager_of?(current_facility)
+  - if current_ability.can?(:new, OrderImport)
```
drench: TODO: Eliminate the remaining `operator_of?`/`manager_of?` checks in views and controllers

drench:
:+1: when green

app/views/admin/shared/_tabnav_order.html.haml
jhanggi: For sure. Also, we should I18n this file at some point.

PR #752, opened 2016-08-19
[#126394] Rename I18n key to use "_" rather than "-"

drench:
`text_helpers` was not translating "-".


jhanggi:
Glad to see that was a quick fix. Do we want to set a separate translation for offline?

Also, would it be worth making sure that I18n raises translation not found exceptions in dev/test? `text` does already, but `t` does not.

drench:
We were using text helpers in this case, but it did not raise an exception. There's something different in how it handles `-`.

PR #753, opened 2016-08-19
[#131810] Permit assigned user parameter

jhanggi:
The assigned user was not being persisted because the parameter was not
being permitted through.


drench:
:+1:

PR #754, opened 2016-08-19
 [#126394] Update public schedule calendar icon tooltips

drench:
This came about because text_helpers was not translating the "-" in "in-use" but "in use" was not accurate given an instrument might be unavailable because it's offline.

With this change:
* If an instrument is available and online, the tooltip should be "Available"
* If an instrument is offline, the tooltip should be "Instrument is down"
* If an instrument is otherwise unavailable, the tooltip should be "Unavailable"

Though it may not have caught this "in-use" missing translation, this change also configures the ~~`test` and~~ `development` environment~~s~~ to raise exceptions on missing translations.

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -58,12 +58,15 @@ module ProductsHelper
   end
 
   def public_calendar_availability_options(product)
-    if product.available? && product.online?
+    if product.offline?
+      { class: ["icon-calendar", "in-use"],
+        title: text("instruments.offline.note") }
+    elsif product.available?
       { class: ["icon-calendar", "available"],
         title: text("instruments.public_schedule.available") }
     else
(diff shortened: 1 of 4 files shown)


jhanggi:
Looks like we've got a bunch of test failures from the missing translation exceptions. Maybe we should do that as a separate PR and find those (first step: make sure they're actually valid errors)

jhanggi:
Should we keep the text as "In Use"? I suppose that it probably also shows that when we're outside of the configured schedule rules, so maybe "Unavailable" is more accurate.

drench:
I backed out the missing translation exceptions for `test` for the time being. Agreed, there's enough there to be a separate PR.

I've left the exceptions on for `development` so we can find legitimate ones as we run across them.

jhanggi:
👍

PR #756, opened 2016-08-22
[#130628] Shorten table name of Sanger ProductGroup

jhanggi:
The original name was too long for oracle.

I couldn’t think of a good way to add a new migration cleanly. This hasn’t been merged downstream anywhere, so it’s just developer environments that may be affected. So before you do a rake db:migrate with this merged, you should do a db:rollback.


drench:
So the Oracle max for table and column names is 30 chars. Another thing to watch out for. :+1:

PR #757, opened 2016-08-23
[#131810] Prevent duplicate users in staff lists

jhanggi:
* Order detail popup assigned user dropdown
* New/In Process Orders assigned user dropdown
* Admin Facility Staff index - Mentioned in NU #126595, but there are a
couple other issues still left around duplicate behavior.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -113,7 +113,6 @@ class User < ActiveRecord::Base
     facility
       .users
       .order("LOWER(user_roles.role), LOWER(users.last_name), LOWER(users.first_name)")
-      .distinct
   end
 
   #
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -109,15 +109,11 @@ class User < ActiveRecord::Base
   end
 
   # Find the users for a facility
-  # TODO: move this to facility?
   def self.find_users_by_facility(facility)
-    find_by_sql(<<-SQL)
-      SELECT u.*
-      FROM #{User.table_name} u
-      LEFT JOIN #{UserRole.table_name} ur ON u.id=ur.user_id
-      WHERE ur.facility_id = #{facility.id}
-      ORDER BY LOWER(ur.role), LOWER(u.last_name), LOWER(u.first_name)
-    SQL
+    facility
+      .users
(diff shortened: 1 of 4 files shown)


spec/models/user_spec.rb
```diff
@@ -271,4 +271,20 @@
...
     end
   end
+
+  describe ".find_users_by_facility" do
+    let!(:facility_director) { create(:user, :facility_director, facility: facility) }
+    let!(:staff) { create(:user, :staff, facility: facility) }
+    let!(:normal_user) { create(:user) }
+    let!(:other_facilitly_director) { create(:user, :facility_director, facility: build_stubbed(:facility)) }
+    let!(:facility_admin_and_director) do
+      create(:user, :facility_director, facility: facility) do |u|
```
drench: Would `FactoryGirl.create(:user, :facility_administrator, :facility_director, facility: facility)` work here?

app/models/user.rb
```diff
@@ -109,15 +109,11 @@ def update_password(params)
...
-    find_by_sql(<<-SQL)
-      SELECT u.*
-      FROM #{User.table_name} u
-      LEFT JOIN #{UserRole.table_name} ur ON u.id=ur.user_id
-      WHERE ur.facility_id = #{facility.id}
-      ORDER BY LOWER(ur.role), LOWER(u.last_name), LOWER(u.first_name)
-    SQL
+    facility
+      .users
+      .order("LOWER(user_roles.role), LOWER(users.last_name), LOWER(users.first_name)")
```
drench: So much nicer than the previous version. I'm still wondering if this could be turned into `Facility#users`. Does `Facility#users` get used anywhere outside this method?

jhanggi: Possibly not, but this method was used in enough places, that I didn't want to deal with moving changing the consumers of this method. I am going to try out moving the `distinct` ont `Facility#users`, though. Then really this is just adding an ordering.

spec/models/user_spec.rb
jhanggi: Good call. I like that.

drench:
:+1:

PR #758, opened 2016-08-23
Fix missing translation errors

jhanggi:
* Facility admin show/edit
* Facility Staff listing

There are probably more, but I ran across these in #757.


drench:
:+1:

PR #760, opened 2016-08-24
[#130628] Fragment Analysis plate files

jhanggi:


vendor/engines/sanger_sequencing/app/presenters/sanger_sequencing/well_plate_presenter.rb
```diff
@@ -2,20 +2,23 @@ module SangerSequencing
...
+    def initialize(well_plate, plate_mode = :default)
       @well_plate = well_plate
+      @plate_mode = plate_mode
     end
 
     def to_csv
       CSV.generate do |csv|
         csv << ["Container Name", "Plate ID", "Description", "ContainerType", "AppType", "Owner", "Operator", "PlateSealing", "SchedulingPref"]
-        csv << ["",               "",         "",            "96-Well",       "Regular", "mb core", "mbcore", "Septa", "1234"]
+        csv << ["",               "",         "",            "96-Well",       "Regular", text("owner"), text("operator"), "Septa", "1234"]
```
drench: Doesn't necessarily have to happen immediately but any benefit in I18n-ing the other fields?

jhanggi: I was thinking about that, and also about putting some of the translations into DC, but that seemed more effort than it was worth for right now. If another school starts using this feature, and they happen to use the same machine and format, we can deal with it then.

drench:
:+1:

PR #761, opened 2016-08-25
[#130754] Add "Export Raw" report for instrument unavailable reservations

drench:
This adds an "Export Raw" link when the "Instrument Unavailable" reporting tab is selected, though I think these changes might help make adding raw exports for other tabs in the future a bit easier.

There's a lot of new extractions here, though there's probably more we could do. I specifically put `TODO` notes on the various report-related HAML files that share a lot and might benefit from consolidation.

The "Export Raw" link itself is also a bit of a `TODO`. On the instrument reporting page, we show and hide the link depending on which report tab is active. Currently that means it's visible for "Instrument Unavailable" and hidden for the others. Because there's just one type of instrument export raw report, it was easier to set the link to point to the instrument unavailable report at page load time and leave it. But if we ever add other types of "raw" instrument reports, we'll need to make this link's href dynamic.


app/controllers/concerns/reports/csv_export_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    private
+
+    def email_to_address
+      params[:email_to_address].presence || current_user.email
+    end
+
+    def generate_report_data_csv
+      ExportRawReportMailer.delay.raw_report_email(email_to_address, raw_report)
```
jhanggi: ActiveJob syntax?

app/controllers/reports/instrument_unavailable_export_raw_reports_controller.rb
```diff
@@ -0,0 +1,23 @@
+module Reports
+
+  class InstrumentUnavailableExportRawReportsController < ReportsController
+
+    include CsvExportController
```
jhanggi: Would it make more sense to have `CsvExportController` be a class and have it inherit from `ReportsController`?

app/models/reports/instrument_unavailable_export_raw.rb
```diff
@@ -0,0 +1,90 @@
...
+        ReservationRowDurationCalculator
+        .new(reservation, start_time: date_start.beginning_of_day, end_time: date_end.end_of_day)
+        .duration_in_seconds
+
+      format("%.2f", (duration_in_seconds / 60.0))
+    end
+
+    def default_report_hash
+      {
+        type: -> (reservation) { reservation.type.sub(/Reservation\z/, "") },
```
jhanggi: Should we use something more like `model_name.human`? Seems like locales would be a better way of handling this, especially since we're displaying in the type in a couple places.

app/models/reports/instrument_unavailable_export_raw.rb
```diff
@@ -0,0 +1,90 @@
...
+        instrument: -> (reservation) { reservation.product.name },
```
jhanggi: Can we do just `:product`? I think it's `to_s` just calls `name`.

app/models/reports/instrument_unavailable_export_raw.rb
```diff
@@ -0,0 +1,90 @@
...
+        start_time: :reserve_start_at,
+        end_time: :reserve_end_at,
+        total_minutes: -> (reservation) { formatted_total_minutes(reservation) },
+        admin_note: :admin_note,
+        category: -> (reservation) { reservation_category_label(reservation.category) },
+      }
+    end
+
+    def reservation_row(reservation)
+      report_hash.values.map do |callable|
```
jhanggi: We should probably share this with the other report.

app/controllers/reports/export_raw_reports_controller.rb
```diff
@@ -24,14 +17,8 @@ def raw_report
...
-    def generate_report_data_csv
-      ExportRawReportMailer.delay.raw_report_email(email_to_address, raw_report)
-      if request.xhr?
-        render nothing: true
-      else
-        flash[:notice] = I18n.t("controllers.reports.mail_queued", email: email_to_address)
-        redirect_to send("#{action_name}_facility_general_reports_path", current_facility)
-      end
+    def success_redirect_path
+      public_send("#{action_name}_facility_general_reports_path", current_facility)
```
jhanggi: Looks like this was actually wrong on master. I'm getting `undefined method export_all_facility_general_reports_path'` if I turn of JS.  `action_name` will always be `export_all` in this case, so this would probably cause infinite redirects.

Doesn't really matter that much because it should always be getting hit via ajax.

jhanggi: I got infinite redirects when I tried `redirect_to action: :export_all, facility: current_facility`

app/views/reports/instrument_reports/report_table.html.haml
```diff
@@ -0,0 +1,20 @@
+-# TODO: this is identical to instrument_day_reports/report_table
+
+.updated_values{style: "display: none"}
+  .date_start= human_datetime(@date_start, date_only: true)
+  .date_end= human_datetime(@date_end, date_only: true)
+
+.export_raw{data: { visible: "false" } }
```
jhanggi: Is this line the only difference between these and `reports/report_table`? Would it be cleaner to set an instance variable in the controller to decide whether to be visible or not and use the same partial everywhere? It would be nice to be able to move `xhr_html_template` back to private.

app/controllers/concerns/reports/csv_export_controller.rb
drench: Good catch; this is a bit more involved since `deliver_later` complains about serializing `raw_report`.

app/models/reports/instrument_unavailable_export_raw.rb
drench: Yes, and this also brings up the fact that there is no defined locale for `AdminReservation`. Fixing that now.

app/controllers/concerns/reports/csv_export_controller.rb
jhanggi: This is kind of a weird one, I think. It probably tries to convert it to Global ID, but we actually need to drop in the full report. I think we're generating the report in the app, and then just delaying the email sending. Ideally, we would delay the report creation. If it's too complicated, it might not be worth it.

drench: I added a `TODO` comment. `git grep '\.delay\.'` shows a few more that could use `deliver_later` conversion.

jhanggi:
👍

PR #762, opened 2016-08-25
Fix Oracle error on loading facility staff (#191)

jhanggi:
`ORA-01791: not a SELECTed expression` because we're not SELECTing the
user_roles table. There's no real reason we're sorting by the role first.

Pulled up from https://github.com/tablexi/nucore-nu/pull/191


drench:
:+1:

PR #763, opened 2016-08-26
[#132121] Handle more than 1000 accounts receivable

jhanggi:
On the cross-facility tab, we were hitting 1000+ errors on oracle. This
view is still slow based on the #, but this will stop the errors.


drench:
:+1:

PR #764, opened 2016-08-26
Update I18n load_path

jhanggi:
The UIC load path was not in the correct order, but only in production. I could not figure out what was happening differently between there and stage, and why it has not been a problem for NU or DC.

I did figure out that `text_helpers` adds the subdirectories `**/*.{rb,yml}`. That happens after this initialization, and appeared to be loading everything again, so we would have `/en.yml` multiple times, which could be after the `override/en.yml`.

This now waits until after that initialization and forces the `override/` directory to happen at higher priority in the load path.

This is already in UIC, and I'm testing it on DC.


drench:
:+1:

PR #765, opened 2016-08-26
[#127094] Normal users could place orders in the past

jhanggi:
The big problem is that the account assignment on `order_detail` was autosaving the reservation without enforcing the non-admin validations (`save_as_user`).

The first idea I tried was doing `update_column`s so callbacks didn't get triggered, but it turns out we do need validations.

Next idea was flipping `save` and `save_with_extended_validations` so the extended validations (in the future and within 14 day window) are enforced by default, but not when purchasing as an admin. However, too many of our tests rely on those validations not being enforced as part of setup.

This refactor makes sure that `reservation` is not built until the last moment possible.

### DIFF
--- a/app/services/reservation_creator.rb
+++ b/app/services/reservation_creator.rb
@@ -1,0 +1,92 @@
+class ReservationCreator
+
+  attr_reader :order, :order_detail, :params, :error
+
+  delegate :merged_order?, :instrument_only_order?, to: :status_q
+
+  def initialize(order, order_detail, params)
+    @order = order
+    @order_detail = order_detail
+    @params = params
+  end
+
+  def save(session_user)
+    if !@order_detail.bundled? && params[:order_account].blank?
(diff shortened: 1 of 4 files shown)


app/services/reservation_creator.rb
```diff
@@ -0,0 +1,92 @@
...
+        elsif @order.order_details.one?
+          @success = :instrument_only_order
+        else
+          @success = :default
+        end
+      rescue ActiveRecord::RecordInvalid => e
+        @error = e.message
+        raise ActiveRecord::Rollback
+      rescue => e
+        @error = I18n.t("orders.purchase.error")
```
drench: Minor: since there are no tests for this class directly I checked code coverage. This `rescue` block looks like the only part of the class that isn't getting tested.

drench:
Nice extraction and cleanup. One comment on test coverage.

drench:
:+1:

PR #766, opened 2016-08-30
[#126393] Update instrument down category list

drench:
This updates the list of offline categories to use the one NU provided. This change also uses locales as the source of the category list.


jhanggi:
👍

PR #767, opened 2016-09-01
403 on Sanger Submission controller should 403 instead of error

jhanggi:
The previous commit was cherry-picked out of NU, which was not running the sanger_sequencing engine specs (which it probably shouldn’t many features have been overridden there). This should fix the failure on master.


drench:
:+1:

PR #768, opened 2016-09-02
[#128102841] Bulk mail recipient search form update

drench:
This updates bulk mail recipient search form and search backend. The main difference is the use of checkboxes to select user classes ("Customers", "Authorized Users", "Account Owners"). This should otherwise work as before: the user sets search filters and is given a list of recipients to view and download as a CSV.

The label alignment with the checkboxes looks slightly off, but this appears to be app-wide.


spec/factories/products.rb
```diff
@@ -7,6 +7,12 @@
     is_hidden false
     initial_order_status_id { FactoryGirl.create(:order_status, name: "New").id }
 
+    after(:build) do |product|
+      if product.facility_account.present?
```
drench: It's common in specs to set both a `facility` and `facility_account` when creating a `Product`. Having to explicitly set `facility` felt redundant, so I added this.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_fields.html.haml
```diff
@@ -2,17 +2,27 @@
 
   .row
     .span3
-      = f.input :search_type, collection: @search_types.invert, selected: params[:search_type], input_html: { name: "search_type" }
+      - @user_types.each do |user_type, label_text|
+        = label_tag(user_type) do
+          = check_box_tag "bulk_email[user_types][]",
```
drench: I did not find a way to get `collection_check_boxes` to generate what this does.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_export.html.haml
```diff
@@ -1,10 +1,12 @@
-= form_tag facility_bulk_email_path, :method => :get, :id => "bulk_email_export" do
+= form_tag facility_bulk_email_path, method: :get, id: "bulk_email_export" do
   = hidden_field_tag :format, :csv
-  = hidden_field_tag :search_type, @search_fields[:search_type]
+  - @search_fields[:bulk_email][:user_types].each do |user_type|
```
jhanggi: Take a look at the HTML output on the Facility or Product edit forms. Most of those checkboxes are aligned properly.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -37,16 +39,26 @@ def init_search_options
...
-      @search_types.delete(:authorized_users) unless @products.exists?(requires_approval: true)
+      @user_types = user_types
+      @user_types.delete(:authorized_users) unless @products.exists?(requires_approval: true)
     end
 
-    def search_types
-      BulkEmailSearcher::SEARCH_TYPES.each_with_object({}) do |search_type, hash|
-        hash[search_type] = I18n.t("bulk_email.search_type.#{search_type}")
+    def datepicker_field_input(form, key)
+      date = @search_fields[key].presence.try(:tr, "-", "/") || ""
```
jhanggi: How do we deal with this elsewhere? I didn't think we needed this replacement in most places.

vendor/engines/bulk_email/app/services/bulk_email/bulk_email_searcher.rb
```diff
@@ -7,15 +7,22 @@ class BulkEmailSearcher
...
+        USER_TYPES & (search_fields[:bulk_email].try(:[], :user_types) || []).map(&:to_sym)
+    end
+
     def do_search
-      return unless SEARCH_TYPES.include? search_fields[:search_type].to_sym
-      public_send(:"search_#{search_fields[:search_type]}")
+      return if user_types.blank?
+      user_types.map do |user_type|
+        public_send(:"search_#{user_type}")
+      end.sum.uniq
```
jhanggi: Would this read a little more cleanly if we said `flatten`? And maybe `collect` instead of `map` here?

vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js
```diff
@@ -1,9 +1,19 @@
 $(function() {
+
+  function selected_user_types() {
```
jhanggi: I _think_ we've been moving towards using camel case for JS function and variable names.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_export.html.haml
drench: Thanks - it turns out the `label` tag having the `checkbox` class is crucial. There are other checkbox labels in the app missing this; `/facilities/:id/facility_accounts/:id/edit` has one example.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
drench: Datepicker disallows typing characters other than `/` and digits, but it doesn't stop pasting. I don't have the reference, but I remember a situation with dates being pasted in "mm-dd-yyyy" format being an issue before.

The searcher service object converts `-` to `/` before parsing dates, but the `param` value remains the same and becomes a blank when displaying the form back to the user after a search.

We're not really consistent: `facilities/:id/instruments/:id/price_policies/new` accepts either "mm-dd-yyyy" or "yyyy-mm-dd" (the latter may be a paste workaround as well).

jhanggi: Ok, I was thinking we had a standard way of dealing with this. I think it's fine.

Would `@search_fields[key].to_s.tr("-", "/")` be equivalent? It would get rid of the `try` and the `||`.

vendor/engines/bulk_email/app/services/bulk_email/bulk_email_searcher.rb
```diff
@@ -7,15 +7,22 @@ class BulkEmailSearcher
...
-    SEARCH_TYPES = [:customers, :account_owners, :customers_and_account_owners, :authorized_users].freeze
+    USER_TYPES = %i(customers authorized_users account_owners).freeze
 
     def initialize(search_fields)
       @search_fields = search_fields
     end
 
+    def user_types
+      @user_types ||=
+        USER_TYPES & (search_fields[:bulk_email].try(:[], :user_types) || []).map(&:to_sym)
```
jhanggi: I feel like the `try`/`||` pattern is not always as clear as it could be. Are we ever using anything in `search_fields` that's not `:bulk_email`? It might be easier to deal with this if we made `@search_fields = Array(search_fields[:bulk_email])` in the contructor. 

It might also be a little clearer here to introduce an intermediate variable for the user inputs.

vendor/engines/bulk_email/app/services/bulk_email/bulk_email_searcher.rb
```diff
@@ -7,15 +7,22 @@ class BulkEmailSearcher
...
+    end
+
     def do_search
-      return unless SEARCH_TYPES.include? search_fields[:search_type].to_sym
-      public_send(:"search_#{search_fields[:search_type]}")
+      return if user_types.blank?
```
jhanggi: Is this check necessary anymore now that `user_types` is always sanitized and returns an array?

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -37,16 +39,26 @@ def init_search_options
...
+      form.input(key, input_html: { value: date, class: "datepicker", name: key })
```
jhanggi: This might get annoying to fix because it looks to be something more global, but right now the maximum date for date_start is today. That's probably right for reporting and transaction searching, but in this scenario, we probably want to be able to select something like "users with reservations next week".

vendor/engines/bulk_email/app/services/bulk_email/bulk_email_searcher.rb
drench: I had taken this out but put it back because if `#do_search` returns an empty array, the template will say "No users found", which we don't want when a user first lands on `/facilities/:id/bulk_email`. There's probably a clearer way to accomplish this.

drench: …which also means if a user deselects all checkboxes and does a search, it will *not* say "No users found" but probably should (or something more specific, like "Choose at least one user type").

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
drench: It turned out not that annoying: we can use `datepicker__data` here with no `min-date` or `max-date` specified.

vendor/engines/bulk_email/app/services/bulk_email/bulk_email_searcher.rb
jhanggi: Could we do a check in the view to look at another parameter (I think `commit` is always there if it's a submitted form)? That might be a cleaner way to determine if someone is just landing or if they're submitting.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -37,16 +39,26 @@ def init_search_options
...
+      @user_types = user_types
+      @user_types.delete(:authorized_users) unless @products.exists?(requires_approval: true)
     end
 
-    def search_types
-      BulkEmailSearcher::SEARCH_TYPES.each_with_object({}) do |search_type, hash|
-        hash[search_type] = I18n.t("bulk_email.search_type.#{search_type}")
+    def datepicker_field_input(form, key)
+      date = @search_fields[key].to_s.tr("-", "/")
+      form.input(key, input_html: { value: date, class: "datepicker datepicker__data", name: key })
```
jhanggi: Do we need the `datepicker` class now? This isn't going to get initialized multiple times, is it?

vendor/engines/bulk_email/app/services/bulk_email/bulk_email_searcher.rb
jhanggi: Any thoughts on this comment? ^

drench: I see this as somewhat related to the blank `@user` issue; the try exists because there's otherwise no way to tell the difference between hitting the search form landing page and actually doing a search. I have a patch coming up.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -15,17 +15,18 @@ class BulkEmailController < ApplicationController
...
+    helper_method :user_type_selected?
+
     def search
-      if params[:search_type]
-        searcher = BulkEmailSearcher.new(@search_fields)
-        @users = searcher.do_search
-        @order_details = searcher.order_details
-      end
+      @searcher = BulkEmailSearcher.new(@search_fields)
+      @users = @searcher.do_search
```
drench: Just a note that nothing appeared to be using the `@order_details` results so I removed it. A lot of the tests focused on orders, but I've updated them to look at `@users` instead.

vendor/engines/bulk_email/spec/services/bulk_email/bulk_email_searcher_spec.rb
```diff
@@ -17,228 +17,346 @@
...
+        ]
+      end
+
+      it_behaves_like "active/inactive users" do
+        let(:user) { purchaser }
+      end
+
+      context "when selecting no specific product" do
+        it "returns users who are customers of all products" do
+          expect(users).to contain_exactly(purchaser, purchaser2, purchaser3)
```
drench: I'm not sure this is the behavior we want (no product selected == all products selected)?

drench: I could see it being useful though, since we don't otherwise offer a blanket "anybody who purchased anything" option.

jhanggi: Yeah, I do think we want this behavior.

drench:

It struck me during this process that the service object could have a better name than `BulkEmailSearcher` (since it doesn't search bulk emails) but for the sake of the diffs I want to save that rename for a future PR.

jhanggi:
Rename sounds good to me, but I agree it's best saved for a separate PR.

👍

PR #770, opened 2016-09-07
[#132075] Increase size of instrument down icon on user-facing view

jhanggi:


drench:
Do we want this larger on the daily view too? If not, :+1:

drench:
:+1:

PR #771, opened 2016-09-09
[#128102967] Allow bulk email recipient de/selection for export

drench:
This adds checkboxes to the bulk email recipient search result list, allowing the user to select or deselect recipients for export. It should be Oracle-1001-proof.

It's tagged WIP while I work out a visual issue that's bugging me where the recipient list table shifts up and down depending on if the "Export" button is visible.


vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js
```diff
@@ -31,4 +31,15 @@ $(function() {
     return false;
   })
 
+  function showHideRecipientExportButton() {
+    if ($('.js--bulk-email-recipient').is(':checked')) {
+      $('.js--bulk-email-export-button').show();
```
jhanggi: What about disabling it so it's grayed out? Then you don't have to worry about the shifting.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -21,15 +21,12 @@ class BulkEmailController < ApplicationController
...
-      respond_to do |format|
-        format.html
-
-        format.csv do
-          filename = "bulk_email_#{params[:bulk_email][:user_types].join('-')}.csv"
-          set_csv_headers(filename)
-        end
-      end
+    def download
+      @users = User.where_ids_in(params[:recipient_ids])
```
jhanggi: I don't think we need to worry about this, but should we add any sort of constraints around this? People could get access to other facilities user lists. Maybe `BulkEmailSearcher` has an `all` method that gives us the widest list within a facility. Not sure if it's worth dealing with, but something to think about.

vendor/engines/bulk_email/spec/controllers/bulk_email/bulk_email_controller_spec.rb
```diff
@@ -112,19 +112,31 @@
...
+    end
 
-      it "sets the filename" do
-        expect(response.headers["Content-Disposition"])
-          .to eq('attachment; filename="bulk_email_customers.csv"')
-      end
+    before(:each) do
+      @action = "download"
+      @method = :post
+      @params.merge!(format: :csv, recipient_ids: recipients.map(&:id))
```
jhanggi: I don't really care about the 

> Use `@params[:format] = :csv; @params[:recipient_ids] = recipients.map(&:id)` instead of `@params.merge!(format: :csv, recipient_ids: recipients.map(&:id))` cop. 

I'd be fine if we disabled that one.

jhanggi:
Looking pretty good so far.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
drench: Yeah, we would need to hang on to more of the search criteria because otherwise users don't really "belong" to a facility. I'm wondering: does this open anything new to a user who also has access to search `/facilities/:id/users`?

drench:
This is ready for review. It does lack feature specs. I looked into testing file downloads using http://stackoverflow.com/a/29544674 but was getting timeout errors and felt like I was hitting a wall. It seems like it would be nice to have though.

vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js
```diff
@@ -31,4 +31,16 @@ $(function() {
     return false;
   })
 
+  function showHideRecipientExportButton() {
+    $downloadButton = $('.js--bulk-email-export-button');
```
drench: Just noticed this global. I guess I thought I was writing coffeescript.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_fields.html.haml
```diff
@@ -19,7 +19,7 @@
   .row
     %h4.span8= t("bulk_email.dates.label")
 
-  .row
+  #dates_between.row
```
drench: I think we lost this `id` in 8c91b7d7a788b3a9def5339e5f3f48069aa48244. It might be better as a `.js--*` class now but this should restore the enable/disable date field behavior without changing the javascript for now (it's going to change in a near-future PR).

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
jhanggi: I don't think it's anything newer than that, so I think we're probably ok. I wouldn't worry about it. I just wanted to raise the idea.

jhanggi:
👍

PR #772, opened 2016-09-14
[#128113219] Compose and send bulk email

drench:
This adds, for global admins, a "Compose Mail" button as an alternate flow to the CSV download of recipients. This allows the user to send mail (through `delayed_job`) to the selected users after specifying a subject line and a custom (text-only) message.

It's tagged "WIP" because there are a few aspects I'd like to revisit:
* If selecting a large number of recipients, the display on the compose page could use some kind of overflow handling.
* ~~Hitting "cancel" on the compose page takes the user back to `/facilities/:id/bulk_email`, losing any search options or selected recipients from the previous step.~~
* The custom text message is wrapped within boilerplate text, though this standard text isn't shown to the user.
* The email boilerplate text is minimal and could use expansion.


vendor/engines/bulk_email/app/views/bulk_email/bulk_email/create.html.haml
```diff
@@ -0,0 +1,29 @@
...
+
+  .span8= f.input :subject, input_html: { class: "input-xxlarge" }
+
+  .span8
+    = f.input :custom_message,
+      as: :text,
+      input_html: { class: "input-xxlarge", rows: 6 }
+
+  .span8
+    = link_to text("shared.cancel"), facility_bulk_email_path, class: "btn"
```
jhanggi: I think our usual styling for "Cancel" is just a link and to the right of the submit button.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/create.html.haml
```diff
@@ -0,0 +1,29 @@
...
+    = f.submit text("bulk_email.send_mail"), class: "btn btn-primary"
```
jhanggi: I would probably preemptively add a `disable_with` to this to avoid double clicks that could cause duplicate emails.

vendor/engines/bulk_email/spec/controllers/bulk_email/bulk_email_controller_spec.rb
```diff
@@ -139,4 +139,41 @@
...
+
+  describe "POST #deliver" do
+    let(:recipients) { FactoryGirl.create_list(:user, 3) }
+    let(:custom_message) { "Custom message" }
+
+    before(:each) do
+      @action = "deliver"
+      @method = :post
+      @params[:bulk_email_delivery_form] = {
+        subject: subject_line,
```
jhanggi: Should we prefix the subject line with anything?

vendor/engines/bulk_email/config/locales/en.yml
```diff
@@ -11,4 +11,20 @@ en:
       start: Starting
       end: Ending
     no_results: No users found
-    export_html: Export
+    compose_mail: Compose Mail
+    export: Export
+    body:
+      greeting: Hello %{recipient_name},
+      signoff: If you have any questions, please contact %{facility_name}.
```
jhanggi: I know we're still figuring this out, but should we link to the facility?

jhanggi: And I'm thinking we should probably say something about being sent on behalf of `!app_name!`. We can punt on that, though.

vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js.coffee
```diff
@@ -0,0 +1,59 @@
...
+    user_types = @selectedUserTypes()
+    user_types.length == 1 && user_types[0] == 'authorized_users'
+
+  selectedUserTypes: -> @$userTypeCheckboxes().filter(':checked').map -> @.value
+
+  toggleNonRestrictedProducts: ->
+    # Hide non-restricted items when doing an authorized_users search
+    isHideNonRestrictedProducts = @authorizedUsersSelectedOnly()
+    @$form.find('#products option[data-restricted=false]').each ->
+      $option = $(this)
```
drench: I changed it, though I'm not totally sold on using `@` in this context.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -11,26 +11,50 @@ class BulkEmailController < ApplicationController
...
+        @users = User.where_ids_in(@delivery_form.recipient_ids)
+        render :create
+      end
     end
 
     private
 
+    def init_delivery_form
+      @delivery_form = DeliveryForm.new(current_facility)
+      @delivery_form.assign_attributes(params) if params[:bulk_email_delivery_form].present?
```
drench: Good catch: it was like this because an earlier version did not have `recipient_ids` nested within `:bulk_email_delivery_form`. It's there now, so there's no longer a need to send all of `params`.

vendor/engines/bulk_email/spec/controllers/bulk_email/bulk_email_controller_spec.rb
drench: Updated to prefix this with "[{app_name} {facility_name}]"

vendor/engines/bulk_email/config/locales/en.yml
drench: Linking to the facility is a good idea. I'd like to get feedback from NU on how they'd like this to read.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -11,26 +11,64 @@ class BulkEmailController < ApplicationController
...
+    def bulk_email_cancel_path
+      if cancel_params.present?
+        facility_bulk_email_path(cancel_params)
+      else
+        facility_bulk_email_path
+      end
+    end
+
+    def cancel_params
+      @cancel_params ||= params.select do |field, _value|
```
jhanggi: Does `params.slice(%w(start_date end_date bulk_email products))` get what you want?

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -11,26 +11,64 @@ class BulkEmailController < ApplicationController
...
+        format.html
+      end
+    end
+
+    def deliver
+      if @delivery_form.deliver_all
+        flash[:notice] = text("bulk_email.delivery.success", count: @delivery_form.recipient_ids.count)
+        redirect_to facility_bulk_email_path
+      else
+        @users = User.where_ids_in(@delivery_form.recipient_ids)
```
jhanggi: Should we have some sort of flash message?

vendor/engines/bulk_email/spec/services/bulk_email/recipient_searcher_spec.rb
```diff
@@ -73,6 +73,33 @@ def place_order(purchaser:, product:, account:)
...
+      {
+        bulk_email: { user_types: [:customers, :account_users] },
+        commit: "Submit",
+        facility_id: facility.id,
+        start_date: "1/1/2009",
+        end_date: "12/31/2016",
+        products: [1, 3, 5],
+      }
+    end
+    let(:fragment) { Nokogiri::HTML(searcher.search_params_as_hidden_fields.join("")) }
```
jhanggi: I just found `Capybara.string`, which in theory _should_ work here to give you Capybara's matchers. For some reason, though, I couldn't make it work. Didn't seem to parse the string correctly.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
drench: It would take something weird like insert into the delayed jobs table failing to make it fail, but if something like that does happen there ought to be a message. Added.

drench:
Just a note that this does not include a mailer preview. It turns out adding previews in an engine is not really supported. It's possible to set `config.action_mailer.preview_path` in an engine, but then you lose the previews in the main app.

There was some activity toward allowing multiple directories here https://github.com/rails/rails/pull/16836 but it doesn't look like it's moved in 2 years.

PR #773, opened 2016-09-14
[#132309] Always use fulfilled at for price calculation

jhanggi:
If you change the account of an order, we were calculating the price policy based on the current time, not the fulfilled_at date. This removes the parameter from the `assigns_price_policy` method to make sure we don't run into this again.

The one time we do want to use current time is when canceling a reservation with fee, because we don't really have a fulfilled_at, so we should use the current time.

This also cleans up an unused argument in `assign_estimated_price`.

I have two pending tests for another issue: if you change the account of an order in the popup, and also modify the cost, we're recalculating the pricing in the before_save hook and overwriting whatever the user put in. That fix will likely come as a separate PR.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -468,7 +468,7 @@ class OrderDetail < ActiveRecord::Base
     end
     change_status!(OrderStatus.complete.first) do |od|
       od.fulfilled_at = event_time
-      od.assign_price_policy(event_time)
+      od.assign_price_policy
     end
   end
 
@@ -655,10 +655,9 @@ class OrderDetail < ActiveRecord::Base
     JournalRowUpdater.new(self).update
   end
 
-  def assign_estimated_price(second_account = nil, date = Time.zone.now)
+  def assign_estimated_price(date = fulfilled_at || Time.current)
     self.estimated_cost    = nil
     self.estimated_subsidy = nil
-    second_account = account unless second_account
 
     # is account valid for facility
     return unless product.facility.can_pay_with_account?(account)
@@ -667,8 +666,8 @@ class OrderDetail < ActiveRecord::Base
     assign_estimated_price_from_policy @estimated_price_policy
   end
 
-  def assign_estimated_price!(second_account = nil, date = Time.zone.now)
-    assign_estimated_price(second_account, date)
+  def assign_estimated_price!
+    assign_estimated_price(Time.current)
     save!
   end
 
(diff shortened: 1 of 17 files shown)


spec/controllers/order_management/order_details_controller_spec.rb
```diff
@@ -495,7 +495,17 @@ def reservation_params(reserve_start_at, actual_start_at = nil)
         expect(order_detail.reload.actual_total).to eq(16.00)
       end
 
-      it "returns an error when trying to set subsidy more than quantity" do
+      xit "updates the price while changing accounts" do
```
drench: As long as this is going to be short-lived I'm OK with `xit` rather than a more formal `pending`.

spec/controllers/order_management/order_details_controller_spec.rb
```diff
@@ -513,6 +537,17 @@ def reservation_params(reserve_start_at, actual_start_at = nil)
       it "updates the quanity" do
         expect { do_request }.to change { order_detail.reload.quantity }.to(2)
       end
+
+      xit "updates the price while changing quantity" do
```
drench: Same comment on `xit` vs. `pending`.

app/models/order_detail.rb
```diff
@@ -856,9 +855,17 @@ def has_completed_reservation?
     !product.is_a?(Instrument) || (reservation && (reservation.canceled_at || reservation.actual_end_at || reservation.reserve_end_at < Time.zone.now))
   end
 
+  def time_for_policy_lookup
+    if fulfilled_at
+      fulfilled_at
+    elsif reservation.try(:canceled?)
+      Time.current
```
drench: Could this method just be `fulfilled_at || Time.current`?
I'm wondering since `Product#cheapest_price_policy` should default to `Time.zone.now` if given a falsy `date`.

jhanggi: This will fall back to nil `nil` to and call `cheapest_price_policy(self, nil)`, so it doesn't take the default. In there, it will not find any `cheapest_price_policy(nil)`, which is what I want. This helps ensure that we don't accidentally do this again, because we should never be finding the actual price policy it's fulfilled or it's canceled with charge (which doesn't have a `fulfilled_at`, but does end up in completed).

spec/controllers/order_management/order_details_controller_spec.rb
jhanggi: I buy that the formal `pending` makes sense. It hopefully will be short-lived, but I'm not sure when we'll be able to get to that piece.

spec/controllers/order_management/order_details_controller_spec.rb
```diff
@@ -505,14 +516,41 @@ def reservation_params(reserve_start_at, actual_start_at = nil)
...
+        order_detail.backdate_to_complete!(Time.current)
         @params[:order_detail] = { quantity: 2 }
       end
 
       it "updates the quanity" do
         expect { do_request }.to change { order_detail.reload.quantity }.to(2)
       end
+
+      it "updates the price while changing quantity" do
+        pending "currently buggy: it uses the price policy's calculation, not the params"
```
jhanggi: Using `pending` was extra helpful because it showed me that my test was buggy (the order detail was not complete).

drench:
:+1:

PR #774, opened 2016-09-29
[#132463] Fix route to updating Scheduling Groups on access list

jhanggi:
Got set to GET instead of PUT in the Rails upgrade.

We'll want to get this to UIC fairly quickly.


drench:
:+1:

PR #775, opened 2016-09-29
[#130782] Adds feature tag for setting default start date

kt32291:
https://pm.tablexi.com/issues/130782

This adds the ability to turn on setting a default start date under the create statements tab in billing.

### DIFF
--- a/app/controllers/facility_statements_controller.rb
+++ b/app/controllers/facility_statements_controller.rb
@@ -29,6 +29,7 @@ class FacilityStatementsController < ApplicationController
   def new_with_search
     @order_details = @order_details.need_statement(@facility)
     @order_detail_action = :send_statements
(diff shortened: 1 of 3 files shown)


app/controllers/facility_statements_controller.rb
```diff
@@ -29,6 +29,7 @@ def index
   def new_with_search
     @order_details = @order_details.need_statement(@facility)
     @order_detail_action = :send_statements
+    set_default_start_date if SettingsHelper.feature_on?(:set_start_date)
```
jhanggi: Looks pretty good, but can we make the feature flag name a little more specific, so we know it applies to the statementing search?

kt32291:
@jhanggi updated

jhanggi:
Looks like the name needs to change in the specs as well.

kt32291:
Updated

PR #776, opened 2016-09-29
[#130583] Updates styles for printing order details

kt32291:
https://pm.tablexi.com/issues/130583

This updates the styles so printing order details (happening specifically in UIC) only takes up a single page.

### DIFF
--- a/vendor/engines/projects/app/views/projects/shared/_select_project.html.haml
+++ b/vendor/engines/projects/app/views/projects/shared/_select_project.html.haml
@@ -1,8 +1,7 @@
 - if order_detail.selectable_projects.any?
-  = f.input :project_id,
-    collection: order_detail.selectable_projects,
-    input_html: { class: "js--chosen", data: { placeholder: t(".placeholder") } },
-    include_blank: true
+  = f.input :project_id do
+    = f.input :project_id, collection: order_detail.selectable_projects, input_html: { class: "js--chosen hide-from-print", data: { placeholder: t(".placeholder") } }, include_blank: true
+    .show-for-print= f.object.project.name if f.object.project
 
   :javascript
     ChosenActivator.activate();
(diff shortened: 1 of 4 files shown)


jhanggi:
I'm using chrome for testing this. It might be worth putting in some targeted selectors for this particular view.

* There's some weird cutting off of the Payment Source value.
* I'm not quite sure what we should do with "Project". If we can, I'd like to hide it if it's just the placeholder text, and show it if there is a value. Might not be worth the effort if it's too difficult. UIC doesn't use that feature, so it might not be a big deal.
* For reservations, the times are out of alignment. I'm not sure I'm that worried about the missing colons there.

jhanggi:
Here's a screenshot of the PDF I got when I printed:

screen shot 2016-09-29 at 5 03 07 pm

vendor/engines/projects/app/views/projects/shared/_select_project.html.haml
```diff
@@ -1,8 +1,7 @@
 - if order_detail.selectable_projects.any?
-  = f.input :project_id,
-    collection: order_detail.selectable_projects,
-    input_html: { class: "js--chosen", data: { placeholder: t(".placeholder") } },
-    include_blank: true
+  = f.input :project_id do
+    = f.input :project_id, collection: order_detail.selectable_projects, input_html: { class: "js--chosen hide-from-print", data: { placeholder: t(".placeholder") } }, include_blank: true
```
jhanggi: Project is still behaving a little weird. There are now two "Project" labels. The input is not being hidden. I think it has to do with the `f.input :project_id` being nested inside another `f.input :project_id`.

screen shot 2016-10-04 at 3 23 10 pm

kt32291:
@jhanggi finally got this fixed!

jhanggi:
Looks good on my mac. Do you have an IE virtual machine to test on since that's the primary place they were seeing this issue?

kt32291:
@jhanggi I do not, but I can try browserstack

kt32291:
@jhanggi just needed one little fix in IE. All good now!

PR #778, opened 2016-09-30
[#124051] Add validation for journal reference length

kt32291:
https://pm.tablexi.com/issues/124051

The journal reference has a database length restriction, but this adds a validation so a reference that is too long for the db can't get past the validation.


jhanggi:
I just pushed up a change that will add the `maxlength` attribute to the form input. I expected it to just be there, but turns out we weren't using the simple_form helpers for those fields.

kt32291:
@jhanggi 👍

PR #779, opened 2016-09-30
[#124334] Restrict amount sum to non-cancelled orders

kt32291:
https://pm.tablexi.com/issues/124334

This restricts the price sum on the orders page based on whether the order status is cancelled.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -126,6 +126,8 @@ class OrderDetail < ActiveRecord::Base
       .where("products.type" => product_type)
   }
 
+  scope :non_cancelled, -> { joins(:order_status).where.not(order_statuses: { name: "Canceled" }) }
+
   def self.for_facility(facility)
     for_facility_id(facility.id)
   end
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -126,7 +126,7 @@ class OrderDetail < ActiveRecord::Base
       .where("products.type" => product_type)
   }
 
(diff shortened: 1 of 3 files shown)


app/models/order_detail.rb
```diff
@@ -126,6 +126,8 @@ def journal_or_statement_date
       .where("products.type" => product_type)
   }
 
+  scope :non_cancelled, -> { joins(:order_status).where.not(order_statuses: { name: "Canceled" }) }
```
jhanggi: I'm actually surprised we don't have a scope for this already. We should probably be using the `state` column for this rather than doing a join to `OrderStatus`. 

When we do need to compare to an order status, we usually use something like `OrderStatus.canceled_status` (you might also see `OrderStatus.complete.first` scattered throughout the code, which is the older way), so this would probably be more like`where.not(order_status: OrderStatus.canceled_status)`.

Also, I know you hate it, but we settled on `canceled`. There used to be both throughout the code, but a couple years ago we decided we wanted consistency (#173).

spec/models/order_detail_spec.rb
```diff
@@ -967,6 +967,16 @@ def assign_price_policy
         end
       end
     end
+
+    describe "non_cancelled" do
+      let!(:order_detail) { create(:order_detail, order_status: OrderStatus.canceled.first, order: Order.last, product: Product.last) }
```
jhanggi: I wonder if it it would be worth adding `canceled` and `complete` traits onto the factory. Not necessary, but just a thought.

kt32291:
@jhanggi updated

PR #780, opened 2016-09-30
[#126724] Prevent order_at from getting set on cart quantity change

jhanggi:
Here's how to replicate:

* Order as another user
* Open the "More options" so you can see the order date fields
* Change the quantity
* Click "Purchase"
* You'll see "Quantities have changed. Please review updated prices then click "Purchase".", however the date set in "More options" will be stored as the ordered_at date, even though the order has not been purchased.

Also, UIC #121346

This is some pretty gnarly code (I’m pretty sure I wrote it :frowning:), and it didn’t seem worth more time than I had already spent on tracking down the issue.

### DIFF
--- a/spec/controllers/orders_controller_spec.rb
+++ b/spec/controllers/orders_controller_spec.rb
@@ -558,6 +558,22 @@ RSpec.describe OrdersController do
         expect(assigns[:order].reload.ordered_at).to match_date Time.zone.now
       end
 
+      it "does not ordered_at when quantities change" do
+        @order.validate_order!
+        maybe_grant_always_sign_in :director
+        switch_to @staff
+
+        @params.merge!(
+          "quantity#{@order_detail.id}" => 5,
+          "order_date" => "09/30/2016",
+          "order_time" => { "hour" =>"4", "minute" => "0", "ampm" =>"PM" }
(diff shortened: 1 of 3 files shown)


spec/controllers/orders_controller_spec.rb
```diff
@@ -558,6 +558,22 @@ def deliver; end
         expect(assigns[:order].reload.ordered_at).to match_date Time.zone.now
       end
 
+      it "does not ordered_at when quantities change" do
```
drench: `it "does not set ordered_at`…?

spec/controllers/orders_controller_spec.rb
```diff
@@ -558,6 +558,22 @@ def deliver; end
...
+        @order.validate_order!
+        maybe_grant_always_sign_in :director
+        switch_to @staff
+
+        @params.merge!(
+          "quantity#{@order_detail.id}" => 5,
+          "order_date" => "09/30/2016",
+          "order_time" => { "hour" =>"4", "minute" => "0", "ampm" =>"PM" }
```
drench: The Code Climate complaints are also nitpicky but probably auto-correctable.

PR #781, opened 2016-10-03
[#128124193 128124365 128124381] Add instrument down send mail button

drench:
This adds a "send mail" button to the instrument admin screen to the right of the online/offline toggle button. When clicked, it sends the user into the bulk email flow with the current instrument selected and a range of the next seven days filled-in as start and end dates.

If canceled from the recipient search screen, the user should be sent back to instrument admin.
After a successful send action, the user should be sent back to instrument admin.
If canceled on the mail composition screen, the user is sent back to the recipient search screen as before, and canceling from there should bring the user back to instrument admin.

As before, the user does not get redirected on CSV export though we may want to rethink that.

### DIFF
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -64,14 +64,16 @@ module BulkEmail
     end
 
     def delivery_success_path
+      return_path_from_params || facility_bulk_email_path
+    end
+
+    def return_path_from_params
       return_path = params[:return_path].presence
-      if return_path && Rails.application.routes.recognize_path(return_path)
-        return_path
-      else
-        facility_bulk_email_path
-      end
+      return_path if return_path &&
+                     return_path.starts_with?("/") &&
+                     Rails.application.routes.recognize_path(return_path)
     rescue ActionController::RoutingError
-      facility_bulk_email_path
+      nil
     end
 
     def init_delivery_form
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -68,6 +68,9 @@ module BulkEmail
     end
 
     def return_path_from_params
+      # This controller accepts an optional return_path param to redirect to
+      # after a successful mail queueup. This returns that param value if it is
+      # a valid application path, or nil if not.
       return_path = params[:return_path].presence
       return_path if return_path &&
                      return_path.starts_with?("/") &&
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -64,7 +64,14 @@ module BulkEmail
     end
 
     def delivery_success_path
-      params[:return_path].presence || facility_bulk_email_path
+      return_path = params[:return_path].presence
+      if return_path && Rails.application.routes.recognize_path(return_path)
+        return_path
+      else
+        facility_bulk_email_path
+      end
+    rescue ActionController::RoutingError
+      facility_bulk_email_path
     end
 
     def init_delivery_form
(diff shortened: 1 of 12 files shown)


vendor/engines/bulk_email/app/helpers/bulk_email/application_helper.rb
```diff
@@ -0,0 +1,18 @@
+module BulkEmail
+
+  module ApplicationHelper
+
+    include Rails.application.routes.url_helpers
+
+    def admin_instrument_send_mail_path(instrument)
+      facility_bulk_email_path(
+        products: [instrument.id],
+        start_date: l(Date.today, format: :usa),
```
jhanggi: If having no user type selected returns nothing, then we should probably default it to "Customers". That way you can come through and all you have to click is "Search".

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_fields.html.haml
```diff
@@ -26,3 +26,6 @@
   .row
     .span8
       = f.submit text("shared.search"), class: "btn"
+      - if params[:return_path].present?
+        = link_to t("shared.cancel"), params[:return_path]
```
jhanggi: I feel like we could improve the layout of this view a little bit. I think the compose mail button shouldn't go further right than the table's right edge. There probably doesn't need to be as much of a gap between the two buttons. And "Cancel" could probably be in a better spot.

screen shot 2016-10-04 at 3 39 36 pm

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_fields.html.haml
```diff
@@ -26,3 +26,6 @@
...
+        = hidden_field_tag "return_path", params[:return_path]
```
jhanggi: We should probably do some sort of sanitization on this parameter. I can modify this in the URL to make me go anywhere I want.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_fields.html.haml
drench: I'm updating this now, but I'm curious if you're still seeing that "Search" button way off on the right. I'm seeing it just to the left of the "Cancel" link right now (Chrome, Safari, and FF).

jhanggi: No, the search is over on the right. I think I was playing with some styling.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -60,7 +60,20 @@ def bulk_email_cancel_path
...
+      @cancel_params ||= params.slice(:start_date, :end_date, :bulk_email, :products, :return_path)
+    end
+
+    def delivery_success_path
+      return_path_from_params || facility_bulk_email_path
+    end
+
+    def return_path_from_params
+      return_path = params[:return_path].presence
+      return_path if return_path &&
```
jhanggi: I think a comment about why this is doing these things would be a good idea so future us knows that this is about sanitization.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -60,7 +60,20 @@ def bulk_email_cancel_path
...
+                     return_path.starts_with?("/") &&
+                     Rails.application.routes.recognize_path(return_path)
```
jhanggi: Did not know about this method. Nice.

jhanggi:
Just one small thing about adding a comment, otherwise 👍

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
drench: Yeah this is the first time I've used it. One gotcha for me was that if you give it a full URL it looks like it ignores the "scheme://hostname" part so something like `https://github.com/` will pass. That's why there's the check that it starts with "/". I'll add that comment.

PR #782, opened 2016-10-06
[#128124279] Add context-aware default text to the bulk email form

drench:
When sending bulk mail via the instrument admin screen, this should fill in defaults to the form for the subject and body content fields. The text is defined in locales under `bulk_email.reason_statements` and is subject to change.

This does add more baggage to the params we're passing around between the steps. I could see, especially if we need to make the default subject and body text different, having to rethink that.

This also includes a style change to handle overflow on the email compose screen's recipient list, so it should handle large recipient lists better than before.

### DIFF
--- a/app/support/products/scheduling_support.rb
+++ b/app/support/products/scheduling_support.rb
@@ -98,6 +98,10 @@ module Products::SchedulingSupport
     offline_reservations.current.any?
   end
 
+  def offline_category
+    current_offline_reservation.try(:category)
+  end
+
   def online!
     offline_reservations.current.update_all(reserve_end_at: Time.current)
   end
@@ -107,6 +111,10 @@ module Products::SchedulingSupport
   end
 
   private
+
+  def current_offline_reservation
+    offline_reservations.current.last
+  end
 
   def reservation_in_week(after, duration, rules, options)
     day_of_week = after.wday
--- a/vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
+++ b/vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
@@ -6,6 +6,11 @@
 
 }
 
+.bulk_email_recipient_list {
+  max-height: 100px;
+  overflow-y: scroll;
+}
+
 #bulk_email_create {
   .select_all_none {
     margin-left: 0;
--- a/vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
+++ b/vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
@@ -6,11 +6,6 @@
 
 }
 
-.bulk_email_recipient_list {
-  max-height: 100px;
-  overflow-y: scroll;
-}
-
 #bulk_email_create {
   .select_all_none {
     margin-left: 0;
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -62,8 +62,8 @@ module BulkEmail
 
     def bulk_email_content_generator
       @content_generator ||=
-        if params[:subject_product_id].present?
-          ContentGenerator.new(current_facility, Product.find(params[:subject_product_id]))
+        if params[:product_id].present?
+          ContentGenerator.new(current_facility, Product.find(params[:product_id]))
         else
           ContentGenerator.new(current_facility)
         end
@@ -75,7 +75,7 @@ module BulkEmail
                                       :bulk_email,
                                       :products,
                                       :return_path,
-                                      :subject_product_id)
+                                      :product_id)
     end
 
     def delivery_success_path
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -60,7 +60,17 @@ module BulkEmail
     end
 
     def cancel_params
-      @cancel_params ||= params.slice(:start_date, :end_date, :bulk_email, :products, :return_path)
+      return @cancel_params if @cancel_params
+      @cancel_params = params.slice(:start_date,
+                                    :end_date,
+                                    :bulk_email,
+                                    :products,
+                                    :return_path,
+                                    :default_text)
+      if params[:bulk_email_delivery_form] && !params[:default_text]
+        @cancel_params[:default_text] = params[:bulk_email_delivery_form][:custom_message]
+      end
+      @cancel_params
     end
 
     def delivery_success_path
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -70,7 +70,12 @@ module BulkEmail
     end
 
     def cancel_params
-      @cancel_params ||= params.slice(:start_date, :end_date, :bulk_email, :products, :return_path)
+      @cancel_params ||= params.slice(:start_date,
+                                      :end_date,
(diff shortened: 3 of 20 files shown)


vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_results.html.haml
```diff
@@ -4,6 +4,12 @@
   - if params[:return_path].present?
     = hidden_field_tag :return_path, params[:return_path]
 
+  - if params[:default_text].present?
```
jhanggi: Maybe I'm missing something. Why do we need to keep these in hidden fields?

jhanggi:
I thought the story was to add some static text to the email, not provide a default starting point.

vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
```diff
@@ -6,6 +6,11 @@
 
 }
 
+.bulk_email_recipient_list {
+  max-height: 100px;
```
kt32291: Why 100? Magic number?

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -60,7 +60,17 @@ def bulk_email_cancel_path
...
-      @cancel_params ||= params.slice(:start_date, :end_date, :bulk_email, :products, :return_path)
+      return @cancel_params if @cancel_params
+      @cancel_params = params.slice(:start_date,
+                                    :end_date,
+                                    :bulk_email,
+                                    :products,
+                                    :return_path,
+                                    :default_text)
+      if params[:bulk_email_delivery_form] && !params[:default_text]
+        @cancel_params[:default_text] = params[:bulk_email_delivery_form][:custom_message]
```
kt32291: +1

vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
drench: I guess it is magic. I want to limit the height of this list and 100 looked reasonable. I'm open to alternatives.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/_results.html.haml
drench: To clarify, I am making changes to display the canned text and making that canned text dynamic though not passed in through hidden params, verbatim.

drench:
Updated. I changed the method for setting canned content by setting an optional subject product param, rather than passing around text. This also updates the mail compose form to expose the canned text.

One outstanding issue: I'm still fighting with the subject line form field (I'd like it to fill 100% of its container) but I didn't want to hold up potentially reviewing the rest.

app/support/products/scheduling_support.rb
```diff
@@ -98,6 +98,10 @@ def offline?
     offline_reservations.current.any?
   end
 
+  def offline_category
```
jhanggi: This fields a little weird in `SchedulingSupport`, but the whole separation into concerns is already a little weird.

app/support/products/scheduling_support.rb
```diff
@@ -108,6 +112,10 @@ def online?
 
   private
 
+  def current_offline_reservation
+    offline_reservations.current.last
```
jhanggi: Does `current` do an ordering? Oracle will not have guaranteed ordering.

vendor/engines/bulk_email/app/mailers/bulk_email/mailer.rb
```diff
@@ -2,16 +2,24 @@ module BulkEmail
 
   class Mailer < BaseMailer
 
-    def send_mail(recipient:, custom_subject:, facility:, custom_message:)
+    def send_mail(recipient:, custom_subject:, facility:, custom_message:, subject_product_id: nil)
```
jhanggi: As a general thing, do we care that the product is tied to the subject? Would `product_id` as a name suffice? Do we already have the product loaded when we call this method? If so, could we just pass `product`? Is this going through ActiveJob/GlobalID? If not, we should probably do `facility` as `facility_id`.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/create.html.haml
```diff
@@ -16,20 +16,32 @@
...
-      input_html: { class: "input-xxlarge", rows: 6 }
-
-  .span8
-    = f.submit text("bulk_email.send_mail"),
-      class: "btn btn-primary",
-      data: { disable_with: t("bulk_email.sending") }
-
-    = link_to text("shared.cancel"), bulk_email_cancel_path
+  - if params[:subject_product_id].present?
+    = hidden_field_tag :subject_product_id, params[:subject_product_id]
```
jhanggi: Would this all be cleaner if we just always had the tags and dealt with empties later? Just curious.

vendor/engines/bulk_email/app/views/bulk_email/mailer/send_mail.html.haml
```diff
@@ -1,3 +1 @@
-%p= html "bulk_email.body.greeting", recipient_name: @recipient.full_name
-%p= @custom_message
-%p= html "bulk_email.body.signoff", facility_name: @facility.name
+= @body.gsub(/\n/, "<br>").html_safe
```
jhanggi: This seems dangerous here. Could we use `simple_format`?

vendor/engines/bulk_email/app/inputs/bulk_email/subject_input.rb
```diff
@@ -0,0 +1,14 @@
+module BulkEmail
+
+  class SubjectInput < SimpleForm::Inputs::TextInput
+
+    def input(_wrapper_options)
+      template.content_tag(:div, class: "input-prepend") do
+        template.content_tag(:span, options[:prefix], class: "add-on") +
```
jhanggi: Is there a way we can append these or `safe_join` so that we don't need the `html_safe` that's triggering a rubocop warning?

jhanggi:
Tested it out a little bit, seems to work pretty well. Only concern, which probably doesn't matter is a race condition where I send a bulk email, it queues them up in delayed_job, but before it can process them, I mark the instrument back online. I _think_the email will go out without the reason. It's probably low enough likelihood that we don't need to worry about it.

app/support/products/scheduling_support.rb
drench: There ought to be only one current offline reservation for an Instrument. It could be `.first` even.

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/create.html.haml
drench: Yeah, empties wouldn't hurt and would make the haml a bit nicer if they were gone.

app/support/products/scheduling_support.rb
jhanggi: Makes sense.

vendor/engines/bulk_email/app/inputs/bulk_email/subject_input.rb
drench: The strings this produces turn out to already be html-safe! Removing…

vendor/engines/bulk_email/app/views/bulk_email/mailer/send_mail.html.haml
drench: Thanks - forgot about `simple_format`

vendor/engines/bulk_email/app/mailers/bulk_email/mailer.rb
drench: I had meant it in the sense that the product is the subject of the bulk email run, but the param name isn't really clear.

This does use ActiveJob with globalid, so it could just be `product`.

PR #783, opened 2016-10-06
[#128124251] Add links to select date ranges for bulk email recipient search

drench:
This includes #782; the feature is in 59046c3.

This adds 3 links that fill in the bulk email recipient search form with some possibly useful date ranges: the last 90 days, the last year, and next month. I expect these to change after client feedback.

These are real links, not JavaScript-driven. I considered JS, but making them standard links updates the recipient list more or less for free.

### DIFF
--- a/vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
+++ b/vendor/engines/bulk_email/app/assets/stylesheets/bulk_email/application.scss
@@ -6,6 +6,10 @@
 
 }
 
+.bulk_email_search_actions {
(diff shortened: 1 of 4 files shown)


jhanggi:
The new change seems good. I'd like to see if we can make #782 a little cleaner and simpler.

jhanggi:
Once #782 is in, rebase this and I'll take a quick look, but assuming there are no conflicts, this should be good to go.

drench:
Rebased and ready for review.

jhanggi:
👍

PR #784, opened 2016-10-06
[#125701] External user statements view

kt32291:
https://pm.tablexi.com/issues/125701

This gives users a column to download their statements for transactions if available. I made this part of the transactions table, so the users still had the benefits of search to find their statements. The link only shows if they have a statement to view.

### DIFF
--- a/app/controllers/statements_controller.rb
+++ b/app/controllers/statements_controller.rb
@@ -34,7 +34,7 @@ class StatementsController < ApplicationController
 
   def init_account
     # CanCan will make sure that we're authorizing the account
-    @account = Account.find(params[:id] || params[:account_id])
+    @account = Account.find(params[:account_id])
   end
 
   #
--- a/app/controllers/statements_controller.rb
+++ b/app/controllers/statements_controller.rb
@@ -15,7 +15,7 @@ class StatementsController < ApplicationController
 
   # GET /accounts/:id/statements
   def index
-    @statements = @account.statements.paginate(page: params[:page])
+    @statements = @account.statements.uniq.paginate(page: params[:page])
   end
 
   # GET /accounts/:account_id/statements/:id
@@ -42,10 +42,8 @@ class StatementsController < ApplicationController
     @account = Account.find(params[:account_id])
   end
 
-  #
-  # Override CanCan's find -- it won't properly search by 'recent'
   def init_statement
-    @statement = Statement.find_by(id: params[:id], account_id: @account.id)
+    @statement = Statement.find_by!(id: params[:id], account_id: @account.id)
     @facility = @statement.facility
   end
 
--- a/app/controllers/statements_controller.rb
+++ b/app/controllers/statements_controller.rb
@@ -4,7 +4,7 @@ class StatementsController < ApplicationController
   before_action :authenticate_user!
   before_action :check_acting_as
   before_action :init_account
-  before_action :init_statement
+  before_action :init_statement, only: [:show]
 
   load_and_authorize_resource
 
@@ -13,22 +13,12 @@ class StatementsController < ApplicationController
     super
   end
 
-  # GET /accounts/:account_id/facilities/:facility_id/statements/:id
+  # GET /accounts/:account_id/statements/:id
   def show
     action = "show"
     @active_tab = "accounts"
 
-    case params[:id]
-    when "recent"
-      @order_details = @account.order_details.for_facility_with_price_policy(@facility)
-      @order_details = @order_details.paginate(page: params[:page])
-    when "list"
-      action = "list"
-      @statements = @statements.paginate(page: params[:page])
-    end
-
     respond_to do |format|
-      format.html { render action: action }
       format.pdf do
         @statement_pdf = StatementPdfFactory.instance(@statement, params[:show].blank?)
         render action: "show"
@@ -44,21 +34,14 @@ class StatementsController < ApplicationController
 
   def init_account
     # CanCan will make sure that we're authorizing the account
-    @account = Account.find(params[:account_id])
+    @account = Account.find(params[:id] || params[:account_id])
   end
 
   #
   # Override CanCan's find -- it won't properly search by 'recent'
   def init_statement
-    @facility = Facility.find_by_url_name!(params[:facility_id])
-    @statements = @account.statements.where(facility_id: @facility.id)
-
-    @statement = if params[:id] =~ /\w+/i
-                   @statements.blank? ? Statement.find_by_facility_id(@facility.id) : @statements.first
-                 else
-                   @account.statements.find(params[:id])
-                 end
-    @statement = Statement.new if @statement.nil?
+    @statement = Statement.find(params[:id])
+    @facility = @statement.facility
   end
 
 end
--- a/app/controllers/statements_controller.rb
+++ b/app/controllers/statements_controller.rb
@@ -13,6 +13,12 @@ class StatementsController < ApplicationController
     super
   end
 
+  # GET /accounts/:id/statements
+  def index
+    @order_details = @account.order_details.has_statement.by_statemented_at
+    @date_range_field = "journal_or_statement_date"
+  end
+
   # GET /accounts/:account_id/statements/:id
   def show
     action = "show"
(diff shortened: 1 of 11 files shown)


jhanggi:
This is an interesting approach in that it makes it easy to see which statement a particular order detail is on. What I believe the client is looking for is a reverse-chronological listing of the statements for an account. It might be nice to have a all-my-statements-across-all-my-accounts listing, but that may be tricky to implement, so just per-account is fine. I'm open to putting the listing somewhere else, but as another tab next to Transactions and Members seems like a good place.

One additional note, this tab should only appear for statementable accounts (see `Account.config.statement_account_types`).

app/views/shared/transactions/_table.html.haml
```diff
@@ -27,6 +27,6 @@
     .row
       .span12= yield :action_instructions
     .row
-      .span12= render :partial => 'shared/transactions/table_inside', :locals => { :order_details => order_details }
+      .span12= render :partial => 'shared/transactions/table_inside', :locals => { :order_details => order_details, :show_statements => local_assigns[:show_statements] }
```
jhanggi: Can we clean up hash syntax as we come across the old styles?

app/views/shared/transactions/_table_inside.html.haml
```diff
@@ -58,4 +60,6 @@
           - order_detail.send(:extend, PriceDisplayment)
           = order_detail.wrapped_total
         %td.nowrap= order_detail.order_status
+        - if local_assigns[:show_statements]
+          %td= link_to "Download", account_statement_path(@account, order_detail.statement, format: :pdf) if order_detail.statement
```
jhanggi: Can we use just the `order_detail.statement_id` rather than the `statement`? That'll avoid an extra call to the database on each row.

app/controllers/statements_controller.rb
```diff
@@ -13,22 +13,18 @@ def initialize
     super
   end
 
-  # GET /accounts/:account_id/facilities/:facility_id/statements/:id
+  # GET /accounts/:id/statements
+  def index
+    @order_details = @account.order_details.has_statement.by_statemented_at
```
jhanggi: I think we're looking for a list of statements, not a list of order details. We don't need to see the individual order details in this view.

app/views/shared/_tabnav_account.html.haml
```diff
@@ -3,4 +3,5 @@
   = tab t('shared.tabnav_account.transact'), transactions_account_path(@account), (secondary_tab == 'transactions')
   - if @facility && @account.class.using_statements?
     = tab t('shared.tabnav_account.statement'), account_facility_statement_path(@account, @facility, 'list'), (secondary_tab == 'statements')
+  = tab t('shared.tabnav_account.statement'), account_statements_path(@account), (secondary_tab == 'statements')
```
jhanggi: Is this view shared with the admin/facility side? If not, I think we can kill off the previous line. However, we should use `@account.class.using_statements?` (I believe this calls into `Account.config` to only show this tab for Purchase Orders and Credit Cards. Chart Strings don't (for now at least, UIC will want them in an upcoming story, but NU does not).

app/controllers/statements_controller.rb
jhanggi: Seeing a total $ amount for the statement does make sense.

kt32291:
@jhanggi updated

app/controllers/statements_controller.rb
```diff
@@ -50,15 +45,8 @@ def init_account
   #
   # Override CanCan's find -- it won't properly search by 'recent'
```
jhanggi: I think this comment is out of date since `recent` is gone. Not quite sure.

app/controllers/statements_controller.rb
```diff
@@ -50,15 +45,8 @@ def init_account
...
-    @facility = Facility.find_by_url_name!(params[:facility_id])
-    @statements = @account.statements.where(facility_id: @facility.id)
-
-    @statement = if params[:id] =~ /\w+/i
-                   @statements.blank? ? Statement.find_by_facility_id(@facility.id) : @statements.first
-                 else
-                   @account.statements.find(params[:id])
-                 end
-    @statement = Statement.new if @statement.nil?
+    @statement = Statement.find_by(id: params[:id], account_id: @account.id)
```
jhanggi: `find_by!` might be a good idea here so it goes straight to a 404 rather than 500ing on the next line if it's not found.

app/models/order_detail.rb
```diff
@@ -96,6 +96,7 @@ def journal_or_statement_date
   ## TODO validate which fields can be edited for which states
 
   scope :by_ordered_at, -> { joins(:order).order("orders.ordered_at DESC") }
+  scope :by_statemented_at, -> { joins(:statement).order("statements.created_at DESC") }
```
jhanggi: I think this scope is no longer needed.

app/models/order_detail.rb
```diff
@@ -282,6 +283,7 @@ def can_be_viewed_by?(user)
       .where(orders: { facility_id: facility.id })
       .where.not(statement_id: nil)
   }
+  scope :has_statement, -> { where.not(statement_id: nil) }
```
jhanggi: Is this still needed? I don't see any references to it anymore?

app/controllers/statements_controller.rb
```diff
@@ -13,22 +13,17 @@ def initialize
     super
   end
 
-  # GET /accounts/:account_id/facilities/:facility_id/statements/:id
+  # GET /accounts/:id/statements
+  def index
+    @statements = @account.statements.paginate(page: params[:page])
```
jhanggi: Just found an interesting behavior. Because `Account` `has_many :statements, through: :order_details` rather than being a direct relation, I'm seeing duplicate statements, one for each order detail on the statement.

Making it a direct relationship based on `Statement#account_id` seems like the right choice, but I'm not sure what effect that'll have on other places in the app. Maybe try it out and see if any specs break?

Other option is to do a `uniq`/`distinct` on the result.

app/views/shared/_statements_table.html.haml
```diff
@@ -0,0 +1,48 @@
...
+                - if order_detail.reservation
+                  %br
+                  %em= order_detail.reservation
+              - if order_detail.note.present?
+                .order-detail-extra.order-detail-note= "Note: #{order_detail.note}"
+          %td.currency=h s.order_details.count
+          %td.currency= number_to_currency(s.total_cost)
+          %td= t("statements.reconciled.#{s.reconciled?}")
+          %td
+            = link_to "Download", account_statement_path(s.account, s, format: :pdf)
```
jhanggi: I think just the link in the first column is enough.

app/views/shared/_statements_table.html.haml
```diff
@@ -0,0 +1,48 @@
+- if @statements.empty?
+  %p.notice= t(".no_statements")
+- else
+  %table.table.table-striped.table-hover
```
jhanggi: I'm fine sharing this table, but I think the only difference is the admin view sees the Account column and the user view shows the facility 

Worth noting, right now all statement-able accounts (POs and CCs) are tied to a specific facility. This might change in the future, so I don't think anything should change because of it.

app/views/shared/_statements_table.html.haml
```diff
@@ -0,0 +1,48 @@
...
+            = "##{s.invoice_number}"
+            %br
+            = link_to t("statements.pdf.download"), statement_path(s)
+          %td= human_datetime(s.created_at)
+          %td= s.account.notify_users.map(&:full_name).join(', ')
+          - if local_assigns[:admin_view]
+            %td= link_to s.account, facility_account_path(current_facility, s.account)
+          - else
+            %td= s.facility.name
+            %td.user-order-detail
```
jhanggi: I don't think we need any order detail information on this view.

kt32291:
@jhanggi updated

jhanggi:
As an account owner/business admin, I am unable to download the statement. When I click the "Download" link, I get a 403. 

This might be a tricky one since it might involve delving into the bowels of the Ability class. Or, it could be that we just need to route the link to a different place (it's going to `FacilityAccountsController#show_statement` right now). I'll leave it to your judgement whether it ends up being simpler to do a conditional in the shared view or if it's easier to use different table views for regular users vs admins.

kt32291:
@jhanggi idk why that code climate thing suddenly popped up, but this should be fine.

jhanggi:
Looks good! I don't quite get the code climate issue, and I wouldn't worry about it.

PR #785, opened 2016-10-10
[#130689] Updates grouped options for select

kt32291:
https://pm.tablexi.com/issues/130689

This updates the grouped options for select for the bundle products to properly group them under a single parent.

### DIFF
--- a/app/models/bundle.rb
+++ b/app/models/bundle.rb
@@ -5,23 +5,13 @@ class Bundle < Product
 
   def products_for_group_select
     products = facility.products.where.not(type: "Bundle").order(:type, :name)
-    current_group = []
-    current_opts  = []
-    groups = []
+    groups = products.map {|x| x.class.name.pluralize}
+    options = {}
+    groups.map {|x| options[x] = []}
     products.each do |p|
-      if p.class.name != current_group[0]
-        unless current_opts.empty?
-          current_group << current_opts
-          groups << current_group
-          current_group = []
-          current_opts  = []
-        end
-        current_group << p.class.name.pluralize
-      end
-      current_opts << [p.to_s_with_status, p.id]
+      options[p.class.name.pluralize] << [p.to_s_with_status, p.id]
     end
-    groups << (current_group << current_opts) unless current_opts.empty?
-    groups
+    options
   end
 
   def products_active?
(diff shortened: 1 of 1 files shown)


app/models/bundle.rb
```diff
@@ -5,23 +5,13 @@ class Bundle < Product
 
   def products_for_group_select
     products = facility.products.where.not(type: "Bundle").order(:type, :name)
-    current_group = []
-    current_opts  = []
-    groups = []
+    groups = products.map {|x| x.class.name.pluralize}
```
jhanggi: Is there a way we could use `group_by { |p| p.class.name.pluralize }`? It might make the conversion to `[p.to_s_with_status, p.id]` a little trickier, though.

kt32291:
@jhanggi updated

app/models/bundle.rb
```diff
@@ -5,23 +5,12 @@ class Bundle < Product
...
-          current_group = []
-          current_opts  = []
-        end
-        current_group << p.class.name.pluralize
-      end
-      current_opts << [p.to_s_with_status, p.id]
+    options = {}
+    products.group_by { |p| p.class.name.pluralize }.each do |cname, products|
+      options[cname] ||= []
+      products.map {|p| options[cname] << [p.to_s_with_status, p.id]}
```
jhanggi: I think we can invert this somewhat since we're already grouped:
`options[cname] = products.map { |p| [p.to_s_with_status, p.id] }` I think this reads a little clearer, and doesn't need the default anymore.

jhanggi: We could also probably do some sort of `inject`/`each_with_object`, but I don't know that it would be easier to read.

kt32291:
@jhanggi this should be good now

app/models/bundle.rb
```diff
@@ -5,23 +5,11 @@ class Bundle < Product
...
-        unless current_opts.empty?
-          current_group << current_opts
-          groups << current_group
-          current_group = []
-          current_opts  = []
-        end
-        current_group << p.class.name.pluralize
-      end
-      current_opts << [p.to_s_with_status, p.id]
+    options = Hash.new { |h, k| h[k] = [] }
```
jhanggi: I'm thinking the default might not be necessary since we're doing direct assignment, but it doesn't hurt anything.

PR #786, opened 2016-10-10
[#132076] Allows admin to add optional category to reservation

kt32291:
https://pm.tablexi.com/issues/132076

This allows adding a category on the admin side to a reservation, to indicate blocking off an instrument for a certain period of time.

### DIFF
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -80,8 +80,6 @@ class FacilityReservationsController < ApplicationController
       end
     end
 
-    @reservation.assign_category_attribute(session_user, params[:reservation])
-
     Reservation.transaction do
       begin
         @reservation.save_as_user!(session_user)
@@ -157,6 +155,7 @@ class FacilityReservationsController < ApplicationController
     set_windows
 
     @reservation.assign_times_from_params(params[:admin_reservation])
+    @reservation.assign_category_attribute(session_user, params[:reservation])
     @reservation.admin_note = params[:admin_reservation][:admin_note]
 
     if @reservation.save
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -80,6 +80,8 @@ class FacilityReservationsController < ApplicationController
       end
     end
 
+    @reservation.assign_category_attribute(session_user, params[:reservation])
+
     Reservation.transaction do
       begin
         @reservation.save_as_user!(session_user)
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -155,8 +155,8 @@ class FacilityReservationsController < ApplicationController
     set_windows
 
     @reservation.assign_times_from_params(params[:admin_reservation])
-    @reservation.assign_category_attribute(session_user, params[:reservation])
     @reservation.admin_note = params[:admin_reservation][:admin_note]
+    @reservation.category = params[:admin_reservation][:category]
 
     if @reservation.save
       flash[:notice] = "The reservation has been updated successfully."
--- a/app/helpers/facility_reservations_helper.rb
+++ b/app/helpers/facility_reservations_helper.rb
@@ -5,12 +5,11 @@ module FacilityReservationsHelper
   end
 
   def admin_category_collection
-    I18n.t("reservations.categories").invert.to_a
+    I18n.t("admin_reservations.categories").invert.to_a
   end
 
   def reservation_category_label(reservation)
-    scope = reservation.is_a?(OfflineReservation) ? "offline_reservations.categories" : "reservations.categories"
-    I18n.t(reservation.category.presence, scope: scope, default: "")
+    I18n.t(reservation.category.presence, scope: "#{reservation.class.name.underscore}s.categories", default: "")
   end
 
   def reservation_links(reservation)
--- a/app/helpers/facility_reservations_helper.rb
+++ b/app/helpers/facility_reservations_helper.rb
@@ -4,8 +4,13 @@ module FacilityReservationsHelper
     I18n.t("offline_reservations.categories").invert.to_a
   end
 
-  def reservation_category_label(category)
-    I18n.t(category.presence, scope: "offline_reservations.categories", default: "")
+  def admin_category_collection
+    I18n.t("reservations.categories").invert.to_a
+  end
+
+  def reservation_category_label(reservation)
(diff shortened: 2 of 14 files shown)


app/controllers/facility_reservations_controller.rb
```diff
@@ -80,6 +80,8 @@ def update
       end
     end
 
+    @reservation.assign_category_attribute(session_user, params[:reservation])
```
jhanggi: I think we should actually be setting this in the `update_admin` action.

jhanggi: Also, when you're in this context, I'm 95% the user is already authorized, so I don't think we need the additional check.

app/models/reservation.rb
```diff
@@ -159,6 +168,10 @@ def assign_actuals_off_reserve
     self.actual_end_at   ||= reserve_end_at
   end
 
+  def assign_category_attribute(user, params)
+    assign_attributes(category: params[:category]) if user.operator_of?(product.facility)
```
jhanggi: I said in a previous note that I don't think this check will be necessary, but if we do need it, we're preferring to use cancan over explicit use of of `operator_of?`.

app/support/reservations/validations.rb
```diff
@@ -23,7 +23,7 @@ module Reservations::Validations
     validate :duration_is_interval
 
     validates :category,
-              inclusion: { in: -> (r) { r.class::CATEGORIES }, allow_blank: false },
+              inclusion: { in: -> (r) { r.class::CATEGORIES }, allow_blank: true },
```
jhanggi: I'd check with @drench on this. I'm pretty sure we need `allow_blank` to be false for instrument not available reservations.

app/views/facility_reservations/_reservation_fields.html.haml
```diff
@@ -1,4 +1,7 @@
 - if @reservation.admin_editable?
+  .control-group
```
jhanggi: We need to be careful. This is ending up on the normal facility reservation edit page (accessed by the New/In-process tab under Reservations). We want it to only be there for admin reservations.

app/support/reservations/validations.rb
drench: For the base `Reservation` model, I'd leave this as-is. I'd put the category presence validation in `AdminReservation`, much like it is for `OfflineReservation`.

kt32291:
@jhanggi updated

app/controllers/facility_reservations_controller.rb
```diff
@@ -155,6 +155,7 @@ def update_admin
     set_windows
 
     @reservation.assign_times_from_params(params[:admin_reservation])
+    @reservation.assign_category_attribute(session_user, params[:reservation])
```
jhanggi: Is the extra check here on `session_user` necessary? I'm 95% sure you're already validated as being at operator at this point.

app/models/reservation.rb
jhanggi: When updating an admin reservation (everything worked fine for creation), I'm getting an error here because `params` is `nil`.

jhanggi:
There are also a good chunk of test failures.

app/views/facility_reservations/_reservation_fields.html.haml
```diff
@@ -1,4 +1,8 @@
 - if @reservation.admin_editable?
+  - if @reservation.admin?
+    .control-group
+      = f.input :category do
+        = f.select :category, options_for_select(Reservation::CATEGORIES.map {|c| [t("reservations.categories.#{c}"), c]}, f.object.category), include_blank: "Select a Category"
```
jhanggi: See `FacilityReservationsHelper#offline_category_collection`. That might allow you to clean this up a little and keep things consistent.

app/helpers/facility_reservations_helper.rb
```diff
@@ -4,8 +4,13 @@ def offline_category_collection
...
   end
 
-  def reservation_category_label(category)
-    I18n.t(category.presence, scope: "offline_reservations.categories", default: "")
+  def admin_category_collection
+    I18n.t("reservations.categories").invert.to_a
+  end
+
+  def reservation_category_label(reservation)
+    scope = reservation.is_a?(OfflineReservation) ? "offline_reservations.categories" : "reservations.categories"
```
jhanggi: In this case, the reservation should be either an `AdminReservation` or `OfflineReservation`, correct? Can we move the locales to be under `admin_reservations` since that's really what the categories apply to and use "#{reservation.class.name.underscore}`?

I'm also wondering if we should remove the default so we'd get errors in development if we're looking at an invalid category.

app/models/reservation.rb
```diff
@@ -9,6 +9,14 @@ class Reservation < ActiveRecord::Base
   include Reservations::RelaySupport
   include Reservations::MovingUp
 
+  CATEGORIES = %w(
```
jhanggi: @drench can probably speak to this better, but it feels like this definition should be on `AdminReservation`.

app/helpers/facility_reservations_helper.rb
kt32291: `class.name` in this case will return Reservation I believe - I was having this issue when I moved the categories constant to AdminReservation, the validation for the absence of a category in ReservationValidation was failing because it claimed categories wasn't defined on the class.

app/models/reservation.rb
kt32291: I tried moving it and the validation in ReservationValidations failed. See comment above.

app/support/reservations/validations.rb
```diff
@@ -23,12 +23,12 @@ module Reservations::Validations
...
               if: -> (r) { r.class.const_defined?(:CATEGORIES) }
 
     validates :category,
               absence: true,
-              unless: -> (r) { r.class.const_defined?(:CATEGORIES) }
+              unless: -> (r) { r.class.const_defined?(:CATEGORIES) || r.admin? }
```
jhanggi: I'd like @drench to weigh in on this. I think the inheritance of these validations  I would vote from removing the `absence` is a little funky. I would vote for moving the `inclusion` off of the superclass and define them in `AdminReservation` and `OfflineReservation` (I think the `allow_blank` logic is different between the two). And I'm not really sure the `absence` validation actually adds any value. Could we remove this validation altogether?

drench: I think moving the `category` inclusions to the subclasses is a good idea. I kind of like enforcing no category for plain `Reservation` but if removing it makes these validations less complicated then let's take it out.

PR #787, opened 2016-10-11
[#130581] Ability to view and send cross-facility statements

kt32291:
https://pm.tablexi.com/issues/130581

This allows a billing admin to view and send statements across facilities, under the Billing tab on the right side.


app/controllers/facility_statements_controller.rb
```diff
@@ -18,16 +18,13 @@ def initialize
 
   # GET /facilities/:facility_id/statements
   def index
-    @statements =
-      current_facility
-      .statements
-      .order(created_at: :desc)
-      .paginate(page: params[:page])
+    statements = current_facility.cross_billing? ? Statement.all : current_facility.statements
```
jhanggi: I'm wondering if this is a good opportunity for turning the cross-facility "facility" into a true null object. I think it's something that @drench has been wanting to do.

app/controllers/facility_statements_controller.rb
```diff
@@ -18,16 +18,13 @@ def initialize
...
+    @statements = statements.order(created_at: :desc).paginate(page: params[:page])
   end
 
   # GET /facilities/:facility_id/statements/new
   def new_with_search
-    @order_details = @order_details.need_statement(@facility)
+    @order_details = current_facility.cross_billing? ? @order_details.need_any_statement : @order_details.need_statement(@facility)
```
jhanggi: This might tie into the null object idea, but I think this might be better if we do the cross-facility checking into `need_statement` rather than need to do the check several places.

app/models/facility.rb
```diff
@@ -73,6 +73,10 @@ def destroy
     # super
   end
 
+  def cross_billing?
```
jhanggi: We have a `cross_facility?` and `single_facility?` method we should be able to use.

app/models/order_detail.rb
```diff
@@ -266,6 +265,11 @@ def can_be_viewed_by?(user)
       .where("dispute_at IS NULL OR dispute_resolved_at IS NOT NULL")
   }
 
+  scope :need_statement, lambda { |facility|
```
jhanggi: I believe if you use `OrderDetail.for_facility(facility)`, it'll already handle the case of cross-facility.

spec/controllers/facility_statements_controller_spec.rb
```diff
@@ -30,10 +30,14 @@ def create_order_details
 
     before(:each) do
       @authable = FactoryGirl.create(:facility)
+      @other_facility = FactoryGirl.create(:facility)
       @user = FactoryGirl.create(:user)
+      @other_user = FactoryGirl.create(:user)
       UserRole.grant(@user, UserRole::ADMINISTRATOR)
       @account = FactoryGirl.create(@account_sym, account_users_attributes: account_users_attributes_hash(user: @user), facility_id: @authable.id)
+      @other_account = FactoryGirl.create(@account_sym, account_users_attributes: account_users_attributes_hash(user: @other_user), facility_id: @other_facility.id)
```
jhanggi: Do we need these additional objects in all of the tests or can we make them more local to the tests that specifically need them?

kt32291:
@jhanggi I addressed most of the comments but I'd like to do the null object work in a separate PR if that's alright, because I started it briefly, and I think it's going to touch a few parts of the app.

app/controllers/facility_statements_controller.rb
```diff
@@ -56,7 +54,7 @@ def send_statements
 
       @account_statements = {}
       to_statement.each do |account, order_details|
-        statement = Statement.create!(facility: current_facility, account_id: account.id, created_by: session_user.id)
+        statement = Statement.create!(facility: order_details.first.facility, account_id: account.id, created_by: session_user.id)
```
jhanggi: Are the changes here in `send_statements` more related to the cross-facility statementing feature? If so, can we move them out of this PR?

app/models/order_detail.rb
jhanggi: Can we use `OrderDetail.for_facility(facility)` instead of needing extra methods?

app/controllers/facility_statements_controller.rb
kt32291: This is that feature.

kt32291:
@jhanggi that method was already there, I just adjusted it by adding the other one to use for this scenario, and then using it in this one. That existing method is used in other parts of the code I didn't write so I'd rather not remove it.

app/models/order_detail.rb
jhanggi: What I'm trying to suggest is that I believe we can replace`.where("products.facility_id" => facility.id)` with `for_facility(facility)`, and that method should take care of it for us, then we don't need the new `need_any_statement` scope. Does that make sense? Do you think that might work?

jhanggi:
The statement creation functionality looks really good. One last thing: can we add the facility to the Statement History view? We should only display it when you're in the cross-facility context.

kt32291:
@jhanggi updated

app/views/shared/_statements_table.html.haml
```diff
@@ -9,7 +9,7 @@
         %th= Statement.human_attribute_name(:sent_to)
         - if local_assigns[:admin_view]
           %th= Account.model_name.human
-        - else
+        - elsif cross_facility?
```
jhanggi: I'm not sure these should be `elsif`s. Depending on where this partial is shared, this should probably be nested under the `if local_assigns[:admin_view]`, or be a complete independent `if`.

kt32291:
@jhanggi updated

app/views/shared/_statements_table.html.haml
```diff
@@ -9,7 +9,7 @@
         %th= Statement.human_attribute_name(:sent_to)
         - if local_assigns[:admin_view]
           %th= Account.model_name.human
-        - else
+        - if cross_facility?
```
jhanggi: This is not directly related to this ticket, and probably should have been caught in the my statements stories. Can we add the account number to the My Accounts tab to it looks like the transactions/members tabs?

screen shot 2016-10-27 at 3 53 27 pm
screen shot 2016-10-27 at 3 53 19 pm

drench:
:+1:

PR #788, opened 2016-10-14
[#132794] Get rid of unnecessary error and bang method

jhanggi:
The signature of the `AASM::InvalidTransition` changed in the new
version, but we weren’t hitting this code (because it should be truly
exceptional) so we didn’t know. Then we tried to update a quantity to
zero, and this blew up.

The save! happens shortly after this, so here we only need to do one
save (hence getting rid of the bang).

### DIFF
--- a/spec/models/order_detail_spec.rb
+++ b/spec/models/order_detail_spec.rb
@@ -727,6 +727,16 @@ RSpec.describe OrderDetail do
       end
 
     end
+
+    describe "invalid quantity" do
+      it "should not transition to complete" do
+        @order_detail.quantity = 0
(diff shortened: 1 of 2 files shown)


spec/models/order_detail_spec.rb
```diff
@@ -727,6 +727,16 @@ def assign_price_policy
       end
 
     end
+
+    describe "invalid quantity" do
+      it "should not transition to complete" do
```
drench: Extreme nitpick: "should" could be "does"

drench:
:+1:

PR #790, opened 2016-10-18
[#98768] Hiding cancelled orders with cost

kt32291:
https://pm.tablexi.com/issues/98768

We want cancelled orders with cost to have a canceled status, not complete, and this fixes that; then they no longer show in the daily view schedule.

### DIFF
--- a/app/helpers/timeline_helper.rb
+++ b/app/helpers/timeline_helper.rb
@@ -12,7 +12,7 @@ module TimelineHelper
     classes << "admin" if reservation.admin?
     classes << "behalf_of" if reservation.ordered_on_behalf_of?
     classes << "in_progress" if reservation.can_switch_instrument?
-    classes << "status_#{status(reservation)}" if reservation.order_detail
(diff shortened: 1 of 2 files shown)


jhanggi:
Let's chat about this tomorrow because this piece of business logic is a little weird. Cancel with fee has some interesting behavior, going back a long time, even before I started working on the project. In order to make the whole cancel-with-fee work without severe changes to journaling/statementing portions of the code, canceled-with-fee are considered "Complete". Because of this, this ticket is really more of a display issue. In this scenario, we should ignore the state of the OrderDetail, but rather look at `Reservation#canceled?`.

kt32291:
Ok @jhanggi I buy that.

app/helpers/timeline_helper.rb
```diff
@@ -67,4 +67,8 @@ def public_timeline?
     @public_timeline ||= false
   end
 
+  def status(reservation)
```
jhanggi: I'm a little wary of using this generic of a method name in a helper. If another helper implements the same method, I could see them clobbering each other.

PR #793, opened 2016-10-20
[#132384] Adds reconciled filter to date range field

kt32291:
https://pm.tablexi.com/issues/132384

This allows you to use `reconciled_at` for an option on the date range search for order details.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -345,7 +345,7 @@ class OrderDetail < ActiveRecord::Base
     search = search.joins(:journal) if action.to_sym == :journal_date
 
     # If we're searching on fulfilled_at, ignore any order details that don't have a fulfilled date
-    search = search.where("fulfilled_at IS NOT NULL") if action.to_sym == :fulfilled_at
+    search = search.where("#{action} IS NOT NULL") if action.to_sym == :fulfilled_at || :reconciled_at
 
     if start_date
       search = search.where("#{action} >= ?", start_date.beginning_of_day)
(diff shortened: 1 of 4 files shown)


app/models/order_detail.rb
```diff
@@ -345,7 +345,7 @@ class << self
     search = search.joins(:journal) if action.to_sym == :journal_date
 
     # If we're searching on fulfilled_at, ignore any order details that don't have a fulfilled date
-    search = search.where("fulfilled_at IS NOT NULL") if action.to_sym == :fulfilled_at
+    search = search.where("#{action} IS NOT NULL") if action.to_sym == :fulfilled_at || :reconciled_at
```
jhanggi: I think this actually compiles down to `if (action.to_sym == :fulfilled_at) || :reconciled_at`, so it would always be true.

It's whitelisted, so it's not a real SQL injection possiblity, but it would be nice to just avoid that error altogether, either have this as two separate lines or I think we could do `search.where.not(action =>nil)`.

Part of me almost wants to further refactor this scope to use a `case` statement (as a possible first step to a further refactoring).

app/services/transaction_search/date_range_searcher.rb
```diff
@@ -3,7 +3,7 @@ module TransactionSearch
   class DateRangeSearcher < BaseSearcher
 
     include DateHelper
-    FIELDS = %w(ordered_at fulfilled_at journal_or_statement_date).freeze
+    FIELDS = %w(ordered_at fulfilled_at reconciled_at journal_or_statement_date).freeze
```
jhanggi: Journaling/Statementing happens before reconciliation. Can we put `reconciled_at` at the end of the list so the dates are chronological.

jhanggi: One other thing I thought of, but don't need to change right now is that it doesn't make sense to select reconciled or journal/statement while you're sending notifications. We're always using the same list, but a future enhancement might be to hide or disable the unusable options.

kt32291:
@jhanggi updated

PR #794, opened 2016-10-20
[#132462] Fix pagination count issue

kt32291:
https://pm.tablexi.com/issues/132462

I am guessing that this is occurring because of an ActiveRecord Relation size/count problem, due to the complex query that returns the users being counted. This seems to be an issue within the will_paginate gem (see: https://github.com/mislav/will_paginate/issues), so I don't know how much I want to do beyond this, unless we want to swap the gem out with kaminari. 

This is a hack that should fix the pagination page count issue, by turning the ActiveRecord relation into an array.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -34,7 +34,6 @@ class UsersController < ApplicationController
       User
       .with_recent_orders(current_facility)
       .order(:last_name, :first_name)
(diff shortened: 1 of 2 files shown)


jhanggi:
Put this in slack the other day, but to make sure you saw it. On NU, I tried out
```

PR #796, opened 2016-10-21
Fix missing translation

jhanggi:


app/views/facility_accounts_reconciliation/_table.html.haml
```diff
@@ -3,9 +3,9 @@
     %tr
       %th
       %th= Statement.human_attribute_name(:invoice_number)
-      %th= t(".th.order")
+      %th= Order.model_name.human
       %th= Statement.human_attribute_name(:total_cost)
-      %th= t(".th.reconcile_note")
+      %th= Statement.human_attribute_name(:reconcile_note)
```
drench: Should we `titlecase` this? We're not really consistent on this; as an example just above we've got `Statement.human_attribute_name(:invoice_number)` not titlecased.

jhanggi: Since we're localizing it, I think we should title-case it within the locale rather than explicitly call `titleize`. Looks like we need to update it for `invoice_number`.

PR #797, opened 2016-10-21
[#132878] Fix default Sanger batch download

jhanggi:
The plate_mode argument is coming in as either nil or a blank string
because it’s coming from the `group` field on batches. That was causing
`NoMethodError: undefined method `keys' for #<String:0x00000009d8b6f8>`
on production, and a missing translation in development.


drench:
:+1: and I'm going to merge so I can get it out in the next NU stage release

jhanggi:
I actually don't think this was a problem for NU since they don't download CSVs, but still good to get it in.

PR #798, opened 2016-10-24
[#132896] Update date ranges without a page refresh

drench:
This changes the behavior of the "quick links" for selecting date ranges on the bulk email recipient search screen to just update the date fields without a page reload, to preserve other form values.


jhanggi:
👍

PR #801, opened 2016-10-26
[#132660 132661] Log bulk email senders, recipients, subject lines, and search criteria

drench:
On a bulk email send, this stores information about the job: its custom subject line and the email addresses of the recipients.

This also includes support for storing recipient search criteria, but does not do that yet. It also does not yet expose the log in the UI.


db/schema.rb
```diff
@@ -64,6 +64,14 @@
     t.datetime "expires_at",            null: false
   end
 
+  create_table "bulk_email_jobs", force: :cascade do |t|
```
jhanggi: Should we store sent by? Or is that another story?

vendor/engines/bulk_email/app/models/bulk_email/delivery_form.rb
```diff
@@ -41,6 +46,14 @@ def recipients
       @recipients ||= User.where_ids_in(recipient_ids)
     end
 
+    def record_delivery
+      Job.create!(
```
jhanggi: It might be worth prefixing this with `BulkEmail::` just to make sure it's clear which type of `Job` this is.

vendor/engines/bulk_email/app/models/bulk_email/job.rb
```diff
@@ -0,0 +1,20 @@
+module BulkEmail
+
+  class Job < ActiveRecord::Base
+
+    self.table_name = "bulk_email_jobs"
+
+    validates_presence_of :subject
+    validates_with JobJsonFieldValidator
+
+    def search_criteria=(value)
```
jhanggi: I don't think you're setting this yet, but could we use ActiveRecord's built-in serialization? `serialize :search_criteria, JSON`

http://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Serialization/ClassMethods.html

vendor/engines/bulk_email/app/validators/bulk_email/job_json_field_validator.rb
```diff
@@ -0,0 +1,30 @@
+module BulkEmail
+
+  class JobJsonFieldValidator < ActiveModel::Validator
```
jhanggi: Since there's no real user input here, are these validations really necessary? What's the value? They feel like extra overhead without a lot of upside.

db/schema.rb
drench: No, I left the field out. Thanks for catching it.

vendor/engines/bulk_email/app/validators/bulk_email/job_json_field_validator.rb
drench: This was here to hopefully avoid future `rescue`s when reading these back, but if I refactor to use activerecord serialization, then yes these can probably go away.

drench:
Updated; I wrote the search criteria storage parts while this was in review and it was easier to incorporate those changes than split them off at that point.

db/schema.rb
```diff
@@ -64,6 +64,19 @@
...
+    t.integer  "facility_id",     limit: 4,     null: false
+    t.string   "sender",          limit: 255,   null: false
```
jhanggi: I feel like sender should probably be a direct reference to the user. We don't delete users, so there shouldn't be that worry.

vendor/engines/bulk_email/app/models/bulk_email/job.rb
```diff
@@ -0,0 +1,21 @@
+module BulkEmail
+
+  class Job < ActiveRecord::Base
+
+    self.table_name = "bulk_email_jobs"
+
+    belongs_to :facility, foreign_key: :facility_id
+
+    serialize :recipients, Array
+    serialize :search_criteria, Hash
```
jhanggi: This serializes to YAML, correct? I don't have a problem with that; just want to confirm.

drench: Correct; both `:recipients` and `:search_criteria` are serialized to YAML.

vendor/engines/bulk_email/app/models/bulk_email/facility_extension.rb
```diff
@@ -0,0 +1,17 @@
+module BulkEmail
+
+  module FacilityExtension
```
jhanggi: I'm not opposed to it, but do we really need this extension? Could we get away with searching `BulkEmail::Job.where(facility: current_facility)` when we need it? I'm just wondering if it's worth the additional footprint on Facility.

vendor/engines/bulk_email/app/models/bulk_email/user_extension.rb
```diff
@@ -0,0 +1,17 @@
+module BulkEmail
+
+  module UserExtension
+
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :bulk_email_jobs,
```
jhanggi: Same here. Is this relation necessary and worth the overhead? I mean overhead in terms of API, not performance. And it looks like we might not even be extending this onto the User yet.

vendor/engines/bulk_email/db/migrate/20161024213800_create_bulk_email_jobs.rb
```diff
@@ -0,0 +1,15 @@
+class CreateBulkEmailJobs < ActiveRecord::Migration
+
+  def change
+    create_table :bulk_email_jobs do |t|
+      t.references :facility, null: false, foreign_key: true
```
jhanggi: Let's make sure to double check that this actually adds the regular index in Oracle. It's weird about creating a foreign key does not implicitly create an index like it does in MySQL.

jhanggi:
Can we look at the code climate/rubocop configuration and see if we can disable that cop for the specs?

vendor/engines/bulk_email/app/models/bulk_email/facility_extension.rb
drench: We could get by without it, and the `UserExtension` too.

jhanggi:
👍

PR #802, opened 2016-10-27
[#126819] Standardize on a general bulk email instrument offline message

drench:
This change updates the canned text to be the same for all instrument down categories, but leaves the door open for different messages based on category if needed.


jhanggi:
If we add categories in the future, we would need to remember that adding a locale is necessary. What if we just set up `other` and test the `default` in the locale lookup to be other? That way you could override any categories you wanted, but would default to this standard text.

jhanggi:
👍

PR #803, opened 2016-10-27
[#132895] Remove global admin check from the admin reservation listing

drench:
This was a missed `TODO` that should have been done as part of #795.


jhanggi:
👍

PR #804, opened 2016-11-02
[#132927] Prevent double click on Reconcile Orders

jhanggi:
Pulled up from https://github.com/tablexi/nucore-nu/pull/206

### DIFF
--- a/app/views/facility_accounts_reconciliation/_action_row.html.haml
+++ b/app/views/facility_accounts_reconciliation/_action_row.html.haml
@@ -7,4 +7,4 @@
       .controls
         = text_field_tag :reconciled_at, format_usa_date(Time.current), class: :datepicker__data, data: { min_date: unreconciled_order_details.map(&:journal_or_statement_date).min.iso8601, max_date: Time.current.iso8601 }
 
-  .span1.pull-right= submit_tag t("facility_journals.show.submit"), class: ["btn", "btn-primary"]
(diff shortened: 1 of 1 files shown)


app/views/facility_accounts_reconciliation/_action_row.html.haml
```diff
@@ -7,4 +7,4 @@
       .controls
         = text_field_tag :reconciled_at, format_usa_date(Time.current), class: :datepicker__data, data: { min_date: unreconciled_order_details.map(&:journal_or_statement_date).min.iso8601, max_date: Time.current.iso8601 }
 
-  .span1.pull-right= submit_tag t("facility_journals.show.submit"), class: ["btn", "btn-primary"]
+  .span1.pull-right= submit_tag t("facility_journals.show.submit"), class: ["btn", "btn-primary"], data: { disable_with: t("facility_journals.show.submit") }
```
drench: This threw me a bit at first since usually we put different text on the button when it's disabled, but this pattern makes sense.

jhanggi: I'm usually not a fan of changing the text when disabling it, mainly because it usually makes the dimensions of the button change.  I really wish you could just say `disable_with: true`, and it would use the same text for both states. Looks like we're doing a little bit of both ways in this app.

PR #809, opened 2016-11-11
[#130583] Creates separate print view

kt32291:
https://pm.tablexi.com/issues/130583

This creates a separate print view for the details normally displayed in the form for printing. This also removes the now unneeded styles from the print file.

### DIFF
--- a/app/views/order_management/order_details/_print_details.html.haml
+++ b/app/views/order_management/order_details/_print_details.html.haml
@@ -1,23 +1,27 @@
 %div.manage_order_detail
   .banner-list
     .row
-      %dl
+      %dl.span8
         %dt= @order_detail.class.human_attribute_name(:account)
         %dd= @order_detail.account
     = render_view_hook "after_account_print", order_detail: @order_detail
     .row
       = banner_label @order_detail, :order_status
       = banner_date_label @order_detail, :reconciled_at if @order_detail.reconciled_at
-    .row
-      = banner_date_label @order_detail.reservation, :reserve_start_at, OrderDetail.human_attribute_name(:reserve_start_date).titleize
-      %dl.span2
-        %dt= t("activerecord.attributes.reservation.duration_mins")
-        %dd.timeinput= @order_detail.reservation.try(:duration_mins)
-    .row
-      = banner_date_label @order_detail.reservation, :actual_start_at, t("activerecord.attributes.reservation.actual_start_date")
-      %dl.span2
-        %dt= t("activerecord.attributes.reservation.actual_duration_mins")
-        %dd.timeinput= @order_detail.reservation.try(:actual_duration_mins)
+    - if @order_detail.reservation
+      .row
+        = banner_date_label @order_detail.reservation, :reserve_start_at, OrderDetail.human_attribute_name(:reserve_start_date).titleize
+        %dl.span2
+          %dt= t("activerecord.attributes.reservation.duration_mins")
+          %dd.timeinput= @order_detail.reservation.try(:duration_mins)
+      .row
+        = banner_date_label @order_detail.reservation, :actual_start_at, t("activerecord.attributes.reservation.actual_start_date")
+        %dl.span2
+          %dt= t("activerecord.attributes.reservation.actual_duration_mins")
+          %dd.timeinput= @order_detail.reservation.try(:actual_duration_mins)
+    - else
+      .row
+        = banner_label @order_detail, :quantity
     .row
       - prefix = @order_detail.actual_cost? ? "actual" : "estimated"
       - @order_detail.send(:extend, PriceDisplayment)
@@ -29,7 +33,6 @@
       = banner_label @order_detail, :note
       - if @order_detail.reconciled_note.present?
         = banner_label @order_detail, :reconciled_note
-      - if @order_detail.complete?
-        %dl.span2
-          %dt Assigned Staff
-          = @order_detail.assigned_user ? @order_detail.assigned_user.full_name : 'Unassigned'
+      %dl.span2
+        %dt Assigned Staff
+        = @order_detail.assigned_user ? @order_detail.assigned_user.full_name : 'Unassigned'
--- a/app/views/order_management/order_details/_print_details.html.haml
+++ b/app/views/order_management/order_details/_print_details.html.haml
@@ -1,19 +1,15 @@
-.show-for-print
+%div.manage_order_detail
   .banner-list
     .row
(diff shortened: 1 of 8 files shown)


kt32291:
This looked good in broswerstack simulating Windows.

app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -0,0 +1,38 @@
+.show-for-print
+  .banner-list
+    .row
+      %dl
+        %dt= @order_detail.class.human_attribute_name(:account)
+        %dd= @order_detail.account
+    - if @order_detail.project
```
jhanggi: We should probably use a view hook for this. In the forks that don't have the projects engine enabled, `OrderDetail` won't respond to `project`.

app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -0,0 +1,38 @@
...
+      .row
+        %dl
+          %dt= @order_detail.class.human_attribute_name(:project)
+          %dd= @order_detail.project.name
+    .row
+      = banner_label @order_detail, :order_status
+      = banner_date_label @order_detail, :reconciled_at if @order_detail.reconciled_at
+    .row
+      = banner_date_label @order_detail.reservation, :reserve_start_at, t("activerecord.attributes.reservation.reserve_start_date")
```
jhanggi: Could we call `OrderDetail.human_attribute_name(:reserve_start_date)`? I feel like that might be a little cleaner.

app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -0,0 +1,38 @@
...
+        %dt= t("activerecord.attributes.reservation.duration_mins")
+        %dd.timeinput= @order_detail.reservation.try(:duration_mins)
+    .row
+      = banner_date_label @order_detail.reservation, :actual_start_at, t("activerecord.attributes.reservation.actual_start_date")
+      %dl.span2
+        %dt= t("activerecord.attributes.reservation.actual_duration_mins")
+        %dd.timeinput= @order_detail.reservation.try(:actual_duration_mins)
+    .row
+      - @order_detail.send(:extend, PriceDisplayment)
+      = banner_label @order_detail, :display_cost, t("activerecord.attributes.order_detail.actual_cost")
```
jhanggi: I feel like we should probably distinguish between actual and estimated price in the labels like we do in the edit view.

app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -0,0 +1,38 @@
...
+      = banner_label @order_detail, :display_subsidy, t("activerecord.attributes.order_detail.actual_subsidy")
+      = banner_label @order_detail, :display_total, t("activerecord.attributes.order_detail.actual_total")
+
+    .row
+      = banner_label @order_detail, :note
+      - if @order_detail.can_reconcile? || @order_detail.reconciled?
```
jhanggi: Could this be condensed down to `if @order_detail.reconciled_note.present?`?

kt32291:
@jhanggi I decided to do it this way because I want them to be able to access it the same way, and not have to use a different route.

jhanggi:
I think keeping it on the same route is fine, but I feel like it's now made the view more complicated to understand.

I'm throwing this idea for organization out there. I'm open to other ideas as well.

```haml
# edit.html.haml
.banner-list
  -# All the stuff that's shared
.hide-from-print
  = render "form"
.show-for-print
  = render "print_details

```

kt32291:
@jhanggi I dig that. I'll go with that approach.

kt32291:
Updated! @jhanggi

app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -0,0 +1,35 @@
...
+    .row
+      %dl
+        %dt= @order_detail.class.human_attribute_name(:account)
+        %dd= @order_detail.account
+    = render_view_hook "after_account_print", order_detail: @order_detail
+    .row
+      = banner_label @order_detail, :order_status
+      = banner_date_label @order_detail, :reconciled_at if @order_detail.reconciled_at
+    .row
+      = banner_date_label @order_detail.reservation, :reserve_start_at, OrderDetail.human_attribute_name(:reserve_start_date).titleize
```
jhanggi: We need a conditional in here for if there is a reservation include the times, but if it's not a reservation, then show the quantity.

vendor/engines/projects/lib/projects/engine.rb
```diff
@@ -22,6 +22,10 @@ class Engine < Rails::Engine
                         "after_order_status",
                         "projects/shared/select_project"
 
+      ViewHook.add_hook "order_management.order_details.print_details",
```
jhanggi: This actually applies to the view hook above this, but since the partial the view hook is in changed, it needs to change. Now, we're not showing the project on the form anymore.

app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -0,0 +1,35 @@
...
+      - @order_detail.send(:extend, PriceDisplayment)
+      = banner_label @order_detail, :display_cost, t("activerecord.attributes.order_detail.#{prefix}_cost")
+      = banner_label @order_detail, :display_subsidy, t("activerecord.attributes.order_detail.#{prefix}_subsidy")
+      = banner_label @order_detail, :display_total, t("activerecord.attributes.order_detail.#{prefix}_total")
+
+    .row
+      = banner_label @order_detail, :note
+      - if @order_detail.reconciled_note.present?
+        = banner_label @order_detail, :reconciled_note
+      - if @order_detail.complete?
```
jhanggi: I think the assigned staff should either always be shown in the print view or if there is someone assigned. The form side displays the assigned user if the order is complete and a dropdown if it's not--the dropdown may be populated, and is editable until the order is complete.

jhanggi:
Would it be possible to generate a couple of PDFs from different browsers that we can attach to the ticket to make sure the client is happy with it?

kt32291:
@jhanggi updated! 

Chrome:
[nucore-chrome.pdf](https://github.com/tablexi/nucore-open/files/598392/nucore-chrome.pdf)

Safari:
[nucore-safari.pdf](https://github.com/tablexi/nucore-open/files/598393/nucore-safari.pdf)

Screenshot of IE print preview (from browserstack):
screen shot 2016-11-17 at 2 29 22 pm

PR #810, opened 2016-11-11
[#133123] Disable sending of statements with feature flag

kt32291:
https://pm.tablexi.com/issues/133123

This allows you to set a feature flag that will skip the sending of statements to all the payment sources, because Dartmouth does that manually.


app/controllers/facility_statements_controller.rb
```diff
@@ -25,15 +25,15 @@ def index
...
-    @order_detail_action = :send_statements
+    @order_detail_action = :create_statements
     set_default_start_date if SettingsHelper.feature_on?(:set_statement_search_start_date)
     @layout = "two_column_head"
   end
 
-  # POST /facilities/:facility_id/statements/send_statements
-  def send_statements
+  # POST /facilities/:facility_id/statements/create_statements
+  def create_statements
```
jhanggi: I fully support this change in naming, although would simply `create` work as well and make it a little more RESTful?

config/locales/en.controllers.yml
```diff
@@ -44,10 +44,10 @@ en:
 
     facility_statements:
       errors_html: We experienced the following errors. Please try again.<br/>%{errors}
-      send_statements:
+      create_statements:
         order_error: Order %{order_detail_id} was either not found or cannot be statemented at this time.
         no_selection: No orders selected
-        success_html: Notifications sent successfully to:<br/>%{accounts}
+        success_html: Statements made successfully for:<br/>%{accounts}
```
jhanggi: I wonder if we should have separate success messages here as well like we do for the instructions.

config/settings.yml
```diff
@@ -112,6 +112,7 @@ feature:
   # results file notifications requires that my_files be on as well
   results_file_notifications_on: true
   set_statement_search_start_date_on: false
+  skip_statement_emails_on: false
```
jhanggi: What would you think of inverting the naming here? `send_statement_emails_on` or more explicitly `send_emails_on_statement_create_on` (I find the `_on` suffix really annoying, but I don't think it's worth the effort to change)? Then everyone except DC is set to `true`. I think that might clarify some logic.

app/controllers/facility_statements_controller.rb
```diff
@@ -67,11 +67,13 @@ def send_statements
...
-        @account_statements.each do |account, statement|
-          account.notify_users.each { |u| Notifier.delay.statement(user: u, facility: statement.facility, account: account, statement: statement) }
+        unless SettingsHelper.feature_on?(:skip_statement_emails)
+          @account_statements.each do |account, statement|
+            account.notify_users.each { |u| Notifier.delay.statement(user: u, facility: statement.facility, account: account, statement: statement) }
+          end
         end
         account_list = @account_statements.map { |a, _s| a.account_list_item }
-        flash[:notice] = I18n.t("controllers.facility_statements.send_statements.success_html", accounts: account_list.join("<br/>")).html_safe
+        flash[:notice] = I18n.t("controllers.facility_statements.create_statements.success_html", accounts: account_list.join("<br/>")).html_safe
```
jhanggi: I wonder if it would be worth taking the time while we're in to switch this to use `text_helpers`.

app/views/facility_statements/new.html.haml
```diff
@@ -9,7 +9,10 @@
   = render :partial => 'shared/transactions/top', :locals => { :tab => 'statements' }
```
jhanggi: While we're in this file, can we clean up the style? Hash syntax, double quotes, prefer `render "shared/transactions/top", tab: "statements"` to using the `partial` option.

app/views/facility_statements/new.html.haml
```diff
@@ -9,7 +9,10 @@
...
 - if @order_details.any?
   = content_for :action_instructions do
-    %p.notice.notice-info= t("facility_statements.new.instructions")
+    - if SettingsHelper.feature_on?(:skip_statement_emails)
+      %p.notice.notice-info= t("facility_statements.new.instructions_without_email")
```
jhanggi: Maybe use `text_helpers` and get it better organized so we can just say `text("instructions_without_email")`?

config/settings.yml
kt32291: So I did this because I believe the default should be to send them. Turning a feature on feels like overriding a default. But maybe I'm thinking about it the wrong way.

spec/controllers/facility_statements_controller_spec.rb
```diff
@@ -133,19 +134,40 @@ def create_order_details
...
-        expect(@order_detail1.dispute_at).to be_nil
-
-        grant_and_sign_in(@user)
-        do_request
-        expect(flash[:error]).to be_nil
-        expect(assigns(:account_statements)).to have_key(@account)
-        expect(assigns(:account_statements)).to have_key(@account2)
-        expect(response).to redirect_to action: :new
+      context "when statement emailing is on", feature_setting: { skip_statement_emails: false } do
+        it "should create and send statements" do
```
drench: This and its opposite (lines 155-170) are nearly identical. Pulling the common parts out into shared examples would highlight the differences.

config/routes.rb
```diff
@@ -281,7 +281,7 @@
 
     resources :statements, controller: "facility_statements", only: [:index, :new, :show] do
```
jhanggi: I believe we can add `create` to the `only` list here. That will change the path to be `POST /facilities/:facility_id/facility_statements` rather than `/create`.

spec/models/statement_creator_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+  let(:order_detail_1) { place_and_complete_item_order(user, facility, account, true) }
+  let(:order_detail_2) { place_and_complete_item_order(user, facility, account, true) }
+  let(:order_detail_3) { place_and_complete_item_order(user, facility, account, false) }
+
+  def initialize_creator
+    @creator = described_class.new(order_detail_ids: [order_detail_1.id, order_detail_2.id], session_user: user, current_facility: facility)
+  end
+
+  describe "#new" do
+    before { initialize_creator }
```
jhanggi: Could this be collapsed to `let(:creator) { described_class.new(...) }`?

spec/support/facility_statements_shared_examples.rb
```diff
@@ -0,0 +1,14 @@
+RSpec.shared_examples "it sets up order_detail and creates statements" do
```
jhanggi: This is more of a personal preference, but I prefer to leave the shared examples in the spec file if it's only used in a single file. One less file to track down.

spec/models/statement_creator_spec.rb
```diff
@@ -0,0 +1,98 @@
+require "rails_helper"
+
+RSpec.describe StatementCreator do
+  let(:user) { create(:user) }
+  let(:facility) { create(:facility) }
+  let(:account) { FactoryGirl.create(:nufs_account,
```
jhanggi: Few rubocop things on this. I suspect it could be autocorrected.

spec/controllers/facility_statements_controller_spec.rb
```diff
@@ -133,19 +135,26 @@ def create_order_details
...
-        expect(assigns(:account_statements)).to have_key(@account2)
-        expect(response).to redirect_to action: :new
+      context "when statement emailing is on", feature_setting: { send_statement_emails: true } do
+        include_examples "it sets up order_detail and creates statements"
+
+        it "sends statements" do
+          grant_and_sign_in(@user)
+          do_request
+
+          expect(ActionMailer::Base.deliveries).not_to be_empty
```
jhanggi: I could be wrong, but I don't believe we do `ActionMailer::Base.deliveries.clear` always, so this it's possible that this test is inaccurate. For these kind of things, I like `expect { do_request }.not_to change(ActionMailer::Base.deliveries, :count)` and `expect { do_request }.to change(ActionMailer::Base.deliveries, :count).by(1)`

config/locales/views/en.facility_statements.yml
```diff
@@ -0,0 +1,10 @@
+en:
```
jhanggi: Can we move this into `config/locales/views/admin`?

app/models/statement_creator.rb
```diff
@@ -0,0 +1,63 @@
...
+
+  def create
+    OrderDetail.transaction do
+      set_order_details_to_statement
+      setup_statement_from_details
+    end
+  end
+
+  def formatted_errors
+    errors.join("<br/>")
```
jhanggi: I could go either way, but part of me feels like the `join` belongs more in the controller than the service object, especially since we have two of these, and they're HTML specific. I don't have a strong feeling, though.

app/models/statement_creator.rb
```diff
@@ -0,0 +1,63 @@
+class StatementCreator
+  attr_accessor :order_detail_ids, :errors, :to_statement, :account_statements, :session_user, :current_facility
```
jhanggi: Rubocop: Empty line missing at class body beginning.

app/models/account.rb
```diff
@@ -138,6 +138,10 @@ def notify_users
     [owner_user] + business_admin_users
   end
 
+  def send_notification(statement)
```
jhanggi: Another thing where I don't necessarily feel strongly on, but this is seems like it's part of the responsibility of the StatementCreator.

app/models/statement_creator.rb
```diff
@@ -0,0 +1,63 @@
+class StatementCreator
```
jhanggi: We tend to put these kind of objects in `app/services`.

app/controllers/facility_statements_controller.rb
```diff
@@ -25,61 +25,38 @@ def index
...
-        flash[:error] = I18n.t("controllers.facility_statements.errors_html", errors: @errors.join("<br/>")).html_safe
+  # POST /facilities/:facility_id/statements/create
+  def create
+    @statement_creator = StatementCreator.new(order_detail_ids: params[:order_detail_ids], session_user: session_user, current_facility: current_facility)
+    if @statement_creator.order_detail_ids.nil? || @statement_creator.order_detail_ids.empty?
+      flash[:error] = text("no_selection")
+    elsif @statement_creator.create
+      if @statement_creator.errors.any?
+        flash[:error] = text("errors_html", errors: @statement_creator.formatted_errors).html_safe
         raise ActiveRecord::Rollback
```
jhanggi: This rollback needs to stay inside the transaction.

app/controllers/facility_statements_controller.rb
```diff
@@ -25,61 +25,38 @@ def index
...
-      end
-
-      if @errors.any?
-        flash[:error] = I18n.t("controllers.facility_statements.errors_html", errors: @errors.join("<br/>")).html_safe
+  # POST /facilities/:facility_id/statements/create
+  def create
+    @statement_creator = StatementCreator.new(order_detail_ids: params[:order_detail_ids], session_user: session_user, current_facility: current_facility)
+    if @statement_creator.order_detail_ids.nil? || @statement_creator.order_detail_ids.empty?
+      flash[:error] = text("no_selection")
+    elsif @statement_creator.create
```
jhanggi: Would it make sense for `create` to return false if there are any errors and not have the nested conditional?

app/controllers/facility_statements_controller.rb
```diff
@@ -25,61 +25,38 @@ def index
...
-        end
-        @account_statements[account] = statement
-      end
-
-      if @errors.any?
-        flash[:error] = I18n.t("controllers.facility_statements.errors_html", errors: @errors.join("<br/>")).html_safe
+  # POST /facilities/:facility_id/statements/create
+  def create
+    @statement_creator = StatementCreator.new(order_detail_ids: params[:order_detail_ids], session_user: session_user, current_facility: current_facility)
+    if @statement_creator.order_detail_ids.nil? || @statement_creator.order_detail_ids.empty?
```
kt32291: Blank!! I was completely BLANK-ing (pun intended) when I was trying to think of the method that covers both :P

kt32291:
@jhanggi updated again!

app/services/statement_creator.rb
```diff
@@ -0,0 +1,71 @@
...
+  end
+
+  private
+
+  def set_order_details_to_statement
+    order_detail_ids.each do |order_detail_id|
+      od = nil
+      begin
+        od = OrderDetail.need_statement(current_facility).readonly(false).find(order_detail_id)
+        @to_statement[od.account] ||= []
```
jhanggi: I think if we change the intializer to `@to_statement = Hash.new([])` so it defaults to empty array, I think we can get rid of this. Don't necessarily need to do this now. I would be comfortable holding off on further refactors until we get to DC's one-email-per-generation.

app/services/statement_creator.rb
```diff
@@ -0,0 +1,71 @@
+class StatementCreator
+
+  attr_accessor :order_detail_ids, :errors, :to_statement, :account_statements, :session_user, :current_facility
+
+  def initialize(params)
+    @order_detail_ids = params[:order_detail_ids]
```
jhanggi: One option to consider for the future is to make these keyword args.

config/locales/views/en.facility_notifications.yml
```diff
@@ -0,0 +1,15 @@
...
+  views:
+    facility_notifications:
+      in_review:
+        head: "Orders In Review"
+        no_orders: "No orders are currently in review"
+        instructions: |
+          The orders listed below are still in review by the customer.  The time the review window will close is listed.
+          You may manually mark an order as reviewed, allowing you to create a statement for or journal the order immediately.
+        submit: "Mark As Reviewed"
+      index:
```
jhanggi: Can this chunk be removed from `en.yml` since it's been moved here?

config/locales/views/en.shared.yml
```diff
@@ -0,0 +1,6 @@
+en:
+  views:
+    shared:
+      sidenav_billing:
+        statements:
+          head: "Send Notifications"
```
jhanggi: sidenav_billing is within the admin context, so we should move it there.

config/locales/views/admin/en.facility_statements.yml
```diff
@@ -0,0 +1,10 @@
+en:
+  views:
+    facility_statements:
+      new:
+        instructions: Please review the following payment sources. A statement will be generated for each of the payment sources and an email with a link to the statement will be sent.
+        instructions_without_email: Please review the following payment sources. A statement will be generated for each of the payment sources.
+        no_orders: There are no completed orders that require statements.
+      index:
+        head: Statement History
```
jhanggi: Not quite sure how this happened, but in the sidebar, the "Statement History" is appearing as "Send Notifications".

config/locales/views/en.facility_notifications.yml
jhanggi: Actually, should this file have been included in this PR? I don't see any changes to the referencing views.

app/services/statement_creator.rb
```diff
@@ -0,0 +1,71 @@
...
+    @current_facility = params[:current_facility]
+    @errors = []
+    @to_statement = {}
+  end
+
+  def create
+    OrderDetail.transaction do
+      set_order_details_to_statement
+      setup_statement_from_details
+      raise ActiveRecord::Rollback if errors.any?
```
kt32291: It did not return true by default, I had to explicitly state a return value (hence the last line)

kt32291:
@jhanggi updated. I skipped the changes that weren't necessary for this go-round so we can get this merged in.

jhanggi:
👍

PR #812, opened 2016-11-14
[#130772] Keep bulk email state on compose form errors

drench:
This should solve the problem of losing bulk email "state" when running into a form error on the mail compose screen. It should also clean up the issue where a flash error message generated in that case was lingering beyond where it made sense to display.

I'm not super-pleased about `#populate_flow_state_params`. I added doc, but does it deserve an extraction?

### DIFF
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -61,15 +61,12 @@ module BulkEmail
     def populate_flow_state_params
       return if params[:bulk_email_delivery_form].blank?
       search_criteria = JSON.parse(params[:bulk_email_delivery_form][:search_criteria])
-      params[:product_id] ||= params[:bulk_email_delivery_form][:product_id]
-      params[:start_date] ||= search_criteria["start_date"]
-      params[:end_date] ||= search_criteria["end_date"]
-      params[:bulk_email] ||= {}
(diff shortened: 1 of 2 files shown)


vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -45,14 +45,33 @@ def deliver
...
 
     private
 
+    # There are several hidden parameters in the bulk email compose form that
+    # hold state on how the user got here, what product(s) the mail is about,
+    # and how the recipients were chosen. If there is a form error, these params
+    # need to be set to keep that state in the redisplayed form.
+    def populate_flow_state_params
+      return if params[:bulk_email_delivery_form].blank?
+      search_criteria = JSON.parse(params[:bulk_email_delivery_form][:search_criteria])
```
drench: You're right on the `||=`; they don't need to be there.

The problem I run into with `bulk_email_recipient_search_params` is that in this context (when form validation fails), those values only exist in the `search_criteria` as JSON. They could conceivably exist as added hidden form fields, though that feels like it would clutter the form with extra params that the "happy path" would just throw away.

jhanggi: could we do something with `slice` or `merge`? 

`params = search_criteria.merge(params[:bulk_email_delivery_form])`. Just some way of not needing to specify every single field?

jhanggi:
👍

PR #813, opened 2016-11-14
[#132656] Add bulk mail history views

drench:
This adds a "History" tab to the Bulk Email section. The index view is a paginated list of bulk email jobs, most recent first. Clicking through to a show view has complete details about the job: what user triggered it, when, who the recipients were, the search criteria used to generate the recipient list, along with the subject and email body content.

The subject and body content are as they are stored in the database. Currently that means it does not include any of the "canned" content parts. If we want to include those extras (and I think we do), that will be a separate change to bulk email logging.

### DIFF
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -59,7 +59,8 @@ module BulkEmail
     end
 
     def job
-      @bulk_email_job = BulkEmail::Job.where(facility: current_facility).find(params[:id])
+      job = BulkEmail::Job.where(facility: current_facility).find(params[:id])
+      @bulk_email_job = BulkEmail::JobDecorator.new(job)
     end
 
     private
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -51,6 +51,13 @@ module BulkEmail
       end
     end
 
+    def history
+      @bulk_email_jobs = current_facility
+                         .bulk_email_jobs
+                         .order(created_at: :desc)
+                         .paginate(page: params[:page])
+    end
+
     private
 
     def bulk_email_cancel_path
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -52,10 +52,14 @@ module BulkEmail
     end
 
     def history
-      @bulk_email_jobs = current_facility
-                         .bulk_email_jobs
+      @bulk_email_jobs = BulkEmail::Job
+                         .where(facility: current_facility)
                          .order(created_at: :desc)
                          .paginate(page: params[:page])
+    end
+
+    def job
+      @bulk_email_job = BulkEmail::Job.where(facility: current_facility).find(params[:id])
     end
 
     private
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -43,13 +43,33 @@ module BulkEmail
         flash[:notice] = text("bulk_email.delivery.success", count: @delivery_form.recipient_ids.count)
         redirect_to delivery_success_path
       else
-        flash[:error] = text("bulk_email.delivery.failure")
+        flash.now[:error] = text("bulk_email.delivery.failure")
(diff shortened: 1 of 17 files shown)


vendor/engines/bulk_email/app/views/bulk_email/bulk_email/history.html.haml
```diff
@@ -0,0 +1,32 @@
+= content_for :tabnav do
+  = render "nav_tab", secondary_tab: "history"
+
+= content_for :h1 do
+  = current_facility
+
+%h2= t("bulk_email.title")
```
jhanggi: Should we use `text_helper`s?

vendor/engines/bulk_email/config/routes.rb
```diff
@@ -3,5 +3,7 @@
     get "bulk_email", to: "bulk_email/bulk_email#search"
     post "bulk_email", to: "bulk_email/bulk_email#create"
     post "bulk_email/deliver", to: "bulk_email/bulk_email#deliver"
+    get "bulk_email/history", to: "bulk_email/bulk_email#history"
+    get "bulk_email/job/:id", to: "bulk_email/bulk_email#job", as: "bulk_email_job"
```
jhanggi: The BulkEmailController has started to get pretty un-RESTful. Would it be worth splitting these out into `index` and `show` methods on a BulkEmail::JobsController?

vendor/engines/bulk_email/app/views/bulk_email/bulk_email/job.html.haml
```diff
@@ -0,0 +1,21 @@
...
+= content_for :sidebar do
+  = render "admin/shared/sidenav_users", sidenav_tab: "bulk_email"
+
+= readonly_form_for :bulk_email_job do |f|
+  = f.input :sender
+  = f.input :created_at
+  = f.input :recipient_list
+  = f.input :start_date
+  = f.input :end_date
+  = f.input :user_types
```
jhanggi: This may be getting picky, but should we display the fields in the same order they're on the form? So user_types, products, start_date, end_date.

jhanggi: And part of me thinks it might be good to group the search criteria and the content/recipient_list.

vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
```diff
@@ -45,14 +43,26 @@ def deliver
         flash[:notice] = text("bulk_email.delivery.success", count: @delivery_form.recipient_ids.count)
         redirect_to delivery_success_path
       else
-        flash[:error] = text("bulk_email.delivery.failure")
+        flash.now[:error] = text("bulk_email.delivery.failure")
```
drench: I cherry-picked #812 so the diff is a little redundant from here to line 64.

vendor/engines/bulk_email/app/decorators/bulk_email/job_decorator.rb
```diff
@@ -0,0 +1,42 @@
...
+    def sender
+      user.email
+    end
+
+    def recipient_list
+      recipients.join(", ")
+    end
+
+    def products
+      Product.find(selected_product_ids).map(&:name).join(", ")
```
jhanggi: I created an email where I selected no products (so not through instrument not available). This is causing the details view to 404 because it's trying to do `SELECT  `products`.* FROM `products` WHERE `products`.`id` = 0 LIMIT 1`

jhanggi: Seems like it might have more to do with a blank `product_id`. If I have a product in the search_criteria, but a blank `product_id`, then I'm getting the same error.

vendor/engines/bulk_email/app/views/bulk_email/jobs/show.html.haml
```diff
@@ -0,0 +1,23 @@
...
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_users", sidenav_tab: "bulk_email"
+
+= readonly_form_for :bulk_email_job do |f|
+  = f.input :sender
+  = f.input :created_at
+  = f.input :recipient_list
+  = f.input :subject
+  = f.input :body
```
jhanggi: Should we `simple_format` this like we do the actual email?

I did discover that `simple_format` will allow some HTML through (like `<strong>`, `<a>`, and `<h1>` tags), but does sanitize for script. I think I'm ok with that. Or, according to the docs, we should invoke `h` before passing to simple_format (both here and the email itself).

http://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html

vendor/engines/bulk_email/app/decorators/bulk_email/job_decorator.rb
drench: Good catch; it's legit to have no products associated with an email.

jhanggi:
👍

PR #814, opened 2016-11-16
[#133346] Include boilerplate and user-entered content when logging bulk email content

drench:
This changes bulk email logging so that it includes the unalterable "canned" parts of the subject and body along with user-entered text.

As a side-effect it ended up simplifying `BulkEmail::Mailer` quite a bit.

### DIFF
--- a/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
+++ b/vendor/engines/bulk_email/app/controllers/bulk_email/bulk_email_controller.rb
@@ -112,7 +112,7 @@ module BulkEmail
     end
 
     def init_delivery_form
-      @delivery_form = DeliveryForm.new(current_user, current_facility)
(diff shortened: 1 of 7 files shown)


jhanggi:
Not sure where it's happening, but now, the greeting it not getting interpolated; it's sending the email as "Hello Firstname Lastname,"

drench:
> the greeting it not getting interpolated

Good catch; I think I found it. The injected `BulkEmail::ContentGenerator` takes a `recipient` but this isn't setting it, so it's falling back to the default `#recipient_name`.

drench:
Updated

jhanggi:
If I send via instrument offline, the "Hello" is present in both the emails and the log, but the canned "Example Instrument has been taken offline. You may receive additional messages regarding the status of this instrument." is not.

drench:
Give it another test. I didn't bisect to be sure, but I think this loss of the `product_id` parameter (which is the root of the message change) happened earlier.

jhanggi:
👍

PR #816, opened 2016-11-18
[#133375] Fix access to statements for facility admin/director

jhanggi:
Related to #784 and #787, facility staff were getting 403s when trying
to download statements. `statement_path` is a alias to
`facility_account_statement_path`.

When testing, we should make sure all of these views have access to
their statements:
* as a PI looking through “My Payment Sources” (this should be the one
place we use `account_statement_path`.
* as a facility admin, director
* Under “Statement History”
* Under “Payment Sources”…”Statements”
* as a billing admin looking under the cross facility tab under
“Statement History”

### DIFF
--- a/spec/controllers/facility_accounts_controller_spec.rb
+++ b/spec/controllers/facility_accounts_controller_spec.rb
@@ -325,19 +325,37 @@ RSpec.describe FacilityAccountsController, feature_setting: { edit_accounts: tru
       is_expected.to render_template "show_statement_list"
     end
 
-    it "should show statement PDF" do
-      @params[:statement_id] = @statement.id
-      @params[:format] = "pdf"
-      maybe_grant_always_sign_in :director
-      do_request
-      expect(assigns(:account)).to eq(@account)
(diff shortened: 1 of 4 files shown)


spec/controllers/facility_accounts_controller_spec.rb
```diff
@@ -325,19 +325,37 @@
...
+      end
 
+      it "shows the statement PDF for a director" do
+        maybe_grant_always_sign_in :director
+        do_request
+        expect(assigns(:account)).to eq(@account)
+        expect(assigns(:facility)).to eq(@authable)
+        expect(assigns(:statement)).to eq(@statement)
+        expect(response.content_type).to eq("application/pdf")
+        expect(response.body).to match(/%PDF-1.\d/)
```
drench: Old issue, but it probably should be looking at the start of the string. `match(/\A%PDF-1\.\d+\b/)` maybe?

jhanggi: Good call. Updated.

PR #817, opened 2016-11-18
[#132637] Cart should not include merge orders

jhanggi:
* As a facility staff/admin, add a service requiring an order form or a
reservation to an existing order
* You will now have a “The following order details need your
attention.“ listing
* Go to your cart
* The attention-needed order appears as your cart
* Remove the order detail
* Get `orders#remove (RuntimeError) "Can't modify frozen hash”` error

What was happening:
* The Order is destroyed by OrderDetailObserver#after_destroy
* In OrdersController#remove, we trying to set `order.account_id = nil`
on an already destroyed order.

The solution is to make sure the merge order is not considered when
pulling up your cart.

This also fixes some translation missing errors.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -21,7 +21,7 @@ class Order < ActiveRecord::Base
   end
 
   def self.carts
-    where(ordered_at: nil)
+    where(ordered_at: nil, merge_with_order_id: nil)
   end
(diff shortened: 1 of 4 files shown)


kt32291:
Very nice :)

PR #821, opened 2016-11-21
[#126780] Change scope for AvailableAccountFinder

kt32291:
https://pm.tablexi.com/issues/126780

This uses the user that the order was placed for instead of the acting user to scope payment sources for the order detail editing.

### DIFF
--- a/app/views/order_details/edit.html.haml
+++ b/app/views/order_details/edit.html.haml
@@ -10,7 +10,7 @@
 = simple_form_for([@order, @order_detail]) do |f|
   %p.alert.alert-warning= text(".warning")
 
-  = f.input :account_id, collection: AvailableAccountsFinder.new(@order_detail.order.user, @order_detail.facility, current: @order_detail.account), include_blank: false, label: Account.model_name.human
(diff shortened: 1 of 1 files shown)


app/views/order_details/edit.html.haml
```diff
@@ -10,7 +10,7 @@
 = simple_form_for([@order, @order_detail]) do |f|
   %p.alert.alert-warning= text(".warning")
 
-  = f.input :account_id, collection: AvailableAccountsFinder.new(acting_user, @order_detail.facility, current: @order_detail.account), include_blank: false, label: Account.model_name.human
+  = f.input :account_id, collection: AvailableAccountsFinder.new(@order_detail.order.user, @order_detail.facility, current: @order_detail.account), include_blank: false, label: Account.model_name.human
```
jhanggi: We delegate `OrderDetail#user` to Order, so we don't need that in there.

kt32291:
@jhanggi but that should be the caveat, right?

jhanggi:
Yeah, I don't think we need to change anything there, I just want to make sure that we're aware of it.

kt32291:
@jhanggi updated

PR #822, opened 2016-11-22
[#133079] Hide grey background when archived/hidden instruments

jhanggi:
On the public timeline, if there were archived or hidden instruments,
the grey background was appearing even if there was only one visible
instrument. On the facility timeline view, archived instruments were
causing the same problem.

This also removes an unused method on Instrument.

### DIFF
--- a/app/views/shared/timeline/_instrument.html.haml
+++ b/app/views/shared/timeline/_instrument.html.haml
@@ -1,26 +1,25 @@
-- unless instrument.is_archived? || (instrument.is_hidden? && public_timeline?)
-  .timeline_instrument
-    %h4
-      - if public_timeline?
-        = link_to instrument,
-          facility_instrument_path(current_facility, instrument)
-      - else
-        = link_to instrument,
-          facility_instrument_schedule_path(current_facility, instrument)
+.timeline_instrument
+  %h4
+    - if public_timeline?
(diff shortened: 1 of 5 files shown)


app/views/shared/timeline/_instrument.html.haml
```diff
@@ -1,26 +1,25 @@
...
+  .timeline
+    .unit_container
+      = render partial: "shared/timeline/reservation",
+        collection: (instrument.visible_reservations(@display_datetime) + ScheduleRule.unavailable_for_date(instrument, @display_datetime)),
+        as: :reservation,
+        locals: { product: instrument }
 
-        - if @display_datetime.beginning_of_day == Time.zone.now.beginning_of_day
-          .current_time{style: "left: #{datetime_left_position(@display_datetime, Time.zone.now)}"}
+      - if @display_datetime.beginning_of_day == Time.zone.now.beginning_of_day
```
drench: Minor: we're favoring `Time.current` now, I think

jhanggi: Yeah, we are preferring that. The only thing I did in this file was remove the top conditional (since it's now taken care of by the scope in`schedule`) and un-indented everything. But I'll make the change.

kt32291:
🎉

PR #823, opened 2016-11-28
[#132298] Ensure actuals are not set for past on-behalf orders

kt32291:
https://pm.tablexi.com/issues/132298

Actuals should not be set for past orders, and this adds a param to the method that controls the setting of actuals. Also adds spec to the reservations controller spec to check actuals are not being set for orders in the past.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -461,11 +461,11 @@ class OrderDetail < ActiveRecord::Base
     update(state: "complete", order_status: OrderStatus.complete_status)
   end
 
-  def backdate_to_complete!(event_time = Time.zone.now, set_actuals = true)
+  def backdate_to_complete!(event_time = Time.zone.now)
     # if we're setting it to compete, automatically set the actuals for a reservation
     if reservation
       raise NUCore::PurchaseException.new(t_model_error(Reservation, "cannot_be_completed_in_future")) if reservation.reserve_end_at > event_time
-      reservation.assign_actuals_off_reserve if set_actuals
+      reservation.assign_actuals_off_reserve unless reservation.product.reservation_only?
       reservation.save!
     end
     change_status!(OrderStatus.complete.first) do |od|
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -461,11 +461,11 @@ class OrderDetail < ActiveRecord::Base
     update(state: "complete", order_status: OrderStatus.complete_status)
   end
 
-  def backdate_to_complete!(event_time = Time.zone.now)
+  def backdate_to_complete!(event_time = Time.zone.now, set_actuals = true)
     # if we're setting it to compete, automatically set the actuals for a reservation
     if reservation
       raise NUCore::PurchaseException.new(t_model_error(Reservation, "cannot_be_completed_in_future")) if reservation.reserve_end_at > event_time
-      reservation.assign_actuals_off_reserve
+      reservation.assign_actuals_off_reserve if set_actuals
       reservation.save!
     end
     change_status!(OrderStatus.complete.first) do |od|
--- a/spec/controllers/orders_controller_spec.rb
+++ b/spec/controllers/orders_controller_spec.rb
@@ -694,7 +694,7 @@ RSpec.describe OrdersController do
         end
 
         context "for a non reservation-only instrument" do
-          let(:instrument) { FactoryGirl.create(:setup_instrument, :timer, facility: @authable, facility_account: @facility_account) }
+          let(:instrument) { FactoryGirl.create(:setup_instrument, :timer, facility: facility, facility_account: @facility_account) }
 
           before { setup_past_reservation }
 
@@ -736,8 +736,6 @@ RSpec.describe OrdersController do
             it "sets the canceled time on the reservation" do
               assigns[:order].order_details.all? { |od| expect(od.reservation.canceled_at).not_to be_nil }
               expect(reservation.reload.canceled_at).not_to be_nil
-              # Should this match the date put in the form, or the date when the action took place
-              # @reservation.canceled_at.should match_date @submitted_date
             end
           end
         end
--- a/spec/controllers/orders_controller_spec.rb
+++ b/spec/controllers/orders_controller_spec.rb
@@ -667,10 +667,7 @@ RSpec.describe OrdersController do
       end
 
       context "backdating a reservation" do
-        before :each do
-          @instrument = FactoryGirl.create(:instrument,
-                                           facility: @authable,
-                                           facility_account: @facility_account)
+        def setup_past_reservation
           @instrument_pp = create :instrument_price_policy, price_group: @price_group, start_date: 7.days.ago, expire_date: 1.day.from_now, product: @instrument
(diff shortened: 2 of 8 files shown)


app/models/order_detail.rb
```diff
@@ -461,11 +461,11 @@ def force_complete!
     update(state: "complete", order_status: OrderStatus.complete_status)
   end
 
-  def backdate_to_complete!(event_time = Time.zone.now)
+  def backdate_to_complete!(event_time = Time.zone.now, set_actuals = true)
```
drench: How do you feel about changing the interface to this method so it could be called as `@order_detail.backdate_to_complete!(@reservation.reserve_end_at, set_actuals: false)` below, to make it a bit more self-documenting?

spec/controllers/orders_controller_spec.rb
```diff
@@ -683,46 +680,70 @@ def deliver; end
...
         end
 
-        it "sets the actual times to the completed reservation times" do
-          do_request
-          expect(@reservation.reload.actual_start_at).to match_date @reservation.reserve_start_at
-          expect(@reservation.actual_end_at).to match_date(@reservation.reserve_start_at + 60.minutes)
-        end
+        context "for a non reservation-only instrument" do
+          before :each do
+            @instrument = FactoryGirl.create(:instrument,
```
jhanggi: Can we make these two contexts use the same base factory, either `:instrument` and `:setup_instrument` and use either `no_relay` on `:instrument` or your new `:not_reservation_only` on `:setup_instrument`?

spec/factories/accessories.rb
```diff
@@ -7,6 +7,20 @@
     after(:create) do |instrument, evaluator|
       instrument.accessories << evaluator.accessory
     end
+
+    trait :not_reservation_only do
```
jhanggi: Because `:setup_instrument` is the parent and you've added the `:not_reservation_only` trait there, I believe you can just call `not_reservation_only` here and it should work.

jhanggi: And can we call this trait something non-negated? Like `timer` or `timer_without_relay`?

spec/controllers/orders_controller_spec.rb
```diff
@@ -683,46 +680,70 @@ def deliver; end
...
+          end
 
-        it "assigns a price policy and cost" do
-          do_request
-          expect(@order_detail.reload.price_policy).not_to be_nil
-          expect(@order_detail.actual_cost).not_to be_nil
-        end
+          it "sets the order_detail states to completed because it's in the past" do
+            do_request
+            assigns[:order].order_details.all? { |od| expect(od.state).to eq("complete") }
```
drench: Could this be something like `expect(assigns[:order].order_details).to all(be_complete)` ?

spec/controllers/orders_controller_spec.rb
```diff
@@ -683,46 +680,70 @@ def deliver; end
...
 
-        it "sets the actual times to the completed reservation times" do
-          do_request
-          expect(@reservation.reload.actual_start_at).to match_date @reservation.reserve_start_at
-          expect(@reservation.actual_end_at).to match_date(@reservation.reserve_start_at + 60.minutes)
-        end
+        context "for a non reservation-only instrument" do
+          before :each do
+            @instrument = FactoryGirl.create(:instrument,
+                                             facility: @authable,
```
drench: I'd use the `let()`-assigned values wherever possible.

spec/controllers/orders_controller_spec.rb
```diff
@@ -683,46 +680,70 @@ def deliver; end
...
-            # Should this match the date put in the form, or the date when the action took place
-            # @reservation.canceled_at.should match_date @submitted_date
+          context "canceled" do
+            before :each do
+              @params[:order_status_id] = OrderStatus.canceled.first.id
+              do_request
+            end
+
+            it "is able to set order_detail states to canceled" do
+              assigns[:order].order_details.all? { |od| expect(od.state).to eq("canceled") }
```
drench: As above, would this work? `expect(assigns[:order].order_details).to all(be_canceled)`

spec/controllers/orders_controller_spec.rb
kt32291: This whole file uses @ variables, and I'm try to set the instrument and everything for that request. Unless I change the whole file, I don't know if this change is worth it right now (though I agree it should be done).

kt32291:
@drench see my comment about switching to let variables

spec/controllers/orders_controller_spec.rb
```diff
@@ -683,46 +680,71 @@ def deliver; end
           @submitted_date = 2.days.ago.change(hour: 14, min: 27)
         end
 
-        it "sets the order_detail states to completed because it's in the past" do
-          do_request
-          assigns[:order].order_details.all? { |od| expect(od.state).to eq("complete") }
-        end
+        context "for a reservation-only instrument" do
+          before :each do
+            @instrument = FactoryGirl.create(:setup_instrument,
```
jhanggi: We've started the transitioning this file, so it would be nice to switch what we can (without going too overboard, of course). You'll see at the top of the file we have things like (simplified):

```
let(:facility) { create(:facility) }

before { @authable = facility }
```

I think it would be nice to at least use the `facility` from `let` here rather than `@authable`.

It might also be worth considering trying to switch to an `instrument` let (might involve a little changing to `setup_past_reservation`). 

On the other hand, this file is a mess, and is probably more effort than it's worth to clean up right now.

kt32291:
@jhanggi Did some minor refactors, but didn't get too crazy. Let me know if this is good!

spec/controllers/orders_controller_spec.rb
```diff
@@ -667,62 +667,78 @@ def deliver; end
...
           end
         end
 
-        it "sets the actual times to the completed reservation times" do
-          do_request
-          expect(@reservation.reload.actual_start_at).to match_date @reservation.reserve_start_at
-          expect(@reservation.actual_end_at).to match_date(@reservation.reserve_start_at + 60.minutes)
-        end
+        context "for a non reservation-only instrument" do
+          let(:instrument) { FactoryGirl.create(:setup_instrument, :timer, facility: @authable, facility_account: @facility_account) }
```
jhanggi: It's not a big deal, but we have a `facility` `let`, so we should probably use that over `@authable`. Not sure if it's worth creating a let for `@facility_account`.

spec/controllers/orders_controller_spec.rb
```diff
@@ -667,62 +667,78 @@ def deliver; end
...
+            end
+
+            it "is able to set order_detail states to canceled" do
+              expect(assigns[:order].order_details).to all(be_canceled)
+            end
+
+            it "sets the canceled time on the reservation" do
+              assigns[:order].order_details.all? { |od| expect(od.reservation.canceled_at).not_to be_nil }
+              expect(reservation.reload.canceled_at).not_to be_nil
+              # Should this match the date put in the form, or the date when the action took place
```
jhanggi: I see this was already there, but we should either fix this spec or remove the comment.

kt32291:
@jhanggi updated!

PR #824, opened 2016-11-28
[#133457] Adds ability to search by statement id

kt32291:
https://pm.tablexi.com/issues/133457

This adds the ability to only enter the statement id and use the global search to find statements that way. If the search query does not have a hyphen, it looks by only id.


jhanggi:
👍 Looks good.

PR #825, opened 2016-11-29
[#125724] Track ordered_by on the order detail level

kt32291:
https://pm.tablexi.com/issues/125724

This sets the order detail ordered_by accurately for orders that are added to, using the current user instead of just defaulting to the user who placed the original order.


jhanggi:
Feel free to squash & merge, @kt32291.

PR #827, opened 2016-11-30
[#133349] Add cross-account transactions-in-review view

drench:
This adds a new view showing transactions-in-review for all accounts the current user manages (that is, is an Owner or Business Administrator of). Its path is `/transactions` (a route that was used in the past for transaction history) though there are no links to it yet.

The controller and view are very similar to the single-account transactions-in-review code. It uses the same standard `TransactionSearch`, with the same limitations: the search only applies to the list of "in-review" orders, not the "Recently Reviewed" list. I'm not sure how to work around it without either altering `TransactionSearch` or moving controller logic into the view.

This also includes a refactor of `it_should_support_searching` into rspec shared examples. The existing specs that use `it_should_support_searching` are still in the old controller spec helper style so I left the method, but it now wraps the shared examples.


app/controllers/transactions_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def index_with_search
+    @recently_reviewed = administered_order_details.recently_reviewed.paginate(page: params[:page])
+    @order_details = administered_order_details.all_in_review
+    @extra_date_column = :reviewed_at
+    @order_detail_link = {
+      text: text("shared.dispute"),
+      display?: proc { |order_detail| order_detail.can_dispute? },
+      proc: proc { |order_detail| order_order_detail_path(order_detail.order, order_detail) },
```
jhanggi: This is kind of a bad name. Could we come up with something better? (I realize this is copied out of another controller). What if we made this hash a presenter/decorator? Worth doing?

app/views/transactions/index.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  = text("transactions.in_review")
+
+#header_billing.grid_9= render "shared/transactions/search"
+
+.clearfix
+
+- if @order_details.any?
+  .grid_12= render "shared/transactions/table", order_details: @order_details
+
+- if @recently_reviewed.any?
```
jhanggi: Do we want to do anything if both `@order_details` and `@recently_reviewed` are both empty?

app/views/transactions/index.html.haml
```diff
@@ -0,0 +1,21 @@
+-# TODO: app/views/accounts/transactions_in_review.html.haml is similar
+
+= render "shared/transactions/headers"
+
+- content_for :h1 do
+  = text("transactions.in_review")
+
+#header_billing.grid_9= render "shared/transactions/search"
```
jhanggi: I think we may want to hide this search form if there are no `@order_details`. When that's empty, you can type in all the search fields, but there are no options.

app/controllers/transactions_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+  before_action :check_acting_as
+
+  include TransactionSearch
+
+  def initialize
+    @active_tab = "accounts"
+    super
+  end
+
+  def index_with_search
```
jhanggi: I wonder if we should call this `/transactions/in_review` and reserve `/transactions` for the cross-account "my transactions" view we were talking about with NU today. Maybe not worth doing unless we actually do that feature.

drench: My first thought was to consider changing the URL if and when we decide to have separate all-transactions and in-review views. But considering this the the URL we're going to be emailing, it would make sense to keep it stable. Let's go with `/transactions/in_review` now.

app/controllers/transactions_controller.rb
drench: I took some time on this, and it's leading to a bigger refactor than I expected, mostly due to the views involved. I think it's worth doing, but I'd like to put it in a separate PR.

.rubocop-txi.yml
```diff
@@ -28,6 +28,13 @@ Metrics/LineLength:
 Metrics/MethodLength:
   Max: 20
 
+Rails/HttpPositionalArguments:
+  Exclude:
```
jhanggi: Would it be better to disable this cop altogether?

app/controllers/accounts_controller.rb
```diff
@@ -61,4 +62,8 @@ def ability_resource
     @account
   end
 
+  def administered_order_details
+    @administered_order_details ||= current_user.administered_order_details
```
jhanggi: Does this get used more than once? 1) I'm wondering if it's necessary to extract out into its own method, and 2) since I believe it's returning an ActiveRecord::Relation, is the memoization necessary?

app/controllers/transactions_controller.rb
jhanggi: I'm cool pushing this off.

app/controllers/transactions_controller.rb
```diff
@@ -0,0 +1,31 @@
...
+      text: text("shared.dispute"),
+      display?: proc { |order_detail| order_detail.can_dispute? },
+      proc: proc { |order_detail| order_order_detail_path(order_detail.order, order_detail) },
+    }
+  end
+
+  private
+
+  def administered_order_details
+    @administered_order_details ||= current_user.administered_order_details
```
jhanggi: Same question here, especially around the memoization.

.rubocop-txi.yml
drench: Maybe, though it would be weird to run into this outside of controller specs, wouldn't it?

PR #829, opened 2016-12-01
[#132637] Fix behavior of cancel on merge order reservation

jhanggi:
When adding a reservation to an existing order (i.e. by facility
staff), the “Cancel” button was not returning you to the expected
location and was instead causing an error.

`orders#remove (RuntimeError) "Can't modify frozen hash”`

### DIFF
--- a/app/views/reservations/new.html.haml
+++ b/app/views/reservations/new.html.haml
@@ -55,7 +55,9 @@
 
     - url_after_action = facility_path(@instrument.facility)
     %li
-    - if @order_detail.bundled?
+    - if @order.to_be_merged?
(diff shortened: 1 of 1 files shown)


kt32291:
🎉 but I do agree with @drench. If the only part that varies is the path, maybe extract a method to return the path.

jhanggi:
Pulling it into a helper method is not a bad idea, but I've been starting to lean towards fewer single-use view helper methods because I'm not convinced the extra indirection is worth the cleanliness of the view code (if we were also using this logic elsewhere, I would likely pull it up for the sake of DRY).

PR #830, opened 2016-12-05
Gem updates

drench:
There are more that still could use updating, but this bumps a few. My particular concern was making sure we keep up with aasm.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -89,7 +89,7 @@ group :development, :test do
   gem "factory_girl_rails", "~> 4.5.0"
   gem "guard-rspec", require: false
   gem "guard-teaspoon", require: false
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -89,10 +89,10 @@ group :development, :test do
   gem "factory_girl_rails", "~> 4.5.0"
   gem "guard-rspec", require: false
   gem "guard-teaspoon", require: false
-  gem "pry-rails",         "~> 0.3.2"
-  gem "pry-byebug",        "~> 2.0.0"
+  gem "pry-rails", "~> 0.3.4"
```
jhanggi: Are the pry gems even worth locking at all?

drench: Probably not.

PR #831, opened 2016-12-05
[#133351] Send one notification per account admin

drench:
~~~This includes #827; I'll rebase once that's merged.~~~

Rather than sending one notification per account, this sends one notification per user, noting which accounts the notification is about, and links the user to `/transactions/in_review`.

The notification sender spec got a bit of an overhaul and this adds a new mailer preview for `Notifier.review_orders`.

### DIFF
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -50,7 +50,6 @@ class Notifier < ActionMailer::Base
 
   def review_orders(args)
     @user = User.find(args[:user_id])
-    @facility = Facility.find(args[:facility_id])
     @accounts = Account.where(id: args[:account_ids]).pluck(:description).map { |s| %("#{s}") }.to_sentence
     send_nucore_mail @user.email, text("views.notifier.review_orders.subject")
   end
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -51,7 +51,7 @@ class Notifier < ActionMailer::Base
   def review_orders(args)
     @user = User.find(args[:user_id])
     @facility = Facility.find(args[:facility_id])
-    @account = Account.find(args[:account_id])
+    @accounts = Account.where(id: args[:account_ids]).pluck(:description).map { |s| %("#{s}") }.to_sentence
     send_nucore_mail @user.email, text("views.notifier.review_orders.subject")
   end
 
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -50,7 +50,7 @@ class Notifier < ActionMailer::Base
 
   def review_orders(args)
     @user = User.find(args[:user_id])
-    @accounts = Account.where(id: args[:account_ids]).pluck(:description).map { |s| %("#{s}") }.to_sentence
+    @accounts = Account.find(args[:account_ids]).map(&:account_list_item).to_sentence
     send_nucore_mail @user.email, text("views.notifier.review_orders.subject")
   end
 
--- a/app/models/account_user.rb
+++ b/app/models/account_user.rb
@@ -7,6 +7,10 @@ class AccountUser < ActiveRecord::Base
   ACCOUNT_OWNER = "Owner".freeze
   ACCOUNT_ADMINISTRATOR = "Business Administrator".freeze
 
+  scope :administrators, lambda {
+    User.where(id: where(user_role: admin_user_roles).select(:user_id))
+  }
+
   def self.read_only_user_roles
     [ACCOUNT_PURCHASER]
   end
--- a/app/models/account_user.rb
+++ b/app/models/account_user.rb
@@ -7,16 +7,16 @@ class AccountUser < ActiveRecord::Base
   ACCOUNT_OWNER = "Owner".freeze
   ACCOUNT_ADMINISTRATOR = "Business Administrator".freeze
 
-  scope :administrators, lambda {
-    User.where(id: where(user_role: admin_user_roles).select(:user_id))
-  }
-
   def self.read_only_user_roles
     [ACCOUNT_PURCHASER]
   end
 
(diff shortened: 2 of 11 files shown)


app/support/notification_sender.rb
```diff
@@ -74,10 +74,18 @@ def reviewed_at
   class AccountNotifier
 
     def notify_accounts(accounts_to_notify)
+      notifications = {}
       accounts_to_notify.each do |account_id, facility_id|
-        account = Account.find(account_id)
-        account.notify_users.each do |u|
-          Notifier.review_orders(user_id: u.id, facility_id: facility_id, account_id: account_id).deliver_now
+        user_ids = AccountUser.where(account_id: account_id, user_role: ["Business Administrator", "Owner"]).pluck(:user_id)
```
jhanggi: Can we use `AccountUser.admin_user_roles` here? Should we have another scope?

app/support/notification_sender.rb
```diff
@@ -74,10 +74,18 @@ def reviewed_at
...
+        user_ids.each do |user_id|
+          notifications[user_id] ||= { facility_id => [] }
+          notifications[user_id][facility_id] << account_id
+        end
+      end
+
+      notifications.each do |user_id, facility_ids|
```
jhanggi: This method is a whole lot more complicated than the old version. Is there something we can do further up the stack to clean this up? Is there a clearer way to write this? What if we sent one email per user instead of one email per user per facility? The new view is cross-facility as well, correct?

app/views/notifier/review_orders.text.erb
```diff
@@ -1,3 +1,3 @@
-<%= text(".body", facility: @facility.to_s, account: @account.to_s) %>
+<%= text(".body", facility: @facility.to_s, accounts: @accounts.html_safe) %>
```
jhanggi: Why is `@accounts` unsafe at this point?

drench: I was seeing entity encoding here, which we don't want in plain text. That said, I'm thinking I should have wrapped it in `raw` instead. Updating.

app/support/notification_sender.rb
drench: This feels like a scope that should have already existed. I'm adding it.

app/views/notifier/review_orders.text.erb
jhanggi: What's getting encoded? Is there html inside `@accounts`? That doesn't feel right. Are they `join`ed?

app/models/account_user.rb
```diff
@@ -7,6 +7,10 @@ class AccountUser < ActiveRecord::Base
   ACCOUNT_OWNER = "Owner".freeze
   ACCOUNT_ADMINISTRATOR = "Business Administrator".freeze
 
+  scope :administrators, lambda {
```
jhanggi: This feels a little weird to call it a `scope` since it appears to return `User`s instead of `AccountUser`. Maybe just a class method?

app/support/notification_sender.rb
```diff
@@ -7,8 +7,17 @@ def initialize(current_facility, order_detail_ids)
     @order_detail_ids = order_detail_ids
   end
 
+  def init_account_ids_to_notify
+    @account_ids_to_notify =
```
jhanggi: Would this be cleaned up at all if we turned `account_ids_to_notify` into a memoized method rather than an `attr_reader` with an initialization?

app/support/notification_sender.rb
```diff
@@ -65,29 +62,33 @@ def order_details
...
-          Notifier.review_orders(user_id: u.id, facility_id: facility_id, account_id: account_id).deliver_now
+    def notify_accounts(account_ids_to_notify)
+      notifications_hash(account_ids_to_notify).each do |user_id, account_ids|
+        Notifier.review_orders(user_id: user_id, account_ids: account_ids).deliver_now
+      end
+    end
+
+    private
+
+    def notifications_hash(account_ids_to_notify)
```
jhanggi: This method name is a little vague. Not sure what better to call it, though. Maybe add a comment about the structure it's expected to return?

app/support/notification_sender.rb
drench: I tried this earlier, but it's complicated. The tricky part is that if we wait to set `account_ids_to_notify` when it's used, that happens *after* the order details in question are marked as reviewed, meaning we'll get an empty set.

jhanggi: Gotcha. Makes sense.

app/mailers/notifier.rb
```diff
@@ -50,8 +50,7 @@ def order_receipt(args)
 
   def review_orders(args)
     @user = User.find(args[:user_id])
-    @facility = Facility.find(args[:facility_id])
-    @account = Account.find(args[:account_id])
+    @accounts = Account.where(id: args[:account_ids]).pluck(:description).map { |s| %("#{s}") }.to_sentence
```
jhanggi: Or even just `to_s`?

app/views/notifier/review_orders.text.erb
jhanggi: What about the `smart: false` option? Would that avoid the conversion "smart" quotes and apostrophes?

app/support/notification_sender.rb
```diff
@@ -17,24 +26,19 @@ def perform
...
+    account_ids_to_notify.count
   end
 
   def accounts_notified
-    Account.find(account_ids_notified)
-  end
-
-  def accounts_notified_size
-    account_ids_to_notify.count
+    Account.where(id: account_ids_to_notify)
```
jhanggi: I don't think we'll hit it, but should we protect from Oracle's 1000+ restriction?

PR #832, opened 2016-12-07
[#125660] Reservation user action updates

drench:
This makes two changes to reservation user actions:

* If the `session_user` is the user that made the reservation (or the reservation was made on this user's behalf), then a "Begin Now" link can appear. Mainly we want this so administrators are not able to "Begin Now" another user's reservation from the `/facilities/:id/users/:id/reservations` screen.

* When an administrator cancels a reservation from this same screen, it should impose the same cancellation fee that would apply if the user canceled it directly. Afterward, it should redirect the user back to the screen they were on (it was sending operators to `/reservations/upcoming`).


app/controllers/order_details_controller.rb
```diff
@@ -72,6 +72,27 @@ def dispute
...
+      facility_user_reservations_path(order_detail.facility, order_detail.user)
+    else
+      reservations_path
+    end
+  end
+
+  def cancel_with_fee(order_detail)
+    if operator_cancellation?(order_detail)
+      # When this is an operator who is not the ordering user; admin cancel with fee:
+      order_detail.cancel_reservation(session_user, OrderStatus.canceled_status, true, true)
```
kt32291: Maybe switch to using params here? The boolean values you're passing in are a bit confusing,  so doing ```admin_cancellation: true, admin_with_cancel_fee: true``` instead.

drench: Agreed it's not a good interface. I didn't want to change it since the method is used elsewhere, but now that I'm looking for usage the cleanup would be less than I thought. I'll update.

app/controllers/order_details_controller.rb
```diff
@@ -72,6 +72,27 @@ def dispute
...
+    else
+      order_detail.cancel_reservation(session_user)
+    end
+  end
+
+  def operator_cancellation?(order_detail)
+    order_detail.user != session_user && current_ability.can?(:cancel, order_detail)
```
jhanggi: This is probably an edge case, but if I'm an operator who has placed an order for myself, I think it will always bring me back to the operator view, even if I cancel from the user view.

app/presenters/reservation_user_action_presenter.rb
```diff
@@ -23,7 +23,7 @@ def user_actions
 
     if can_switch_instrument?
```
jhanggi: Does this ever return true for the operator? Also, it's hidden here, but are we hiding the accessories link as well from the operator?

drench: Since `Reservation#can_switch_instrument_on?` doesn't consider `current_user`, it could be true for operators. Same goes for `accessories?`. Stepping back, it's looking more like all the sharing between the user and admin interfaces made things more complicated.

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  def index
+    @order_details = user_order_details
+                     .where("orders.ordered_at IS NOT NULL")
+                     .order("orders.ordered_at DESC")
+                     .paginate(page: params[:page])
+  end
+
+  # PUT /facilities/:facility_id/users/:user_id/reservations/:reservation_id/cancel
+  def cancel
+    order_detail = user_order_details.find(params[:reservation_id])
```
drench: I hope it isn't too confusing to call this `reservation_id` when it's really an `order_detail_id`.

drench:
This has the update we discussed, where the user-facing and admin-facing reservation listings are no longer sharing views or controller methods (they used to share the `cancel` controller action and reservation listing view).

There are changes I made to the admin-facing view and order detail presenter that could apply to the user-facing code but I've left the non-admin pieces mostly as they were for now.

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+  before_action :load_user
+
+  load_and_authorize_resource class: Reservation
+
+  layout "two_column"
+
+  # GET /facilities/:facility_id/users/:user_id/reservations
+  def index
+    @order_details = user_order_details
+                     .where("orders.ordered_at IS NOT NULL")
```
jhanggi: I believe we have a scope for this. `purchased` (which is an alias of `ordered`, but I think we concluded that that got confusing against the idea of sorting).

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+                     .order("orders.ordered_at DESC")
```
jhanggi: We also have a `by_ordered_at` scope.

app/controllers/facility_user_reservations_controller.rb
jhanggi: I think it is confusing to call it `reservation_id` when it's actually an `order_detail_id`. What's the advantage of calling it `reservation_id`?

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+                     .paginate(page: params[:page])
+  end
+
+  # PUT /facilities/:facility_id/users/:user_id/reservations/:reservation_id/cancel
+  def cancel
+    order_detail = user_order_details.find(params[:reservation_id])
+    raise ActiveRecord::RecordNotFound unless order_detail.reservation.can_cancel?
```
jhanggi: Should we have a `try` or something in case there is no reservation? (which would really only happen in weird edge cases or someone mucking with the URLs)

app/models/order_detail.rb
```diff
@@ -741,11 +744,11 @@ def disputed?
     dispute_at.present? && !canceled?
   end
 
-  def cancel_reservation(canceled_by, order_status = OrderStatus.canceled.first, admin_cancellation = false, admin_with_cancel_fee = false)
+  def cancel_reservation(canceled_by, order_status: OrderStatus.canceled_status, admin: false, admin_with_cancel_fee: false)
```
jhanggi: I like this API change.

app/presenters/order_detail_presenter.rb
```diff
@@ -21,6 +21,10 @@ def description_as_html
     end.join(" &mdash; ").html_safe
   end
 
+  def description_as_html_with_facility_prefix
+    "#{facility.abbreviation} / #{description_as_html}".html_safe
```
jhanggi: Does this really need to be `html_safe`d? Or is it something that goes through the markdown smart-quote filter that we could turn off?

app/services/cancellation_fee_calculator.rb
```diff
@@ -0,0 +1,28 @@
...
+  private
+
+  def calculated_fee
+    if @calculated_fee.blank?
+
+      # This is wrapped in a transaction and rolled back because
+      # #cancellation_fee can reassign price policies and costs.
+      @reservation.class.transaction do
+        @reservation.canceled_at = Time.current
+        @calculated_fee = @reservation.order_detail.cancellation_fee
```
jhanggi: I see that `cancellation_fee` will try to assign a price policy (and possibly pricing) if there isn't already one, but it doesn't look like it would actually hit the database.

app/presenters/order_detail_presenter.rb
```diff
@@ -49,6 +53,12 @@ def survey_url
     survey_completed? ? external_service_receiver.show_url : ""
   end
 
+  def wrapped_total
```
jhanggi: Why do we need to do this wrapping here? I get that the object might need the `PriceDisplayment`, but why does it need a `reload`?

app/views/facility_user_reservations/index.html.haml
```diff
@@ -0,0 +1,66 @@
...
+        %th.centered= OrderDetail.human_attribute_name(:id)
+        %th.centered= text("users.reservations.actions")
+        %th= text("users.reservations.timespan")
+        %th= OrderDetail.human_attribute_name(:product)
+        %th.centered= OrderDetail.human_attribute_name(:status)
+        %th.currency= OrderDetail.human_attribute_name(:actual_total)
+
+    %tbody
+      - OrderDetailPresenter.wrap(@order_details).each do |order_detail|
+        - reservation_presenter = ReservationUserActionPresenter.new(controller, order_detail.reservation)
```
jhanggi: I looks like we're only using the presenter for the edit link. Would it make sense to remove the use of the presenter and just use the path helper? Or really, do we need to make it link to the view/edit link on this admin view, and have them use the order detail popup for modifying times.

I think if we get rid of the use of the presenter here, we can clean up the additional conditionals in the presenter that check if it needs to be an admin link or not.

app/views/facility_user_reservations/index.html.haml
```diff
@@ -0,0 +1,66 @@
...
+          %td.centered= link_to order_detail, order_detail.show_order_path
+
+          %td.centered
+            - if order_detail.reservation.try(:can_cancel?)
+              - fee = CancellationFeeCalculator.new(order_detail.reservation).fee
+              - confirm_text = fee > 0 ? text("reservations.delete.confirm_with_fee", fee: number_to_currency(fee)) : text("reservations.delete.confirm")
+
+              = link_to text("reservations.delete.link"),
+                facility_user_reservation_cancel_path(current_facility, @user, order_detail),
+                method: :put,
```
jhanggi: Would this not be a `delete`?

drench: I see the `reservation` as getting updated when canceled. More of a state change than a removal.

app/presenters/order_detail_presenter.rb
drench: We'd have to go back and rethink `description_as_html`, which does its own escaping. Without the `html_safe` here, the `description_as_html` part gets displayed double-encoded when rendered.

app/presenters/order_detail_presenter.rb
drench: I traced the `reload` back to 957cbce10f78205802c76c0131c9da84a4d49528.
The `reload` remained when it became`od.reload.wrapped_cost` when introducing `PriceDisplayment` in 0c845be58e61d7c5eac79165131d9397c1b883e6, and in 95e07461bb709de591f6fd6cbb6896ce11b923eb when the method name was renamed to `#wrapped_total`.

It doesn't seem to need the `reload` in testing but maybe it did at one point? I'm removing it.

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+                     .purchased
+                     .by_ordered_at
+                     .paginate(page: params[:page])
+  end
+
+  # PUT /facilities/:facility_id/users/:user_id/reservations/:id/cancel
+  def cancel
+    order_detail = user_order_details.find(params[:id])
+    raise ActiveRecord::RecordNotFound unless order_detail.reservation.try(:can_cancel?)
+    order_detail.transaction do
```
jhanggi: I see this is coming from `OrderDetailsController`, but is the transaction necessary? Seems like if it failed, then nothing should have hit the database.

drench: I don't know what scenario could lead to it, but what if there's no cancellation fee so it hits `clear_statement` and then `save!` fails?

drench:
@kt32291 any feedback after the updates?

kt32291:
👍

PR #819, opened 2016-11-21
[#132887] Fix links on shared schedules for admin reservations

kt32291:
https://pm.tablexi.com/issues/132887

This fixes the link for reservations on shared schedules by using the actual instrument the reservation is attached to for the link instead of the instrument page you are on.

PR #834, opened 2016-12-08
[#132887] Fix the schedule links so they still stay with instrument

kt32291:
https://pm.tablexi.com/issues/132856

This links the edit reservation to still have the path of the instrument you are looking at if it is a shared schedule, and then returns to the path you were at if you edit or cancel instead of the base instrument for the reservation.

This is a little non-rails-y but the behavior they want is a bit weird to me, so this was the best way in my opinion.

### DIFF
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -141,7 +141,7 @@ class FacilityReservationsController < ApplicationController
   # GET /facilities/:facility_id/instruments/:instrument_id/reservations/:id/edit_admin
   def edit_admin
     @instrument  = current_facility.instruments.find_by_url_name!(params[:instrument_id])
-    @reservation = @instrument.reservations.find(params[:reservation_id])
+    @reservation = Reservation.where(product_id: @instrument.schedule.products.map(&:id)).find(params[:reservation_id])
     raise ActiveRecord::RecordNotFound unless @reservation.order_detail_id.nil?
     set_windows
     render layout: "two_column"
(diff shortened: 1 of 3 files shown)


kt32291:
The code climate issue was an existing one.

app/controllers/facility_reservations_controller.rb
```diff
@@ -141,7 +141,7 @@ def create
   # GET /facilities/:facility_id/instruments/:instrument_id/reservations/:id/edit_admin
   def edit_admin
     @instrument  = current_facility.instruments.find_by_url_name!(params[:instrument_id])
-    @reservation = @instrument.reservations.find(params[:reservation_id])
+    @reservation = Reservation.where(product_id: @instrument.schedule.products.map(&:id)).find(params[:reservation_id])
```
jhanggi: I think this can be written more simply as `@instrument.schedule.reservations.find`. That's what `InstrumentController#schedule` uses. We might also want to add the `non_user` scope here to prevent modifying actual orders.

jhanggi: It might also be worth extracting this into a private method since it's used in several places in this controller.

app/views/facility_reservations/edit_admin.html.haml
```diff
@@ -18,7 +18,7 @@
 = content_for :tabnav do
   = render :partial => 'admin/shared/tabnav_product', :locals => {:secondary_tab => 'reservations'}
 
-%h2= @instrument
+%h2= @reservation.product
```
jhanggi: This feels a little strange to me. Let's say the admin reservation was originally created on Instrument #1, and I'm on the schedule page for Instrument #2, and click edit. This is going to show me Instrument #1 and then when I submit it takes me back to #2.

jhanggi:
We decided with the client that we are instead going to only display the offline and admin reservations on the instrument they were created on. There's probably also a little in https://github.com/tablexi/nucore-open/pull/819 that can be undone.

PR #835, opened 2016-12-08
[#126780] Fixes scope to return overlap of accounts

kt32291:
https://pm.tablexi.com/issues/126780

This changes the scope so it only returns accounts that both the acting as user and the order detail user have access to.

### DIFF
--- a/app/services/available_accounts_finder.rb
+++ b/app/services/available_accounts_finder.rb
@@ -1,18 +1,27 @@
 class AvailableAccountsFinder
 
-  def initialize(user, facility, current: nil)
+  def initialize(user, facility, current: nil, current_user: nil)
     @user = user
+    @current_user = current_user
     @facility = facility
     @current_account = current
   end
 
   def accounts
-    accounts = @user.accounts.for_facility(@facility).active
+    accounts = available_accounts
     if @current_account && !accounts.include?(@current_account)
       accounts += [@current_account]
     end
     accounts
   end
   alias to_a accounts
+
+  private
+
+  def available_accounts
+    available_accounts = @user.accounts.for_facility(@facility).active
+    available_accounts = available_accounts & @current_user.accounts.for_facility(@facility).active if @current_user
+    available_accounts
+  end
 
 end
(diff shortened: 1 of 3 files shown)


app/services/available_accounts_finder.rb
```diff
@@ -1,18 +1,27 @@
...
     accounts
   end
   alias to_a accounts
 
+  private
+
+  def available_accounts
+    available_accounts = @user.accounts.for_facility(@facility).active
+    available_accounts &= @current_user.accounts.for_facility(@facility).active if @current_user
+    available_accounts
```
drench: At this point, I'd look at moving the code that the adds `@current_account` here as well, since it could be written as a single line. It would leave nothing in `#accounts`… meaning you could rename this method to `#accounts` and have no `private` method at all.

spec/services/available_accounts_finder_spec.rb
```diff
@@ -30,6 +30,19 @@
...
+    context "with a different current user" do
+      subject(:accounts) { described_class.new(user, facility, current: current, current_user: current_user).accounts }
+
+      let(:current_user) { FactoryGirl.create(:user) }
+      let!(:other_chartstring) { FactoryGirl.create(:nufs_account, :with_account_owner, owner: current_user) }
+
+      before { FactoryGirl.create(:account_user, :business_administrator, user: user, account: other_chartstring) }
+
+      describe "for any facility" do
+        it { is_expected.to match_array([chartstring, other_chartstring]) }
```
drench: A minor style thing to consider (which I often forget): this could be written as `is_expected.to contain_exactly(chartstring, other_chartstring)` to avoid the `[]`.

app/services/available_accounts_finder.rb
jhanggi: Another option would be something like `Account.for_facility(@facility).for_user(@user, @current_user).active` and refactor `for_user` to support multiple users. nbd, though

jhanggi: I don't think we're doing any sorting, but that would allow us to do it in SQL if we wanted.

kt32291:
@jhanggi @drench updated

PR #836, opened 2016-12-12
[#121328] Allow editing of started reservation in the grace period

kt32291:
https://pm.tablexi.com/issues/121328

This allows you to edit the duration of a started reservation while you are within the grade period (which is automatically added to the reserve_start_at when you Create & Start a reservation right away.

### DIFF
--- a/spec/models/reservation_spec.rb
+++ b/spec/models/reservation_spec.rb
@@ -415,15 +415,16 @@ RSpec.describe Reservation do
             end
 
             context "the instrument has a reservation lock window" do
+              let(:current_time) { Time.zone.now }
+
               before :each do
-                @current_time = Time.zone.now
                 instrument.update_attribute(:lock_window, 12)
                 reservation.reload
               end
 
               context "within the grace period" do
                 before do
-                  reservation.assign_attributes(reserve_start_at: @current_time + 5.minutes, actual_start_at: @current_time)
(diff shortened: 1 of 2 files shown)


spec/models/reservation_spec.rb
```diff
@@ -413,6 +413,23 @@
             context "there is no following reservation" do
               it_behaves_like "a customer is allowed to edit"
             end
+
+            context "the instrument has a reservation lock window" do
+              before :each do
+                @current_time = Time.zone.now
```
jhanggi: This is kind of picky, but I think we should put this in a `let` block instead of an instance variable.

kt32291:
@jhanggi updated

spec/models/reservation_spec.rb
```diff
@@ -413,6 +413,24 @@
...
+
+              before :each do
+                instrument.update_attribute(:lock_window, 12)
+                reservation.reload
+              end
+
+              context "within the grace period" do
+                before do
+                  reservation.assign_attributes(reserve_start_at: current_time + 5.minutes, actual_start_at: current_time)
+                  reservation.save(validate: false)
```
drench: This pattern exists elsewhere, but I'm wondering if we could get away with `reservation.update_columns` on line 427 and make the non-validating `#save` unnecessary.

kt32291:
@drench updated

PR #837, opened 2016-12-13
[#120321] Update reservation auto-cancel minutes field messaging

drench:
This updates the field "hint" text and adds a warning message when the field is zero or blank, explaining what that means.

It looks like this:
auto-cancel-warning

I fought with `simple_form` for a bit to try to get the warning to appear just to the right of the field instead but did not arrive there.

### DIFF
--- a/app/assets/javascripts/app/manage_instruments.coffee
+++ b/app/assets/javascripts/app/manage_instruments.coffee
@@ -10,7 +10,7 @@ $ ->
       else
         $(this).addClass 'interval-error'
 
-  $('#instrument_auto_cancel_mins').change (event)->
+  $('#instrument_auto_cancel_mins').change (event) ->
     $warning_node = $(".js--auto_cancel_mins-zero-warning")
     minutes = $(event.target).val() || 0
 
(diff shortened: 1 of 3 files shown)


app/views/instruments/_instrument_fields.html.haml
```diff
@@ -1,20 +1,27 @@
...
   = f.input :schedule, as: :readonly, value_method: :display_name, hint: false
 
--# - selection = @instrument.control_mechanism || :manual
-= f.input :control_mechanism, collection: options_for_control_mechanism, include_blank: false, selected: @instrument.control_mechanism || :manual
+= f.input :control_mechanism,
+  collection: options_for_control_mechanism,
+  include_blank: false,
+  selected: @instrument.control_mechanism || :manual
 
 - # TODO
```
jhanggi: Any idea what this `TODO` is here for?

app/assets/javascripts/app/manage_instruments.coffee
```diff
@@ -10,3 +10,13 @@ $ ->
       else
         $(this).addClass 'interval-error'
 
+  $('#instrument_auto_cancel_mins').change (event)->
```
jhanggi: Arrow spacing (from code climate).

PR #838, opened 2016-12-14
[#133417] Grant Account Managers un/suspend Account in cross-facility context

drench:
Account Managers previously had `:manage` rights to `Account` (covering all actions) with exceptions for `:suspend` and `:unsuspend`. This removes those exceptions.

### DIFF
--- a/lib/ability.rb
+++ b/lib/ability.rb
@@ -52,8 +52,6 @@ class Ability
         can :manage, [Account, AccountUser, User]
         cannot :switch_to, User
       end
-
-      cannot [:suspend, :unsuspend], Account
     end
 
     if user.billing_administrator?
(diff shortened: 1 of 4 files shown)


lib/ability.rb
```diff
@@ -29,6 +29,7 @@ def initialize(user, resource, controller = nil)
       end
 
       cannot(:switch_to, User) { |target_user| !target_user.active? }
+      ability_extender.extend(user, resource)
```
drench: This doesn't strictly need to be here to make this feature work, but I feel like it should have it. The fact that ability extensions weren't applied in all scenarios confused me until I remembered the early `return` for global admins.

jhanggi: Sounds good.

PR #839, opened 2016-12-15
[#133594] Implemented reservation cutoff time

kt32291:
https://pm.tablexi.com/issues/133594

This allows users to set a cutoff time on an instrument that prevents users from making reservations that are less than that number of hours before the reservation (so the staff has time to prepare). You can still cancel the reservation.

### DIFF
--- a/app/support/products/scheduling_support.rb
+++ b/app/support/products/scheduling_support.rb
@@ -155,7 +155,8 @@ module Products::SchedulingSupport
     attr_accessor :time, :rule, :day_start, :day_end, :options
 
     def initialize(time, rule, options = {})
-      self.time = time
+      first_available_time = Time.zone.now + rule.instrument.cutoff_hours.hours
+      self.time = (time < first_available_time) ? first_available_time : time
       self.rule = rule
       self.day_start = Time.zone.local(time.year, time.month, time.day, rule.start_hour, rule.start_min, 0)
       self.day_end   = Time.zone.local(time.year, time.month, time.day, rule.end_hour, rule.end_min, 0)
--- a/app/support/products/scheduling_support.rb
+++ b/app/support/products/scheduling_support.rb
@@ -155,17 +155,12 @@ module Products::SchedulingSupport
     attr_accessor :time, :rule, :day_start, :day_end, :options
 
     def initialize(time, rule, options = {})
-      self.time = rule.instrument.cutoff_hours > 0 ? time_with_cutoff_hours(time, rule.instrument) : time
+      self.time = rule.instrument.cutoff_hours > 0 ? [time, rule.instrument.cutoff_hours.hours.from_now].max : time
       self.rule = rule
       self.day_start = Time.zone.local(time.year, time.month, time.day, rule.start_hour, rule.start_min, 0)
       self.day_end   = Time.zone.local(time.year, time.month, time.day, rule.end_hour, rule.end_min, 0)
       self.options   = options
       adjust_time
-    end
-
-    def time_with_cutoff_hours(time, instrument)
-      first_available_time = Time.zone.now + instrument.cutoff_hours.hours
-      time < first_available_time ? first_available_time : time
     end
 
     def adjust_time
--- a/app/support/products/scheduling_support.rb
+++ b/app/support/products/scheduling_support.rb
@@ -155,13 +155,17 @@ module Products::SchedulingSupport
     attr_accessor :time, :rule, :day_start, :day_end, :options
 
     def initialize(time, rule, options = {})
-      first_available_time = Time.zone.now + rule.instrument.cutoff_hours.hours
-      self.time = (time < first_available_time) ? first_available_time : time
+      self.time = rule.instrument.cutoff_hours > 0 ? time_with_cutoff_hours(time, rule.instrument) : time
       self.rule = rule
       self.day_start = Time.zone.local(time.year, time.month, time.day, rule.start_hour, rule.start_min, 0)
       self.day_end   = Time.zone.local(time.year, time.month, time.day, rule.end_hour, rule.end_min, 0)
       self.options   = options
       adjust_time
+    end
+
+    def time_with_cutoff_hours(time, instrument)
+      first_available_time = Time.zone.now + instrument.cutoff_hours.hours
+      time < first_available_time ? first_available_time : time
     end
 
     def adjust_time
--- a/app/support/reservations/validations.rb
+++ b/app/support/reservations/validations.rb
@@ -181,7 +181,7 @@ module Reservations::Validations
   private
 
   def admin_or_ordered_by_staff?
-    admin? || order_detail.created_by_user.operable_facilities.include?(facility)
+    admin? || order_detail.try(:created_by_user) && order_detail.created_by_user.operable_facilities.include?(facility)
   end
 
   def requires_cutoff_validation?
--- a/app/support/reservations/validations.rb
+++ b/app/support/reservations/validations.rb
@@ -19,7 +19,7 @@ module Reservations::Validations
       end
     end
 
-    validate :starts_before_cutoff, if: :requires_cutoff_validation?, unless: :admin_or_created_by_admin?, on: [:create, :update]
+    validate :starts_before_cutoff, if: :requires_cutoff_validation?, unless: :admin_or_ordered_by_staff?, on: [:create, :update]
     validate :starts_before_ends
     validate :duration_is_interval
   end
@@ -180,8 +180,8 @@ module Reservations::Validations
 
   private
 
-  def admin_or_created_by_admin?
-    admin? || order_detail.created_by_user.administrator?
+  def admin_or_ordered_by_staff?
+    admin? || order_detail.created_by_user.operable_facilities.include?(facility)
   end
 
   def requires_cutoff_validation?
--- a/app/support/reservations/validations.rb
+++ b/app/support/reservations/validations.rb
@@ -36,7 +36,7 @@ module Reservations::Validations
   end
 
   def starts_before_cutoff
-    errors.add(:reserve_start_at, :after_cutoff) if reserve_start_at < product.cutoff_hours.hours.from_now
+    errors.add(:reserve_start_at, :after_cutoff, hours: product.cutoff_hours) if reserve_start_at < product.cutoff_hours.hours.from_now
   end
 
   def duration_is_interval
--- a/app/support/reservations/validations.rb
+++ b/app/support/reservations/validations.rb
@@ -19,7 +19,7 @@ module Reservations::Validations
       end
     end
 
-    validate :starts_before_cutoff, if: :requires_cutoff_validation?, unless: :admin?, on: [:create, :update]
+    validate :starts_before_cutoff, if: :requires_cutoff_validation?, unless: :admin_or_created_by_admin?, on: [:create, :update]
     validate :starts_before_ends
     validate :duration_is_interval
   end
@@ -179,6 +179,10 @@ module Reservations::Validations
   end
 
   private
+
+  def admin_or_created_by_admin?
+    admin? || order_detail.created_by_user.administrator?
+  end
 
   def requires_cutoff_validation?
     product.cutoff_hours && reserve_start_at && in_the_future? && reserve_start_at_changed?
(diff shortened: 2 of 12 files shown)


app/support/reservations/validations.rb
```diff
@@ -34,6 +35,12 @@ def starts_before_ends
     end
   end
 
+  def starts_before_cutoff
+    if product.cutoff_time && reserve_start_at && in_the_future? && reserve_start_at_changed?
+      errors.add("reserve_start_at", "must be before the cutoff time for this instrument") if reserve_start_at < Time.now + product.cutoff_time.hours
```
drench: I'm going to agree with Code Climate: go with `Time.current`.

drench: This file hasn't been localizing messages generally, but it probably should. Something like `errors.add(:reserve_start_at, :after_cutoff)`?

app/views/instruments/_instrument_fields.html.haml
```diff
@@ -70,3 +70,6 @@
   = f.input :lock_window,
     label: text("instruments.instrument_fields.reservation.label.lock_window"),
     hint: text("instruments.instrument_fields.reservation.instruct.lock_window")
+  = f.input :cutoff_time,
+    label: t("instruments.instrument_fields.reservation.label.cutoff_time"),
+    hint: t("instruments.instrument_fields.reservation.instruct.cutoff_time")
```
drench: I'd go `text()` over `t()` here, for consistency if nothing else.

config/locales/en.yml
```diff
@@ -1091,6 +1091,7 @@ en:
             This field should be blank for instruments managed by reservation only.
             When blank or set to 0, the system will not cancel unstarted reservations.
           lock_window: "Window of time within which a customer may no longer move their reservation. Cancellations are permitted but may invoke a fee."
+          cutoff_time: "Minimum amount of hours before reservation instrument can be reserved."
```
drench: I'd take "number of hours" over "amount of hours" here.

spec/models/reservations_validations_spec.rb
```diff
@@ -24,4 +24,24 @@
...
   end
+
+  describe "starts_before_cutoff" do
+    context "when there is a cutoff" do
+      before :each do
+        reservation.product.update_attribute(:cutoff_time, 2)
+      end
+
+      context "when reservation is after the cutoff" do
+        before { reservation.assign_attributes(reserve_start_at: Time.now + 3.hours, reserve_end_at: Time.now + 4.hours) }
```
drench: If you set `reserve_start_at` using the `reservation` factory, it will automatically set `reserve_end_at` to 1 hour later. Something like `FactoryGirl.build(:setup_reservation, reserve_start_at: 3.hours.from_now)`.

spec/models/reservations_validations_spec.rb
```diff
@@ -24,4 +24,24 @@
...
+
+        it { is_expected.to be_valid }
+      end
+
+      context "when reservation is before the cutoff" do
+        before { reservation.assign_attributes(reserve_start_at: Time.now + 1.hour, reserve_end_at: Time.now + 2.hours) }
```
drench: Same as above on having the factory set `reserve_end_at` for you.

db/migrate/20161213173410_add_cutoff_time_to_products.rb
```diff
@@ -0,0 +1,7 @@
+class AddCutoffTimeToProducts < ActiveRecord::Migration
+
+  def change
+    add_column :products, :cutoff_time, :integer, null: false, default: 0
```
drench: `cutoff_time` almost sounds like it would be a datetime. `cutoff_hours` maybe?

spec/models/reservations_validations_spec.rb
```diff
@@ -24,4 +24,24 @@
     reservation.update_attributes reserve_start_at: now, reserve_end_at: now + 1.hour + 5.minutes
     expect(reservation.errors[:base]).to be_present
   end
+
+  describe "starts_before_cutoff" do
+    context "when there is a cutoff" do
+      before :each do
+        reservation.product.update_attribute(:cutoff_time, 2)
```
drench: Does anything weird happen if `:cutoff_time` is negative?

app/support/reservations/validations.rb
jhanggi: Or `product.cutoff_time.hours.from_now`. 

I like the idea of localizing the error. That also should allow a test for `expect(reservation.errors).to be_added(:reserve_start_at, :after_cutoff)` to make sure that it's invalid for the right reason.

I also don't love the conditional setup, especially since the important one probably falls off to the right of the screen (might be better after localization) and we also have a conditional in the `before_action` declaration. 

I'm not sure the best setup is. I think `reserve_start_at` and `reserve_start_at_changed?` could move to the `before_action`. Maybe `in_the_future?` as well. Those are sort of the should the validation run, and then `product.cutoff_time` and the time check are the validation itself. That might be the wrong line, though, so I'm open to ideas.

config/locales/en.yml
jhanggi: Maybe remove the "reservation". "Minimum number of hours before instrument can be reserved." NU has a lower priority ticket for this, but I think they might have a good idea of how to describe this.

spec/models/reservations_validations_spec.rb
jhanggi: We should probably add a validation on instrument for numericality equal or greater than 0.

kt32291:
@jhanggi it's handled, it doesn't add errors in the window if user is an admin. I'll work on these comments tomorrow, wanted you to know I see them though 👍

kt32291:
@jhanggi @drench updated

config/locales/en.models.yml
```diff
@@ -76,6 +76,8 @@ en:
             base:
               conflict: The reservation conflicts with another reservation.
               conflict_in_cart: The reservation conflicts with another reservation in your cart. Please purchase or remove it then continue.
+            reserve_start_at:
+              after_cutoff: must be before the cutoff time for this instrument
```
jhanggi: I think it might be nice to include the # of hours in this message. So something like "must be at least %{hours} hours in the future".

app/support/products/scheduling_support.rb
```diff
@@ -155,14 +155,19 @@ class ReservationFinder
     attr_accessor :time, :rule, :day_start, :day_end, :options
 
     def initialize(time, rule, options = {})
-      self.time = time
+      self.time = rule.instrument.cutoff_hours > 0 ? time_with_cutoff_hours(time, rule.instrument) : time
```
jhanggi: Do we actually need this `> 0` check? If we used `to_i`, then `nil` and zero would covert to `0`.

app/support/products/scheduling_support.rb
```diff
@@ -155,14 +155,19 @@ class ReservationFinder
...
       self.rule = rule
       self.day_start = Time.zone.local(time.year, time.month, time.day, rule.start_hour, rule.start_min, 0)
       self.day_end   = Time.zone.local(time.year, time.month, time.day, rule.end_hour, rule.end_min, 0)
       self.options   = options
       adjust_time
     end
 
+    def time_with_cutoff_hours(time, instrument)
+      first_available_time = Time.zone.now + instrument.cutoff_hours.hours
```
jhanggi: I think we like to prefer `hours.from_now`. And in general, we're preferring `Time.current` to `Time.zone.now`, but that's not a big deal.

app/support/products/scheduling_support.rb
```diff
@@ -155,14 +155,19 @@ class ReservationFinder
...
+      time < first_available_time ? first_available_time : time
```
jhanggi: What about one-lining it to something like `[time, instrument.cutuff_hours.hours.from_now].max`. Then we might not even need the extracted method.

app/support/reservations/validations.rb
```diff
@@ -19,6 +19,7 @@ module Reservations::Validations
       end
     end
 
+    validate :starts_before_cutoff, if: :requires_cutoff_validation?, unless: :admin_or_ordered_by_staff?, on: [:create, :update]
```
jhanggi: Do we need the `on: [:create, :update]`?

app/support/reservations/validations.rb
```diff
@@ -175,6 +180,14 @@ def longest_reservation_window(groups = [])
 
   private
 
+  def admin_or_ordered_by_staff?
+    admin? || order_detail.try(:created_by_user) && order_detail.created_by_user.operable_facilities.include?(facility)
```
jhanggi: This is a obscure use case, but it's definitely possible, and I'm not quite sure this is behaving like it should in that situation.

* An instrument has a 12 hour cutoff
* An admin places a reservation on behalf of a user for 6 hours from now
* The user logs in and goes in to My Reservations and clicks on the reservation
* They can move the reservation

I'm not sure they should be able to do that.

I think that's part of why we use the `save_extended_validations` which saves as the acting user rather than checks how it was purchased.

app/views/instruments/_instrument_fields.html.haml
```diff
@@ -70,3 +70,6 @@
   = f.input :lock_window,
     label: text("instruments.instrument_fields.reservation.label.lock_window"),
     hint: text("instruments.instrument_fields.reservation.instruct.lock_window")
+  = f.input :cutoff_hours,
+    label: text("instruments.instrument_fields.reservation.label.cutoff_hours"),
```
jhanggi: Would it be worth moving some of these locales onto the model so simple_form can use `human_attribute_name`, which is one of the things it tries to use for `label`.

app/support/products/scheduling_support.rb
kt32291: Huh? Yes, because we don't want to try to apply the zero, because we want the time that's passed in if it's zero (if they don't have cutoff hours), not the current time plus 0.

kt32291:
@jhanggi updated

spec/models/reservations_validations_spec.rb
```diff
@@ -24,4 +24,40 @@
...
+      before :each do
+        reservation.product.update_attribute(:cutoff_hours, 2)
+      end
+
+      context "when reservation is after the cutoff" do
+        let(:reservation) { build :setup_reservation, reserve_start_at: Time.zone.now + 3.hours }
+        let(:user) { reservation.order_detail.created_by_user }
+
+        it "saves the reservation" do
+          expect(reservation.save_as_user(user)).to eq(true)
```
drench: Minor (and not show-stopping): this project goes more for `be true` over `eq(true)` style.
Though in this particular case I might have even considered `expect { reservation.save_as_user!(user) }.not_to raise_error`

spec/models/instrument_spec.rb
```diff
@@ -488,6 +488,15 @@
       assert_equal 45, @next_reservation.reserve_start_at.min
     end
 
+    it "should find next available reservation with cutoff hours" do
+      expect(@rule.instrument.update_attribute :cutoff_hours, 10).to be true
```
jhanggi: Not a big deal, but I generally put these kind of setup steps outside an `expect`. I would consider putting it inside a `before` block.

spec/models/instrument_spec.rb
```diff
@@ -488,6 +488,15 @@
...
+
+      @next_reservation = @instrument.next_available_reservation(after = Time.zone.now.beginning_of_day)
```
jhanggi: Do we need the `after` assignment? Does `next_reservation` need to be an instance variable?

spec/models/reservations_validations_spec.rb
```diff
@@ -24,4 +24,40 @@
...
   end
+
+  describe "starts_before_cutoff" do
+    context "when there is a cutoff" do
+      before :each do
+        reservation.product.update_attribute(:cutoff_hours, 2)
+      end
+
+      context "when reservation is after the cutoff" do
+        let(:reservation) { build :setup_reservation, reserve_start_at: Time.zone.now + 3.hours }
```
jhanggi: `3.hours.from_now`

spec/models/instrument_spec.rb
drench: If you're going to do more edits, go with `it "finds the next available…"` instead of `it "should find…"`

kt32291:
@jhanggi updated

PR #840, opened 2016-12-16
Update orders-in-review tests to disregard array ordering

drench:
This is cherry-picked from https://github.com/tablexi/nucore-nu/pull/216. The results coming back from Oracle are correct but not necessarily in the same order as from MySQL.


jhanggi:
👍

PR #841, opened 2016-12-20
[#133667] Replace TimeCop with ActiveSupport's TimeHelpers

drench:
TimeHelpers, either by accident or design, isn't a full drop-in replacement for TimeCop. The changes here are mostly due to:

* `travel`/`travel_to` blocks restoring the real time, not the time it may have been `travel`ed to before the block
* `travel`/`travel_to` freezing time

The first feels like it may be a bug. The second means that, since the entire test suite is run under a `travel_to`, the clock never moves except when changed through `travel`ing. There are places where there was an assumption of clock movement. Example: taking an instrument's order detail to `complete` checks if `reservation.reserve_end_at < Time.zone.now`. Some specs that worked under timecop did so because `Time.zone.now` was a fraction of a second after `reserve_end_at`. There are workarounds here for that situation.

When updating specs in the future, we should check for `:time_travel` tagging and decide if it's still necessary, since some may have it only to stop the clock.


spec/models/order_detail_spec.rb
```diff
@@ -475,13 +474,14 @@ def create_price_policy(params)
                order_detail: order_detail_without_price_policy,
               )
 
-        Timecop.travel(2.days.from_now) do
-          order_details.each do |order_detail|
-            order_detail.change_status!(OrderStatus.find_by_name("In Process"))
-            order_detail.change_status!(OrderStatus.find_by_name("Complete"))
-            order_detail.reload
-          end
+        current_time = Time.current
```
jhanggi: This seems like something we could wrap up in our own helper and keep a block API similar to what Timecop had. Something like:

```ruby
def travel_and_return(interval)
  current_time = Time.current
  travel(interval)
  yield
  travel(current_time)
end
```

We could also have a `travel_to_and_return` if we need it. I'm up for different names.

jhanggi: This might even allow us to deal with nesting if we always use this method instead of the built-in ones that reset to actual current time. Not sure, though.

jhanggi:
I have the codeclimate chrome plugin installed and that's where I'm seeing the finder warnings. Don't see them in codeclimate proper.

app/models/order_status.rb
```diff
@@ -15,7 +15,7 @@ class OrderStatus < ActiveRecord::Base
     end
   end
 
-  scope :for_facility, -> (facility) { where(facility_id: [nil, facility.id]).order(:lft) }
+  scope :for_facility, ->(facility) { where(facility_id: [nil, facility.id]).order(:lft) }
```
jhanggi: Did rubocop do this? I thought we usually put a space there.

spec/controllers/reports/general_reports_controller_spec.rb
```diff
@@ -249,7 +249,7 @@ def assert_report_params_init
 
   def assert_report_init(_label)
     expect(response).to be_success
-    expect(assigns(:total_quantity)).to be_instance_of Fixnum
+    expect(assigns(:total_quantity)).to be_instance_of Fixnum # rubocop:disable Lint/UnifiedInteger
```
jhanggi: Should this be disabled, or can we change this to `Integer`?

spec/models/instrument_spec.rb
```diff
@@ -718,7 +718,7 @@
         context "but it was canceled" do
           let(:user) { FactoryGirl.build :user }
           before :each do
-            Timecop.travel(60.minutes.ago) do
+            travel(-60.minutes) do
```
jhanggi: I don't mind `travel(x.minutes)` to `travel_to(x.minutes.from_now)`, but I kind of prefer `travel_to(60.minutes.ago)` to using negative minutes. What do you think?

spec/models/reservation_spec.rb
```diff
@@ -1023,13 +1016,19 @@
...
-    it "sets actual start time", :timecop_freeze do
+    it "sets actual start time", :time_travel do
       reservation.start_reservation!
       expect(reservation.actual_start_at).to eq(Time.current)
     end
 
     context "with a running reservation" do
-      let!(:running) { create :setup_reservation, product: instrument, reserve_start_at: 1.hour.ago, reserve_end_at: Time.current, actual_start_at: 1.hour.ago }
+      let!(:running) do
+        start_time = 1.hour.ago - 1.second
```
jhanggi: What does the one second do? Why didn't we need it before?

spec/models/reservation_spec.rb
```diff
@@ -1115,8 +1114,8 @@
         @instrument.schedule_rules.destroy_all
         @instrument.schedule_rules.reload
         @instrument.update_attribute :reserve_interval, 15
-        @rule_9_to_5 = @instrument.schedule_rules.create(FactoryGirl.attributes_for(:schedule_rule, start_hour: 9, end_hour: 17))
-        @rule_5_to_7 = @instrument.schedule_rules.create(FactoryGirl.attributes_for(:schedule_rule, start_hour: 17, end_hour: 19))
+        @rule_9to5 = @instrument.schedule_rules.create(FactoryGirl.attributes_for(:schedule_rule, start_hour: 9, end_hour: 17))
```
jhanggi: Is this variable used anywhere?

spec/spec_helper.rb
```diff
@@ -98,12 +101,16 @@
...
+
+  config.around(:each, :time_travel) do |example|
+    current_time = Time.current
     # freeze time to specific time by defining let(:now)
-    time = defined?(now) ? now : Time.zone.now
-    Timecop.freeze time do
-      example.call
-    end
+    time = defined?(now) ? now : Time.current
+    travel_to(time)
```
jhanggi: Could this use the `travel_to_and_return` form?

app/models/order_status.rb
drench: This is Rubocop. This spacing got flagged in #843 too.

spec/controllers/reports/general_reports_controller_spec.rb
drench: `@total_quantity` is not an `instance_of?(Integer)`, but it is a `kind_of?(Integer)`.Updating.

spec/models/reservation_spec.rb
drench: It's not used, but its companion `@rule_5to7` is. Assigning it feels like documentation, kind of like how we sometimes have a `let!(:something)` without referencing the `:something` in newer tests. I'm OK with leaving it until we update this section.

spec/models/reservation_spec.rb
drench: We didn't need it because the clock used to move. There could be a clearer way to do this, but the problem came down to `OrderDetail#has_completed_reservation?` being `false` because `reserve_end_at` had not passed, and never will.

PR #842, opened 2016-12-21
[#133446] Adds reply to for facility email

kt32291:
https://pm.tablexi.com/issues/133446

This adds the reply_to option to all facility related emails so they can communicate with the facility in a reply email. 

I *believe* this is all the spots that need it, but I can add it to any other they'd like the reply_to to direct to the facility.

### DIFF
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -36,7 +36,7 @@ class Notifier < ActionMailer::Base
 
   def order_notification(order, recipient)
     @order = order
-    send_nucore_mail recipient, text("views.notifier.order_notification.subject"), template_name: "order_receipt", reply: @order.facility.email
+    send_nucore_mail recipient, text("views.notifier.order_notification.subject"), "order_receipt", reply_to_email(@order, @order.facility)
   end
 
   # Custom order forms send out a confirmation email when filled out by a
@@ -45,7 +45,7 @@ class Notifier < ActionMailer::Base
     @user = args[:user]
     @order = args[:order]
     @greeting = text("views.notifier.order_receipt.intro")
-    send_nucore_mail args[:user].email, text("views.notifier.order_receipt.subject"), reply: @order.facility.email
+    send_nucore_mail args[:user].email, text("views.notifier.order_receipt.subject"), nil, reply_to_email(@order, @order.facility)
   end
 
   def review_orders(args)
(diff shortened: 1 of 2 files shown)


app/mailers/notifier.rb
```diff
@@ -87,8 +87,8 @@ def statement_pdf
     @statement_pdf ||= StatementPdfFactory.instance(@statement)
   end
 
-  def send_nucore_mail(to, subject, template_name = nil)
-    mail(subject: subject, to: to, template_name: template_name)
+  def send_nucore_mail(to, subject, options = {})
```
drench: I'd lean toward keyword args (with `nil` defaults) rather than an options hash to be more self-documenting.

app/mailers/notifier.rb
```diff
@@ -87,8 +87,8 @@ def statement_pdf
...
+    mail(subject: subject, to: to, template_name: options[:template_name], reply_to: options[:reply])
```
drench: I'd go with `:reply_to` for the `options` key, for consistency.

drench:
I like the idea of using `Product#contact_email` and falling back to `Facility#email`.

I don't know that we need a feature flag because they're already sort of doing it. UIC has `contact_email` set for all of its products and NU has it set for none. UIC has also set `email` for all facilities while NU has set it for all facilities but 8 (and at least one is a 'test' facility).

So we will need a fallback for those 8 NU facilities, but that might be enough.

jhanggi:
I'd be fine without the feature flag. We should probably just skip adding a reply-to if there isn't an email to use. ActionMailer may already treat nil/blank in that way. If so, then we just need to do the product->facility fallback.

app/mailers/notifier.rb
```diff
@@ -87,8 +87,12 @@ def statement_pdf
     @statement_pdf ||= StatementPdfFactory.instance(@statement)
   end
 
-  def send_nucore_mail(to, subject, template_name = nil)
-    mail(subject: subject, to: to, template_name: template_name)
+  def send_nucore_mail(to, subject, template_name = nil, reply_to = nil)
+    mail(subject: subject, to: to, template_name: template_name, reply_to: reply_to)
```
drench: If `reply_to` is `nil` it looks like we get a `Reply-To:` header with no value. GMail is smart enough to ignore it, but it would be nicer to not generate the header at all if it's blank.

app/mailers/notifier.rb
```diff
@@ -87,8 +87,12 @@ def statement_pdf
     @statement_pdf ||= StatementPdfFactory.instance(@statement)
   end
 
-  def send_nucore_mail(to, subject, template_name = nil)
-    mail(subject: subject, to: to, template_name: template_name)
+  def send_nucore_mail(to, subject, template_name = nil, reply_to = nil)
```
kt32291: Ahh sorry no I totally misunderstood what you commented before, now I got you.

kt32291:
@drench updated!

PR #844, opened 2016-12-23
Update travel/travel_to to use TimeHelper methods

drench:
A few more execution-order time issues emerged after merging #841. I searched the code for `/\btravel(_to)?\b/` and believe this takes care of everything that should be using the `travel_to_and_return`/`travel_and_return` helpers instead.


jhanggi:
👍

PR #846, opened 2017-01-09
[#132371] Adds task for reporting on overlapping price rules

kt32291:

This adds a task that returns a basic list for which instruments and price groups have overlapping price policies, and the ids of those price policies (in case we want to manually fix them in some way).

### DIFF
--- a/lib/tasks/reports.rake
+++ b/lib/tasks/reports.rake
@@ -2,7 +2,7 @@ namespace :reports do
   desc "Gives report for concatenated price rules"
   task concatenated_price_rules: :environment do
     policies = {}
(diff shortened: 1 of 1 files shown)


lib/tasks/reports.rake
```diff
@@ -0,0 +1,15 @@
+namespace :reports do
+  desc "Gives report for concatenated price rules"
+  task concatenated_price_rules: :environment do
+    policies = {}
+    PricePolicy.current.group_by { |p| [p.product_id, p.price_group_id] }.map { |k, ps| policies[k] = ps if ps.count > 1 }
```
drench: I'd go with `each` over `map` here since you're not directly capturing its result. Either that or consider refactoring using `each_with_object`.

jhanggi:
I'm cool with this just to see how many of these there are.

kt32291:
@drench updated!

PR #848, opened 2017-01-10
Oracle-related query updates and refactoring

drench:
Issues in Oracle queries emerged after switching from TimeCop to Rails' built-in Time Helpers (because `travel`/`travel_to` always round to the nearest second).

These changes came out of https://github.com/tablexi/nucore-nu/pull/217.


jhanggi:
👍

PR #849, opened 2017-01-10
[#55313] Added expiration date column to accounts

kt32291:
https://pm.tablexi.com/issues/55313

This adds the expiration date column to facility accounts index view.


config/locales/en.yml
```diff
@@ -200,6 +200,7 @@ en:
         account: "Payment Source"
         owner: "Owner"
         total: "Unreconciled Total (actual + estimated)"
```
jhanggi: Do we need this locale anymore?

app/views/facility_accounts/_account_balances.html.haml
```diff
@@ -7,14 +7,13 @@
...
-        %th.currency= t(".th.total")
+        %th= t(".th.expiration")
     %tbody
       - @accounts.each do |account|
         %tr
           %td= payment_source_link_or_text(account)
           %td= account.owner_user.full_name if account.owner_user
-          %td.currency
-            = number_to_currency(account.unreconciled_total(current_facility))
+          %td= human_date(account.expires_at)
```
jhanggi: Don't know if nows the time to do it, but could `humam_date`'s implementation be replaced with `date.strftime("%B %e, %Y")` or move that format into `en.yml` and use Rail's `l` helper?

app/views/facility_accounts/_account_balances.html.haml
```diff
@@ -7,14 +7,13 @@
       %tr
         %th= t(".th.account")
         %th= t(".th.owner")
-        %th.currency= t(".th.total")
+        %th= t(".th.expiration")
```
jhanggi: Now that this file doesn't include balances anymore, should we change the filename?

Also, should we use `Account.human_attribute_name(:expires_at)` to keep this header consistent?

PR #850, opened 2017-01-10
[#132371] Fix view to only show newest price policies

kt32291:

This fixes the view to only show newest price policies instead of the mistakenly overlapping ones together.

### DIFF
--- a/app/models/price_policy.rb
+++ b/app/models/price_policy.rb
@@ -33,13 +33,13 @@ class PricePolicy < ActiveRecord::Base
   end
 
   def self.current_and_newest
-    # TODO: Fix bug that causes this to be neccessary (in truncate_existing_policies)
+    # TODO: Fix bug that allows overlapping price policies (in truncate_existing_policies)
     # This method returns the newest price policy for when price policies accidentally overlap.
     current.newest
   end
 
   def self.newest
-    ids = group(:price_group_id).maximum(:id).collect(&:last)
+    ids = group(:price_group_id).maximum(:id).values
     where(id: ids)
   end
 
--- a/app/models/price_policy.rb
+++ b/app/models/price_policy.rb
@@ -35,7 +35,12 @@ class PricePolicy < ActiveRecord::Base
   def self.current_and_newest
     # TODO: Fix bug that causes this to be neccessary (in truncate_existing_policies)
     # This method returns the newest price policy for when price policies accidentally overlap.
-    current.group_by(&:price_group_id).map { |_, policies| policies.sort { |x, y| y.id <=> x.id }.first }
+    current.newest
+  end
+
+  def self.newest
+    ids = group(:price_group_id).maximum(:id).collect(&:last)
+    where(id: ids)
   end
 
   def self.current_for_date(date)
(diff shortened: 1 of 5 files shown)


app/models/price_policy.rb
```diff
@@ -32,6 +32,12 @@ def self.current
     current_for_date(Time.zone.now)
   end
 
+  def self.current_and_newest
+    # TODO: Fix bug that causes this to be neccessary (in truncate_existing_policies)
+    # This method returns the newest price policy for when price policies accidentally overlap.
+    current.group_by(&:price_group_id).map { |_, policies| policies.sort { |x, y| y.id <=> x.id }.first }
```
drench: It's a weird case. Could you add a test, preferably one that would fail once the bug is fixed?

drench: There are things you could do to tighten up the enumerable method chaining, but once you have tests I'd like to see this as a scope returning an activerecord relation.

jhanggi: I'd definitely like to see tests as well. Is the "newest" price policy the one that will get applied?

kt32291:
@drench @jhanggi updated! I added a couple specs, and at least one will fail when we fix the issue (depending on how we fix it)

spec/models/price_policy_spec.rb
```diff
@@ -34,6 +34,29 @@
...
+    let(:overlapping_price_policy) do
+      FactoryGirl.create(
+        :instrument_price_policy,
+        start_date: (price_policy.start_date + 1.day).beginning_of_day,
+        product: price_policy.product,
+        price_group: price_policy.price_group,
+      )
+    end
+
+    before { overlapping_price_policy.update_attributes(start_date: 1.day.ago.beginning_of_day) }
```
jhanggi: I'm a little confused by what's happening here in these specs. Why are we setting the `start_date` for this policy twice? And why are we checking for the validity?

spec/models/price_policy_spec.rb
```diff
@@ -34,6 +34,29 @@
...
+
+    it "saves updated price policy" do
+      expect(overlapping_price_policy).to be_valid
+    end
+
+    it "has two overlapping policies" do
```
jhanggi: I think it might be smart to do a spec at a higher level, where we have two policies: and older one with a cheaper price and a newer one. Then on the order detail, find the do `assign_price_policy`, which will call `cheapest_price_policy`  and make sure it chooses the correct policy.

app/models/price_policy.rb
```diff
@@ -32,6 +32,17 @@ def self.current
...
   end
 
+  def self.current_and_newest
+    # TODO: Fix bug that causes this to be neccessary (in truncate_existing_policies)
+    # This method returns the newest price policy for when price policies accidentally overlap.
+    current.newest
+  end
+
+  def self.newest
+    ids = group(:price_group_id).maximum(:id).collect(&:last)
```
jhanggi: I don't have a strong preference, but I had to try out `group(:price_group_id).maximum(:id)` to see that it returns a hash of `{ price_group_id => max_id }`. We could call `values` on it. Might read a little clearer than the `collect` and make it more obvious what the earlier methods return.

app/models/price_policy.rb
```diff
@@ -32,6 +32,17 @@ def self.current
     current_for_date(Time.zone.now)
   end
 
+  def self.current_and_newest
+    # TODO: Fix bug that causes this to be neccessary (in truncate_existing_policies)
```
jhanggi: I think expanding on what the bug is in this comment will be helpful when we stumble back on this in the future. Something like `Fix bug that allows overlapping price policies`.

kt32291:
@drench @jhanggi Updated!

PR #851, opened 2017-01-12
[#133731] Increase the size of the delayed_jobs.handler column for MySQL

drench:
This increases the size limit from 2^8 characters to 2^32 for MySQL.

For Oracle, this column is already a `CLOB`, with a size limit of 2**31 characters, so the migration makes no changes.


jhanggi:
👍 Will we need a separate migration on oracle? Or is it already big enough, and we just needed this for UIC?

drench:
We should be fine with Oracle for a while. I don't believe we've seen anyone hit the `CLOB` limit yet.
https://docs.oracle.com/javadb/10.10.1.2/ref/rrefclob.html

PR #852, opened 2017-01-13
Set and restore billing review_period in specs through rspec metadata

drench:
We need to set a review period in the `#in_review` tests since we can't count on a review period being set in `Settings`.

### DIFF
--- a/spec/controllers/transactions_controller_spec.rb
+++ b/spec/controllers/transactions_controller_spec.rb
@@ -6,8 +6,7 @@ RSpec.describe TransactionsController do
   let(:product) { FactoryGirl.create(:setup_item, :with_facility_account) }
   let(:user) { FactoryGirl.create(:user) }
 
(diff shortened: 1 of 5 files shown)


spec/controllers/transactions_controller_spec.rb
```diff
@@ -7,6 +7,8 @@
   let(:user) { FactoryGirl.create(:user) }
 
   describe "GET #in_review" do
+    before(:all) { Settings.billing.review_period = 7.days }
```
jhanggi: We should probably make sure to reset this after the test so we don't run into spec order issues.

drench:
I've updated this to use `billing_review_period` metadata wherever specs need to change the review period value.

drench:
I held back on major refactoring of `facility_notifications_controller_spec.rb` for now. A lot of the changes in that file are just indentation.

PR #853, opened 2017-01-16
Update demo seed to pass NU validations

drench:
This wasn't so much an issue with this fork as with NU's due to its account validations. I was able to seed a clean database on the NU fork with this patch.

### DIFF
--- a/lib/tasks/demo_seed.rake
+++ b/lib/tasks/demo_seed.rake
@@ -298,15 +298,21 @@ namespace :demo do
       account_owner_attributes,
       account_purchaser_attributes,
     ]
+    nufs_account_attributes = {
+      expires_at: 1.year.from_now,
+      created_by: user_director.id,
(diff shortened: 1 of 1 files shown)


lib/tasks/demo_seed.rake
```diff
@@ -298,12 +298,12 @@ namespace :demo do
     end
 
     # create a second nufsaccount for split accounts
-    nufsaccount2 = NufsAccount.find_by_account_number("111-2222222-44444444-01")
+    nufsaccount2 = NufsAccount.find_by_account_number("123-1234567-12345678-01")
```
jhanggi: Code climate is suggesting switching to `find_by(account_number: "xxx")`.

drench: Good idea. Finder updates are overdue. I cleaned up this one along with the others Rubocop found.

jhanggi:
👍

PR #854, opened 2017-01-16
[#133966] Join orders to transactions in-review search queries

drench:
The `OrderDetail.all_in_review` scope needs to `JOIN` the orders table to for references to `ordered_at` in transaction searches to work.

This also includes some scope refactoring and cleanup.


jhanggi:
👍

PR #855, opened 2017-01-16
[#133446] Reply to bulk email

kt32291:
https://pm.tablexi.com/issues/133446

This adds the reply_to field to the bulk email and removes it from the emails where I mistakenly added it. 

We don't actually have the product id here unless we parse it from the search criteria, so I'm using the facility email. Another option I came across would be adding it to the bulk email form, so it's always customizable (totally doable), would love to hear if anyone has a preference.

### DIFF
--- a/vendor/engines/bulk_email/app/models/bulk_email/delivery_form.rb
+++ b/vendor/engines/bulk_email/app/models/bulk_email/delivery_form.rb
@@ -43,9 +43,13 @@ module BulkEmail
       content_generator.wrap_text(custom_message, recipient_name)
     end
 
+    def reply_to
+      facility.email
+    end
+
     def deliver(recipient)
       Mailer
-        .send_mail(recipient: recipient, subject: subject, body: body(recipient.full_name))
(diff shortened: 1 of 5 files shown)


jhanggi:
There's a `product` method on line 67 of `DeliveryForm`. `product_id` might not be there for general bulk email. In that case, we should fall back to using the facility email. In the case of the instrument not available feature, or just the plain "Send Mail" feature from the instrument page (Products > Instruments > Instrument Name, e.g. https://nucore-staging.northwestern.edu/facilities/test-facility/instruments/gen_inst/schedule), the I believe `product_id` should be available. Dan can probably provide a little more information on that.

vendor/engines/bulk_email/app/models/bulk_email/delivery_form.rb
```diff
@@ -43,9 +43,13 @@ def body(recipient_name = nil)
       content_generator.wrap_text(custom_message, recipient_name)
     end
 
+    def reply_to
+      facility.email
```
drench: Agreed, you could do something like `product.try(:contact_email).presence || facility.email`

It's a good catch on `#product` because that may be an otherwise dead method.

PR #856, opened 2017-01-19
[#133383] Add optional note, fulfilled_at, and status fields to the facility order screen

drench:
This includes general updates to the related controller, views and specs. Enough was happening in private controller methods when adding-to-order that it seemed appropriate to extract it to a service object. I think Code Climate might say it hasn't gone far enough.

A few other notes:

* The addition of a bottom margin on `chosen-container` was specifically to better align the fields on this form, but looking across the app I thought this tweak would make sense elsewhere, like in the report filters.

Without the margin:
report filter without margin

With the margin:
report filter with margin

* `fulfilled_at` is a date in the form, but a timestamp in the database, set to midnight local (app) time. Would something like noon be more reasonable?

* The order status select offers all available status values, but the state machine won't necessarily allow them all to be set. This is facility-facing so I went with presuming an operator would know what the appropriate state should be.

### DIFF
--- a/app/assets/javascripts/app/merge_order_form.js.coffee
+++ b/app/assets/javascripts/app/merge_order_form.js.coffee
@@ -1,0 +1,16 @@
+class window.MergeOrderForm
+  constructor: (@$form) ->
+    @$fulfilled_at = @$form.find("#fulfilled_at")
+    @$order_status_id = @$form.find("#order_status_id")
+    @initializeListener()
+    @$order_status_id.change()
+
+  initializeListener: ->
+    @$order_status_id.change (event) =>
+      if $(event.target).find("option:selected").text() == "Complete"
+        @$fulfilled_at.show()
+      else
+        @$fulfilled_at.hide()
+
+$ ->
+  new MergeOrderForm($("form.edit_order"))
--- a/app/assets/javascripts/app/merge_order_form.js.coffee
+++ b/app/assets/javascripts/app/merge_order_form.js.coffee
@@ -13,4 +13,4 @@ class window.MergeOrderForm
         @$fulfilled_at.hide()
 
 $ ->
-  $(".js--edit-order").each -> new MergeOrderForm($(this))
+  $(".js--edit-order").each (_, form) -> new MergeOrderForm($(form))
--- a/app/assets/javascripts/app/merge_order_form.js.coffee
+++ b/app/assets/javascripts/app/merge_order_form.js.coffee
@@ -13,4 +13,4 @@ class window.MergeOrderForm
         @$fulfilled_at.hide()
 
 $ ->
-  new MergeOrderForm($("form.edit_order"))
+  $(".js--edit-order").each -> new MergeOrderForm($(this))
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -69,10 +69,16 @@ class FacilityOrdersController < ApplicationController
     if quantity <= 0
       flash[:notice] = I18n.t "controllers.facility_orders.update.zero_quantity"
     else
-      if OrderAppender.new(product, quantity, original_order, current_user).add!
-        flash[:error] = I18n.t "controllers.facility_orders.update.notices", product: product.name
-      else
-        flash[:notice] = I18n.t "controllers.facility_orders.update.success", product: product.name
+      order_appender = OrderAppender.new(product, quantity, original_order, current_user)
+      begin
+        if order_appender.add!
+          flash[:error] = I18n.t "controllers.facility_orders.update.notices", product: product.name
+        else
+          flash[:notice] = I18n.t "controllers.facility_orders.update.success", product: product.name
+        end
+      rescue => e
+        Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+        flash[:error] = I18n.t "controllers.facility_orders.update.error", product: product.name
       end
     end
 
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -60,8 +60,9 @@ class FacilityOrdersController < ApplicationController
     end
   end
 
+  # PUT/PATCH /facilities/:facility_id/orders/:id
   def update
-    product = Product.find(params[:product_add].to_i)
+    product = current_facility.products.find(params[:product_add])
     original_order = @order
     quantity = params[:product_add_quantity].to_i
 
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -63,13 +63,15 @@ class FacilityOrdersController < ApplicationController
   # PUT/PATCH /facilities/:facility_id/orders/:id
   def update
     product = current_facility.products.find(params[:product_add])
-    original_order = @order
     quantity = params[:product_add_quantity].to_i
 
     if quantity <= 0
       flash[:notice] = I18n.t "controllers.facility_orders.update.zero_quantity"
     else
-      order_appender = OrderAppender.new(product, quantity, original_order, current_user)
+      order_appender = OrderAppender.new(product: product,
+                                         quantity: quantity,
+                                         original_order: @order,
+                                         user: current_user)
       begin
         if order_appender.add!
           flash[:error] = I18n.t "controllers.facility_orders.update.notices", product: product.name
@@ -82,7 +84,7 @@ class FacilityOrdersController < ApplicationController
       end
     end
 
-    redirect_to facility_order_path(current_facility, original_order)
+    redirect_to facility_order_path(current_facility, @order)
   end
 
   protected
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -69,7 +69,11 @@ class FacilityOrdersController < ApplicationController
     if quantity <= 0
       flash[:notice] = I18n.t "controllers.facility_orders.update.zero_quantity"
     else
-      add_to_order(product, quantity, original_order)
+      if OrderAppender.new(product, quantity, original_order, current_user).add!
+        flash[:error] = I18n.t "controllers.facility_orders.update.notices", product: product.name
+      else
+        flash[:notice] = I18n.t "controllers.facility_orders.update.success", product: product.name
+      end
     end
 
     redirect_to facility_order_path(current_facility, original_order)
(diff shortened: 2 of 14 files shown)


app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -1,7 +1,26 @@
...
+
+      = select_tag :product_add,
+        options_for_select(current_facility.products.active_plus_hidden.pluck(:name, :id))
+
+      = text_field_tag :note,
+        nil,
+        placeholder: OrderDetail.human_attribute_name(:note),
+        class: "string optional"
+
+      = text_field_tag :fulfilled_at,
```
jhanggi: Maybe we show/hide the fulfilled at box based on the status?

app/services/order_appender.rb
```diff
@@ -0,0 +1,71 @@
...
+    @note = note
+    @fulfilled_at = fulfilled_at
+    @order_status_id = order_status_id
+  end
+
+  def add!
+    notifications = false
+    order_details.each do |order_detail|
+      order_detail.note = note if note.present?
+      order_detail.fulfilled_at = fulfilled_at if fulfilled_at.present?
```
jhanggi: I think this needs to happen after the `change_status!`. If I set a date and Complete, the fulfilled at is being set to now.

And this should not happen if the status is not complete (or a substatus of complete). Right now, if I put a date in, and leave status blank, it gets created as New, but with a fulfilled_at.

app/views/facility_orders/_merge_order_form.html.haml
jhanggi: Also, I probably shouldn't be able to set a fulfilled date in the future.

app/controllers/facility_orders_controller.rb
```diff
@@ -60,18 +60,34 @@ def send_receipt
...
+                                         order_status_id: params[:order_status_id].presence)
+      begin
+        if order_appender.add!
+          flash[:error] = I18n.t "controllers.facility_orders.update.notices", product: product.name
+        else
+          flash[:notice] = I18n.t "controllers.facility_orders.update.success", product: product.name
+        end
+      rescue => e
+        Rails.logger.error "#{e.message}\n#{e.backtrace.join("\n")}"
+        flash[:error] = I18n.t "controllers.facility_orders.update.error", product: product.name
```
kt32291: Same note here on using ```text``` and a shorter key

app/services/order_appender.rb
```diff
@@ -0,0 +1,71 @@
...
+      account_id: original_order.account_id,
+      user_id: original_order.user_id,
+      created_by: user.id,
+    )
+  end
+
+  def merge?
+    products.any? do |product|
+      product.is_a?(Instrument) ||
+        (product.is_a?(Service) && (product.active_survey? || product.active_template?))
```
kt32291: Is there any way to simplify this method?

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -1,7 +1,26 @@
...
+        nil,
+        placeholder: OrderDetail.human_attribute_name(:fulfilled_at),
+        class: "datepicker__data string optional"
+
+      = select_tag :order_status_id,
```
drench: I'm seeing added products set to "New" on `master`.

drench:
Updated.

The UI should only show the `fulfilled_at` field when the order status selected is "Complete". That's literally what it uses in the JavaScript. Maybe not ideal, but we don't really have anything in the system to convey that "Complete" means setting `fulfilled_at` is OK. The backend should ignore `fulfilled_at` unless the status is "Complete".

Also, while the UI will allow you to select "Complete" for an `Instrument`, this will fail because the state machine requires a completed reservation, which it won't have.

app/assets/javascripts/app/merge_order_form.js.coffee
```diff
@@ -0,0 +1,16 @@
...
+
+  initializeListener: ->
+    @$order_status_id.change (event) =>
+      if $(event.target).find("option:selected").text() == "Complete"
+        @$fulfilled_at.show()
+      else
+        @$fulfilled_at.hide()
+
+$ ->
+  new MergeOrderForm($("form.edit_order"))
```
jhanggi: Should we call this `.js--edit-order`?

app/assets/javascripts/app/merge_order_form.js.coffee
```diff
@@ -0,0 +1,16 @@
+class window.MergeOrderForm
+  constructor: (@$form) ->
+    @$fulfilled_at = @$form.find("#fulfilled_at")
+    @$order_status_id = @$form.find("#order_status_id")
+    @initializeListener()
+    @$order_status_id.change()
+
+  initializeListener: ->
+    @$order_status_id.change (event) =>
+      if $(event.target).find("option:selected").text() == "Complete"
```
jhanggi: I always forget, does `val()` work for `$("select")`?

app/services/order_appender.rb
```diff
@@ -0,0 +1,68 @@
...
+    products = product.is_a?(Bundle) ? product.products : [product]
+
+    note = params[:note].presence
+    fulfilled_at = parse_fulfilled_at(params[:fulfilled_at])
+    order_status = load_order_status(params[:order_status_id].presence, product.facility)
+    order = products.any?(&:mergeable?) ? build_merge_order : original_order
+    order_details = order.add(product, quantity, created_by: user.id)
+
+    notifications = false
+    order_details.each do |order_detail|
```
jhanggi: This is a pretty complicated method. Is there any way we can break it down?

app/services/order_appender.rb
```diff
@@ -0,0 +1,68 @@
...
+      facility_id: original_order.facility_id,
+      account_id: original_order.account_id,
+      user_id: original_order.user_id,
+      created_by: user.id,
+    )
+  end
+
+  def parse_fulfilled_at(fulfilled_at_time)
+    fulfilled_date = parse_usa_date(fulfilled_at_time).try(:to_date)
+    fulfilled_date.beginning_of_day + 12.hours if valid_fulfilled_at?(fulfilled_date)
```
jhanggi: I'm probably fine with it, but if you pass an invalid date, does this mean that it's use `Time.current` rather than come back as telling the user it's invalid?

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -1,7 +1,27 @@
...
+        nil,
+        placeholder: OrderDetail.human_attribute_name(:fulfilled_at),
+        data: { "min-date": SettingsHelper.fiscal_year_beginning.iso8601, "max-date": Date.today.iso8601 },
+        class: "datepicker__data string optional"
+
       %br
-      = submit_tag 'Add To Order', :class => 'btn btn-primary', :id => :product_add_btn
+
+      = label_tag :note, OrderDetail.human_attribute_name(:note)
+      = text_area_tag :note,
```
jhanggi: Can we use simple_form? Then we shouldn't need to decide ourselves what kind of box this should be? If we change it to a `text` column, then it should automatically change, right? And it should pick up the maxlength from the validations.

app/assets/javascripts/app/merge_order_form.js.coffee
drench: It does, but `val()` gives the `order_status_id` value, not the text that I want here.

app/services/order_appender.rb
drench: The issue is that if we get a `fulfilled_at`, that means we are taking the order to "complete" status, which causes the state machine to call `OrderDetail#make_complete` which sets `fulfilled_at` to `Time.zone.now`. We'd have to back that out in the case of an unparseable `fulfilled_at`.

app/views/facility_orders/_merge_order_form.html.haml
drench: The complication is that `note` is on `OrderDetail`, not `Order`. Which makes me think this form should maybe have been a `POST` on `OrderDetail` not a `PUT` on `Order` all along. Worth pursuing now or `TODO`?

app/assets/javascripts/app/merge_order_form.js.coffee
jhanggi: Ok, makes sense.

app/services/order_appender.rb
jhanggi: I think it's probably fine as is if it's going to make things more complicated.

app/views/facility_orders/_merge_order_form.html.haml
jhanggi: Gotcha. I don't think it's worth pursuing changing the HTTP method/location. I haven't looked too closely into the note-expansion PR, but I know I'm going to have some feedback there, so having the textarea here might put us in a weird place temporarily.

app/assets/javascripts/app/merge_order_form.js.coffee
```diff
@@ -0,0 +1,16 @@
...
+        @$fulfilled_at.show()
+      else
+        @$fulfilled_at.hide()
+
+$ ->
+  $(".js--edit-order").each -> new MergeOrderForm($(this))
```
jhanggi: Codeclimate has a couple things to say about `this`.

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -1,7 +1,27 @@
...
+
+      = label_tag :order_status_id, OrderDetail.human_attribute_name(:order_status)
+      = select_tag :order_status_id,
+        options_for_select(OrderStatus.add_to_order_statuses(current_facility).map { |s| [s.name_with_level, s.id] }),
+        class: "js--chosen optional"
+
+      = text_field_tag :fulfilled_at,
+        nil,
+        placeholder: OrderDetail.human_attribute_name(:fulfilled_at),
+        data: { "min-date": SettingsHelper.fiscal_year_beginning.iso8601, "max-date": Date.today.iso8601 },
```
jhanggi: You should be able to use `data: { min_date: xxx, max_date: xxx }` and HAML will convert the underscores into dashes.

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -1,7 +1,27 @@
...
+
       %br
-      = submit_tag 'Add To Order', :class => 'btn btn-primary', :id => :product_add_btn
+
+      = label_tag :note, OrderDetail.human_attribute_name(:note)
+      = text_area_tag :note,
+        nil,
+        class: "string optional wide"
+
+      = submit_tag "Add To Order", class: "btn btn-primary", id: :product_add_btn
```
jhanggi: I18n?

PR #857, opened 2017-01-19
[#133966] Show search fields and message on empty transactions in review search

drench:


app/views/transactions/in_review.html.haml
```diff
@@ -1,12 +1,14 @@
 -# TODO: app/views/accounts/transactions_in_review.html.haml is similar
 
+- if @order_details.blank?
+  - flash.now[:alert] = text("transactions.none_recently_reviewed")
```
jhanggi: Since this is happening in the view, can we use `%p.notice=` instead of setting a flash? I also think the blue`notice` is a little calmer. And can we put it below the filters? That's closer to what we do in other places. And that'll make it clearer that the filters don't apply to recently closed.

kt32291:
I agree with Jason's one comment, otherwise looks ok.

drench:
Here's an updated no-results example:
screen shot 2017-01-23 at 10 15 17 am

I also used this as an opportunity to extract the common transaction in_review parts into a shared partial.

config/locales/en.yml
```diff
@@ -1146,7 +1146,7 @@ en:
 
   transactions:
     in_review: Transactions in Review
-    none: You have no transactions in review, or any recently reviewed.
+    none_recently_reviewed: No transactions in review were found.
```
drench: The `none` message felt redundant after adding the notice.

jhanggi: If this is the message for no transactions currently in review, then the key doesn't feel right.

drench: Agreed. Updating.

jhanggi:
It's a little weird now that if there is nothing in review that the search fields are there, but there's nothing in the chosen dropdowns. Maybe that's fine because the link to that page won't be there in that case. Does that sound about right?

drench:
> if there is nothing in review that the search fields are there, but there's nothing in the chosen dropdowns

This may have been the reasoning behind suppressing the search fields originally.

PR #858, opened 2017-01-20
[#133975] Allow apostrophes in emails

kt32291:
https://pm.tablexi.com/issues/133975

This allows apostrophes in emails, since there are those that have them that need to sign up.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -26,7 +26,7 @@ RSpec.describe User do
   it { is_expected.to be_valid }
 
   context "when user has email address with apostrophe" do
-    subject(:user) { create(:user, email: "o'niel@example.com" ) }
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -25,6 +25,12 @@
 
   it { is_expected.to be_valid }
 
+  context "when user has email address with apostrophe" do
+    subject(:user) { create(:user, email: "o'niel@example.com") }
```
jhanggi: I see it's like this in other tests, but this could probably be `build`, and then we wouldn't need to hit the database for this test.

drench: Agreed on `build`. I prefer `build` for validation tests anyway, since if validation fails, the `create` would blow up before hitting the actual test.

kt32291:
@jhanggi @drench updated

PR #860, opened 2017-01-26
[#125660] Make reservation cancellation idempotent

drench:
When canceling an already-canceled reservation, the controller should behave to a client the same as it did when cancelling initially.

### DIFF
--- a/app/controllers/facility_user_reservations_controller.rb
+++ b/app/controllers/facility_user_reservations_controller.rb
@@ -21,9 +21,9 @@ class FacilityUserReservationsController < ApplicationController
   # PUT /facilities/:facility_id/users/:user_id/reservations/:id/cancel
   def cancel
     order_detail = user_order_details.find(params[:id])
-    raise ActiveRecord::RecordNotFound unless order_detail.reservation.try(:can_cancel?)
+    raise ActiveRecord::RecordNotFound if not_customer_cancelable?(order_detail.reservation)
     order_detail.transaction do
-      if cancel_with_fee(order_detail)
+      if order_detail.reservation.canceled_at? || cancel_with_fee(order_detail)
         flash[:notice] = text("cancel.success")
       else
         flash[:error] = text("cancel.error")
@@ -42,6 +42,10 @@ class FacilityUserReservationsController < ApplicationController
 
   def load_user
     @user = User.find(params[:user_id])
+  end
+
+  def not_customer_cancelable?(reservation)
+    reservation.present? && !reservation.canceled_at? && !reservation.can_cancel?
   end
 
   def user_order_details
--- a/app/controllers/facility_user_reservations_controller.rb
+++ b/app/controllers/facility_user_reservations_controller.rb
@@ -5,6 +5,7 @@ class FacilityUserReservationsController < ApplicationController
   before_action :authenticate_user!
   before_action :check_acting_as
   before_action :load_user
+  before_action :load_order_detail, only: :cancel
 
   load_and_authorize_resource class: "OrderDetail"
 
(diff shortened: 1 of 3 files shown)


app/controllers/facility_user_reservations_controller.rb
```diff
@@ -44,6 +44,10 @@ def load_user
     @user = User.find(params[:user_id])
   end
 
+  def not_customer_cancelable?(reservation)
+    reservation.present? && !reservation.canceled_at? && !reservation.can_cancel?
```
jhanggi: Is there a way we can invert this logic and make it make sense? There's a lot of negation going on.

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -21,9 +21,9 @@ def index
   # PUT /facilities/:facility_id/users/:user_id/reservations/:id/cancel
   def cancel
     order_detail = user_order_details.find(params[:id])
```
jhanggi: Would it be worth changing this parameter to `order_detail_id`?

drench: It used to be `reservation_id` until b87258cd6a608b2c295133a99f4977740db8131a. Updating.

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -18,35 +19,43 @@ def index
...
-    order_detail.transaction do
-      if cancel_with_fee(order_detail)
-        flash[:notice] = text("cancel.success")
-      else
-        flash[:error] = text("cancel.error")
-        raise ActiveRecord::Rollback
+    reservation = @order_detail.reservation
+    raise ActiveRecord::RecordNotFound if reservation.blank?
+
+    unless reservation.canceled_at?
```
jhanggi: What about just `reservation.canceled?`?

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -18,35 +19,39 @@ def index
...
-    order_detail = user_order_details.find(params[:id])
-    raise ActiveRecord::RecordNotFound unless order_detail.reservation.try(:can_cancel?)
-    order_detail.transaction do
-      if cancel_with_fee(order_detail)
-        flash[:notice] = text("cancel.success")
-      else
-        flash[:error] = text("cancel.error")
-        raise ActiveRecord::Rollback
-      end
+    reservation = @order_detail.reservation
```
drench: I refactored, though I don't know if it's as clear as it could be.

The logic I'm going for is:

* If the `OrderDetail` has no `Reservation`, 404
* If the `Reservation` was already canceled, show the success notice and redirect
* If the `Reservation` is not cancelable, 404
* Attempt to cancel the `Reservation`
  - If cancellation succeeds, show the success notice and redirect
  - If cancellation fails, show the failure notice and roll back

app/controllers/facility_user_reservations_controller.rb
```diff
@@ -18,35 +19,39 @@ def index
...
   end
 
   private
 
-  def cancel_with_fee(order_detail)
-    order_detail
-      .cancel_reservation(session_user, admin: true, admin_with_cancel_fee: true)
+  def cancel_with_fee!
+    @order_detail.transaction do
+      unless @order_detail.cancel_reservation(session_user, admin: true, admin_with_cancel_fee: true)
```
jhanggi: Is there anything reasonable that would cause this to fail? Seems like if this fails, then you would get both the success and error flashes.

drench: Possibly, since there's kind of a lot happening in `#cancel_reservation`. I missed `ActiveRecord::Rollback` not short-circuiting it.

app/controllers/facility_user_reservations_controller.rb
drench: I refactored to make it work like:

* If the `OrderDetail` has no `Reservation`, 404
* If the `Reservation` is not canceled,
  - 404 if the `Reservation` can't be canceled
  - Attempt to cancel the `Reservation` and roll back if it fails
* If the `Reservation` is canceled, flash a success notice
* If the `Reservation` is not canceled, flash an error notice
* Redirect to the index

PR #861, opened 2017-01-26
[#134059] More space in notes

kt32291:
https://pm.tablexi.com/issues/134059

This allows more space for order detail notes (no longer restricted to 100 characters, now up to string length of 255). I believe I fixed all views for longer notes, but it's possible I missed some, so if you see anything, just let me know.

### DIFF
--- a/app/assets/javascripts/app/detail_notes.coffee
+++ b/app/assets/javascripts/app/detail_notes.coffee
@@ -1,10 +1,12 @@
 $ ->
-  hiddenText = "See Note"
-  openText = "Hide Note"
+  hiddenText = "More"
+  openText = "Less"
 
   $('.js--toggleNote').on 'click', (e) ->
     e.preventDefault()
-    $(this).next(".js--note").toggle()
+    $parent = $(this).parent()
     text = $(this).text()
+    $parent.find(".js--note").toggle()
+    $parent.find(".js--truncatedNote").toggle()
     $(this).text if text == hiddenText then openText else hiddenText
     return
--- a/app/assets/javascripts/app/detail_notes.coffee
+++ b/app/assets/javascripts/app/detail_notes.coffee
@@ -4,9 +4,9 @@ $ ->
 
   $('.js--toggleNote').on 'click', (e) ->
     e.preventDefault()
-    $parent = $(this).parent()
-    text = $(this).text()
+    $parent = $(@).parent()
+    text = $(@).text()
     $parent.find(".js--note").toggle()
     $parent.find(".js--truncatedNote").toggle()
-    $(this).text if text == hiddenText then openText else hiddenText
+    $(@).text if text == hiddenText then openText else hiddenText
     return
--- a/app/assets/javascripts/app/detail_notes.coffee
+++ b/app/assets/javascripts/app/detail_notes.coffee
@@ -1,0 +1,10 @@
+$ ->
+  hiddenText = "See Note"
+  openText = "Hide Note"
+
+  $('.js--toggleNote').on 'click', (e) ->
+    e.preventDefault()
+    $(this).next(".js--note").toggle()
+    text = $(this).text()
+    $(this).text if text == hiddenText then openText else hiddenText
+    return
--- a/app/views/facility_orders/_merge_order_form.html.haml
+++ b/app/views/facility_orders/_merge_order_form.html.haml
@@ -18,9 +18,8 @@
         class: "datepicker__data string optional"
 
       = label_tag :note, OrderDetail.human_attribute_name(:note)
-      = text_field_tag :note,
+      = text_area_tag :note,
         nil,
-        maxlength: 100,
         class: "string optional wide"
 
       %br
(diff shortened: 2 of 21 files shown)


drench:
I agree we should look at making the column `text`. If we do, Oracle may need special handling in the migration. There's some discussion here: https://asktom.oracle.com/pls/asktom/f?p=100:11:0::::P11_QUESTION_ID:1770086700346491686

The tl;dr is we may need to create a new column of type `CLOB`, copy all of the existing notes into it, delete the old `note` column, and then rename the new `CLOB` column to `note`.

kt32291:
@jhanggi @drench updated

jhanggi:
I'll give another review tomorrow, but in the meantime, master now has a normal text field in `merge_order_form`. That should change back to a textarea in this PR.

config/locales/views/en.order_details.yml
```diff
@@ -5,3 +5,11 @@ en:
...
           Changing the Payment Source could result in adjusted pricing. If you are
           concerned about this or receive incorrect pricing, please contact the !facility_downcase!.
+    shared:
+      order_detail_cell:
+        note:
+          more: More
+      transactions:
+        table_inside:
+          note:
+            more: More
```
drench: Not a blocker, but I could see "More" being in the global shared translations, where we have Create, Cancel, Delete, Save, Search, etc.

jhanggi:
Couple more spots: "My Orders" and "My Reservations". Make sure to check all the tabs.

kt32291:
@jhanggi did you want to add it to views that don't currently show the notes? (My Orders doesn't do it currently)

jhanggi:
I don't want to add notes where we're not currently showing notes. However, My Orders, both Pending and All tabs use the partial `/app/views/shared/_order_details_table.html.haml`, which includes the note. Looks like My Reservations uses a shared view, so we've got that one covered.

And now doing even more testing, we should probably turn the note fields on the cart (need to make sure the note available to user option is enabled for the product) and when placing a reservation into textareas.

app/views/facility_account_orders/index.html.haml
```diff
@@ -31,31 +31,29 @@
...
+            - if order_detail.stored_files_template_result.present?
+              - order_detail.stored_files_template_result.each do |stored_file|
+                .indented
+                  = link_to t(".label.view_order.file"), stored_file.download_url
 
-              - if order_detail.note.present?
-                %li
-                  %i= t(".note", text: order_detail.note)
+            - if order_detail.note.present?
+              = render partial: "shared/order_detail_note", locals: { order_detail: order_detail }
```
jhanggi: I think we're generally moving towards `render "shared/order_detail_note", order_detail: order_detail`, but it's not a huge deal.

app/views/shared/_order_detail_note.html.haml
```diff
@@ -0,0 +1,3 @@
+%i.js--note= "Note: #{order_detail.note}"
```
jhanggi: Worth I18ning?

app/assets/javascripts/app/detail_notes.coffee
```diff
@@ -0,0 +1,12 @@
...
+  openText = "Less"
+
+  $('.js--toggleNote').on 'click', (e) ->
+    e.preventDefault()
+    $parent = $(@).parent()
+    text = $(@).text()
+    $parent.find(".js--note").toggle()
+    $parent.find(".js--truncatedNote").toggle()
+    $(@).text if text == hiddenText then openText else hiddenText
+    return
```
jhanggi: Just curious, what does the bare `return` do in this case vs excluding it (like is common in ruby)?

kt32291:
Updated

kt32291:
@jhanggi this works in oracle - we good to merge?

jhanggi:
There are still two things to address:

> I think this was branched off master before the add to existing order changes were put in place. We either need to rebase so we can swap the input for a textarea there or we can plan to do another PR
> I also think we should widen the text area in the cart.

I'd be fine merging this and doing those in a separate PR, but they need to be done before we can call this ticket ready for stage.

kt32291:
@jhanggi which input needs to be changed?

jhanggi:
`facility_orders/_merge_order_form`. It might not be on this branch. It should pretty much be a revert of #864, which we did because because Dan had had implemented #856 with a textarea, but that went in before this was ready.

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -18,9 +18,8 @@
         class: "datepicker__data string optional"
 
       = label_tag :note, OrderDetail.human_attribute_name(:note)
-      = text_field_tag :note,
+      = text_area_tag :note,
```
jhanggi: I don't think `text_area_tag` automatically adds a `maxlength` the way that simple_form does. We should probably add that here.

kt32291:
@jhanggi updated

app/assets/javascripts/app/detail_notes.coffee
drench: A bare `return` tells the compiler the function should not return anything.

PR #862, opened 2017-01-31
[#125093] Allow Business Admins to use a cross-facility "Users" tab

drench:
https://pm.tablexi.com/issues/125093

This opens the cross-facility "Users" tab to Billing Administrators. It already exists for Account Managers, but opening it to BAs revealed queries that expected a single facility, and partials that weren't checking certain abilities and settings flags. For example, we probably should have been checking the `manage_payment_sources_with_users` setting on the "User" side from the start.

The bulk email feature probably had the most significant changes: it no longer requires a bulk email job to have a facility, and omits the canned "signoff" when in a cross-facility context (since the message normally mentions contacting the facility).

Its controller tests also needed to move away from `controller_spec_helper`; it would have been complicated to make it work with certain products and user roles were associated with a particular facility, while making the requests against the "ALL" facility.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -52,7 +52,7 @@ class Product < ActiveRecord::Base
     where("products.id NOT IN (?)", exclusion_list)
   end
 
-  scope :for_facility, lambda {|facility|
+  scope :for_facility, lambda { |facility|
     return self.class.none if facility.blank?
     return where(facility_id: facility.id) if facility.single_facility?
     return all if facility.cross_facility?
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -53,10 +53,13 @@ class Product < ActiveRecord::Base
   end
 
   scope :for_facility, lambda { |facility|
-    return self.class.none if facility.blank?
-    return where(facility_id: facility.id) if facility.single_facility?
-    return all if facility.cross_facility?
-    self.class.none
+    if facility.blank?
+      self.class.none
+    elsif facility.single_facility?
+      where(facility_id: facility.id)
+    else # cross-facility
+      all
+    end
   }
 
   def self.requiring_approval
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -54,7 +54,7 @@ class Product < ActiveRecord::Base
 
   scope :for_facility, lambda { |facility|
     if facility.blank?
-      self.class.none
+      none
     elsif facility.single_facility?
       where(facility_id: facility.id)
     else # cross-facility
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -52,6 +52,13 @@ class Product < ActiveRecord::Base
     where("products.id NOT IN (?)", exclusion_list)
   end
 
+  scope :for_facility, lambda {|facility|
+    return self.class.none if facility.blank?
+    return where(facility_id: facility.id) if facility.single_facility?
+    return all if facility.cross_facility?
+    self.class.none
+  }
+
   def self.requiring_approval
     where(requires_approval: true)
   end
--- a/app/views/facility_account_orders/index.html.haml
+++ b/app/views/facility_account_orders/index.html.haml
@@ -32,16 +32,14 @@
           %td= order_detail.ordered_at
           %td
             %ul
+              - reservation = order_detail.reservation
               %li= order_detail.description_as_html
-              - if order_detail.reservation.present?
-                - if order_detail.reservation_admin_editable?
-                  %li.indented
-                    = link_to order_detail.reservation,
-                      order_detail.edit_reservation_path
-                - else
-                  %li.indented
-                    = link_to order_detail.reservation,
-                      order_detail.show_reservation_path
+              - if reservation.present?
+                %li.indented
+                  - if order_detail.reservation_admin_editable? && current_ability.can?(:edit, reservation)
+                    = link_to reservation, order_detail.edit_reservation_path
+                  - else
+                    = reservation
 
               - if order_detail.survey_completed?
                 %li.indented
--- a/app/views/facility_account_orders/index.html.haml
+++ b/app/views/facility_account_orders/index.html.haml
@@ -24,25 +24,29 @@
         %th.currency= t(".th.amount")
 
     %tbody
-      - OrderDetailPresenter.wrap(@order_details).each do |order_detail|
-        %tr{class: order_detail.row_class}
+      - @order_details.each do |order_detail|
+        - od = OrderDetailPresenter.new(order_detail)
+        %tr{class: od.row_class}
           %td
-            = link_to order_detail.description, order_detail.show_order_path
+            - if current_ability.can?(:show, order_detail)
+              = link_to od.description, od.show_order_path
+            - else
+              = od.description
 
-          %td= order_detail.ordered_at
+          %td= od.ordered_at
 
           = render "shared/order_detail_cell",
-            od: order_detail,
-            show_reservation: order_detail.reservation.present?
+            od: od,
+            show_reservation: od.reservation.present?
 
-          %td= order_detail.order_status
+          %td= od.order_status
           %td.currency
-            - if order_detail.cost_estimated?
-              %span.estimated_cost= order_detail.estimated_total
-            - elsif order_detail.price_policy.blank?
+            - if od.cost_estimated?
+              %span.estimated_cost= od.estimated_total
+            - elsif od.price_policy.blank?
               = t(".unassigned")
             - else
-              %span.actual_cost= order_detail.actual_total
+              %span.actual_cost= od.actual_total
       %tr
         %td.centered{colspan: 5}= will_paginate(@order_details)
 
--- a/app/views/facility_account_orders/index.html.haml
+++ b/app/views/facility_account_orders/index.html.haml
@@ -30,30 +30,10 @@
             = link_to order_detail.description, order_detail.show_order_path
 
           %td= order_detail.ordered_at
-          %td
-            %ul
-              - reservation = order_detail.reservation
-              %li= order_detail.description_as_html
-              - if reservation.present?
-                %li.indented
-                  - if order_detail.reservation_admin_editable? && current_ability.can?(:edit, reservation)
-                    = link_to reservation, order_detail.edit_reservation_path
-                  - else
-                    = reservation
 
-              - if order_detail.survey_completed?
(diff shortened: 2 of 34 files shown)


app/views/facility_user_reservations/index.html.haml
```diff
@@ -41,8 +41,11 @@
           %td= order_detail.reservation
 
           %td.user-order-detail.order-detail-description
-            = link_to order_detail.description_as_html_with_facility_prefix,
-              facility_instrument_path(order_detail.facility, order_detail.product)
+            - if current_facility.try(:single_facility?)
```
drench: We probably should revisit this: the facility user orders partial just shows a text description with no link. The orders partial is shared in user and facility facing views though, where the reservation equivalents are not shared.

app/models/product.rb
```diff
@@ -52,6 +52,13 @@ def self.exclude(exclusion_list)
     where("products.id NOT IN (?)", exclusion_list)
   end
 
+  scope :for_facility, lambda { |facility|
+    return self.class.none if facility.blank?
+    return where(facility_id: facility.id) if facility.single_facility?
+    return all if facility.cross_facility?
+    self.class.none
```
jhanggi: What's the scenario where it falls back to this?

vendor/engines/bulk_email/spec/controllers/bulk_email/bulk_email_controller_spec.rb
```diff
@@ -143,58 +140,53 @@
...
-        }.to_json,
-      }
-      @params[:return_path] = return_path if return_path.present?
-
-      sign_in @admin
-      do_request
+    let(:user) { FactoryGirl.create(:user, :senior_staff, facility: facility) }
+
+    let(:params) do
+      super().merge(bulk_email_delivery_form: {
```
jhanggi: Do you need the parentheses in `super()`? Also, I didn't know you could do that inside a `let`.

drench: It's something I've used a bit before (in 1b6955a931) that I picked up here: http://stackoverflow.com/a/22368792

The parens unfortunately need to be there, otherwise it raises a `RuntimeError`:

`implicit argument passing of super from method defined by define_method() is not supported. Specify all arguments explicitly`

app/models/product.rb
drench: It should never hit it, but it felt wrong to have a scope potentially return nothing (which I think translates to an implicit `all`).

drench:
After some discussion, these are the TODOs for BAs:
- [ ] Disable product access list CRUD
- [ ] Disable adding products to an existing order
- [ ] Disable facility (admin) order and reservations tabs

TBD: Disable bulk email for BAs?

app/models/product.rb
```diff
@@ -52,6 +52,16 @@ def self.exclude(exclusion_list)
     where("products.id NOT IN (?)", exclusion_list)
   end
 
+  scope :for_facility, lambda { |facility|
+    if facility.blank?
+      self.class.none
```
jhanggi: Minor, but do we need the `self.class` part of this?

app/views/shared/_order_details_table.html.haml
```diff
@@ -1,37 +1,34 @@
...
-            - unless od.note.blank?
+            %li= order_detail.description_as_html_with_facility_prefix
+            - if order_detail.note.present?
               %li
-                %i=h "Note: #{od.note}"
-        %td.centered=h od.order_status.try(:name)
-        %td.currency
-          - od.send(:extend, PriceDisplayment)
-          = od.wrapped_total
+                %i= "Note: #{order_detail.note}"
```
jhanggi: Is this going to conflict with #861?

app/views/users/index.html.haml
```diff
@@ -10,10 +11,13 @@
 #result
   %h3= text(".head")
   %p= text(".instruct")
+
   - if @users.any?
-    = render "search/results_table", order_for: true, users: @users
+    = render "search/results_table",
+      order_for: current_facility.try(:single_facility?),
```
jhanggi: Do we have an ability that we could use? Maybe `:switch_to, User`?

db/schema.rb
```diff
@@ -65,7 +65,7 @@
   end
 
   create_table "bulk_email_jobs", force: :cascade do |t|
-    t.integer  "facility_id",     limit: 4,     null: false
+    t.integer  "facility_id",     limit: 4
```
jhanggi: Since we disabled bulk email, should we bring back the `null: false`?

app/views/users/index.html.haml
jhanggi: If you do a search, it's loading the `results_table` with the "Order For" link.

app/views/shared/_order_details_table.html.haml
drench: It looks like #861 doesn't touch this partial, but maybe it should?

app/views/users/index.html.haml
drench: The `@users` here are `UserPresenter`-wrapped. I'm going to merge in #865 and retest.

drench: Ah: Billing Administrators can `:manage` `User` cross-facility. I'll put in an exception for `switch_to`. What about `create`?

db/schema.rb
drench: I'm inclined to leave it to keep the door open for cross-facility (or no facility?) bulk email support in the future.

app/views/users/index.html.haml
drench: One note on user search: I ran into some unexpected (to me) 304s on search results while testing this. The ajax GETs send (and honor) etags.

app/views/shared/_order_details_table.html.haml
kt32291: Noted, it probably should!

app/models/product.rb
```diff
@@ -52,6 +52,16 @@ def self.exclude(exclusion_list)
...
 
+  scope :for_facility, lambda { |facility|
+    if facility.blank?
+      none
+    elsif facility.single_facility?
+      where(facility_id: facility.id)
+    else # cross-facility
+      all
+    end
+  }
```
kt32291: 👍

app/views/search/_results_table.html.haml
```diff
@@ -1,21 +1,24 @@
-- order_for = local_assigns[:order_for]
+- search_type = @search_type.presence || "manage_user"
+- can_switch_any = users.any? { |user| current_ability.can?(:switch_to, user) }
+- show_order_for_links = search_type == "manage_user" && can_switch_any
```
kt32291: Do these variables need to be set in the view?

app/views/users/access_list.html.haml
```diff
@@ -1,18 +1,19 @@
...
-  = render partial: 'admin/shared/sidenav_users', locals: { sidenav_tab: 'users' }
+  = render "admin/shared/sidenav_users", sidenav_tab: "users"
 
 = content_for :tabnav do
-  = render partial: 'admin/shared/tabnav_users', locals: { secondary_tab: 'access_list' }
+  = render "admin/shared/tabnav_users", secondary_tab: "access_list"
 
 %h1
-  = t(:head, scope: 'users.access_list', name: @user.full_name)
+  = t(".head", name: @user.full_name)
```
kt32291: ```text("head", name: @user.full_name)```

app/views/users/access_list.html.haml
```diff
@@ -1,18 +1,19 @@
...
   %small= mail_to @user.email
 
 - if @products_by_type.blank?
-  %p.notice= t('.no_products', name: @user.full_name)
+  %p.notice= t(".no_products", name: @user.full_name)
```
kt32291: Same comment

lib/ability.rb
```diff
@@ -56,8 +56,12 @@ def initialize(user, resource, controller = nil)
     end
 
     if user.billing_administrator?
-      can :manage, [Account, Journal, Order, OrderDetail, Reservation]
-      cannot :administer, [Order, OrderDetail, Reservation]
+      can :manage, [Account, Journal, OrderDetail]
+      can [:send_receipt, :show], Order
+      if resource == Facility.cross_facility
+        can :manage, [Reservation, User]
```
jhanggi: Would it make more sense to make this can be a whitelist of only what they _can_ do rather than `:manage` and remove abilities?

jhanggi:
Sorry for the terseness of the last comment, I got pulled away, but wanted to get the general idea out there.

I noticed a 403 on "Add Access" to a payment source. Looks like that was the same in master.

Payment Source > Orders clicking a reservation link 403s where it didn't before. It probably makes sense to take away that ability because that view only allows you to make reservations in the future (you get a "The reservation must start at a future time" error message if you try to change the time).

lib/ability.rb
```diff
@@ -60,6 +60,7 @@ def initialize(user, resource, controller = nil)
       can [:send_receipt, :show], Order
       if resource == Facility.cross_facility
         can :manage, [Reservation, User]
+        cannot [:edit, :update], Reservation
```
jhanggi: jhanggi 8 minutes ago Member Pending
I'm not really loving this tendency towards using cannot. I think I'd rather whitelist the actions the user can do (so can [:read, :create], Reservation (not sure if there are other ones we need to add). It's probably a little more testing, but if we have decent ability specs, it should be fine.

app/views/facility_account_orders/index.html.haml
```diff
@@ -32,16 +32,14 @@
...
-              - if order_detail.reservation.present?
-                - if order_detail.reservation_admin_editable?
-                  %li.indented
-                    = link_to order_detail.reservation,
-                      order_detail.edit_reservation_path
-                - else
-                  %li.indented
-                    = link_to order_detail.reservation,
-                      order_detail.show_reservation_path
+              - if reservation.present?
```
jhanggi: What would you think of replacing this whole chunk with the `order_detail_cell` partial? It would address the link problem as well as the extra list item dots.

lib/ability.rb
drench: Good call. It's looking like a BA may not need any explicit `Reservation` abilities, since the `FacilityUserReservationsController` authorizes against `OrderDetail`. Updating.

PR #863, opened 2017-02-01
[#133561] Handle attempted reservation creation for deleted cart items

drench:
If the user attempts to create a reservation for an order detail that has been removed from the cart (possible if using multiple tabs or windows), this redirects the user back to the facility landing page with a message to try creating the reservation again from the beginning.

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -95,7 +95,7 @@ class ReservationsController < ApplicationController
     flash.now[:notice] = existing_notices.concat(notices).join("<br />").html_safe unless notices.empty?
   end
 
-  # POST /orders/1/order_details/1/reservations
+  # POST /orders/:order_id/order_details/:order_detail_id/reservations
   def create
     raise ActiveRecord::RecordNotFound unless @reservation.nil?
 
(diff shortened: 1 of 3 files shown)


app/controllers/reservations_controller.rb
```diff
@@ -282,9 +283,14 @@ def load_basic_resources
...
-    @facility = @instrument.facility
+    if @order_detail.present?
+      @reservation = @order_detail.reservation
+      @instrument = @order_detail.product
+      @facility = @instrument.facility
+    else
+      flash[:error] = text("order_detail_removed")
+      return redirect_to facility_path(@order.facility)
+    end
     nil
```
jhanggi: Is there a reason this need to return `nil`? If we get rid of that, do we need the explicit `return` above?

app/controllers/reservations_controller.rb
```diff
@@ -282,9 +283,14 @@ def load_basic_resources
     @order = Order.find(params[:order_id])
     # It's important that the order_detail be the same object as the one in @order.order_details.first
     @order_detail = @order.order_details.find { |od| od.id.to_i == params[:order_detail_id].to_i }
```
jhanggi: This seems weird. Do you know why we're doing this weird find rather than a `find(params[:order_detail_id]` or a `find_by(id: params[:order_detail_id])`?

drench: I don't. The change that brought it in is here: 4ad3a411820fd29a364745940cce22023af92623
Tests still look good for me with `@order.order_details.find(params[:order_detail_id])` so I'm going ahead with it.

drench: I take that back: `find` of course raises a 404, which we don't want here.

drench: I cleaned this up a bit, but using `find` or `where(id: …)` caused spec failures. The `find`-with-block remains.

jhanggi: Did you try `find_by(id: ...)`? What sort of test failures?

jhanggi: If you `rescue ActiveRecord::RecordNotFound`, then wouldn't the behavior of `find` be exactly what you want?

PR #865, opened 2017-02-02
[#133668] Upgrade the cancancan gem to v1.15.0

drench:
There's a change in how cancancan handles objects decorated with `SimpleDelegator`, in that it no longer does. This adds `UserPresenter` to `lib/ability.rb` along side the `User` `can`/`cannot` calls for this reason.

More discussion here: https://github.com/CanCanCommunity/cancancan/issues/255

### DIFF
--- a/lib/ability.rb
+++ b/lib/ability.rb
@@ -24,11 +24,11 @@ class Ability
           cannot [:manage_accounts, :manage_billing, :manage_users], Facility.cross_facility
         end
         unless user.account_manager?
-          cannot :manage, User unless resource.is_a?(Facility) && resource.single_facility?
+          cannot :manage, [User, UserPresenter] unless resource.is_a?(Facility) && resource.single_facility?
         end
       end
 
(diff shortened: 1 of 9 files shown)


lib/ability.rb
```diff
@@ -24,11 +24,11 @@ def initialize(user, resource, controller = nil)
           cannot [:manage_accounts, :manage_billing, :manage_users], Facility.cross_facility
         end
         unless user.account_manager?
-          cannot :manage, User unless resource.is_a?(Facility) && resource.single_facility?
+          cannot :manage, [User, UserPresenter] unless resource.is_a?(Facility) && resource.single_facility?
```
jhanggi: I don't love this solution. I don't expect us to remember to always add `UserPresenter` whenever we add `User` to an ability.

It looks like we're only using `UserPresenter` in the context of global roles. Could we wrap the users within the views where necessary as opposed to the controller level so most of the time it's a plain `User` object, but when we need the extra presentation logic we know we're not in the context of checking abilities?

PR #866, opened 2017-02-03
Facility account view updates

drench:
This should fix a missing translation on `.notice` I found in testing, and also adds whitespace between the account number and "(inactive)" when the recharge chart string is not active.

### DIFF
--- a/app/views/facility_facility_accounts/index.html.haml
+++ b/app/views/facility_facility_accounts/index.html.haml
@@ -12,7 +12,7 @@
   %li= link_to t("admin.shared.add", model: FacilityAccount.model_name.human), new_facility_facility_account_path, :class => "btn-add"
 
 - if @accounts.empty?
-  %p.notice= t(".notice")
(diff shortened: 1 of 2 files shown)


app/views/facility_facility_accounts/index.html.haml
```diff
@@ -9,18 +9,26 @@
...
 
 %ul.inline
-  %li= link_to t("admin.shared.add", model: FacilityAccount.model_name.human), new_facility_facility_account_path, :class => "btn-add"
+  %li
+    = link_to text("admin.shared.add", model: FacilityAccount.model_name.human),
+      new_facility_facility_account_path, class: "btn-add"
 
 - if @accounts.empty?
-  %p.notice= t(".notice")
+  %p.notice= text(".notice")
```
jhanggi: Since this is now `text`, do we need the leading period? Same note for below in `.inactive`.

kt32291:
👍

PR #867, opened 2017-02-05
[#134051] Admin edit users

kt32291:
https://pm.tablexi.com/issues/134051

This adds functionality for facility operators updating users' email (same as username), first name, and last name, and non-admin users cannot access it.  They can only edit external users.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -142,11 +142,21 @@ class UsersController < ApplicationController
   def email
   end
 
+  # GET /facilities/:facility_id/users/:id/edit
   def edit
     @user = User.find(params[:id])
   end
 
+  # PUT /facilities/:facility_id/users/:id
   def update
+    @user = User.find(params[:id])
+    if @user.update_attributes(params[:user])
+      flash[:notice] = text("update.success")
+      redirect_to facility_user_path(current_facility, @user)
+    else
+      flash[:error] = text("update.error", message: @user.errors.full_messages.to_sentence)
+      render action: "edit"
+    end
   end
 
   private
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -142,6 +142,13 @@ class UsersController < ApplicationController
   def email
   end
 
+  def edit
+    @user = User.find(params[:id])
+  end
+
+  def update
+  end
+
   private
 
   def update_access_list_approvals
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -18,7 +18,8 @@ class UsersController < ApplicationController
   before_action :authenticate_user!, except: [:password_reset]
   before_action :check_acting_as
 
-  load_and_authorize_resource except: [:password, :password_reset], id_param: :user_id
+  load_and_authorize_resource except: [:password, :password_reset, :edit, :update], id_param: :user_id
+  load_and_authorize_resource only: [:edit, :update], id_param: :id
 
   layout "two_column"
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -151,7 +151,7 @@ class UsersController < ApplicationController
   # PUT /facilities/:facility_id/users/:id
   def update
     @user = User.find(params[:id])
-    if @user.update_attributes(params[:user])
+    if @user.update_attributes(edit_user_params)
       flash[:notice] = text("update.success")
       redirect_to facility_user_path(current_facility, @user)
     else
@@ -161,6 +161,10 @@ class UsersController < ApplicationController
   end
 
   private
+
+  def edit_user_params
+    params.require(:user).permit(:email, :first_name, :last_name)
+  end
 
   def update_access_list_approvals
     if update_approvals.grants_changed?
--- a/app/views/users/_form.html.haml
+++ b/app/views/users/_form.html.haml
@@ -1,4 +1,4 @@
-= simple_form_for(@user, url: url) do |f|
+= simple_form_for [current_facility, @user] do |f|
   .form-inputs
     = f.input :first_name
     = f.input :last_name
--- a/app/views/users/_form.html.haml
+++ b/app/views/users/_form.html.haml
@@ -1,0 +1,9 @@
+= simple_form_for(@user, url: url) do |f|
+  .form-inputs
+    = f.input :first_name
+    = f.input :last_name
+    = f.input :email
+    = f.input :username, readonly: true
+    = f.button :submit, button_text, class: ["btn", "btn-primary"]
+    &nbsp;
+    = link_to text("shared.cancel"), facility_users_path
--- a/app/views/users/edit.html.haml
+++ b/app/views/users/edit.html.haml
@@ -9,4 +9,4 @@
 
 %h2= text("head", user: @user.full_name)
 %p= text("subhead")
-= render "form", url: facility_user_path(current_facility, @user), button_text: text("shared.update")
+= render "form", button_text: text("shared.update")
--- a/app/views/users/edit.html.haml
+++ b/app/views/users/edit.html.haml
@@ -1,29 +1,12 @@
 = content_for :head_content do
-  :javascript
-    $(document).ready(function() {
-      $("input#user_email").keyup(function(e) {
-        $("input#user_username").val(this.value);
-      });
-
-      $("input#user_email").change(function(e) {
-        $("input#user_username").val(this.value);
-      });
-    });
+  = render "form_head"
 
 = content_for :h1 do
-  =h current_facility
+  = current_facility
 
 = content_for :sidebar do
   = render "admin/shared/sidenav_users", sidenav_tab: "users"
 
 %h2= text("head", user: @user.full_name)
 %p= text("subhead")
-= simple_form_for(@user, url: facility_user_path(current_facility, @user), method: :put) do |f|
-  .form-inputs
-    = f.input :first_name
-    = f.input :last_name
-    = f.input :email
-    = f.input :username, :readonly => true
-    = f.button :submit, text("update"), class: ["btn", "btn-primary"]
-    &nbsp;
-    = link_to text("shared.cancel"), facility_users_path
+= render "form", url: facility_user_path(current_facility, @user), button_text: text("shared.update")
--- a/app/views/users/edit.html.haml
+++ b/app/views/users/edit.html.haml
@@ -1,0 +1,29 @@
+= content_for :head_content do
+  :javascript
+    $(document).ready(function() {
+      $("input#user_email").keyup(function(e) {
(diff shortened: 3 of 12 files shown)


lib/ability.rb
```diff
@@ -19,6 +19,7 @@ def initialize(user, resource, controller = nil)
         can :manage, UserPriceGroupMember if resource.admin_editable?
         can :manage, AccountPriceGroupMember
       else
+        can([:edit, :update], User) { |target_user| target_user.external? }
```
drench: You're on the right track, but I think you may need to flip this into a `cannot`, and move it after the `can :manage, :all` below. But I'd start with adding tests for these ability in `spec/lib/ability_spec.rb`

kt32291: That's what I did originally, and it wasn't working :( but I will add some tests at least, thanks!

kt32291:
@drench updated

app/controllers/users_controller.rb
```diff
@@ -142,6 +143,23 @@ def access_list_approvals
...
 
+  # GET /facilities/:facility_id/users/:id/edit
+  def edit
+    @user = User.find(params[:id])
+  end
+
+  # PUT /facilities/:facility_id/users/:id
+  def update
+    @user = User.find(params[:id])
+    if @user.update_attributes(params[:user])
```
drench: The code climate warning is legit: if I send a `username` param, it gets updated.

jhanggi: I fully support adding strong params as part of this. I'm pretty sure the only other place that a `User` gets updated via a form would be password change.

app/views/users/edit.html.haml
```diff
@@ -0,0 +1,29 @@
...
+        $("input#user_username").val(this.value);
+      });
+
+      $("input#user_email").change(function(e) {
+        $("input#user_username").val(this.value);
+      });
+    });
+
+= content_for :h1 do
+  =h current_facility
```
jhanggi: We shouldn't be using the `h` anymore. That's a pre-rails 3 vestige that should go away wherever we find it.

app/views/users/edit.html.haml
```diff
@@ -0,0 +1,29 @@
...
+  = render "admin/shared/sidenav_users", sidenav_tab: "users"
+
+%h2= text("head", user: @user.full_name)
+%p= text("subhead")
+= simple_form_for(@user, url: facility_user_path(current_facility, @user), method: :put) do |f|
+  .form-inputs
+    = f.input :first_name
+    = f.input :last_name
+    = f.input :email
+    = f.input :username, :readonly => true
```
jhanggi: Use 1.9 hash syntax.

config/routes.rb
```diff
@@ -148,7 +148,7 @@
         post "access_list/approvals", to: 'users#access_list_approvals'
       end
     else
-      resources :users, except: [:edit, :update, :new, :create], constraints: { id: /\d+/ } do
+      resources :users, except: [:new, :create], constraints: { id: /\d+/ } do
```
jhanggi: I think we want to keep the `:edit, :update` here. This is the case of the `create_users` feature being turned off, which I believe is only switched off for UIC. UIC should not be able to edit/update their users because the users are managed by an external system.

lib/ability.rb
```diff
@@ -20,11 +20,13 @@ def initialize(user, resource, controller = nil)
         can :manage, AccountPriceGroupMember
       else
         can :manage, :all
+        cannot([:edit, :update], User) { |target_user| !target_user.external? }
```
jhanggi: Would it make sense to make an `internal?` method that's just `!external?`? 

Also, part of me wonders if it wouldn't be better to make a new message like `User#admin_updatable?` or `details_updatable?` wrapping `external?` (similar to the `User#password_updatable?` method). We're likely going to be changing the definition of internal vs external users in regards to pricing for NU soon, so it would be nice to not have these concepts be quite so conflated (they already are a little bit, but we should tease them apart where possible).

lib/ability.rb
```diff
@@ -126,6 +128,8 @@ def initialize(user, resource, controller = nil)
         end
 
         can [:administer], User
+        can([:edit, :update], User) { |target_user| target_user.external? }
```
jhanggi: I think code climate is suggesting this line looks like `can [:edit, :update], User, &external?`. I'm not sure if I like that any better or worse. Up to you if you want to ignore this warning.

app/views/users/edit.html.haml
```diff
@@ -0,0 +1,29 @@
+= content_for :head_content do
```
jhanggi: Looks like this view was created by copy & pasting `new_external`. Can we DRY it up some? I think the JS should definitely be shared. I'll leave it up to you to to see where to draw the line for the rest of it.

app/views/users/edit.html.haml
```diff
@@ -0,0 +1,29 @@
...
+
+= content_for :h1 do
+  =h current_facility
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_users", sidenav_tab: "users"
+
+%h2= text("head", user: @user.full_name)
+%p= text("subhead")
+= simple_form_for(@user, url: facility_user_path(current_facility, @user), method: :put) do |f|
```
jhanggi: Rails is actually smart enough to determine the `method` based on whether the `@user` is persisted or not. That also means that if you take out the `method` argument and copy it into `new_external` that form should work.

lib/ability.rb
```diff
@@ -126,6 +128,8 @@ def initialize(user, resource, controller = nil)
         end
 
         can [:administer], User
+        can([:edit, :update], User, &:external?)
```
jhanggi: This is an open question we never really discussed. Who should be allowed to edit name/emails? I'm leaning towards only global admins. Otherwise, I think it opens the possibility of someone messing something up inadvertently. If we do want to open it up to more people, some sort of audit trail would be a good idea.

kt32291: Ok, let's just keep it global admins for now

app/views/users/show.html.haml
```diff
@@ -14,3 +14,7 @@
   = f.input :email
   = f.input :last_sign_in_at, :default_value => 'none'
   = f.input :deactivated_at unless f.object.active?
+
+- if can? :edit, @user
+  %ul.inline
+    %li= link_to text("shared.edit"), edit_facility_user_path(current_facility, @user), :class => 'btn'
```
jhanggi: Use 1.9 syntax and double quotes for `class: "btn"`

lib/ability.rb
```diff
@@ -20,11 +20,13 @@ def initialize(user, resource, controller = nil)
         can :manage, AccountPriceGroupMember
       else
         can :manage, :all
+        cannot([:edit, :update], User) { |target_user| !target_user.admin_editable? }
         unless user.billing_administrator?
           cannot [:manage_accounts, :manage_billing, :manage_users], Facility.cross_facility
         end
         unless user.account_manager?
           cannot :manage, User unless resource.is_a?(Facility) && resource.single_facility?
+          cannot([:edit, :update], User) { |target_user| !target_user.admin_editable? }
```
jhanggi: Do we actually need this since we have it above? It's possible we do because of the `cannot :manage, User` in the previous line, but I just wanted to double check.

app/views/users/_form.html.haml
```diff
@@ -0,0 +1,9 @@
+= simple_form_for(@user, url: url) do |f|
```
jhanggi: If you want, you can call `simple_form_for [current_facility, @user] do |f|` and it'll be smart enough to point you to the right endpoint in both the new and edit cases. Then you don't need to pass in the URL.

lib/ability.rb
kt32291: Yeah we do, at least my tests were failing without it.

spec/lib/ability_spec.rb
```diff
@@ -128,6 +138,8 @@
       it { is_expected.to be_allowed_to(:manage_billing, facility) }
       it { is_expected.to be_allowed_to(:batch_update, Order) }
       it { is_expected.to be_allowed_to(:batch_update, Reservation) }
+
+      it_behaves_like "it can edit users"
```
jhanggi: We should make sure we have some negative test cases in the non-global roles.

kt32291:
@jhanggi @drench updated

app/controllers/users_controller.rb
```diff
@@ -142,8 +143,29 @@ def access_list_approvals
...
 
+  # GET /facilities/:facility_id/users/:id/edit
+  def edit
+    @user = User.find(params[:id])
+  end
+
+  # PUT /facilities/:facility_id/users/:id
+  def update
+    @user = User.find(params[:id])
+    if @user.update_attributes(edit_user_params)
```
jhanggi: CodeClimate is flagging this as unprotected mass assignment. It looks like we're using `require`/`permit` as expected. I think it's because we don't have `include ActiveModel::ForbiddenAttributesProtection` for the user model. 

I'd like to get that included as part of this story, but I'm not sure how broad an impact it would have (it would likely affect the other actions in this controller, but I don't know where else it might have effects).

Do you think you could do a quick estimate of how much work you think this would take? Then we can decide if it's worth pursuing. If we decide to do it, I would be comfortable doing that in a separate PR.

lib/ability.rb
```diff
@@ -126,8 +128,10 @@ def initialize(user, resource, controller = nil)
         end
 
         can [:administer], User
+
         if controller.is_a?(UsersController) || controller.is_a?(SearchController)
           can :manage, User
+          cannot([:edit, :update], User)
```
jhanggi: If we do pursue adding strong params, I'd like to consider changing this and the following permissions (the ones that completely prevent edit/update rather than preventing editing internal users) to use a whitelist instead of `manage`.

spec/lib/ability_spec.rb
```diff
@@ -195,6 +209,7 @@
     it_is_allowed_to([:batch_update, :cancel, :index], Reservation)
     it { is_expected.to be_allowed_to(:administer, User) }
     it { is_expected.to be_allowed_to(:manage, PriceGroup) }
+    it_is_not_allowed_to([:edit, :update], FactoryGirl.create(:user))
```
jhanggi: Does this one need to be a `create`, or could it be the class`User`

app/controllers/users_controller.rb
kt32291: I can add it and see how many tests it breaks but I didn't even think you had to explicitly include it?

jhanggi: Go ahead and merge this and we can figure that out separately.

jhanggi: Usually, strong params is enabled by default in new applications, but since this is an older app, we have it turned off by default and add it in where we can easily.

PR #868, opened 2017-02-07
[#133442] Notify users when assigned to orders

drench:
This adds a new notification, triggered when a user is assigned to one or more orders. This can occur one-at-a-time in the `OrderManagement::OrderDetailsController` through the order detail modal, or in a batch through the `FacilityOrdersController` and `FacilityReservationsController`.

All methods invoke the same mailer, which assumes the order details for each notification are all from the same facility and all assigned to the same user.

There is a mailer preview at `/rails/mailers/order_assignment_mailer/order_assignment` demonstrating one order assignment.

https://pm.tablexi.com/issues/133442

### DIFF
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -172,10 +172,6 @@ class FacilityReservationsController < ApplicationController
 
     msg_hash = batch_updater.update!
 
-    OrderAssignmentMailer
-      .notify_assigned_user(batch_updater.newly_assigned_order_details)
-      .deliver_later
-
     # add flash messages if necessary
     flash.merge!(msg_hash) if msg_hash
   end
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -172,6 +172,10 @@ class FacilityReservationsController < ApplicationController
 
     msg_hash = batch_updater.update!
 
+    OrderAssignmentMailer
+      .notify_assigned_user(batch_updater.newly_assigned_order_details)
+      .deliver_later
+
     # add flash messages if necessary
     flash.merge!(msg_hash) if msg_hash
   end
--- a/app/mailers/order_assignment_mailer.rb
+++ b/app/mailers/order_assignment_mailer.rb
@@ -1,9 +1,9 @@
 class OrderAssignmentMailer < BaseMailer
 
   def notify_assigned_user(order_details)
-    return if order_details.blank?
-
     @order_details = Array(order_details)
+    return if @order_details.blank?
+
     @user = @order_details.first.assigned_user
 
     mail(to: @user.email, subject: text("notify_assigned_user.subject"))
--- a/app/mailers/order_assignment_mailer.rb
+++ b/app/mailers/order_assignment_mailer.rb
@@ -6,13 +6,13 @@ class OrderAssignmentMailer < BaseMailer
     @order_details = Array(order_details)
     @user = @order_details.first.assigned_user
 
-    mail(to: @user.email, subject: subject)
+    mail(to: @user.email, subject: text("notify_assigned_user.subject"))
   end
 
-  private
+  protected
 
-  def subject
-    text("views.order_assignment_mailer.notify_assigned_user.subject")
+  def translation_scope
+    "views.order_assignment_mailer"
   end
 
 end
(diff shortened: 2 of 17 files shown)


app/mailers/order_assignment_mailer.rb
```diff
@@ -0,0 +1,18 @@
+class OrderAssignmentMailer < BaseMailer
+
+  def notify_assigned_user(order_details)
```
jhanggi: Is this using GlobalID? Or should we pass the IDs in?

app/mailers/order_assignment_mailer.rb
```diff
@@ -0,0 +1,18 @@
...
+    return if order_details.blank?
+
+    @order_details = Array(order_details)
```
jhanggi: Does it make sense to do this as the first line? `[].blank?` => true

app/mailers/order_assignment_mailer.rb
```diff
@@ -0,0 +1,18 @@
...
+    @user = @order_details.first.assigned_user
+
+    mail(to: @user.email, subject: subject)
+  end
+
+  private
+
+  def subject
+    text("views.order_assignment_mailer.notify_assigned_user.subject")
```
jhanggi: Can we add something like a `translation_scope` method to make this shorter?

app/services/order_detail_batch_updater.rb
```diff
@@ -121,6 +124,9 @@ def update_order_details_from_params
   def update_all(attribute, value)
     order_details.each do |order_detail|
       order_detail.public_send("#{attribute}=", value)
+      if order_detail.assigned_user_id_changed? && order_detail.assigned_user_id.present?
```
jhanggi: We could conceivably pull this up a level or two and do something like`order_details.select { |od| od.assigned_user_id_changed? && od.assigned_user_id.present? }`. Maybe to the `update!` method right before the `all?(&:save!)`? This seems a little buried to me.

app/controllers/facility_reservations_controller.rb
```diff
@@ -172,6 +172,10 @@ def batch_update
 
     msg_hash = batch_updater.update!
 
+    OrderAssignmentMailer
```
jhanggi: I'm wondering if it would make sense to pull this into the updaters rather than the controllers. Then it just takes care of everything rather than the controller needing to deal with `newly_assigned_order_details`

app/mailers/order_assignment_mailer.rb
drench: It's using GlobalID.

config/settings.yml
```diff
@@ -97,6 +97,7 @@ feature:
   netids_on: true
   limit_short_description_on: true
   manage_payment_sources_with_users_on: true
+  order_assignment_notifications_on: true
```
drench: I went with default `true`; NU and DC will need to have this setting set to `false` on merge.

jhanggi: Sounds good.

app/mailers/order_assignment_mailer.rb
```diff
@@ -0,0 +1,19 @@
+class OrderAssignmentMailer < BaseMailer
+
+  def notify_assigned_user(order_details)
+    @order_details = Array(order_details)
+    return if @order_details.blank?
+
+    @user = @order_details.first.assigned_user
+    return if @user.blank?
```
jhanggi: How would this happen? I could see if the job was delayed a while and the order detail had gotten unassigned in the meantime.

spec/mailers/order_assignment_mailer_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      let(:order_detail) { stubbed_order_detail(order) }
+
+      before { described_class.notify_assigned_user(order_detail).deliver_now }
+
+      it "generates an order assignment notification", :aggregate_failures do
+        expect(email.to).to eq [assigned_user.email]
+        expect(email.subject).to include("Order Assignment Notice")
+        expect(email.html_part.to_s)
+          .to include(order_detail.to_s)
+          .and include("assigned this&nbsp;order")
```
jhanggi: What's the value of doing this check? It's especially weird to me because it's testing for the presence of the &nbsp; which is generated by text_helpers.

app/mailers/order_assignment_mailer.rb
drench: Yes, that's the unlikely scenario this is trying to avoid. Part of what makes it unlikely is I don't believe there's a way to unassign through the modal, only through batch assignment.

spec/mailers/order_assignment_mailer_spec.rb
drench: I wanted to capture the singular vs plural messages here. I'm shortening them.

PR #870, opened 2017-02-14
[#126402] Allow Affiliates to have subaffiliates, aka cartegories

drench:
https://pm.tablexi.com/issues/126402

This change adds a flag to `Affiliate` called `subaffiliates_enabled`. When enabled for an `Affiliate`, an admin may define a category, or subaffiliate when assigning an `Affiliate` to an `Account`. This category is stored in the existing `accounts.affiliate_other` field.

To put it another way, how `affiliate_other` is treated depends on `subaffiliates_enabled`: if `true`, it's considered a category, and if `false`, it's an "other".

This updates `Account#affiliate_to_s` to be subaffiliate-aware, so anything that uses `affiliate_to_s` (such as Export Raw reports, and account "show" views) should show subaffiliates when they exist.


app/views/affiliates/index.html.haml
```diff
@@ -1,21 +1,23 @@
...
+%h2= text("pages.affiliates")
+= link_to text("affiliates.add"), new_affiliate_path, class: "btn-add"
 
 %table.table.table-striped.table-hover
   %thead
     %tr
       %th
-      %th{width: "100%"}= t("affiliates.th")
+      %th= Affiliate.model_name.human
+      %th= Affiliate.human_attribute_name(:subaffiliates_enabled)
```
drench: I don't know how useful displaying this flag on the index is. It would not take much to remove.

app/models/account.rb
```diff
@@ -269,10 +269,25 @@ def to_s(with_owner = false, flag_suspended = true)
   end
 
   def affiliate_to_s
-    return unless affiliate
-    affiliate_name = affiliate.name
-    affiliate_name += ": #{affiliate_other}" if affiliate == Affiliate.OTHER
-    affiliate_name
+    if affiliate
```
jhanggi: I kind of lean towards the guard clause here, but it's not a big deal.

app/services/account_builder.rb
```diff
@@ -164,13 +164,16 @@ def build_account_users
...
+      account.affiliate_other =
+        if affiliate == Affiliate.OTHER
+          account_params[:affiliate_other]
+        elsif affiliate.subaffiliates_enabled?
+          account_params[:subaffiliate]
+        end
     else
-      account.affiliate_id = nil
-      account.affiliate_other = nil
+      account.affiliate_id = account.affiliate_other = nil
```
jhanggi: I thought rubocop didn't like multiple assigns like this, but maybe I'm wrong.

app/views/affiliates/index.html.haml
```diff
@@ -1,21 +1,23 @@
...
   %tbody
-    - Affiliate.all.select(&:destroyable?).each do |affiliate|
+    - Affiliate.destroyable.each do |affiliate|
       %tr
         %td= delete_affiliate_link(affiliate)
         %td.indented_lots
           = affiliate.name
           %span.indented (#{edit_affiliate_link(affiliate)})
```
jhanggi: This table is definitely inconsistent with many of our other index views. Should we change it?

app/views/affiliates/index.html.haml
jhanggi: I'm cool with leaving it for now.

drench:
> Is the affiliate category required when you choose an affiliate that has them enabled?

As it is now, no, but making it required would help consolidate it with "Other"; we could `enable_subaffiliates` on "Other" and lose some of the conditionals.

app/services/account_builder.rb
```diff
@@ -164,13 +164,11 @@ def build_account_users
   # param is present, and the affiliate exists. Also ensure affiliate_other
   # gets wiped out if the affiliate_id does not point to `Other`.
   def set_affiliate
-    if account.class.using_affiliate? && account_params.key?(:affiliate_id)
-      affiliate = Affiliate.find_by_id(account_params[:affiliate_id])
-      account.affiliate_id = affiliate.try(:id)
-      account.affiliate_other = nil if affiliate.try(:id) != Affiliate.OTHER.id
+    if affiliate.present?
+      account.affiliate_id = affiliate.id
+      account.affiliate_other = affiliate.subaffiliates_enabled? ? account_params[:affiliate_other] : nil
```
jhanggi: Here's a thought. Could we move this check into `Affiliate` itself?

```
def affiliate_other=(value)
  self[:affiliate_other] = value if subaffiliates_enabled?
end
```

Or it could assign `nil` if enabled is false.

db/seeds.rb
```diff
@@ -12,8 +12,6 @@
 OrderStatus.create(name: "Complete")
 OrderStatus.create(name: "Reconciled")
 
-Affiliate.find_or_create_by(name: "Other")
```
jhanggi: Do we want to get rid of the default "Other" in seeds? Or is it added somewhere else?

lib/affiliate_account.rb
```diff
@@ -3,7 +3,9 @@ module AffiliateAccount
 
   def self.included(base)
     base.validates_presence_of :affiliate_id
-    base.validates_length_of :affiliate_other, minimum: 1, if: proc { |cc| cc.affiliate == Affiliate.OTHER }
+    base.validates_length_of :affiliate_other,
+                             minimum: 1,
+                             if: proc { |cc| cc.affiliate.try(:other?) }
```
jhanggi: Should this be a check on `subaffiliate_enabled?`?

spec/models/affiliate_spec.rb
```diff
@@ -3,22 +3,35 @@
...
-  it "should allow non-OTHER affiliates to be destroyed" do
+  it "allows non-OTHER affiliates to be destroyed" do
     affiliate = Affiliate.create!(name: "aff1")
     affiliate.destroy
     expect(affiliate).to be_destroyed
   end
 
+  describe "#other?" do
+    context "when it is the 'Other' affiliate" do
+      subject(:affiliate) { Affiliate.OTHER }
```
jhanggi: Should we be using double colon as the formatting for this? `Affiliate::OTHER`. I don't have a strong feeling since we've been using the dot for a while.

vendor/engines/c2po/app/views/facility_accounts/account_fields/_affiliate.html.haml
```diff
@@ -1,4 +1,13 @@
-= f.association :affiliate, collection: Affiliate.ordered_by_name, input_html: { class: "js--affiliate" }, include_blank: true, required: true
+- affiliates = Affiliate.by_name.map { |a| [a.name, a.id, { data: { subaffiliates_enabled: a.subaffiliates_enabled? } } ] }
```
jhanggi: Should we try to push "Other" to the bottom of this list?

vendor/engines/c2po/spec/services/purchase_order_account_builder_spec.rb
```diff
@@ -0,0 +1,26 @@
+require "rails_helper"
+
+RSpec.describe PurchaseOrderAccountBuilder, type: :service do
```
jhanggi: I think this file is missing tests. I think it probably wants `it_behaves_like "AccountBuilder#build"`

db/seeds.rb
drench: The first reference to `Affiliate.OTHER` should create it. We could through in an `Affiliate.OTHER` reference to force it at seed time.

app/models/affiliate.rb
```diff
@@ -7,17 +7,16 @@ class Affiliate < ActiveRecord::Base
   scope :by_name, -> { order(:name) }
 
   def self.OTHER
-    @@other ||= find_or_create_by(name: "Other", subaffiliates_enabled: true)
+    @@other ||= find_or_create_by(name: "Other") { |a| a.subaffiliates_enabled = true }
```
jhanggi: No need to change this since there's only the one extra field to be set, but I want to point you to `create_with` in case you need to do something like this in the future with more attributes.

`create_with(subaffiliates_enabled: true).find_or_create_by(name: "Other")` should be equivalent.

http://apidock.com/rails/v4.2.7/ActiveRecord/Relation/find_or_create_by

jhanggi:
Just curious where we're at on this? The client did confirm that the subafilliate should be required when one of those options is selected.

drench:
Yes, according to Aaron, if subaffiliates are active for an affiliate, then they are required.

jhanggi:
This looks good. I have a few comments still outstanding, but I don't think any of them should necessarily be blocking.

PR #871, opened 2017-02-14
[#134410] Added viewport meta tag

kt32291:
https://pm.tablexi.com/issues/134410

This adds the viewport meta tag to the head so the bootstrap elements are properly responsive, and we can look into what needs to be changed in the order flow. Tested this with ngrok, it does now show the bootstrap elements as responsive now in mobile browsers.

### DIFF
--- a/app/views/shared/_head.html.haml
+++ b/app/views/shared/_head.html.haml
@@ -1,4 +1,5 @@
 %meta{ "http-equiv" => "content-type", "content" => "text/html; charset=UTF-8" }
+%meta{ "name" => "viewport", "content" => "width=device-width, initial-scale=1, maximum-scale=1" }
 - # use gsub to strip out html tags in the title to prevent html from showing #62643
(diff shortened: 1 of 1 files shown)


jhanggi:
How does it look so far? Is it usable? More or less usable than without this?  I'm a little hesitant to actually merge this in if it's going to make things worse. In that case, we should just get some screenshots running locally.

But if it's already better (even if not perfect), then I'm cool with it.

kt32291:
@jhanggi this is probably good if I change it to not apply the viewport setting for admin pages right?

jhanggi:
We should definitely not apply it for any of the admin views, and I think that's a good thing to implement because we will want it eventually. 

I'm thinking we should hold off on merging this until we can spend a little time cleaning up some of the funkiness. 

Did you get a screenshot of the daily view?

jhanggi:
Closing for the time being. We can re-enable this when we move forward with the responsive design.

PR #872, opened 2017-02-14
[#125661] update statement and review order mailer text

meara:
https://pm.tablexi.com/issues/125661

Updated email body text as requested, also added a preview and specs for the statement mailer.


jhanggi:
Looks like there are several test failures. The 404s might be confusing, but I'll bet they stem from some `find` down in one of the mailers. It looks like maybe we're missing the `facility_id` argument from on up the method chain where `Notifier.review_orders` gets called.

One thing we need to take into account is that for "Send Notifications", I think DC uses the cross-facility tab/billing administrator role for that. Which means we need to make sure the email can support both with-a-facility and without-a-facility. I think if they're sent from the cross facility context, then we can probably leave out the facility abbreviation from the subject. If they're sent from a single facility, then we can include it.

Statements should be fine, because there is only one email that goes out per statement (well, to multiple people, but it's the same email), so we can use the statement to get the facility_id.

app/mailers/notifier.rb
```diff
@@ -50,8 +50,9 @@ def order_receipt(args)
 
   def review_orders(args)
     @user = User.find(args[:user_id])
-    @accounts = Account.find(args[:account_ids]).map(&:account_list_item).to_sentence
-    send_nucore_mail @user.email, text("views.notifier.review_orders.subject")
+    @accounts = Account.find(args[:account_ids]).map { |a| "#{a.account_list_item}, #{a.owner.user.full_name}" }.join("\n\n")
```
jhanggi: Part of me feels that the string rendering should happen in the template. The Notifier is kind of like a controller.

app/views/notifier/review_orders.html.haml
```diff
@@ -1 +1 @@
-= html "body_html", accounts: @accounts, login_url: transactions_in_review_url
+= html "body_html", accounts: @accounts, facility: @facility, login_url: transactions_in_review_url
```
jhanggi: This is kind of picky, but I think the email will look better if each account is separated by a single line break rather than blank lines between them. I would consider splitting the localization into separate parts and render the accounts in between them.

app/mailers/notifier.rb
```diff
@@ -50,8 +50,9 @@ def order_receipt(args)
 
   def review_orders(args)
     @user = User.find(args[:user_id])
-    @accounts = Account.find(args[:account_ids]).map(&:account_list_item).to_sentence
-    send_nucore_mail @user.email, text("views.notifier.review_orders.subject")
+    @accounts = Account.find(args[:account_ids])
+    @facility = args[:facility]
```
jhanggi: I'm not 100% of how this is hooked up, but is this method using GlobalID when it gets sent to delayed_job? When delayed_job serializes its arguments, it sometimes will end up serializing the entire facility object, which we like to avoid. I would recommend continuing to pass in the `facility_id`. You could probably do something like `Facility.find_by(id: args[:facility_id])` (I think that will return `nil` if the argument is `nil`) instead of raising an error.

app/mailers/notifier.rb
```diff
@@ -50,8 +50,9 @@ def order_receipt(args)
 
   def review_orders(args)
```
jhanggi: Would it be worth switching to using keyword arguments? `user_id` and `account_ids` are probably required; not sure about facility.

app/support/notification_sender.rb
```diff
@@ -68,9 +68,9 @@ def reviewed_at
 
   class AccountNotifier
 
-    def notify_accounts(account_ids_to_notify)
+    def notify_accounts(account_ids_to_notify, facility)
```
jhanggi: This is another place to check the usage of GlobalID, because I believe this gets delayed.

app/views/notifier/review_orders.text.haml
```diff
@@ -0,0 +1,8 @@
+= text "body_intro", facility: @facility, smart: false
+
+- @accounts.each do |account|
```
jhanggi: Not sure what's doing it (probably HAML), but there's no empty line between the intro and the list of accounts.

meara: @jhanggi Not sure what you mean--do you want it to look different from this?
screen shot 2017-02-21 at 11 12 09 am

jhanggi: HTML version is good. I'm referring to the text version of the email.

meara: Ah gotcha. Swapped to erb.

app/support/notification_sender.rb
meara: Looks like it's being called only with `#deliver_now`, so should be fine according to my understanding.

PR #874, opened 2017-02-14
[#134072] Upgrade factory_girl_rails

drench:


jhanggi:
Looks like we've got some test failures.

PR #877, opened 2017-02-16
[#134071] Upgrade rubocop

drench:


jhanggi:
Rubocop tends to add new cops and change cop names with new versions. Any changes we need to make there?

PR #882, opened 2017-02-16
[#125093] Activate the cross-facilty Users tab for Billing Administrators

drench:
UIC would like to evaluate this on stage. It shouldn't affect NU (the only BAs currently are staff who do testing). One active DC user is set as a BA. Should it get a feature flag?

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -93,6 +93,7 @@ api:
 # For these settings use SettingsHelper#feature_on?
 feature:
   billing_administrator_on: true
+  billing_administrator_users_tab_on: true
   create_users_on: true
   netids_on: true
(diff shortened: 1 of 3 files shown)


jhanggi:
I would vote for giving this a feature flag unless we're sure that everything on the tab works. And make sure it continues to work as expected with `manage_payment_sources_with_users_on` enabled.

Maybe it should be a separate role?

drench:
I put the feature flag in, defaulting to on. I've tested this with `manage_payment_sources_with_users` on and off and haven't run into trouble.

It kind of surprised me that we hadn't used `SettingsHelper` in the ability file before.

jhanggi:
👍

PR #885, opened 2017-02-17
Homogenize product new/edit views

jhanggi:
As I was working on spiking adding a new product type, I found several
discrepancies between the item/service/instrument new/edit views. This
brings them closer together. There is obviously further DRY to do as
item/new and service/new are identical as well as their edit views. and
their show/edit views are now very similar.

ProductsCommonController sets a `@product` variable, so it was trivial
to switch to that. There were also some discrepancies is what was one
level up from the details_field partial which led some views to have
some additional padding that shouldn’t have been there.

The longer-term goal would be to further clean these up as we add new product
types, but this seemed like a good first step.

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -41,9 +41,11 @@ module ProductsHelper
   def product_url_hint(product)
     product_url_name = product.try(:url_name).presence || "url-name"
 
-    url_name_hint = send("facility_#{product.class.name.downcase}_url",
-                         current_facility.url_name,
-                         product_url_name)
+    send(
+      "facility_#{product.class.name.underscore}_url",
+      current_facility.url_name,
+      product_url_name,
+    )
   end
 
   private
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -41,7 +41,7 @@ module ProductsHelper
   def product_url_hint(product)
     product_url_name = product.try(:url_name).presence || "url-name"
 
-    send(
+    public_send(
       "facility_#{product.class.name.underscore}_url",
       current_facility.url_name,
       product_url_name,
--- a/app/views/admin/shared/_tabnav_product.html.haml
+++ b/app/views/admin/shared/_tabnav_product.html.haml
@@ -1,23 +1,22 @@
-- product = @instrument || @service || @item || @bundle || @product
-- return nil if product.nil? || product.id.nil?
+- return nil if @product.nil? || @product.id.nil?
 %ul.nav.nav-tabs
-  = tab "Details", send("manage_facility_#{product.class.name.downcase}_path", current_facility, product), (secondary_tab == 'details')
-  - if product.is_a?(Instrument)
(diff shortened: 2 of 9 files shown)


app/views/admin/shared/_tabnav_product.html.haml
```diff
@@ -1,23 +1,22 @@
-- product = @instrument || @service || @item || @bundle || @product
-- return nil if product.nil? || product.id.nil?
+- return nil if @product.nil? || @product.id.nil?
```
jhanggi: I'm honestly not quite sure what this check does, but as this was a first-step refactor, I left it as-is.

app/views/admin/shared/_tabnav_product.html.haml
```diff
@@ -1,23 +1,22 @@
...
-    = tab 'Scheduling', facility_instrument_schedule_rules_path(current_facility, product), (secondary_tab == 'schedule_rules')
-  - if product.requires_approval?
-    = tab 'Access List', [current_facility, product, :users], (secondary_tab == 'users')
-  - if product.requires_approval? && product.is_a?(Instrument)
-    = tab ProductAccessGroup.model_name.human.pluralize, facility_instrument_product_access_groups_path(current_facility, product), (secondary_tab == 'restriction_levels')
-  - if product.is_a?(Bundle)
-    = tab 'Bundled Products', facility_bundle_bundle_products_path(current_facility, product), (secondary_tab == 'products')
+  = tab "Details", [:manage, current_facility, @product], (secondary_tab == "details")
+  - if @product.is_a?(Instrument)
+    = tab "Scheduling", facility_instrument_schedule_rules_path(current_facility, @product), (secondary_tab == "schedule_rules")
```
jhanggi: I18n is a next step for this, but for a later PR.

app/helpers/products_helper.rb
```diff
@@ -41,9 +41,11 @@ def public_calendar_link(product)
   def product_url_hint(product)
     product_url_name = product.try(:url_name).presence || "url-name"
 
-    url_name_hint = send("facility_#{product.class.name.downcase}_url",
-                         current_facility.url_name,
-                         product_url_name)
+    send(
+      "facility_#{product.class.name.underscore}_url",
```
jhanggi: Change from `downcase` to `underscore` is the result of my spike being on `TestProduct`, which should be `test_product` instead of `testproducts`.

drench: I'd go with `public_send` if possible

app/views/items/new.html.haml
```diff
@@ -2,15 +2,16 @@
...
+  = render "admin/shared/sidenav_product", sidenav_tab: @product.class.name.pluralize.underscore
 
-%h2= text("admin.shared.add", model: Item.model_name.human)
+%h2= text("admin.shared.add", model: @product.class.model_name.human)
 
-= simple_form_for([current_facility, @item]) do |f|
+= simple_form_for([current_facility, @product]) do |f|
   = f.error_messages
-  %ul.form
+  .form-imputs
```
drench: typo

drench: typo

PR #887, opened 2017-02-18
Homogenize price policy views

jhanggi:
Extends from #886. This cleans up some of the price policy views to make adding new products simpler.

### DIFF
--- a/app/controllers/price_policies_controller.rb
+++ b/app/controllers/price_policies_controller.rb
@@ -12,7 +12,7 @@ class PricePoliciesController < ApplicationController
   before_action :set_expire_date_from_params, only: [:create, :update]
   before_action :set_max_expire_date, only: [:edit, :update]
 
-  load_and_authorize_resource
+  load_and_authorize_resource instance_name: :price_policy
 
   layout "two_column"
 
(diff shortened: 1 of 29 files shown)


app/views/admin/shared/_sidenav_product.html.haml
```diff
@@ -1,8 +1,8 @@
 %ul.sidebar-nav
-  = tab "#{Instrument.model_name.human.pluralize} (#{current_facility.instruments.count})", facility_instruments_path(current_facility), sidenav_tab == Instrument.model_name.plural
-  = tab "#{Service.model_name.human.pluralize} (#{current_facility.services.count})", facility_services_path(current_facility), sidenav_tab == Service.model_name.plural
-  = tab "#{Item.model_name.human.pluralize} (#{current_facility.items.count})", facility_items_path(current_facility), sidenav_tab == Item.model_name.plural
-  = tab "#{Bundle.model_name.human.pluralize} (#{current_facility.bundles.count})", facility_bundles_path(current_facility), sidenav_tab == Bundle.model_name.plural
+  - [Instrument, Service, Item, Bundle].each do |product_type|
```
jhanggi: I'm probably going to pull this into some sort of registry once I have a new product type. I wanted to use `Product.descendents`, but that's unreliable: if the subclass hasn't been used yet (and thus not auto-loaded), it's not there, and there's no way to designate order.

app/controllers/price_policies_controller.rb
```diff
@@ -90,43 +90,31 @@ def create_or_update(action)
...
   def init_product
-    @product = current_facility.method(product_var.pluralize)
-                               .call
-                               .find_by_url_name!(params["#{product_var}_id".to_sym])
-    instance_variable_set("@#{product_var}", @product)
-  end
-
-  def model_name
-    self.class.name.gsub(/Controller\z/, "").singularize
+    id_param = params.except(:facility_id).keys.detect { |k| k =~ /_id\z/ }
```
drench: I don't feel as strongly as Code Climate does about using `#end_with?` here.

PR #893, opened 2017-02-20
Move FileUpload routes into product resource so they can be reused more easily.

jhanggi:
* Change "upload" to "index"
* Rename "uploader_create" to "upload_sample_results" to better reflect how its used.

There are still more routes to deal with, but this should cover the main Docs tab.

### DIFF
--- a/app/views/file_uploads/product_survey.html.haml
+++ b/app/views/file_uploads/product_survey.html.haml
@@ -72,13 +72,11 @@
     = f.error_messages
     = f.hidden_field :file_type
 
-    %ul.form
-      %li
-        = f.label :file, "File", class: "required"
-        = f.file_field :file
+    = f.label :file, "File", class: "required"
+    = f.file_field :file
 
     %ul.inline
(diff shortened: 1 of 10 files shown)


app/views/file_uploads/product_survey.html.haml
```diff
@@ -72,13 +72,11 @@
...
-    %ul.form
-      %li
-        = f.label :file, "File", class: "required"
-        = f.file_field :file
+    = f.label :file, "File", class: "required"
+    = f.file_field :file
 
     %ul.inline
-      %li= f.submit "Upload", data: { disable_with: "Uploading..." }, class: "btn"
+      %li= f.submit text("shared.upload"), data: { disable_with: "Uploading..." }, class: "btn"
```
drench: There's a translation for "Uploading..." at `file_uploads.index.submitted` (which is maybe not the ideal spot for it)

jhanggi: Good find. I'm not quite sure where the best place for that to live is. I would vote for cleaning that up when we clean up the rest of the file.

PR #895, opened 2017-02-21
[#132371] Returns helpful message on duplicate start date

kt32291:

This adds a helpful error message if you try to make a price policy for a product with an active applied price policy with the same start date.

### DIFF
--- a/config/locales/en.controllers.yml
+++ b/config/locales/en.controllers.yml
@@ -144,7 +144,7 @@ en:
       errors:
         remove_active_policy: "Sorry, but you cannot remove an active price policy. If you really want to do so move the start date to the future and try again."
         save: There was an error saving the policy
-        same_start_date: "Sorry, but you cannot start a policy the same date as an active policy that has already been applied to an order"
+        same_start_date: "You have chosen a start date that matches an active policy with existing orders. Please choose another start date."
       new:
         success: Price Rules were successfully created.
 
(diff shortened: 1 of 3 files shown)


config/locales/en.controllers.yml
```diff
@@ -144,6 +144,7 @@ en:
       errors:
         remove_active_policy: "Sorry, but you cannot remove an active price policy. If you really want to do so move the start date to the future and try again."
         save: There was an error saving the policy
+        same_start_date: "Sorry, but you cannot start a policy the same date as an active policy that has already been applied to an order"
```
drench: I'm wondering if there's a clearer way to phrase this. Something like "You have chosen a start date that matches an active policy with existing orders. Please choose another start date." maybe?

spec/price_policies_controller_shared_examples.rb
```diff
@@ -222,19 +232,20 @@
...
         end
-        it "should redirect to show on success" do
+
+        it "redirects to show on success" do
           do_request
           is_expected.to redirect_to price_policy_index_path
         end
 
-        it "should create a new price policy for a group that has no fields, but cant purchase" do
+        it "creates a new price policy for a group that has no fields, but cant purchase" do
```
drench: Existing typo: "cant"

config/locales/en.controllers.yml
jhanggi: That sounds good. I'll switch it to that.

PR #896, opened 2017-02-22
[#126401] Update LDAP documentation

drench:
These are updates I made while configuring LDAP authentication.

The most significant change is probably that it is no longer necessary to install an LDAP server on Mac OS. The LDAP server that ships with the OS works. This change also removes the PDF version.


jhanggi:
👍

PR #897, opened 2017-02-22
[#140077551] Deny requests from DoorService

waggles:
* Adds our first route, sending away all who try to enter, regardless of intention (sound familiar??)
* Also, starts adding API documentation to the `SecureRooms` engine README.


vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/base_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+
+RSpec.describe SecureRoomsApi::BaseController do
+  context "scan" do
+    before do
+      encoded_auth_credentials = ActionController::HttpAuthentication::Basic.encode_credentials('dan', 'hodos')
```
waggles: oops 😂

danhodos: Hee hee!

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+The door security hardware communicates through an API with the following endpoints:
+
+#### Cardholder Scans at a Room Entrance
+
+* **Method/URL**
+  `POST /secure_rooms_api/scan`
+
+* **Data Params**
+
+  * `ard_id=[integer]`
```
danhodos: Oooh, lost a "c" here.

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+      { response: "grant" }
+      ```
+
+  **Response 2:** The cardholder has multiple payment methods and must select which they would like to use.
+
+  * **Code:** 200 OK
+  * **Response:**
+
+      ```
+      { response: "deny",
```
danhodos: Looks like this was switched with `select_account` below.

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+#### Cardholder Scans at a Room Entrance
+
+* **Method/URL**
+  `POST /secure_rooms_api/scan`
+
+* **Data Params**
+
+  * `ard_id=[integer]`
+  * `controller_id=[integer]`
+  * `outside_reader_id=[integer]`
```
danhodos: I think this'll probably just come through as `reader_id`

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
 # SecureRooms
 
+The SecureRooms engine adds a product type supporting an access-restricted,
+bill-by-time room.
+
+## Installation
+
+Make sure the `gem "secure_rooms",   "~> 0.0.1", path: "vendor/engines/secure_rooms"`
```
jhanggi: The `"~> 0.0.1"` isn't necessary, and we don't have it in our gemfile. Since it's on the app path, you're always getting the latest version no matter what.

In theory, we should get rid of all the version checks for our engines.

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+* **Data Params**
+
+  * `ard_id=[integer]`
+  * `controller_id=[integer]`
+  * `outside_reader_id=[integer]`
+
+* **Success Response:**
+  There are three types of response:
+
+
```
jhanggi: Extra blank line?

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+  * **Code:** 200 OK
+  * **Body:**
+
+      ```
+      { response: "grant" }
+      ```
+
+  **Response 2:** The cardholder has multiple payment methods and must select which they would like to use.
+
+  * **Code:** 200 OK
```
jhanggi: Part of me feels like we should be using a different HTTP code for denials, but I can't think of a better one. But I think 200 will work fine.

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+      ```
+
+  **Response 3:** The cardholder is not currently allowed to access the SecureRoom, and they are rejected.
+
+  * **Code:** 200 OK
+  * **Response:**
+
+      ```
+      { response:   "select_account",
+          tablet_id:  <integer id>,
```
jhanggi: Alignment with the other keys.

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+          reason:   "I only know how to deny right now." }
+      ```
+
+  **Response 3:** The cardholder is not currently allowed to access the SecureRoom, and they are rejected.
+
+  * **Code:** 200 OK
+  * **Response:**
+
+      ```
+      { response:   "select_account",
```
jhanggi: Do we want to include any sort of message? Or just leave that up to the client?

vendor/engines/secure_rooms/README.md
```diff
@@ -1,2 +1,57 @@
...
+          tablet_id:  <integer id>,
+          accounts:   [] }
```
jhanggi: Have you fleshed out what this array should look like? I'm fine leaving this a TODO if not.

vendor/engines/secure_rooms/config/routes.rb
```diff
@@ -5,4 +5,8 @@
       resources :price_policies, controller: "secure_room_price_policies", except: :show
     end
   end
+
+  namespace :secure_rooms_api do
+    post '/scan', to: 'base#scan'
```
jhanggi: Would it make sense to have a `ScansController#create` and post to that? If we just have the one endpoint, it might not be worth the extra controller, but it currently feels weird with `BaseController`. Usually when I see something like that, I expect it to get subclassed.

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/base_controller_spec.rb
```diff
@@ -0,0 +1,16 @@
...
+      request.env['HTTP_AUTHORIZATION'] = encoded_auth_credentials
+      post :scan
+    end
+
+    it 'denies entry until we add logic' do
```
jhanggi: Picky, but we prefer double quotes.

vendor/engines/secure_rooms/README.md
waggles: `300` and `403` might be more correct in the long run, but I think it'll be easiest to just `200` at this stage of development? I can make some additional notes around some of these WIP rules

vendor/engines/secure_rooms/README.md
waggles: oops I think copy/paste out of my markdown editor needs tweaking 😂

vendor/engines/secure_rooms/README.md
waggles: hmm since this is still sort of internal-facing to us, I think just keys would work until the message hits the eyes of a user (the tablet display)?

vendor/engines/secure_rooms/README.md
waggles: not entirely, I'll update with TODO info

vendor/engines/secure_rooms/config/routes.rb
waggles: Yeah this is not a great name.. lemme ponder a replacement

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    before do
+      name = Settings.secure_rooms_api.basic_auth_name
+      password = Settings.secure_rooms_api.basic_auth_password
+      encoded_auth_credentials = ActionController::HttpAuthentication::Basic.encode_credentials(name, password)
+      request.env['HTTP_AUTHORIZATION'] = encoded_auth_credentials
+      post :scan
+    end
+
+    it "denies entry until we add logic" do
+      expect(response.code).to eq "403"
```
jhanggi: Looks like the issue is more around the credentials than the behavior itself. Given that, part of me likes the readability of `to be_forbidden`, but the error is much easier to troubleshoot this way.

Upon googling this just now, I discovered rspec-rails has a `have_http_status` matcher that should allow us to do `expect(response).to have_http_status(:forbidden)`.

https://www.relishapp.com/rspec/rspec-rails/docs/matchers/have-http-status-matcher

I think that gives us the best of both worlds: "expected the response to have status code :forbidden (403) but it was :unauthorized (401)"

I'm going to use this more often.

waggles: 👌

config/settings/test.yml
```diff
@@ -1,2 +1,7 @@
 devise:
   secret_key: override-me-in-settings.local.yml
+
+# You should change this if you want to be able to access the SecureRooms API
```
jhanggi: I feel like this comment could be more specific for the test environment.

waggles:
we can change the base branch now?? since when woww

PR #898, opened 2017-02-23
Spec updates related to UIC

drench:
This updates the `UsersController` specs to use the `create_users` flag to determine if `User#edit` and `User#update` are available, and changes the `Product` model specs to not rely on user-based price groups, which are not available in the UIC fork.

This change should resolve the spec failures in https://github.com/tablexi/nucore-uic/pull/113 when merged.

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -55,8 +55,8 @@ RSpec.describe UsersController do
       @params[:id] = user.id
     end
 
-    it_should_allow_admin_only do
-      expect(assigns[:user]).to eq(user)
+    context "when create_users is active", feature_setting: { create_users: true } do
+      it_should_allow_admin_only { expect(assigns[:user]).to eq(user) }
     end
   end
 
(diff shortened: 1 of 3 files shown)


spec/controllers/users_controller_spec.rb
```diff
@@ -55,8 +55,8 @@
       @params[:id] = user.id
     end
 
-    it_should_allow_admin_only do
-      expect(assigns[:user]).to eq(user)
+    context "when create_users is active", feature_setting: { create_users: true } do
```
meara: This seems reasonable to me, but I recall someone mentioning that with feature-flags like this, we prefer to test both contexts (with the feature on and the feature off).

What makes this case different?

drench: In general that's how I'd do it. In this case, there was a set of existing specs below starting at line 308 that turn the `create_users` feature off and test that the routes don't exist, so I added `#edit` and `#update` there.

jhanggi: I would probably pull this flag up to the `describe "GET #edit"` line, but it's not a big deal here. But in general, if having the feature off means there is no behavior, it's usually safe to not do the inverse.

meara: Thanks folks!

PR #899, opened 2017-02-23
[#133594] extending reservations for instruments with cutoff hours

meara:
- https://pm.tablexi.com/issues/133594 (Dartmouth)
- https://pm.tablexi.com/issues/28901 (NU)
- https://pm.tablexi.com/issues/121867 (UIC)

Extending reservations behaves appropriately in conjunction with mandated advance reservations (cutoff hours).

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -128,7 +128,7 @@ class ReservationsController < ApplicationController
     raise ActiveRecord::RecordNotFound unless @reservation.nil?
 
     options = current_user.can_override_restrictions?(@instrument) ? {} : { user: acting_user }
-    next_available = @instrument.next_available_reservation(1.minute.from_now, default_reservation_mins.minutes, options)
+    next_available = @instrument.next_available_reservation(after: 1.minute.from_now, duration: default_reservation_mins.minutes, options: options)
     @reservation = next_available || default_reservation
     @reservation.round_reservation_times
     unless @instrument.can_be_used_by?(@order_detail.user)
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -304,8 +304,14 @@ class ReservationsController < ApplicationController
 
   # TODO: you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
-    can_edit = @reservation.admin_editable?
-    can_edit &&= @reservation.can_customer_edit? unless current_user.administrator?
+    if current_user.administrator? && @reservation.admin_editable?
+      can_edit = true
+    elsif @reservation.can_customer_edit?
+      can_edit = true
+    else
+      can_edit = false
+    end
+
     params[:id].to_i != @reservation.id ||
       @reservation.actual_end_at ||
       !can_edit
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -304,17 +304,19 @@ class ReservationsController < ApplicationController
 
   # TODO: you shouldn't be able to edit reservations that have passed or are outside of the cancellation period (check to make sure order has been placed)
   def invalid_for_update?
-    if current_user.administrator? && @reservation.admin_editable?
-      can_edit = true
-    elsif @reservation.can_customer_edit?
-      can_edit = true
-    else
-      can_edit = false
-    end
-
     params[:id].to_i != @reservation.id ||
       @reservation.actual_end_at ||
-      !can_edit
+      !editable_by_current_user?
+  end
+
+  def editable_by_current_user?
+    if current_user.administrator? && @reservation.admin_editable?
+      true
+    elsif @reservation.can_customer_edit?
+      true
+    else
+      false
+    end
   end
 
   def save_reservation_and_order_detail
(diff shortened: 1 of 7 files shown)


app/support/products/scheduling_support.rb
```diff
@@ -152,39 +152,49 @@ def update_schedule_name
...
 
-        start_time = conflict.reserve_end_at
+        next_time = conflict.reserve_end_at
       end
     end
 
+    private
+
+    def set_start_time(time)
+      if !options[:ignore_cutoff] && rule.instrument.cutoff_hours > 0
```
jhanggi: Not a big deal, but this might be easier to read as `if options[:ignore_cutoff] || rule.instrument.cutoff_hours <= 0` and flip the conditional (I think I got that right). Not a fan of double negatives (not & ignore in this case).

app/support/products/scheduling_support.rb
```diff
@@ -152,39 +152,49 @@ def update_schedule_name
...
+        # handle cutoff hours if needed
+        start_time = [time, day_start, rule.instrument.cutoff_hours.hours.from_now].max
```
jhanggi: Another possible way to think about this is
```
possible_start_times = [time, day_start]
possible_start_times << rule.instrument.cutoff_hours.hours.from_now unless options[:ignore_cutoff] || rule.instrument.cutoff_hours <= 0
start_time = possible_start_times.max
```

This is something I'm unsure of: would ignoring the `rule.instrument.cutoff_hours > 0` matter and going with `rule.instrument.cutoff_hours.from_now`, which for the case of 0 is `Time.current`.

app/support/products/scheduling_support.rb
```diff
@@ -152,39 +152,49 @@ def update_schedule_name
...
         return reservation if conflict.nil?
 
-        start_time = conflict.reserve_end_at
+        next_time = conflict.reserve_end_at
       end
     end
 
+    private
+
+    def set_start_time(time)
```
jhanggi: Rubocop recommends not prefixing methods with `set_`. In this case I agree as it's not actually setting anything. It's returning something.

app/support/reservations/moving_up.rb
```diff
@@ -8,11 +8,12 @@ module Reservations::MovingUp
   # if there is no such time slot. For read-only purposes.
   def earliest_possible
     after = 1.minute.from_now
-    next_res = product.next_available_reservation(after,
-                                                  duration_mins.minutes,
-                                                  exclude: self,
-                                                  user: user,
-                                                  until: reserve_start_at)
+    next_res = product.next_available_reservation(after: after,
```
jhanggi: I like the keyword arguments, but now that we're using them, how many `options` are there? Would it be clean enough to make them all top level arguments rather than having an `options` argument?

spec/factories/reservations.rb
```diff
@@ -98,7 +98,7 @@
     transient { user nil }
 
     after(:create) do |reservation, evaluator|
-      reservation.order.update_attribute(:user_id, evaluator.user.id) if evaluator.user
+      reservation.order.update_attributes(user_id: evaluator.user.id) if evaluator.user
```
jhanggi: Is there ever a case where this could fail validations? Or is there some callback we're relying on? In factories like this, I prefer the validation/callback-free `update_attribute` rather than `update_attributes`.

app/support/reservations/moving_up.rb
meara: Will check.

meara: Looks like the options we're passing are:
`:ignore_cutoff`
`:user`
`:exclude`
`:until`

@jhanggi is that too many to have them all top level? Not sure what a reasonable cutoff would be.

jhanggi: I could go either way. My general feeling is once you're going keyword arguments, an `options` parameter seems a little redundant and nested. But if you are using that many keyword arguments, sometimes the options hash is cleaner. I'll leave it up to you.

app/controllers/reservations_controller.rb
```diff
@@ -304,11 +304,19 @@ def ability_resource
...
-    can_edit = @reservation.admin_editable?
-    can_edit &&= @reservation.can_customer_edit? unless current_user.administrator?
     params[:id].to_i != @reservation.id ||
       @reservation.actual_end_at ||
-      !can_edit
+      !editable_by_current_user?
+  end
+
+  def editable_by_current_user?
+    if current_user.administrator? && @reservation.admin_editable?
```
jhanggi: I don't think it's worth the effort of changing right now, but in general, we're preferring switching things like `current_user.administrator?` to a check on abilities.

app/support/products/scheduling_support.rb
```diff
@@ -52,12 +52,12 @@ def can_purchase?(group_ids = nil)
 
   def first_available_hour
     return 0 unless schedule_rules.any?
-    schedule_rules.min { |a, b| a.start_hour <=> b.start_hour }.start_hour
+    schedule_rules.min_by(&:start_hour).start_hour
```
jhanggi: Nice.

app/controllers/reservations_controller.rb
drench: That's a good point: I could see this method being replaced with something like `current_ability.can?(:edit, @reservation)`, if the `Ability` file is set up right. I haven't checked. But I'm good with this version.

app/support/reservations/moving_up.rb
meara: I think because we pass `:options` down through so many methods after this one, I'm not going to change it for now.

PR #909, opened 2017-02-27
user edit link appears only if feature is on

meara:
We saw behavior on UIC where the user edit link was available even though the `:create_users` feature was turned off. The link would then error, since the route did not exist in that context.

This PR corrects that behavior and adds specs verifying that the correct behavior occurs both when the feature is turned on and turned off.

### DIFF
--- a/lib/ability.rb
+++ b/lib/ability.rb
@@ -26,7 +26,11 @@ class Ability
         end
         unless user.account_manager?
           cannot :manage, User unless resource.is_a?(Facility) && resource.single_facility?
-          cannot([:edit, :update], User) { |target_user| !target_user.admin_editable? }
+          if SettingsHelper.feature_on?(:create_users)
+            cannot([:edit, :update], User) { |target_user| !target_user.admin_editable? }
+          else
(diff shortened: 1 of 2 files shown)


lib/ability.rb
```diff
@@ -26,7 +26,11 @@ def initialize(user, resource, controller = nil)
         end
         unless user.account_manager?
           cannot :manage, User unless resource.is_a?(Facility) && resource.single_facility?
-          cannot([:edit, :update], User) { |target_user| !target_user.admin_editable? }
+          if SettingsHelper.feature_on?(:create_users)
+            cannot([:edit, :update], User) { |target_user| !target_user.admin_editable? }
+          else
+            cannot([:edit, :update], User)
```
jhanggi: Should this be more simply `cannot :manage, User`? Or possibly pull the check out a level or two?

meara: If we change this line to `cannot :manage, User` we will also remove indexing and viewing--the Users tab literally will not appear in the admin view.

jhanggi: Got it. Sounds legit.

PR #910, opened 2017-02-27
[#134079] upgrade to jquery-ui-rails 6.0

meara:
https://pm.tablexi.com/issues/134079


Gemfile
```diff
@@ -35,7 +35,7 @@ gem "will_paginate", "~> 3.1.5"
 gem "dynamic_form",     "~> 1.1.4"
 gem "ckeditor", "~> 4.1.6"
 gem "jquery-rails"
-gem "jquery-ui-rails"
+gem "jquery-ui-rails", "~>6.0.1"
```
jhanggi: Picky, but we usually put a space between `~>` and the number.

app/assets/javascripts/application.js
```diff
@@ -3,11 +3,11 @@
 
 //= require jquery
 //= require jquery_ujs
-//= require jquery-ui/datepicker
-//= require jquery-ui/draggable
-//= require jquery-ui/droppable
-//= require jquery-ui/tabs
-//= require jquery-ui/dialog
+//= require jquery-ui/widgets/datepicker
+//= require jquery-ui/widgets/draggable
```
jhanggi: I'm not actually sure we use jquery's draggable/droppable. I think there are some references within timeline.js where there is some interface for rescheduling by drag/dropping reservations, but we don't actually do anything with that interaction.

I think it was a feature that I was experimenting with when we first built the timeline view, but we never used it.

We don't have to right now, but it wouldn't really be a bad idea to remove the dead code.

app/assets/javascripts/application.js
```diff
@@ -3,11 +3,11 @@
...
+//= require jquery-ui/widgets/droppable
+//= require jquery-ui/widgets/tabs
+//= require jquery-ui/widgets/dialog
```
jhanggi: I think it might also be worth seeing if we actually use the dialog widget, and if we can get rid of it. I think we end up using bootstap's modal instead.

app/assets/javascripts/application.js
```diff
@@ -3,11 +3,11 @@
 
 //= require jquery
 //= require jquery_ujs
-//= require jquery-ui/datepicker
-//= require jquery-ui/draggable
-//= require jquery-ui/droppable
-//= require jquery-ui/tabs
-//= require jquery-ui/dialog
+//= require jquery-ui/widgets/datepicker
```
jhanggi: I'm not sure it's a big deal, but the date picker widget has some styling changes. It's much flatter. I think I'm cool with it, but I wanted to mention it.

screen shot 2017-03-02 at 7 31 35 pm
screen shot 2017-03-02 at 7 30 54 pm

PR #914, opened 2017-03-01
[#125661] statement flash notification

meara:
https://pm.tablexi.com/issues/125661

Remove displayed <br/> from statement notification flash message.

### DIFF
--- a/app/controllers/facility_statements_controller.rb
+++ b/app/controllers/facility_statements_controller.rb
@@ -38,9 +38,9 @@ class FacilityStatementsController < ApplicationController
       flash[:error] = text("no_selection")
     elsif @statement_creator.create
       @statement_creator.send_statement_emails
-      flash[:notice] = text(success_message, accounts: @statement_creator.formatted_account_list).html_safe
+      flash[:notice] = text(success_message, accounts: @statement_creator.formatted_account_list)
     else
-      flash[:error] = text("errors_html", errors: @statement_creator.formatted_errors).html_safe
+      flash[:error] = text("errors_html", errors: @statement_creator.formatted_errors)
     end
 
     redirect_to action: :new
--- a/app/controllers/facility_statements_controller.rb
+++ b/app/controllers/facility_statements_controller.rb
@@ -38,9 +38,9 @@ class FacilityStatementsController < ApplicationController
       flash[:error] = text("no_selection")
     elsif @statement_creator.create
       @statement_creator.send_statement_emails
-      flash[:notice] = text(success_message, accounts: @statement_creator.formatted_account_list).html_safe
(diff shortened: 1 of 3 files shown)


app/controllers/facility_statements_controller.rb
```diff
@@ -38,9 +38,9 @@ def create
       flash[:error] = text("no_selection")
     elsif @statement_creator.create
       @statement_creator.send_statement_emails
-      flash[:notice] = text(success_message, accounts: @statement_creator.formatted_account_list).html_safe
+      flash[:notice] = text(success_message, accounts: @statement_creator.formatted_account_list.html_safe).html_safe
```
jhanggi: Would using the `html` text helper allow us to get rid of one of the `html_safe`s?

meara: Empirically, no. Somewhat confusingly, the inner `.html_safe` call is actually just marking the account list as BEING html safe (un-escaping it, basically).

From the [text_helper docs](https://github.com/ahorner/text-helpers/blob/master/README.md):

```text and html will escape all arguments passed to it in order to prevent XSS attacks. If you want to pass html content, you should ensure you mark it as .html_safe

Example: text('welcome_user', username) will escape html characters in username

Welcome &lt;b&gt;Bob&lt;/b&gt;
Example: text('welcome_user', username.html_safe) will output html characters in username

Welcome <b>Bob</b>```

Comment (jhanggi) [inline app/controllers/facility_statements_controller.rb]: The description from the text_helper docs is helpful for me understanding this.

I just played around with it. It's looking to me like we don't need the outer html_self. I tried it both with `text` and `html` helpers.

I could be convinced otherwise, but might it be a little cleaner to move the inner `html_safe` into `StatementCreator`?

Comment (jhanggi) [inline app/controllers/facility_statements_controller.rb]: I still think we can get rid of the outer html_safe.

Comment (meara) [inline app/controllers/facility_statements_controller.rb]: ...I started to write out why I think we still need it and then realized I agree, we don't. Will take it out. 😸

File: spec/controllers/facility_statements_controller_spec.rb [hunk 2]
```diff
@@ -167,6 +167,14 @@ def create_order_details
         end
       end
 
+      context "with multiple payment sources" do
+        it "should display properly formatted flash message" do
```
drench: Nitpick: we're shunning "should" in specs.

spec/controllers/facility_statements_controller_spec.rb
```diff
@@ -167,6 +167,14 @@ def create_order_details
...
+          sign_in(@user)
+          do_request
+          expect(flash[:notice]).to match(%r{Notifications sent successfully to:<br\/>.*account description<br\/.*})
```
drench: Since you're using `%r{}` to surround your regex, you don't need to escape `/`.

PR #915, opened 2017-03-02
[#134996] fix for tab nav

meara:
https://pm.tablexi.com/issues/134996

Resolved issue: tabs disappear when on either Scheduling or Scheduling Groups tab of an instrument on Stage

### DIFF
--- a/app/controllers/product_access_groups_controller.rb
+++ b/app/controllers/product_access_groups_controller.rb
@@ -2,6 +2,7 @@ class ProductAccessGroupsController < ApplicationController
 
   before_action :authenticate_user!
   before_action :init_current_facility
+  before_action :init_current_product
 
   admin_tab :all
   load_and_authorize_resource :facility, find_by: :url_name
@@ -16,7 +17,6 @@ class ProductAccessGroupsController < ApplicationController
   end
 
   def index
-    @product = @instrument # required for tabnav_product
   end
 
   def edit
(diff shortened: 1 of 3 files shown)


app/controllers/product_access_groups_controller.rb
```diff
@@ -16,6 +16,7 @@ def initialize
   end
 
   def index
+    @product = @instrument # required for tabnav_product
```
jhanggi: Have you verified the other actions on this controller? It seems like this might live better as a before_action.

It's probably not worth refactoring this controller/views to use `@product` right now, but that should probably be the better longer-term solution.

app/controllers/schedule_rules_controller.rb
```diff
@@ -20,6 +20,7 @@ def index
     @start_at       = Time.zone.at(params[:start].to_i)
     @end_at         = Time.zone.at(params[:end].to_i)
     @schedule_rules = @instrument.schedule_rules
+    @product        = @instrument # required for tabnav_product
```
jhanggi: Same with this, I think the new/edit are losing the tabs as well.

app/controllers/product_access_groups_controller.rb
```diff
@@ -2,6 +2,7 @@ class ProductAccessGroupsController < ApplicationController
 
   before_action :authenticate_user!
   before_action :init_current_facility
+  before_action :init_current_product
```
jhanggi: Pretty sure this needs to happen after the `load_and_authorize_resource :instrument`, otherwise `@instrument isn't set yet.

meara: Oh good catch, thanks.

PR #917, opened 2017-03-02
[#133442] Redirect to order assignment when not logged in

meara:
https://pm.tablexi.com/issues/133442

As a non-logged in user, the staff member is redirected appropriately when clicking an order link from an assigned staff email.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -144,12 +144,6 @@ class ApplicationController < ActionController::Base
     render "/acting_error", status: 403, layout: "application"
   end
 
-  #
-  # Customize Devise redirect after login
-  def after_sign_in_path_for(resource)
-    session[:requested_params] || super
-  end
-
   def after_sign_out_path_for(_)
     if current_facility.present?
       facility_path(current_facility)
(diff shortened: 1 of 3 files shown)


meara:
@jhanggi -- do we want to do a more global fix or are we good with this?

jhanggi:
I just looked into this because it doesn't feel like quite the right solution. I'm writing this because otherwise I'll forget by Monday.

For this particular page, the `authorize!` raises `CanCan::AccessDenied` if you're not authorized (this could also mean you're not logged in).
In ApplicationController, we rescue from that with `render_403`, which if you're not logged in redirects you to the `new_user_session_path`.
I'm not quite sure why, but your location is not being stored automatically. It actually works just fine for most other pages: you go to a specific link, it gets stored, you log in, you get redirected back to the page you initially tried to go to.

I found 3 possible solutions, and I'm not sure which is the best.

1. Add `before_action :authenticate_user!` to the controller. This ends up being a little faster because then it never loads the facility, order, etc. It's also consistent with most other controllers.
2. In `render_403`, add `store_location_for(:user, request.fullpath)` before we redirect. This is a Devise-provided method (there might be something better, but I'm already getting too deep into Devise). This has the advantage of solving similar problems in the future.
3. Link to the order show view rather than manage. This page wasn't _really_ meant to be accessed directly outside the popup.

PR #918, opened 2017-03-03
[#140077159] Associate card with user

waggles:
* Adds migration for indala_number on User to engine
* Adds view hook for additional fields on User show
* Adds separate controller+ability to handle edit and update of indala field


app/controllers/users_controller.rb
```diff
@@ -163,7 +163,7 @@ def update
   private
 
   def edit_user_params
-    params.require(:user).permit(:email, :first_name, :last_name)
+    params.require(:user).permit(:email, :first_name, :last_name, :indala_number)
```
jhanggi: I think this might be left over from when you were trying to update in the same user edit form.

vendor/engines/secure_rooms/app/controllers/secure_rooms/indala_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
+module SecureRooms
+
+  class IndalaNumbersController < ApplicationController
+
+    include TextHelpers::Translation
+
+    customer_tab :password
+    admin_tab     :all
```
jhanggi: Rubocop/Codeclimate is complaining about this. A lot of existing code still uses argument alignment like this, but we're moving more towards not aligning arguments, so I would recommend closing up the spacing here.

vendor/engines/secure_rooms/app/controllers/secure_rooms/indala_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    before_action :custom_authorize
+    before_action :check_acting_as
+
+    def initialize
+      @active_tab = "admin_users"
+      super
+    end
+
+    def edit
+      @user = User.find(params[:id])
```
jhanggi: I would recommend pulling this into a before_action. Not a blocker, though.

vendor/engines/secure_rooms/app/controllers/secure_rooms/indala_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    # UsersController within which this IndalaNumber form is placed.
+    def current_ability
+      Ability.new(current_user, current_facility, UsersController.new)
+    end
+
+    private def update_user_params
+      params.require(:user).permit(:indala_number)
+    end
+
+    private def custom_authorize
```
jhanggi: Can we come up with a more specific name for this before_action?  Also, in this project we never put the `private` and `def` on the same lines.

vendor/engines/secure_rooms/app/controllers/secure_rooms/indala_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      ability = SecureRooms::IndalaAbility.new(current_user, current_facility)
+      head :not_authorized unless ability.can? :edit, User
```
jhanggi: I'm guessing that using `authorize!` doesn't work here because of the custom ability. Can we instead `raise CanCan::AccessDenied` and let `ApplicationController` handle it? This isn't using ajax, so a `head` without a body feels a little weird.

vendor/engines/secure_rooms/app/controllers/secure_rooms/indala_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+      else
+        flash[:error] = text("controllers.users.update.error", message: @user.errors.full_messages.to_sentence)
+        render action: :edit
+      end
+    end
+
+    # During rendering, we want to use the Ability as if we were using the
+    # UsersController within which this IndalaNumber form is placed.
+    def current_ability
+      Ability.new(current_user, current_facility, UsersController.new)
```
jhanggi: I'm not sure there's a better way, but this feels a little weird to me, especially the using `UsersController.new`. But just to be on the safe side, I would probably call this as `::Ability` to make sure we don't accidentally get the wrong thing at some point.

vendor/engines/secure_rooms/app/models/secure_rooms/indala_ability.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  class IndalaAbility
+
+    include CanCan::Ability
+
+    def initialize(user, facility)
+      return unless user && facility
+
+      if user.administrator? || user_has_role_at_facility?(user, facility)
+        can :edit, User
```
jhanggi: This can be shortened to `can [:edit, :update], User`. And I thought that CanCanCan aliased the actions, so can we get away with only one or the other?

vendor/engines/secure_rooms/app/models/secure_rooms/indala_ability.rb
```diff
@@ -0,0 +1,23 @@
...
+        can :update, User
+      end
+    end
+
+    private def user_has_role_at_facility?(user, facility)
```
jhanggi: I would probably just use `user.operator_of?(facility)` here. There are still a few rogue usages of this in controllers and views, but ideally we restrict usage of `operator_of?` to ability checks.

vendor/engines/secure_rooms/app/views/secure_rooms/indala_numbers/edit.html.haml
```diff
@@ -0,0 +1,17 @@
+= content_for :head_content do
+  = render "users/form_head"
```
jhanggi: `users/form_head` is just setting up some JS for keeping username and email address synced up as you type. Is it necessary here?

vendor/engines/secure_rooms/app/views/secure_rooms/indala_numbers/edit.html.haml
```diff
@@ -0,0 +1,17 @@
...
+
+= content_for :h1 do
```
jhanggi: Not sure why it's not showing, but the tabnav (Details, Orders, etc) is disappearing in this view.

vendor/engines/secure_rooms/app/views/secure_rooms/indala_numbers/edit.html.haml
```diff
@@ -0,0 +1,17 @@
...
+  = current_facility
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_users", sidenav_tab: "users"
+
+%h2= text("views.users.edit.head", user: @user.full_name)
+%p= text("views.users.edit.subhead")
```
jhanggi: I think we can kill this text. In this context "You may use this to edit external users. The email address and username must match." isn't relevant.

vendor/engines/secure_rooms/app/views/secure_rooms/shared/_indala_form_field.html.haml
```diff
@@ -0,0 +1,5 @@
+.control-group
+  %label.control-label
+    = User.human_attribute_name(:indala_number)
+    = link_to "Update", edit_facility_indala_number_path(current_facility, @user), class: "normal-weight"
```
jhanggi: Can we use `text("shared.update")` or another locale here?

vendor/engines/secure_rooms/config/routes.rb
```diff
@@ -4,5 +4,7 @@
       facility_product_routing_concern
       resources :price_policies, controller: "secure_room_price_policies", except: :show
     end
+
+    resources :indala_numbers, controller: "secure_rooms/indala_numbers", only: [:edit, :update]
```
jhanggi: I could be persuaded, especially if it's too complicated, but could we make the path for this `/facilities/:facility_id/users/:id/indala_numbers`?

vendor/engines/secure_rooms/spec/features/edit_indala_number_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+RSpec.describe "User's Indala Number" do
+  let(:facility) { FactoryGirl.create(:facility) }
+  let(:indala_user) { FactoryGirl.create(:user) }
+  let(:staff_user) { FactoryGirl.create(:user, :staff, facility: facility) }
+
+  describe "update" do
+    before do
+      login_as staff_user
+      visit facility_user_path(facility, indala_user)
+      click_link "Update"
```
jhanggi: Minor thing: is there a way to scope this finder down? It seems like this test is somewhat brittle because it could break if there were another "Update" link on the page. If there's not an easy way, it's fine.

vendor/engines/secure_rooms/spec/models/secure_rooms/indala_ability_spec.rb
```diff
@@ -0,0 +1,37 @@
+require "rails_helper"
+
+RSpec.describe SecureRooms::IndalaAbility do
+  subject(:ability) { described_class.new(current_user, facility) }
+  let(:indala_user) { FactoryGirl.create(:user) }
```
jhanggi: Can we get away with `build` or `build_stubbed` for the factories in this test? I like to start with one of the non-db builders until it turns out that it's trying to hit the DB for some reason or relations break.

vendor/engines/secure_rooms/db/migrate/20170220203130_add_indala_number_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndalaNumberToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :indala_number, :string
```
jhanggi: One thought I had is around the name of the database field. Indala seems very specific to NU (though, it could be more universal than I expect). What would you think of naming it something like `scan_card_number` or `card_number`. We would then override the locale/human_attribute_name in NU to "Indala Number". And on that note, even if we leave it as is, we should probably set the human_attribute_name to capitalize "Number".

vendor/engines/secure_rooms/db/migrate/20170220203130_add_indala_number_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddIndalaNumberToUser < ActiveRecord::Migration
+  def change
```
jhanggi: Code climate is flagging missing lines at the beginning/end of the class. I don't care so much in migrations, but I also think a happy rubocop is a benign rubcop.

vendor/engines/secure_rooms/app/controllers/secure_rooms/indala_numbers_controller.rb
waggles: Yeah I'm with you, but I couldn't find a better way to deal with the `if controller.is_a?(FacilityUsersController)` and as a result, the sidenav kept disappearing 😞

app/views/users/edit.html.haml
```diff
@@ -7,6 +7,9 @@
 = content_for :sidebar do
   = render "admin/shared/sidenav_users", sidenav_tab: "users"
 
+= content_for :tabnav do
+  = render :partial => 'admin/shared/tabnav_users', :locals => {:secondary_tab => 'details'}
```
jhanggi: Picky thing (and I'm assuming this was copy/paste from somewhere else): we're trying to transition to the shorter style for `render`: `render "admin/shared/tabnav_users", secondary_tab:  "details"` Also, using double quotes and 1.9 hash syntax.

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
+module SecureRooms
+
+  class CardNumbersController < ApplicationController
+
+    include TextHelpers::Translation
```
jhanggi: I thought this was included automatically. Is it not?

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+        redirect_to facility_user_path(current_facility, @user)
+      else
+        flash[:error] = text("controllers.users.update.error", message: @user.errors.full_messages.to_sentence)
+        render action: :edit
+      end
+    end
+
+    # During rendering, we want to use the Ability as if we were using the
+    # UsersController within which this card number form is placed.
+    def current_ability
```
jhanggi: Does this method need to be public?

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+
+    def edit; end
+
+    def update
+      if @user.update_attributes(update_user_params)
+        flash[:notice] = text("controllers.users.update.success")
+        redirect_to facility_user_path(current_facility, @user)
+      else
+        flash[:error] = text("controllers.users.update.error", message: @user.errors.full_messages.to_sentence)
+        render action: :edit
```
jhanggi: Can this be `render :edit` without the action?

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    private
+
+    def update_user_params
+      params.require(:user).permit(:card_number)
+    end
+
+    def load_and_authorize_user_with_card_number
+      @user = User.find(params[:user_id])
+      ability = SecureRooms::CardNumberAbility.new(current_user, current_facility)
+      raise CanCan::AccessDenied unless ability.can? :edit, User
```
jhanggi: I'm not sure it matters too much, but I think `ability.can? :edit, @user` is slightly more granular.

vendor/engines/secure_rooms/app/views/secure_rooms/card_numbers/edit.html.haml
```diff
@@ -0,0 +1,16 @@
+= content_for :h1 do
+  = current_facility
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_users", sidenav_tab: "users"
+
+= content_for :tabnav do
+  = render :partial => 'admin/shared/tabnav_users', :locals => {:secondary_tab => 'details'}
```
jhanggi: Use newer `render` style.

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_numbers_controller.rb
```diff
@@ -0,0 +1,54 @@
...
+    def initialize
+      @active_tab = "admin_users"
+      super
+    end
+
+    def edit; end
+
+    def update
+      if @user.update_attributes(update_user_params)
+        flash[:notice] = text("controllers.users.update.success")
```
jhanggi: We possibly could override `translation_scope` to return `"controllers.users"`, then we can can use text_helpers a little more succinctly.

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_numbers_controller.rb
waggles: yep, `NoMethodError` even at `protected`

PR #436, opened 2016-02-03
Docker oracle

benlinton:
WIP.  Please provide feedback.

PR #919, opened 2017-03-03
Instructions and scripts on running Oracle with Docker

jhanggi:
This is a pared down version of #436 from way back when. That branch sought to do a full environment with Docker. This only has instructions for getting the database set up and the necessary scripts. It assumes you'll be running the app in your native environment.

It also includes some updates for running `rake db:migrate` from scratch. It does not include the resulting Oracle-based `schema.rb`.

Meara and I can use this as a base for getting her NU environment up and running.

### DIFF
--- a/config/database.yml.oracle.template
+++ b/config/database.yml.oracle.template
@@ -1,15 +1,15 @@
 defaults: &defaults
   adapter: oracle_enhanced
-  database: 10.114.22.3:1521/xe
-  encoding: utf8
+  username: nucore_open_development
+  password: password
+  database: localhost:1521/xe
 
 development:
   <<: *defaults
   username: nucore_open_development
-  password:
+  password: password
 
 test:
   <<: *defaults
   username: nucore_open_test
-  password:
-
+  password: password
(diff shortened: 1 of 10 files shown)


docker/oracle/setup.sql
```diff
@@ -0,0 +1,20 @@
+alter profile default limit password_life_time unlimited;
+
+CREATE USER nucore_open_development
+    IDENTIFIED BY password
+    DEFAULT TABLESPACE users
```
jhanggi: This file will change somewhat in NU. We'll want to change the usernames, and NU should use `bc_nucore` as the tablespace.

jhanggi:
@meara any additional comments here? I'd like to merge this in if possible.

config/database.yml.oracle.template
```diff
@@ -1,15 +1,15 @@
 defaults: &defaults
   adapter: oracle_enhanced
-  database: 10.114.22.3:1521/xe
-  encoding: utf8
+  username: nucore_open_development
+  password: password
+  database: localhost:1521/xe
 
 development:
   <<: *defaults
```
meara: Did we ever figure out why the `defaults` inheritance wasn't working? On mine I just changed it to not use inheritance at all. That's my only concern really.

PR #920, opened 2017-03-03
Deployment configuration for open

jhanggi:
Still to do in separate PRs:

* Set up automated deployments from CI
* Configure rollbar

One thing that will be tricky is merging this down into the other branches. We'll want their versions of deploy/staging.rb,  config/deploy.rb,  and unicorn/stage.rb (although some are still using older capistrano formats, so those could probaby use an upgrade anyways.


Gemfile
```diff
@@ -82,7 +82,9 @@ group :development, :deployment do
   gem "capistrano-rails",   require: false
   gem "capistrano-rvm",     require: false
   gem "capistrano-bundler", require: false
+  gem "eye-patch",          require: false
   gem "whenever",           require: false
+  gem "unicorn",            require: false
```
jhanggi: Ignore the codeclimate here. I'll be taking care of that in the rollbar PR.

jhanggi:
@mlineen I think it might make sense for you to review this.

config/eye.yml.erb.template
```diff
@@ -0,0 +1,69 @@
+<%% app_root = "/home/nucore/nucore.stage.tablexi.com/current" %>
+<%% rails_env = ENV["RAILS_ENV"] %>
+---
+name: <%= @app_name.underscore.camelize %> (<%%= rails_env %>)
```
mlineen: @jhanggi Is the quoting correct here? This might be pulled from the .tt file which converts these into actual values?

config/unicorn/stage.rb
```diff
@@ -0,0 +1,36 @@
+worker_processes 4
```
mlineen: should we reduce this

mlineen:
@jhanggi pretty good; I think as long as the eye file is correct we can proceed

Capfile
```diff
@@ -1,15 +1,18 @@
...
 
-# Includes tasks from other gems included in your Gemfile
+# Include tasks from other gems included in your Gemfile
 require "capistrano/rvm"
-require "capistrano/bundler"
-require "capistrano/rails/assets"
-require "capistrano/rails/migrations"
+require "capistrano/rails"
+require "eye/patch/capistrano3"
+# require "rollbar/capistrano3"
```
meara: Why is this commented out and not just removed altogether?

circle.yml
```diff
@@ -7,6 +7,8 @@ machine:
 dependencies:
   bundler:
     without: [oracle]
+  pre:
+    - cp config/secrets.yml.template config/secrets.yml
```
meara: Confused what this is for--I thought the usual way to set secrets for circle was through their UI.

config/deploy/prod_server.rb.template
```diff
@@ -1,44 +0,0 @@
-set :application, "nucore"
```
meara: Are we removing this template because there is sufficient customization for each school that a template doesn't make sense? Or is this supplanted by other files?

Capfile
jhanggi: I'll be uncommenting it in the next PR which sets up rollbar.

circle.yml
jhanggi: We tend to use secrets/settings files over environment variables (which is what circle can set). In some setups, you commit the secrets file, but have it load environment variables. We've never gone down this route.

One thing this does is validates the template has all the things we need. We don't want nil errors, especially if you're going multi levels down in the secrets tree.

config/deploy/prod_server.rb.template
jhanggi: Mainly it's supplanted by `deploy.rb` and the individual deploy environments. Ultimately, this is a much older capistrano setup, and the dsl has become much simpler these days.

PR #921, opened 2017-03-03
[#140699067] Simplify the secure room new/edit form

jhanggi:
Other product types have many more options than secure rooms need (for
example, they should always be hidden and always require an access
list).

This moves some views around and pulls out some new partials. This
seemed to be about as clean as I could get it and still maintain the
views/admin/products hierarchy (which helps distinguish it from the
front-end views).


app/controllers/order_management/order_details_controller.rb
```diff
@@ -2,6 +2,7 @@ class OrderManagement::OrderDetailsController < ApplicationController
 
   include OrderDetailFileDownload
 
+  # before_action :authenticate_user!
```
jhanggi: Oops. Belongs to another branch.

app/helpers/products_helper.rb
```diff
@@ -39,12 +39,10 @@ def public_calendar_link(product)
   end
 
   def product_url_hint(product)
-    product_url_name = product.try(:url_name).presence || "url-name"
-
     public_send(
       "facility_#{product.class.name.underscore}_url",
       current_facility.url_name,
-      product_url_name,
+      "URL_NAME".html_safe,
```
jhanggi: Yeah, this is a hint. It had kind of a weird behavior where on the new form, it's "url-name", but the edit form it was the existing url_name. And then it would update if you changed it after submit, but not before. I wanted to use `<url-name>` or something like that, but Rails was URL-espaping the brackets, e.g. `%3Curl_name%3E`. That's why I went with all caps.

jhanggi: Any other suggestions?

waggles:
👍

PR #922, opened 2017-03-06
[#140077551] Add "Not Found" responses when identifiers don't match

waggles:
* Adds reader and controller as `CardReader` and `ControlDevice` (to avoid naming confusion with MVC controllers, hopefully)
* builds on #897 , when that's in I can clean up the commits a bit


db/schema.rb
```diff
@@ -88,6 +88,19 @@
   add_index "bundle_products", ["bundle_product_id"], name: "fk_bundle_prod_prod", using: :btree
   add_index "bundle_products", ["product_id"], name: "fk_bundle_prod_bundle", using: :btree
 
+  create_table "card_readers", force: :cascade do |t|
```
jhanggi: I'm wondering if we should prefix these with `secure_room_`. Part of me feels we should do the model classes could be namespaced (`SecureRooms::CardReader`). We do the namespace part for Projects (the model is `Projects::Project`), which I confess is a little awkward, much like `SecureRooms::Room` would have been. What do you think?

db/schema.rb
```diff
@@ -88,6 +88,19 @@
...
+    t.integer  "control_device_id", limit: 4, null: false
+    t.boolean  "entrance",                    null: false
+    t.datetime "created_at",                  null: false
+    t.datetime "updated_at",                  null: false
+  end
+
+  create_table "control_devices", force: :cascade do |t|
+    t.integer  "secure_room_id", limit: 4, null: false
```
jhanggi: I'd maybe like this to be genericized to `product_id`. I could go either way, though.

But we should add an index and a foreign key on it (fyi, mysql automatically creates an index when you create a foreign key, but oracle doesn't, so you'll need to explicitly `add_index` as well as adding the foreign key.

db/schema.rb
```diff
@@ -88,6 +88,19 @@
   add_index "bundle_products", ["bundle_product_id"], name: "fk_bundle_prod_prod", using: :btree
   add_index "bundle_products", ["product_id"], name: "fk_bundle_prod_bundle", using: :btree
 
+  create_table "card_readers", force: :cascade do |t|
+    t.integer  "control_device_id", limit: 4, null: false
+    t.boolean  "entrance",                    null: false
```
jhanggi: Hmm, I'm not crazy about this name, but I can't come up with anything better. I originally thought `direction`, but then true/false doesn't really make sense with it. It's probably fine for now and we can change it if we come up with something better.

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -14,4 +16,10 @@ def scan
     render json: response_json, status: :forbidden
   end
 
+  def load_models
+    @user = User.find_by!(card_number: params[:card_id])
+    @control_device = ControlDevice.find(params[:controller_id])
+    @card_reader = CardReader.find(params[:reader_id])
```
jhanggi: Since we have the foreign key, we should probably `@card_reader = @control_devise.card_readers.find` to scope it down a little better.

vendor/engines/secure_rooms/app/models/card_reader.rb
```diff
@@ -0,0 +1,5 @@
+class CardReader < ActiveRecord::Base
+
+   belongs_to :control_device
```
jhanggi: Rubocop is flagging a spacing issue ("Use 2 (not 3) spaces for indentation.")

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
```diff
@@ -7,11 +7,44 @@
...
+           controller_id: control_device_id,
+           reader_id: card_reader_id
+    end
+
+    subject { response }
+
+    let(:card_reader) { create :card_reader }
+    let(:card_user) { create :user, card_number: '123456' }
+
+    let(:card_reader_id) { card_reader.id }
```
jhanggi: I'm not really sure these `let`s are carrying their weight. I think you'd be fine if you used the dot version in the `post`.

vendor/engines/secure_rooms/spec/factories/secure_room.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :secure_room, class: SecureRoom do
```
jhanggi: I'm not sure the `class` option is necessary. Is there a way we could do `parent: :product` to get these attributes from what already exist?

db/schema.rb
```diff
@@ -688,6 +701,7 @@
     t.datetime "reset_password_sent_at"
     t.integer  "uid",                    limit: 4
     t.datetime "deactivated_at"
+    t.string   "card_number",            limit: 255
```
jhanggi: We should probably add an index on `card_number`

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
jhanggi: We might also need to add a custom `rescue_from` to these controllers. We don't want to send back HTML if we can avoid it. This might be a question for Dan, but should we return in the body why it's 404ing?

db/schema.rb
waggles: actually yeah I prefer that

db/schema.rb
waggles: haha yeah I actually could have just left it out for now, after like 30s of trying to think about what a good name was I just moved on

db/schema.rb
waggles: I'm cool with `product_id` I'll update

db/schema.rb
jhanggi: On second thought (aka shower thought): do we need two models? We can probably get away with a single model with a controller id and a reader id (and those might have been what you were thinking as opposed to database ids, which is how I read them). I suspect that would make it much easier to build the crud interface and then even add to a product. Having to enter the controller id twice in the interface isn't a big deal in my mind.

vendor/engines/secure_rooms/app/models/card_reader.rb
waggles: I just realized I could probably just run rubocop on `engines/secure_rooms` and spot these much more easily 😝

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
waggles: it just made the `context`s easier since there's no calling accidentally id on `nil`

db/schema.rb
waggles: oh wow yeah I hadn't fully thought about what the id's actually are--do we have examples?

db/schema.rb
waggles: Updated this over in #918 since it seemed to fit

db/schema.rb
waggles: since I'm not sure yet what data is going in these, I'd like to just leave it as is until I'm sure--it should be easy enough to migrate into whatever we need at that point

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
waggles: I added the error message for now and updated the README

vendor/engines/secure_rooms/spec/factories/secure_room.rb
waggles: ah okay I added in the `parent:` but I think it also needs the `class:` to generate the right model

db/schema.rb
jhanggi: What do you think of collapsing the two tables into one? If you really want to do that as a separate PR, that's fine. It just seems like that would be extra work as you would then need to write data migrations as opposed to just schema changes.

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
jhanggi: Ok, gotcha.

waggles:
@jhanggi oops I was pushing some updates there might be some oddities

db/schema.rb
waggles: ok yeah based on chat, I'll go ahead and reduce to one table

vendor/engines/secure_rooms/db/migrate/20170301120659_create_card_reader_and_control_device.rb
```diff
@@ -0,0 +1,19 @@
+class CreateCardReaderAndControlDevice < ActiveRecord::Migration
+
+  def change
+    create_table :secure_rooms_control_devices do |t|
+      t.references :product, null: false
+      t.foreign_key :products
```
jhanggi: So funny story: Oracle doesn't automatically create an index on this (or at least it didn't the last time I tried). We can leave it out for now to see what happens when we pull it down into NU, but I wanted to call that out.

waggles: oy, I continue to underestimate oracle's situation 😵 might as well add it if I'm updating here

jhanggi: We can leave it out for now; I'm actually curious if it's gotten fixed recently.

waggles:
@jhanggi ready for 👀

vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
```diff
@@ -0,0 +1,9 @@
+module SecureRooms
+
+  class CardReader < ActiveRecord::Base
+
+    belongs_to :secure_room, foreign_key: :product_id
```
jhanggi: Will validation come in the next PR (which I'm assuming will be the crud for these)? I'm cool if so, just want to check.

vendor/engines/secure_rooms/app/models/secure_room.rb
```diff
@@ -1,3 +1,6 @@
 class SecureRoom < Product
 
+  has_many :secure_rooms_control_devices
```
jhanggi: Looks like these `has_many` haven't been updated for the new model yet.

vendor/engines/secure_rooms/db/migrate/20170301120659_create_card_readers.rb
```diff
@@ -0,0 +1,15 @@
+class CreateCardReaders < ActiveRecord::Migration
+
+  def change
+    create_table :secure_rooms_card_readers do |t|
+      t.references :product, null: false
```
jhanggi: Still need an explicit index set because Oracle.

vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
waggles: yeah, this PR is primarily just to spit out the "deny" so these are only here for their finders to fail

PR #925, opened 2017-03-09
[#140891951] Return list of accounts for SecureRoom

waggles:
* Currently just stubs out accounts_for_product cause I didn't have a chance to set up full accounts/pricing. I'd vote can make a separate story for it if necessary.

### DIFF
--- a/vendor/engines/secure_rooms/README.md
+++ b/vendor/engines/secure_rooms/README.md
@@ -18,14 +18,14 @@ The door security hardware communicates through an API with the following endpoi
 
 * **Data Params**
 
-  * `card_id=[integer]`
-  * `controller_id=[integer]`
-  * `reader_id=[integer]`
+  * `card_number=[integer]`
+  * `controller_identifier=[integer]`
+  * `reader_identifier=[integer]`
   * `some_sort_of_account_id`
 
 * **Current Responses:**
 
-  There are three expected types of response from a properly formatted request. The following are implemented:
+  There are three expected types of response from a properly formatted request.
 
   **Access Denied:** The cardholder is not currently allowed to access the SecureRoom, and they are rejected.
 
@@ -34,7 +34,19 @@ The door security hardware communicates through an API with the following endpoi
 
       ```
       { response: "deny",
-        reason:   "I only know how to deny right now." }
+        reason:   "No accounts found." }
+      ```
+
+  **Access Granted:** We can infer the cardholder's payment method, so they are cleared immediately.
+
+  * **Code:** 200 OK
+  * **Body:**
+
+      ```
+      { response: "grant",
+        tablet_identifier: <id of associated tablet>,
+        name: "Full Nameofuser",
+        accounts:  [<list with one set of account model attributes>] }
       ```
 
   **Must Select Account:** The cardholder has multiple payment methods and must select which they would like to use.
@@ -44,6 +56,8 @@ The door security hardware communicates through an API with the following endpoi
 
       ```
       { response:  "select_account",
+        tablet_identifier: <id of associated tablet>,
+        name: "Full Nameofuser",
         accounts:  [<list of all account model attributes>] }
       ```
 
@@ -57,15 +71,4 @@ The door security hardware communicates through an API with the following endpoi
       ```
       { response: "deny",
         reason:   <Error Message stating missing record> }
-      ```
-
-* **Planned Responses:**
-
-  **Access Granted Immediately:** We can infer the cardholder's payment method, so they are cleared immediately.
-
-  * **Code:** 200 OK
-  * **Body:**
-
-      ```
-      { response: "grant" }
       ```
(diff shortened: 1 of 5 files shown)


waggles:
@danhodos can you double check the api data here to make sure it matches what we decided?

vendor/engines/secure_rooms/README.md
```diff
@@ -34,30 +34,19 @@ The door security hardware communicates through an API with the following endpoi
...
-
-  **Access Granted Immediately:** We can infer the cardholder's payment method, so they are cleared immediately.
+  **Access Granted:** We can infer the cardholder's payment method, so they are cleared immediately.
 
   * **Code:** 200 OK
   * **Body:**
 
       ```
-      { response: "grant" }
+      { response: "grant",
```
danhodos: With a `grant`, I don't think I need these extra details (e.g. `name`, `tablet_identifier`, etc.). You can send them if you want, but I think I can ignore them!

vendor/engines/secure_rooms/README.md
```diff
@@ -67,6 +56,19 @@ The door security hardware communicates through an API with the following endpoi
...
+        tablet_identifier: <id of associated tablet>,
+        name: "Full Nameofuser",
+        accounts:  [<list of all account model attributes>] }
+      ```
+
+  In addition, sending identifiers that do not map to existing records will generate a not found:
+
+  **Not Found:** One of the input identifiers does not match an existing record.
+
+  * **Code:** 404 Not Found
```
danhodos: Nice. I've coded `200` as `grant` and `300` as `select_account`; I'm treating _anything else_ as a deny. For a `403` I just straight `deny, and anything else (e.g. 404, 400, 500) I log a error and then deny.

Do you think instead of logging an error, though, I should report to rollbar? I think maybe yes...

vendor/engines/secure_rooms/app/presenters/secure_rooms/account_presenter.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def to_json
+      {
+        id: id,
+        type: type,
+        description: description,
+        account_number: account_number,
+        expiration_month: expiration_month,
+        expiration_year: expiration_year,
+        name: owner_user.try(:full_name)
```
danhodos: The only thing in this whole hash that I was surprised at was the `try`; why would there not be an owner name? Is this gonna be common?

If this is unavoidable and _not_ common: could we put something useful (so there's no weird blank space in the UI), like maybe the facility name or something?

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -10,19 +10,38 @@ class SecureRoomsApi::ScansController < ApplicationController
   before_action :load_models
 
   def scan
-    response_json = {
-      response: "deny",
-      reason: "I only know how to deny right now.",
-    }
-
-    render json: response_json, status: :forbidden
+    user_accounts = @user.accounts_for_product(@card_reader.secure_room)
```
danhodos: I think we're already starting to see the need for some kind of service object to keep this controller fresh and clean. Dunno that you have to do it in this PR, but I wouldn't wait too long past it, since this code is only going to get hairier (and it's gonna be nice to test all its various permutations outside the context of a controller environment).

jhanggi: I concur. Getting this into a service object will be very helpful.

vendor/engines/secure_rooms/app/presenters/secure_rooms/account_presenter.rb
jhanggi: There's a validation on Account so there should always be an owner. However, I have seen it happen a couple times where there is no owner (due to a manual deletion of a user--which can only be done from the console/directly in the database, or the result of a bug).

waggles: I found a method `owner_user_name` actually

vendor/engines/secure_rooms/README.md
waggles: yeah at the present time the code is shared.. so just ignore them 😄

vendor/engines/secure_rooms/README.md
waggles: by default probably yes, "until we can figure out what's going on" 😳

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
waggles: totally agreed, but this is one of those things that seems "helpful" here and "obviously necessary" for another working branch I have already so if we don't add it right now I don't think there's a risk of forgetting it needs to be done

PR #928, opened 2017-03-13
Display percent discount on schedule rule calendar instead of interval

jhanggi:
The reserve interval doesn't really make sense anymore now that the setting has
moved from the reserve interval has moved from the schedule rule to the instrument
(way back when in #80). Better to show the percent discount.

This will also make more sense when we genericize schedule rules to apply to
other product types like secure rooms which do not have a reserve interval.

This also adds the scheduling groups to the ScheduleRule#index view. I've intentionally
not done I18n as I plan to do that as part of genericizing this view for secure rooms.

In addition, it does some renaming to make it a presenter, which is closer to what it
actually is.

Before:
screen shot 2017-03-13 at 4 36 22 pm

After:
screen shot 2017-03-13 at 4 38 56 pm


spec/models/schedule_rule_spec.rb
```diff
@@ -297,7 +297,7 @@
       assert @rule2.valid?
 
       # find past tuesday, and build calendar objects
-      @tuesday    = ScheduleRule.sunday_last + 2.days
+      @tuesday = Time.current.beginning_of_week(:sunday).to_date + 2
 
       # times should be tue 1 am - 3 am
       @calendar1  = @rule1.as_calendar_object
```
meara: CodeClimate appears to be objecting to the extra space before the `=` here.

spec/models/schedule_rule_spec.rb
```diff
@@ -357,7 +357,7 @@
       assert @rule2.valid?
 
       # find past tuesday, and build calendar objects
-      @tuesday    = ScheduleRule.sunday_last + 2.days
+      @tuesday = Time.current.beginning_of_week(:sunday).to_date + 2
```
meara: If we're no longer using `.day` in this line, shouldn't we take it out in the line below as well?

jhanggi:
This is ready for a re-review.

PR #929, opened 2017-03-14
Improve setup

meara:
- update codeclimate/rubocop config
- update setup.sh to use `db:schema:load` rather than `db:migrate`
- update oracle setup info in readme
- remove unused `roles` table from initial migration

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -22,6 +22,9 @@ engines:
       - javascript
   rubocop:
     enabled: true
+    checks:
+      Rubocop/Style/EmptyMethod:
+        enabled: false
 ratings:
   paths:
   - "app/**"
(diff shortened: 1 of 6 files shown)


.codeclimate.yml
```diff
@@ -22,6 +22,9 @@ engines:
       - javascript
   rubocop:
     enabled: true
+    checks:
```
jhanggi: Can this work if we move this into `.rubocop.yml`?

jhanggi:
One question about rubocop and if we need the setting in codeclimate or if rubocop.yml is fine, otherwise looks good.

meara:
Both experiment and documentation suggest that the change in rubocop.yml is sufficient, which is cool. https://docs.codeclimate.com/v1.0/docs/rubocop

Thanks for the question, I learned a thing. :)

PR #931, opened 2017-03-16
[#140850111] Schedule Rules for SecureRooms

jhanggi:
This generalizes the ScheduleRulesController to support multiple product types as well as pulls some the model functionality into a concern.

I did a lot of moving around, so some code that existed before that's less than great might still be around without change.

Schedule groups will come in a following PR.


app/controllers/schedule_rules_controller.rb
```diff
@@ -16,83 +16,89 @@ def initialize
...
+
+  def product_id
+    params[product_key]
   end
 
-  def init_current_product
-    # required for tabnav_product
-    @product = @instrument
+  def product_key
+    params.except(:facility_id).keys.find { |k| k.end_with?("_id") }
```
mlineen: maybe check explicitly for one of [:instrument_id, :item_id, :service_id, :bundle_id] so that foo_id can't spoof this?

app/models/concerns/products/schedule_rule_support.rb
```diff
@@ -0,0 +1,36 @@
+module Products::ScheduleRuleSupport
+
+  extend ActiveSupport::Concern
+
+  included do
+    has_many :schedule_rules, foreign_key: :product_id, inverse_of: :product
+  end
+
+  def can_purchase?(group_ids = nil)
+    if schedule_rules.empty?
```
mlineen: do we prefer the guard clause style? not sure if rubocop would find this `return false if schedule_rules.empty?`

mlineen:
👍 I really like that we're adding behaviors to products

meara:
@jhanggi This is a ton of code--can we go through this together, maybe, and I'll give you the thumbs-up?

spec/factories/schedule_rules.rb
```diff
@@ -23,6 +23,10 @@
       on_sat true
     end
 
+    factory :weekend_schedule_rule do
+      weekend
```
meara: I like this refactoring, may implement this pattern in future. 😄

spec/controllers/schedule_rules_controller_spec.rb
```diff
@@ -1,178 +1,162 @@
...
 
     context "with restriction levels" do
-      before :each do
-        @restriction_levels = []
-        3.times do
-          @restriction_levels << FactoryGirl.create(:product_access_group, product_id: @instrument.id)
-        end
-        sign_in(@admin)
-      end
+      let!(:restriction_levels) { FactoryGirl.create_list(:product_access_group, 3, product_id: product.id) }
```
meara: Why are we calling these restriction levels? Can we use a name that is more in line with our terminology elsewhere in the app?

jhanggi:
@meara updated. Also had to rebase after merging in #928.

PR #936, opened 2017-03-18
[#140076489] Manage CardReaders

waggles:


vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+    def index
+      @card_readers = @product.card_readers
+    end
+
+    def new; end
+
+    def edit; end
+
+    def create
+      @card_reader.assign_attributes(card_reader_params)
```
jhanggi: To be consistent with our other controllers, we would normally write this as
```ruby
if @card_reader.update_attributes(card_reader_params)
  redirect_to facility_secure_room_card_readers_path(current_facility, @product)
else
  render :new
end
```

Same structure applies to `update`

jhanggi: Also, we would normally throw a flash message in here.

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/index.html.haml
```diff
@@ -0,0 +1,37 @@
...
+
+%p= link_to text("admin.shared.add", model: SecureRooms::CardReader.model_name.human), new_facility_secure_room_card_reader_path(current_facility, @product), class: "btn-add"
+
+- if @card_readers.empty?
+  %p.notice= text("views.admin.secure_rooms.card_readers.index.none")
+- else
+  %table.table.table-striped.table-hover.product_list
+    %thead
+      %tr
+        %th= "Description"
```
jhanggi: We should I18n these headers using `CardReader.human_attribute_name`.

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/index.html.haml
```diff
@@ -0,0 +1,37 @@
...
+      - @card_readers.each do |card_reader|
+        %tr
+          %td
+            = card_reader.description
+          %td
+            = card_reader.card_reader_number
+          %td
+            = card_reader.control_device_number
+          %td
+            = link_to "Edit", edit_facility_secure_room_card_reader_path(current_facility, @product, card_reader)
```
jhanggi: I18n (I think we have a `shared.edit` key for this). Same with `Remove` below.

vendor/engines/secure_rooms/db/migrate/20170316090613_add_unique_index_to_card_readers.rb
```diff
@@ -0,0 +1,12 @@
+class AddUniqueIndexToCardReaders < ActiveRecord::Migration
+
+  def change
+    add_index(
+      :secure_rooms_card_readers,
+      [:product_id, :card_reader_number, :control_device_number],
+      name: "index_secure_rooms_card_readers_on_product_id_and_both_numbers",
```
jhanggi: To support Oracle, this name can only be a maximum of 30 characters.

vendor/engines/secure_rooms/db/migrate/20170316090613_add_unique_index_to_card_readers.rb
```diff
@@ -0,0 +1,12 @@
+class AddUniqueIndexToCardReaders < ActiveRecord::Migration
+
+  def change
+    add_index(
+      :secure_rooms_card_readers,
+      [:product_id, :card_reader_number, :control_device_number],
```
jhanggi: I might need to think about this a little more, but should `[:card_reader_number, :control_device_number`]` be globally unique, not just within the product?

vendor/engines/secure_rooms/spec/features/manage_card_reader_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe "Managing CardReaders" do
+  let(:facility) { create(:setup_facility) }
+  let(:secure_room) { create(:secure_room, facility: facility) }
+  let(:admin) { create(:user, :administrator) }
+  before { login_as admin }
+
+  it "can create, edit, and destroy a card reader" do
```
jhanggi: This is probably fine, but I think I'd rather see one spec for each of the three, one for new/create, one for edit/update, and one for destroy.

vendor/engines/secure_rooms/spec/features/manage_card_reader_spec.rb
```diff
@@ -0,0 +1,42 @@
+require "rails_helper"
+
+RSpec.describe "Managing CardReaders" do
+  let(:facility) { create(:setup_facility) }
+  let(:secure_room) { create(:secure_room, facility: facility) }
+  let(:admin) { create(:user, :administrator) }
+  before { login_as admin }
```
jhanggi: Can we do this as the lowest possible permission level? I think we settled on facility admins and directors, but I could be wrong.

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/new.html.haml
```diff
@@ -0,0 +1,15 @@
...
+  = current_facility
+= content_for :sidebar do
+  = render "admin/shared/sidenav_product", sidenav_tab: "secure_rooms"
+= content_for :tabnav do
+  = render "admin/shared/tabnav_product", secondary_tab: "card_readers"
+
+%h2= @product
+%h3= text("views.admin.secure_rooms.card_readers.new.title")
+
+= simple_form_for @card_reader, url: facility_secure_room_card_readers_path(current_facility, @product) do |f|
```
jhanggi: Rails is smart enough that you should be able to (and I just tested this out) to use `= simple_form_for [current_facility, @product, @card_reader] do |f|` for both new and edit. It switches the URL based on if the card_reader is persisted or not.

That might also lead to some more sharing between the new and edit partials. At the very least, the input part should be shared. We sometimes do that in a `_form.html.haml` partial.

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/new.html.haml
```diff
@@ -0,0 +1,15 @@
...
+  = f.input :description
+  = f.input :card_reader_number
+  = f.input :control_device_number
+  = f.submit
```
jhanggi: To maintain visual consistency, this should have `class: "btn btn-primary"`.

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    private
+
+    def init_product
+      @product = current_facility.products(SecureRoom).find_by!(url_name: params[:secure_room_id])
+    end
+
+    def load_card_reader
+      scope = @product.card_readers
+      @card_reader = params[:id] ? scope.find(params[:id]) : scope.build
```
jhanggi: I think I like this pattern. Where did it come from?

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,63 @@
+module SecureRooms
+
+  class CardReadersController < ApplicationController
+
+    admin_tab :all
+    customer_tab :password
```
waggles: yeah haha I don't really know what these mean so some might sneak in along the way

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
waggles: hmm no place in particular just refactoring along the way

vendor/engines/secure_rooms/db/migrate/20170316090613_add_unique_index_to_card_readers.rb
waggles: good question.. that sort of makes sense to me.. should I update?

vendor/engines/secure_rooms/db/migrate/20170316090613_add_unique_index_to_card_readers.rb
waggles: 😡 of course haha

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/new.html.haml
waggles: updated, but re: sharing code--one the one hand, these two files are now comically similar, but on the other hand, this is only duplication not triplication of a small amount of code and we have no budget, but then I guess even thinking about it is burning money soo ugh I'll just change it

vendor/engines/secure_rooms/db/migrate/20170316090613_add_unique_index_to_card_readers.rb
waggles: I changed it, we can add the other later if needed

jhanggi: When we're getting the data from the reader, we're just getting a controller ID and a reader ID, right? So if we allowed duplicates across products, we could be talking to the wrong product, right? I think globally unique is the way to go.

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/new.html.haml
jhanggi: I'm a little less concerned with the `simple_form_for` definition, but I've found that having the actual form fields be separate between new and edit to actually cost more in the long run because of how easy it is for them to get out of sync. But thanks for changing it.

db/schema.rb
```diff
@@ -612,9 +612,10 @@
     t.string   "control_device_number", limit: 255
     t.datetime "created_at",                        null: false
     t.datetime "updated_at",                        null: false
+    t.string   "description",           limit: 255
   end
 
-  add_index "secure_rooms_card_readers", ["product_id"], name: "index_secure_rooms_card_readers_on_product_id", using: :btree
+  add_index "secure_rooms_card_readers", ["card_reader_number", "control_device_number"], name: "index_identifying_numbers", unique: true, using: :btree
```
jhanggi: It's probably fine, but this index now seems overly general. What about something like `idx_secure_room_reader`?

vendor/engines/secure_rooms/app/views/secure_rooms/shared/_tabnav_secure_room.html.haml
```diff
@@ -0,0 +1,2 @@
+- if @product.is_a?(SecureRoom)
+  = tab "Card Readers", facility_secure_room_card_readers_path(current_facility, @product), (secondary_tab == "card_readers")
```
jhanggi: I18n

db/schema.rb
jhanggi: Looks like we use `i_` in a few places, so `i_secure_room_reader_ids`?

jhanggi: Also, we need model validations against this. Currently there are no validations, even presence.

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+        render :new
+      end
+    end
+
+    def update
+      if @card_reader.update_attributes(card_reader_params)
+        flash[:notice] = text("update.success")
+        redirect_to facility_secure_room_card_readers_path(current_facility, @product)
+      else
+        flash.now[:error] = text("update.error", message: @user.errors.full_messages.to_sentence)
```
jhanggi: Can we avoid adding errors to the flash message, and add them either through `= f.error_messages` or just relying on simple_form's errors on the form elements themselves?

It's probably fine in this case, but it's caused cookie overflow problems in other models.

waggles:
@jhanggi is the code climate complaint worth addressing? it's similar I guess but I can't currently think of a refactor that's worth it

waggles:
me right now:
image

jhanggi:
I would say not worth addressing.

waggles:
ok cool ready for another look

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,82 @@
...
+      else
+        flash[:error] = text("destroy.error")
+      end
+      redirect_to facility_secure_room_card_readers_path(current_facility, @product)
+    end
+
+    protected
+
+    def translation_scope
+      "controllers.secure_rooms/card_readers"
```
jhanggi: What is the translation_scope by default here?

db/schema.rb
```diff
@@ -612,9 +612,10 @@
     t.string   "control_device_number", limit: 255
     t.datetime "created_at",                        null: false
     t.datetime "updated_at",                        null: false
+    t.string   "description",           limit: 255
```
jhanggi: What do we have in mind for the description field?

vendor/engines/secure_rooms/app/views/secure_rooms/card_readers/index.html.haml
```diff
@@ -0,0 +1,37 @@
...
+          %td
+            = card_reader.description
+          %td
+            = card_reader.card_reader_number
+          %td
+            = card_reader.control_device_number
+          %td
+            = link_to text("shared.edit"), edit_facility_secure_room_card_reader_path(current_facility, @product, card_reader)
+          %td
+            = link_to text("shared.remove"), facility_secure_room_card_reader_path(current_facility, @product, card_reader), data: { confirm: "Are you sure?" }, method: :delete
```
jhanggi: I18n the confirmation.

vendor/engines/secure_rooms/app/views/secure_rooms/shared/_tabnav_secure_room.html.haml
```diff
@@ -0,0 +1,2 @@
+- if @product.is_a?(SecureRoom)
```
jhanggi: Probably not a big deal for now, but to future-proof this, `@product.respond_to?(:card_readers)` might be better.

db/schema.rb
waggles: I was thinking optional string like "NW door" or something--assuming just the two numbers aren't the most useful to a human

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
waggles: apparently exactly that, I'll remove

vendor/engines/secure_rooms/app/views/secure_rooms/shared/_tabnav_secure_room.html.haml
waggles: I'm unsure of this, since the secure room name is baked into the route below and the file name, so it feels a little like we're asking the wrong question

vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
```diff
@@ -4,6 +4,8 @@ class CardReader < ActiveRecord::Base
 
     belongs_to :secure_room, foreign_key: :product_id
 
+    validates :card_reader_number, :control_device_number, presence: true
```
jhanggi: Should we also have a uniqueness validation?

vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
```diff
@@ -4,6 +4,9 @@ class CardReader < ActiveRecord::Base
...
+    validates :card_reader_number, uniqueness: true
```
jhanggi: Should this be only be unique within the scope of `control_device_number`?

waggles: I don't think so, I would assume this is like a serial number and thus unique to every reader, did we figure this out though at some point?

jhanggi: That might be a question for @danhodos. I thought reader_number would be something like 1, 2, or 3. I thought the controller number was more like a serial.

This validation should match the index, though, which is on both fields.

vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
```diff
@@ -4,6 +4,9 @@ class CardReader < ActiveRecord::Base
 
     belongs_to :secure_room, foreign_key: :product_id
 
+    validates :product_id, :card_reader_number, :control_device_number, presence: true
+    validates :card_reader_number, uniqueness: [:control_device_number]
```
jhanggi: I've never seen this style before, and I'm not seeing a reference for it. I would expect it to be `validates :card_reader_number, uniqueness: { scope: :control_device_number }`.

waggles: yep no idea what happened there

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+
+  class CardReadersController < ApplicationController
+
+    admin_tab :all
+
+    layout "two_column"
+
+    before_action :authenticate_user!
+    before_action :check_acting_as
+    before_action :authorize_card_reader
```
jhanggi: I think we're getting pretty close. I think we can use CanCan's `authorize_resource` method after we load it in `load_card_reader`, or possibly, even getting rid of `load_card_reader` and using `load_and_authorize_resource through: :product`. Then we don't need to add our own explicit authorization method.

vendor/engines/secure_rooms/app/controllers/secure_rooms/card_readers_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+    before_action :init_current_facility
+    before_action :init_product
+    before_action :load_card_reader, except: [:index]
+
+    def initialize
+      @active_tab = "secure_rooms"
+      super
+    end
+
+    def index
```
jhanggi: Here's a question, and I'm not 100% sure the answer to: should facility staff still be able to see the index page?

If so, then we need to hide their ability to see the edit/remove links on the index page.

If we decide they don't need to see the index view, then we should probably hide the tab, otherwise it will 403.

waggles: I think it will work better for this project if we defer some of these questions to the validation phase, where they can be discussed separately and their effect on budget can be more easily tracked.

waggles: In the meantime, let's just hide the tab.

vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
```diff
@@ -0,0 +1,17 @@
+module SecureRooms
+
+  class AbilityExtension
```
jhanggi: I was wondering if we could forego this ability extension and rely on the `Product` permissions, but that probably doesn't make sense. This probably gives us better control. We can always change it later.

PR #940, opened 2017-03-22
[#140076473] Add * to required fields in Product new/edit

jhanggi:
URL Name, Account, and Recharge Chart String on Product new/edit forms were missing the `*` marking them as required.

This also pulls the facility_account validation up from the individual products into `Product` so SecureRooms will also have the validation.


mlineen:
👍

app/views/admin/products/_account_fields.html.haml
```diff
@@ -1,5 +1,13 @@
 - if SettingsHelper.feature_on? :recharge_accounts
-  = f.input :facility_account_id, collection: current_facility.facility_accounts.active.collect {|fa| [ fa, fa.id ] }, hint: text("hints.deposit_account"), include_blank: false, label: FacilityAccount.model_name.human
+  -# include_blank: false is ignored if you use required: true
+  = f.input :facility_account_id,
+    label: FacilityAccount.model_name.human,
+    label_text: -> (label, _required, _) { "* #{label}" },
+    include_blank: false,
```
meara: Why are we keeping `include_blank: false` at all if it's just ignored?

lib/templates/haml/scaffold/_form.html.haml
```diff
@@ -1,10 +0,0 @@
-= simple_form_for(@<%= singular_table_name %>) do |f|
```
meara: What was this file used for? Not sure why we ever had it.

jhanggi: Looks like it's from default Rails generator. We don't use it anywhere. That's why I deleted it.

app/views/admin/products/_account_fields.html.haml
jhanggi: Maybe my comment wasn't clear. If I added `required: true`, then it would ignore `include_blank: false`. The comment was meant to explain why I wasn't using `required: true`, when that's the more obvious choice (`label_text` is kind of an obscure simple_form method). If you have a suggestion for a better comment, I'm all for it.

meara: Ohhhhh my eyes scanning github smushed the `required: true` from line 13 into this form. That makes way more sense.

For the comment, I would maybe explicitly note that you're using `label_text` instead of `required: true` in order to avoid ignoring `include_blank: false`. That's a little more obvious to me.

lib/templates/haml/scaffold/_form.html.haml
meara: 👍

PR #942, opened 2017-03-27
install haml_lint

meara:
- runs rubocop (and a few other linters) on haml views


Gemfile
```diff
@@ -73,6 +73,7 @@ gem "synaccess_connect", "0.2.2", github: "tablexi/synaccess"
 
 group :development do
   gem "coffeelint"
+  gem "haml_lint", github: "brigade/haml-lint", ref: "1384ec"
```
jhanggi: What's the reason for using the github ref instead of the released version of the gem?

jhanggi:
Is there a way to make this run within codeclimate? So it'll warn us of new issues?

Gemfile
meara: The `0.23.2` version had a bug that prevented us from excluding items in the `.haml-lint_todo.yml` from the build. The issue was fixed on master, but a new version hadn't yet been released.

Looks like they just released `0.24.0` this morning with the fix, so I am updating to use that release.

meara:
Looks like we can't run `haml-lint` on CodeClimate unless we built and released an entire engine for it. It would be much easier to run `haml-lint` on Circle, but I'm not sure if we want a linter to cause Circle to fail.

jhanggi:
I agree, we shouldn't fail circle because of a style issue. Hound could be a good idea, but CodeClimate does 90% of that, and I'm not sure if we want to add another integration that's outside of our standard toolkit.

Here's a question: if we half-update one of those files in the exclude file, will we get any warnings if we run it. Should we clear out the exclude file, and only run the linter on the files we change.

meara:
Ugh it is actually really complicated to write a script to compare only changed files.

`git diff --name-only develop | grep .haml | xargs haml-lint` works, but will error if there are any deleted files, and I suspect renamed files as well.

I feel like I'm starting to put in more time than is worth here in trying to optimize this. Thoughts on just merging this in as is, and we can just use it as a tool locally for now? Using the exclude file isn't optimal, but it's still better than nothing, and then exclude file is helpful if nothing else as a way to log to-dos. We could do period cleanups on some of the worst offending files and gradually cut down on the style to_dos. Thoughts?

jhanggi:
I think we can go ahead and merge this. Let's try to be vigilant about running this when we do PRs, as well as removing files we've fixed from the todo list (i.e. if a haml file gets changed in the process of a PR, it should be removed from todo if it was there). 👍

PR #943, opened 2017-03-28
Extract ScanAuthorizer

waggles:
Moves scanning logic


vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -5,51 +5,21 @@ class SecureRoomsApi::ScansController < ApplicationController
...
 
-  def report_missing_ids(error)
-    response_json = {
-      response: "deny",
-      reason: error.message,
-    }
-
-    render json: response_json, status: :not_found
+  def load_authorizer
+    @scan_authorizer = SecureRooms::ScanAuthorizer.new(
```
jhanggi: My general initial feeling is that we're now packing too much into the authorizer. I think I'd much rather see the authorizer take a User and a Product and return a grant/deny info. Then the controller is responsible for building the JSON response. Maybe there's another object in there that does the translating from the Authorizer to responses.

vendor/engines/secure_rooms/app/services/secure_rooms/scan_authorizer.rb
```diff
@@ -0,0 +1,85 @@
...
+      @response_status = :forbidden
+      @response_json = {
+        response: "deny",
+        reason: denial_reasons,
+      }
+    end
+
+    def denial_reasons
+      [].tap do |reasons|
+        reasons << text("errors.no_accounts") if user_accounts.empty?
```
jhanggi: Think we're actually going to want something that's more as soon as you get a denial return. I don't think we need to check and return every problem. Only the first.

danhodos:
I think it may be beneficial to change the software design here to be something like (and pardon my brevity, I'm typing this from a phone):

```ruby
class CheckAccess
  DEFAULT_RULES = [
    AllowFacilityOperatorRule.new,
    DenyAllRule.new,
  ]

  def initialize(rules = DEFAULT_RULES)
    @rules = rules
  end

  def authorize(user, other, params, here)
    answer = @rules.each do |rule|
      result = rule.call(user, other, params)
      break result unless result == :pass
  end
end

class AllowFacilityOperatorRule
  def call(user, other, params)
    return :grant if user.operator?
    :pass
  end
end
```

With each Rule object responsible for a single rule. I imagine each Rule would implement `call` and either return a `:grant`, a `deny`, or a `:pass` to continue processing the rules. It's kind of like a middleware pattern (like Rack) and could even be implemented as such if it makes sense to add some complexity.

In this way, each Rule becomes testable independently (yay!) and the service object can be seeded with dummy objects to just test its ability to process a set of rules.

Does this make sense? Typing code on the phone sucks.

danhodos:
And obviously you could return a :deny with a reason, etc.

Probably would say also keep your current Service object to interpret the result (:grant or whatever) and turn it into JSON and deal with the other pieces. Not sure about where to split that apart, but I'm a fan of small objects that do a cohesive bit of work

jhanggi:
Word, @danhodos. That's similar to what I was envisioning. How would you pass out a reason? Maybe there's a `Grant`, `Deny`, `Pass` object that has a `reason`?

What is the `here` argument?

waggles:
hm okay

vendor/engines/secure_rooms/config/locales/en.yml
```diff
@@ -1,4 +1,11 @@
 en:
+  services:
+    secure_rooms/scan_authorizer:
+      errors:
+        no_accounts: "No Accounts Found"
+        card_not_found: "User's Card not recognized"
+        reader_not_found: "CardReader not recognized"
+
```
waggles: oops this needs to be moved

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
```diff
@@ -31,7 +31,7 @@
           let(:card_user) { build :user }
 
           it { is_expected.to have_http_status(:not_found) }
-          it "is expected to contain the corresponding reason" do
+          xit "is expected to contain the corresponding reason" do
```
waggles: these are pending in this commit because there aren't yet any response objects with errors

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/selected_account_rule.rb
```diff
@@ -0,0 +1,15 @@
+module SecureRooms
+
+  module AccessRules
+
+    class SelectedAccountRule
+
+      def self.call(_user, _card_reader, _accounts, selected)
+        return :ok if selected.present?
```
waggles: this ends up looking weird, since the existence of `selected` means it was found in the array returned by `accounts_for_product`

waggles:
@danhodos when you have a chance can we chat about how you'll handle the "reason" field? these changes alter how that was set up and I'd like to clarify how the value will be used before updating its handling

waggles:
@jhanggi updated, next I'll merge in the latest master and push

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
```diff
@@ -0,0 +1,19 @@
...
+  module AccessRules
+
+    class AccountSelectionRule < BaseRule
+
+      def self.condition(_user, _card_reader, accounts, selected)
+        if selected.present?
+          Verdict.new(:grant)
+        elsif accounts.present? && selected.blank?
+          Verdict.new(:pending, "Must select Account")
+        end
```
jhanggi: Is there a possibility of this not hitting either of the conditions? Does this need to return a default `pass` verdict?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
```diff
@@ -0,0 +1,20 @@
+module SecureRooms
+
+  module AccessRules
+
+    class BaseRule
+
+      def self.call(user, card_reader, accounts, selected)
+        conditions_verdict = condition(user, card_reader, accounts, selected)
+        conditions_verdict || Verdict.new(:pass)
```
jhanggi: I guess this handles it if the `condition` returns nil. I'm starting to feel like the extra method isn't really carrying its weight anymore, and we might as well have each of them call `call` directly rather than delegating out to `condition`.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
jhanggi: What about the case of `accounts.present?` && `accounts.one?`? Shouldn't we grant at that point?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
```diff
@@ -0,0 +1,43 @@
...
+        @result_code == :pass
+      end
+
+      def has_result_code?(code)
+        @result_code == code
+      end
+
+      private
+
+      def status_for_code(result_code)
```
jhanggi: Part of me wonders if the HTTP translation should be part of the `Verdict` class or not. It sort of seems like a separate concern. I'm fine leaving it for now as we refactor further (I'd still prefer to see separate `GrantVerdict`, `DenyVerdict` and `PassVerdict` classes over this initialize with a symbol).

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
jhanggi: I'm also still leaning a bit towards doing the account fetching here, if only to save the `accounts` argument. You could always return the `accounts` as part of the `pending` verdict.

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -5,51 +5,40 @@ class SecureRoomsApi::ScansController < ApplicationController
...
-  def report_missing_ids(error)
-    response_json = {
-      response: "deny",
-      reason: error.message,
+  def build_json(access_verdict, accounts)
+    {
+      # TODO: (#140895375) return actual tablet_identifier
+      tablet_identifier: "abc123",
+      name: @user.full_name,
+      reason: access_verdict.reason,
```
jhanggi: Should we keep the `response` field as well so you can see it in the JSON as well as the HTTP status?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
```diff
@@ -0,0 +1,20 @@
+module SecureRooms
+
+  module AccessRules
+
+    class BaseRule
+
+      def self.call(user, card_reader, accounts, selected)
```
jhanggi: Do we think it might actually make more sense to pass the `product` rather than the `reader`?

vendor/engines/secure_rooms/app/services/secure_rooms/check_access.rb
```diff
@@ -0,0 +1,24 @@
+module SecureRooms
+
+  class CheckAccess
+
+    DEFAULT_RULES = [
+      AccessRules::OperatorRule,
+      AccessRules::AccountSelectionRule,
+      AccessRules::DefaultRestrictionRule,
```
jhanggi: Is there a strong reason to use the class method `call` rather than an instance method, and load a `new` as we loop over them?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
waggles: this is just here to avoid ending every `Rule` with `Verdict.new(:pass)` since that's more of a <default response>/<requirement for traversing the rules list> within `CheckAccess#authorize`  than a notable piece of the logic within each rule

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
waggles: Yeah I don't really like it here either TBH, let's see what we can do

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
waggles: that makes sense. @danhodos any preference/additional thoughts on this?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
waggles: hmmm we'd lose the ability to check the not-yet-implemented `direction` which might eventually be needed depending on the Occupancy code

vendor/engines/secure_rooms/app/services/secure_rooms/check_access.rb
waggles: nah, I'll move them over

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
waggles: oh, we did actually get the official word from the client that one account = instant grant, right? I forgot

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
jhanggi: Hmm, that's an interesting point. I think we only need to run through this list of rules if we're scanning in. Scanning out will probably be a different thing?

waggles:
@jhanggi updated

vendor/engines/secure_rooms/app/services/secure_rooms/check_access.rb
jhanggi: I think we should call the last item `DenyRule` or `DenyAllRule`. To me, since we're not handling the last case inside of `authorize`, explicitly saying what the last rule does will probably make it clearer than `default`.

vendor/engines/secure_rooms/app/presenters/secure_rooms/scan_response_presenter.rb
```diff
@@ -0,0 +1,40 @@
...
+    def response
+      {
+        status: status_for_code(@access_verdict.result_code),
+        json: {
+          # TODO: (#140895375) return actual tablet_identifier
+          tablet_identifier: "abc123",
+          name: @user.full_name,
+          response: @access_verdict.result_code,
+          reason: @access_verdict.reason,
+          accounts: SecureRooms::AccountPresenter.wrap(@accounts),
```
jhanggi: Do we need to return `accounts` for `grant` and `deny`? I'm not sure it hurt anything, so it might be fine for now.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
```diff
@@ -0,0 +1,26 @@
...
+      def initialize(result_code, reason = nil)
+        @result_code = result_code
+        @reason = reason
+      end
+
+      def pass?
+        @result_code == :pass
+      end
+
+      def has_result_code?(code)
```
jhanggi: Is this method used anymore?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
```diff
@@ -0,0 +1,26 @@
+module SecureRooms
+
+  module AccessRules
+
+    class Verdict
```
jhanggi: I still think I want these to separate classes, but I'm willing to punt on that until we get to some of the next rules and see if becomes necessary.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
```diff
@@ -0,0 +1,21 @@
+module SecureRooms
+
+  module AccessRules
+
+    class AccountSelectionRule < BaseRule
+
+      def evaluate
+        if @selected.present?
+          grant!
```
jhanggi: I like the `grant!`, `deny!` pattern you've set up here.

vendor/engines/secure_rooms/app/presenters/secure_rooms/scan_response_presenter.rb
waggles: unless it's specifically unwanted I'd just leave it in to reduce complexity for now

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
waggles: it's only used in the specs... it's not _exactly_ code just for testing purposes but it does kinda feel like it 😳 what's your opinion on keeping it in for just that readability?

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
waggles: they feel so similar to me right now, I'd prefer to wait and see on this. it does seems like it might become much more useful if we add just a few more rules eventually

waggles:
I also thought about moving "accounts" around but it still didn't quite feel right with what's in place right now. I'd prefer to just leave it for now and maybe revisit later if a separate change would also benefit from it.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
jhanggi: Ah, ok. I did a find on this method and didn't see it anywhere (because it's `have`). I'm cool with keeping it.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
jhanggi: I'm fine leaving them for now.

jhanggi:
I'm cool with everything here for now. We'll see how things change as we move forward.

vendor/engines/secure_rooms/app/services/secure_rooms/check_access.rb
jhanggi: Actually, this is the one thing that I would like to change before merging in.

waggles: oops missed the comment, I'll update

PR #944, opened 2017-03-29
[#134980] responsive daily view - wrap schedules in scrollable div

yoon:
[134980](https://pm.tablexi.com/issues/134980)


jhanggi:
We need those 10px back because we share the stylesheet with the admin's timeline view, which sometimes has toggle switches for relays. 

screen shot 2017-03-30 at 11 36 58 am

I tried adding it back, but then the left hand side of the text got cut off.

yoon:
@jhanggi I modified the width and the switch's position as an alternative.

screen shot 2017-04-26 at 11 15 48 am

yoon:
@jhanggi Added some space around toggle switch. Using more width means that the timeline is less "centered" on large screens (it wasn't actually centered, was just in a span10 with offset1).

I made the scrollable behavior happen only below the 980px breakpoint for the hamburger menu.

Finally, I adjusted the date selector above to work on the iPhone5 in portrait mode.

jhanggi:
I'm cool with the timelines no longer being centered-ish on the page, but is there a way we can center the date over the timelines? Right now it feels off-balance at large widths with the date header centered, but not the timelines.

PR #945, opened 2017-03-29
[#134672] responsive - reserve an instrument

yoon:
[134672](https://pm.tablexi.com/issues/134672)

- The dropdown arrows are set by the OS, and would be quite difficult to modify (it can be done, it just takes a lot of styling). I won't do it unless there's a strong opinion.
- Looks like there's already a tooltip set up in `reservations.js`, which works when `instrument.show_details?` is `true`.

screen shot 2017-03-29 at 10 41 28 am
screen shot 2017-03-29 at 10 41 39 am


jhanggi:
Overall looks pretty good in the main case. There's another use-case for this view where a user can edit a future reservation for extend an currently-ongoing reservation. 

First scenario: 
* Create a reservation in the future. 
* From "My Reservations", click the time in the "Reserved For" column.

Second scenario:
* "Create & Start" a reservation. You should be able to do this as long as the instrument is currently open.
* Then from "My Reservations", click the time in the "Reserved For" column.

There's a double well happening in both those cases.

screen shot 2017-03-29 at 5 55 10 pm

yoon:
@jhanggi thanks for catching that. I checked all the places where the `reservation_fields` partial is being used, and this was the only one with a well remaining.

screen shot 2017-04-26 at 11 22 38 am

jhanggi:
With the responsive feature flag disabled, I'm seeing a red box under the duration:

screen shot 2017-04-26 at 1 04 57 pm

yoon:
Apologies - I hadn't merged our recent changes to re-include `boostrap-responsive.scss` in regardless of the feature flag.

PR #946, opened 2017-03-29
restore use of bootstrap-responsive stylesheet without triggering hamburger menu

yoon:


app/assets/stylesheets/responsive.scss
```diff
@@ -1,2 +1 @@
-@import "bootstrap-responsive";
 @import "bootstrap-responsive-overrides";
```
jhanggi: Don't need to do it now, but would it make sense to now pull whatever is in the `-overrides` file straight into this one?

yoon: Hmm. I was thinking about that... yeah, and I'll split them out again if needed.

jhanggi: Do you want to do that as part of this PR, or later? I'm good either way, just wanting to confirm if I should merge this or not.

yoon: Done, please merge.

PR #950, opened 2017-04-04
[#134989] basic email logger

meara:
- https://pm.tablexi.com/issues/134989

### DIFF
--- a/config/initializers/mail_logger.rb
+++ b/config/initializers/mail_logger.rb
@@ -1,7 +1,9 @@
 class MailLogger < ActiveSupport::LogSubscriber
   def deliver(event)
     info do
-      event.payload
+      pairs = ["sent_at: #{event.time.to_s}"]
+      event.payload.select { |k, v| [:mailer, :subject, :to, :from].include?(k) }.each_pair { |k, v| pairs << "#{k}: #{v}" }
+      "\n>>>MailLogger>>> " + pairs.join(", ")
     end
   end
 end
--- a/config/initializers/mail_logger.rb
+++ b/config/initializers/mail_logger.rb
@@ -1,0 +1,9 @@
+class MailLogger < ActiveSupport::LogSubscriber
+  def deliver(event)
+    info do
+      event.payload
+    end
+  end
+end
+
+MailLogger.attach_to :action_mailer
--- a/config/initializers/mail_logger.rb
+++ b/config/initializers/mail_logger.rb
@@ -1,11 +1,13 @@
 class MailLogger < ActiveSupport::LogSubscriber
+
   def deliver(event)
     info do
       pairs = ["sent_at: #{event.time.to_s}"]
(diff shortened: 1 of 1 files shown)


meara:
@jhanggi some questions on how we want to proceed here.

As far as I can tell, there's a pretty strict limit on what information we can get out of an event. On the event itself, basically the only useful info is the timestamp and the `payload` ([docs](http://api.rubyonrails.org/classes/ActiveSupport/Notifications/Event.html)).

What is in the payload is determined by the type of event, and as far as I can tell, can't be changed or added to. For actionmailer#deliver, it's fairly basic: http://guides.rubyonrails.org/active_support_instrumentation.html#deliver-action-mailer

Back when we originally paired on this, we discussed wanting to add additional data to the email logger, like the `facility_id` or `statement_id`, if available. I haven't been able to come up with a good way of doing this. I do have thoughts on two not-great options.

1) We create our own custom event rather than using the actionmailer#deliver event and trigger it every time we send an email. This seems a little gross, as it risks decoupling our logging event from the actual email sending, and also I think we would have to manually kick it off from each specific mailer in order to have access to the data that we're interested in.

2) We add data to the actual body of the email and somehow hide it, so that we can access it from the encoded form of the email that is present in the payload. This seems similarly impractical as well as a security risk.

Thoughts?

config/initializers/mail_logger.rb
```diff
@@ -0,0 +1,11 @@
+class MailLogger < ActiveSupport::LogSubscriber
+  def deliver(event)
+    info do
+      pairs = ["sent_at: #{event.time.to_s}"]
+      event.payload.select { |k, v| [:mailer, :subject, :to, :from].include?(k) }.each_pair { |k, v| pairs << "#{k}: #{v}" }
```
jhanggi: Can we use `event.payload.slice(:mailer, ...).each_pair`?

jhanggi:
We mainly send either through something that inherits from `BaseMailer#mail` or through `Notifier#send_nucore_mail` (which is the older dumping ground for all emails). We could possibly hook into those two points. We could check if instance variables like `@facility` or `@account` exist. Inside the mailer, you can get the instance variables with `view_assigns`. 

I think starting with at least seeing subject, etc will be a good start. We could even log the body. If we decide to go that route, I'd probably recommend we do that only on UIC. It might even be a good idea to do that now, and then once we've sorted out what we need, we can decide to pull it up into open.

meara:
I'm a little confused about your comment--are you suggesting triggering a custom event from within `Notifier#send_nucore_mail` and logging that? If I'm understanding you correctly, that's basically plan #1 from my comment above. We can certainly do that, but as you say it wouldn't catch emails that aren't sent through the notifier.  Do we want to try something like that in a separate PR on UIC and merge this one on open?

jhanggi:
I was kind of exploring how dirty it would be. I think we go with this as is, and see if it helps us track down problems before we spend more time on it. 

👍

PR #951, opened 2017-04-05
[#140076489] Add direction to SecureRoom card readers

jhanggi:


vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
```diff
@@ -9,6 +9,26 @@ class CardReader < ActiveRecord::Base
 
     delegate :facility, to: :secure_room
 
+    alias_attribute :ingress, :direction_in
+
+    def egress?
+      !ingress?
```
waggles: 👌

vendor/engines/secure_rooms/spec/models/secure_rooms/card_reader_spec.rb
```diff
@@ -0,0 +1,34 @@
+require "rails_helper"
+
+RSpec.describe SecureRooms::CardReader do
+  it { is_expected.to validate_presence_of :product_id }
+  it { is_expected.to validate_presence_of :card_reader_number }
+  it { is_expected.to validate_presence_of :control_device_number }
```
waggles: do we want to require direction?

db/schema.rb
```diff
@@ -607,15 +607,17 @@
...
-    t.integer  "product_id",            limit: 4,   null: false
+    t.integer  "product_id",            limit: 4,                  null: false
     t.string   "card_reader_number",    limit: 255
     t.string   "control_device_number", limit: 255
-    t.datetime "created_at",                        null: false
-    t.datetime "updated_at",                        null: false
+    t.datetime "created_at",                                       null: false
+    t.datetime "updated_at",                                       null: false
     t.string   "description",           limit: 255
+    t.boolean  "direction_in",                      default: true, null: false
```
waggles: oh I guess since we have `null: false` we should add the validation

jhanggi: We default to `true`, and there's no way to set it to anything but true/false in the interface, so if we hit the db null constraint, it means someone's up to no good. Basically, I don't feel like a validation is necessary.

vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
jhanggi: I initially was just going to call it `in`, but `in?` conflict's with `x.in?(array)`.

db/schema.rb
waggles: fine by me 👍

PR #952, opened 2017-04-05
[#140895375] Add tablet token to card readers

jhanggi:
I went with `tablet_token` since that feels closer to what it is rather than an identifier. I left the API key the same since the doors service looks for that.

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
@@ -4,12 +4,14 @@ module SecureRooms
 
     belongs_to :secure_room, foreign_key: :product_id
 
-    validates :product_id, :card_reader_number, :control_device_number, presence: true
-    validates :card_reader_number, uniqueness: { scope: :control_device_number }
-
     delegate :facility, to: :secure_room
 
     alias_attribute :ingress, :direction_in
+
+    validates :product_id, :card_reader_number, :control_device_number, presence: true
+    validates :card_reader_number, uniqueness: { scope: :control_device_number }
+
+    before_create :set_tablet_token
 
     def egress?
       !ingress?
@@ -27,6 +29,10 @@ module SecureRooms
 
     def attribute_translation_scope
       "#{self.class.i18n_scope}.attributes.#{model_name.i18n_key}"
+    end
+
+    def set_tablet_token
+      self.tablet_token ||= ("A".."Z").to_a.sample(12).join
     end
 
   end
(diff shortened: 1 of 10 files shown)


vendor/engines/secure_rooms/app/models/secure_rooms/card_reader.rb
```diff
@@ -29,6 +31,10 @@ def attribute_translation_scope
       "#{self.class.i18n_scope}.attributes.#{model_name.i18n_key}"
     end
 
+    def set_tablet_token
+      self.tablet_token ||= ("A".."Z").to_a.sample(12).join
```
waggles: do we want any checks of uniqueness of this token?

jhanggi: Yeah, that's probably a good idea.

vendor/engines/secure_rooms/db/migrate/20170405182534_add_table_token_to_card_readers.rb
```diff
@@ -0,0 +1,12 @@
+class AddTableTokenToCardReaders < ActiveRecord::Migration
+
+  class SecureRooms::CardReader < ActiveRecord::Base
+  end
+
+  def change
+    add_column :secure_rooms_card_readers, :tablet_token, :string
```
waggles: since we are now looking up by this field it might be useful to have it indexed

jhanggi: Sure.

And I just caught that I misspelled the filename/class name. TableToken.

jhanggi: And updated.

vendor/engines/secure_rooms/db/migrate/20170405182534_add_tablet_token_to_card_readers.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  class SecureRooms::CardReader < ActiveRecord::Base
+  end
+
+  def change
+    add_column :secure_rooms_card_readers, :tablet_token, :string
+    add_index :secure_rooms_card_readers, :tablet_token, unique: true
+
+    SecureRooms::CardReader.reset_column_information
+    SecureRooms::CardReader.find_each { |cr| cr.update_column(:tablet_token, ("A".."Z").to_a.sample(12).join) }
```
jhanggi: Conceivably, this could cause a uniqueness violation, but it seemed unlikely enough here, and it didn't seem worth bringing in the extra complexity I added to the main version of the model.

jhanggi:
This has been updated and rebased to fix merge conflicts.

waggles:
👍

PR #953, opened 2017-04-05
[#140895509], [#140849799], [#140849739] Add additional rules

waggles:
* Adds a new `Rule` specifically denying access if no accounts are found.
* Assumes `params[:account_identifier]` is our internal id (which was returned by initial scan) and uses it to select an account.
* +some cleanup


vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
```diff
@@ -43,20 +43,32 @@
 
     describe "positive responses" do
       context "with multiple accounts" do
+        let(:accounts) { create_list(:account, 3, :with_account_owner, owner: card_user) }
```
jhanggi: I think this will conflict with my #952, but we'll see who gets in first!

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/valid_account_rule.rb
```diff
@@ -0,0 +1,15 @@
+module SecureRooms
+
+  module AccessRules
+
+    class ValidAccountRule < BaseRule
+
+      def evaluate
+        deny! "User has no valid accounts for this Product" if @accounts.blank?
```
jhanggi: Not a big deal, but would it be simpler/clearer if this were added into the existing account checking class?

jhanggi:
Can you add the ticket number to the PR title?

waggles:
@jhanggi updated, plus I added the extra rules in as well

jhanggi:
It's probably too late to break this up into multiple commits, one for each story?

waggles:
I can break it up, since it's fairly separate by file

waggles:
See #954

PR #955, opened 2017-04-06
[#140850873] Restrict secure rooms by schedule rule

jhanggi:
* Users cannot log in outside of available schedule rules
* If schedule groups are enabled, the user needs to be part of a group for the current rule
* I also cleaned up a little bit on the access rules like making the account list and selected account optional.

### DIFF
--- a/vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
+++ b/vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
@@ -11,7 +11,8 @@ RSpec.describe SecureRoomsApi::ScansController do
 
     subject { response }
 
-    let(:card_reader) { create(:card_reader, tablet_token: "TABLETID") }
+    let(:secure_room) { create(:secure_room, :with_schedule_rule) }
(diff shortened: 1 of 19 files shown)


vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
```diff
@@ -11,7 +11,8 @@
 
     subject { response }
 
-    let(:card_reader) { create(:card_reader, tablet_token: "TABLETID") }
+    let(:secure_room) { create(:secure_room, :with_schedule_rule) }
+    let(:card_reader) { create(:card_reader, tablet_token: "TABLETID", secure_room: secure_room) }
```
waggles: should we add `tablet_token` to the factory?

waggles:
👍

vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/scans_controller_spec.rb
jhanggi: I decided not to add it to  the factory because it gets generated automatically. I chose to explicitly set it here so it's a little clearer the value we expect to show up in the json does.

PR #956, opened 2017-04-07
[#135296] hide reviewed_at from views if zero-day review period

meara:
- https://pm.tablexi.com/issues/135296

### DIFF
--- a/app/models/reports/export_raw.rb
+++ b/app/models/reports/export_raw.rb
@@ -88,7 +88,7 @@ module Reports
       if SettingsHelper.has_review_period?
         hash
       else
-        hash.delete_if { |k, _v| k == :reviewed_at }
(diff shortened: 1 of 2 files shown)


app/models/reports/export_raw.rb
```diff
@@ -85,6 +85,11 @@ def default_report_hash
         reconciled_note: :reconciled_note,
         reconciled_at: :reconciled_at,
       }
+      if SettingsHelper.has_review_period?
+        hash
+      else
+        hash.delete_if { |k, _v| k == :reviewed_at }
```
jhanggi: I think we should also kill off the `disputed_` fields as well since they also go away with a zero-day review period. 

I think this might be a little more concise for getting rid of the fields.

```
hash.except(:reviewed_at, ...)
```

jhanggi:
👍

PR #957, opened 2017-04-10
[#140895509] Update readme to current API status

waggles:


waggles:
not sure about that random spec failure, should we be looking into that

PR #958, opened 2017-04-10
[WIP] [#140077839] Update operator access restrict admin

waggles:
I had a small chunk of time so I took a peek at this since it seemed small--if this requires too much additional discussion beyond just a quick chat about roles, we can drop it since it's not above the priority line

* On first add of a spec for `create :user, :administrator` it was immediately green, so I switched the `build` for a `create` to avoid the false positive
* `Role#operator_of?`, which we currently use, includes the check `administrator`? but is otherwise helpful, so I added a check against that role specifically. I'm not sure business-wise why we use a special case in our situation, so I'm not quite sure if this is the correct fix.

### DIFF
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/operator_rule.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/operator_rule.rb
@@ -5,7 +5,7 @@ module SecureRooms
     class OperatorRule < BaseRule
 
       def evaluate
-        grant! if user.operator_of?(card_reader.facility)
+        grant! if user.operator_of?(card_reader.facility) && !user.administrator?
       end
 
     end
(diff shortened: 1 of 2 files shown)


vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/operator_rule.rb
```diff
@@ -5,7 +5,7 @@ module AccessRules
     class OperatorRule < BaseRule
 
       def evaluate
-        grant! if user.operator_of?(card_reader.facility)
+        grant! if user.operator_of?(card_reader.facility) && !user.administrator?
```
jhanggi: I don't think this is quite the right fix. There's the case where a user is an operator_of _and_ is a global admin. I think in that case, we would want to grant.

PR #959, opened 2017-04-10
[#140849913] Add events

waggles:
**In Progress**: Adds event concept to scanning and updates `SecureRoomsApi::ScansController` with an `AccessManager` that handles validating access and creating events.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170405182534) do
+ActiveRecord::Schema.define(version: 20170406162317) do
 
   create_table "account_users", force: :cascade do |t|
     t.integer  "account_id", limit: 4,  null: false
@@ -620,6 +620,20 @@ ActiveRecord::Schema.define(version: 20170405182534) do
   add_index "secure_rooms_card_readers", ["card_reader_number", "control_device_number"], name: "i_secure_room_reader_ids", unique: true, using: :btree
   add_index "secure_rooms_card_readers", ["product_id"], name: "index_secure_rooms_card_readers_on_product_id", using: :btree
   add_index "secure_rooms_card_readers", ["tablet_token"], name: "index_secure_rooms_card_readers_on_tablet_token", unique: true, using: :btree
+
+  create_table "secure_rooms_events", force: :cascade do |t|
+    t.integer  "card_reader_id", limit: 4
+    t.integer  "product_id",     limit: 4
+    t.integer  "user_id",        limit: 4
+    t.datetime "occurred_at"
+    t.string   "outcome",        limit: 255
+    t.datetime "created_at",                 null: false
+    t.datetime "updated_at",                 null: false
+  end
+
+  add_index "secure_rooms_events", ["card_reader_id"], name: "index_secure_rooms_events_on_card_reader_id", using: :btree
+  add_index "secure_rooms_events", ["product_id"], name: "index_secure_rooms_events_on_product_id", using: :btree
+  add_index "secure_rooms_events", ["user_id"], name: "index_secure_rooms_events_on_user_id", using: :btree
 
   create_table "splits", force: :cascade do |t|
     t.integer "parent_split_account_id", limit: 4,                         null: false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -622,17 +622,16 @@ ActiveRecord::Schema.define(version: 20170406162317) do
   add_index "secure_rooms_card_readers", ["tablet_token"], name: "index_secure_rooms_card_readers_on_tablet_token", unique: true, using: :btree
 
   create_table "secure_rooms_events", force: :cascade do |t|
-    t.integer  "card_reader_id", limit: 4
-    t.integer  "product_id",     limit: 4
-    t.integer  "user_id",        limit: 4
+    t.integer  "card_reader_id",  limit: 4
+    t.integer  "user_id",         limit: 4
     t.datetime "occurred_at"
-    t.string   "outcome",        limit: 255
-    t.datetime "created_at",                 null: false
-    t.datetime "updated_at",                 null: false
+    t.string   "outcome",         limit: 255
+    t.string   "outcome_details", limit: 255
+    t.datetime "created_at",                  null: false
+    t.datetime "updated_at",                  null: false
   end
 
   add_index "secure_rooms_events", ["card_reader_id"], name: "index_secure_rooms_events_on_card_reader_id", using: :btree
-  add_index "secure_rooms_events", ["product_id"], name: "index_secure_rooms_events_on_product_id", using: :btree
   add_index "secure_rooms_events", ["user_id"], name: "index_secure_rooms_events_on_user_id", using: :btree
 
   create_table "splits", force: :cascade do |t|
--- a/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
+++ b/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
@@ -8,18 +8,17 @@ class SecureRoomsApi::ScansController < ApplicationController
   before_action :load_user_and_reader
 
   def scan
-    accounts = @user.accounts_for_product(@card_reader.secure_room)
     requested_id = params[:account_identifier].to_i
-    selected_account = accounts.find { |account| account.id == requested_id }
 
-    access_verdict = SecureRooms::CheckAccess.new.authorize(
+    access_manager = SecureRooms::AccessManager.new(@user, @card_reader, requested_id)
+    access_manager.process
+
+    render SecureRooms::ScanResponsePresenter.new(
       @user,
       @card_reader,
-      accounts: accounts,
-      selected_account: selected_account,
-    )
-
-    render SecureRooms::ScanResponsePresenter.new(@user, @card_reader, access_verdict, accounts).response
+      access_manager.verdict,
+      access_manager.accounts,
+    ).response
   end
 
   private
--- a/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
+++ b/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
@@ -8,21 +8,15 @@ class SecureRoomsApi::ScansController < ApplicationController
   before_action :load_user_and_reader
 
   def scan
-    requested_account_id = params[:account_identifier].to_i
-
     verdict = SecureRooms::CheckAccess.new.authorize(
       @user,
       @card_reader,
-      requested_account_id: requested_account_id,
+      requested_account_id: params[:account_identifier],
     )
 
-    SecureRooms::AccessManager.new(
-      verdict,
-    ).process
+    SecureRooms::AccessManager.process(verdict)
 
-    render SecureRooms::ScanResponsePresenter.new(
-      verdict,
-    ).response
+    render SecureRooms::ScanResponsePresenter.present(verdict)
   end
 
   private
--- a/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
+++ b/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
@@ -8,16 +8,20 @@ class SecureRoomsApi::ScansController < ApplicationController
   before_action :load_user_and_reader
 
   def scan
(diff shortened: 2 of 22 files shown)


db/schema.rb
```diff
@@ -621,6 +621,20 @@
   add_index "secure_rooms_card_readers", ["product_id"], name: "index_secure_rooms_card_readers_on_product_id", using: :btree
   add_index "secure_rooms_card_readers", ["tablet_token"], name: "index_secure_rooms_card_readers_on_tablet_token", unique: true, using: :btree
 
+  create_table "secure_rooms_events", force: :cascade do |t|
+    t.integer  "card_reader_id", limit: 4
+    t.integer  "product_id",     limit: 4
+    t.integer  "user_id",        limit: 4
+    t.datetime "occurred_at"
```
jhanggi: Do we actually need a separate field for this? Can we use `created_at` where we need it?

vendor/engines/secure_rooms/app/services/secure_rooms/access_manager.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def process
+      @verdict = create_verdict
+      event = create_event(verdict)
+    end
+
+    # TODO: Extract
+    def create_verdict
+      SecureRooms::CheckAccess.new.authorize(
```
jhanggi: Would it be worth building this in the controller and passing the access checker into the initializer for this as a dependency injection? Not sure if it would help or not.

vendor/engines/secure_rooms/app/services/secure_rooms/access_manager.rb
```diff
@@ -0,0 +1,48 @@
...
+    attr_reader :verdict
+
+    def initialize(user, card_reader, requested_account_id)
+      @user = user
+      @card_reader = card_reader
+      @requested_account_id = requested_account_id
+    end
+
+    def process
+      @verdict = create_verdict
```
jhanggi: This is just an idea: what if the verdict could house a selected account (for grants) and a list of accounts (for pending)? Then this method could return a verdict, so you wouldn't need to use the instance of this class anymore since there would be less state.

vendor/engines/secure_rooms/app/services/secure_rooms/access_manager.rb
```diff
@@ -0,0 +1,48 @@
...
+      )
+    end
+
+    # TODO: Extract
+    def create_event(verdict)
+      Event.create!(
+        occurred_at: Time.current,
+        card_reader: @card_reader,
+        user: @user,
+        outcome: verdict,
```
jhanggi: What actually gets written to the DB? I don't think we have a `to_s` on `Verdict` yet, so I'd be surprised if it writes the `result_code`.

db/schema.rb
waggles: Since this timestamp is related to billing for use of the product, I'm less comfortable with depending on `created_at` to store it.

vendor/engines/secure_rooms/app/services/secure_rooms/access_manager.rb
waggles: nothing useful, this is certainly missing a `TODO`

waggles:
In this update, `Verdict` represents the access verdict for a `User` to a `CardReader` with selected `Account` (maybe). `Verdict` now holds any additional information to be passed back to the client; therefore `ScanResponsePresenter` is now actually a `VerdictPresenter`.

In `ScansController`, we generate a `Verdict` and then present it. In between, we call out to the `AccessManager` which is now responsible internally for creating events and then managing occupancies. At this point, updating that logic shouldn't affect the controller's job of generating and returning the `Verdict`.

I'll likely rename `ScanResponsePresenter` and then update the `AccessManager` now.

db/schema.rb
jhanggi: 🆗

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -8,18 +8,21 @@ class SecureRoomsApi::ScansController < ApplicationController
   before_action :load_user_and_reader
 
   def scan
-    accounts = @user.accounts_for_product(@card_reader.secure_room)
-    requested_id = params[:account_identifier].to_i
-    selected_account = accounts.find { |account| account.id == requested_id }
+    requested_account_id = params[:account_identifier].to_i
```
jhanggi: Part of me feels this shouldn't get `to_i`. That implies that account_id `0` has meaning. I think `nil` has more meaning that `0` in this case. Either way, it can probably be pulled down to where it's actually used.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
```diff
@@ -5,25 +5,24 @@ module AccessRules
...
       def evaluate
+        accounts = user.accounts_for_product(card_reader.secure_room)
+        selected_account = accounts.find { |account| account.id == requested_account_id }
+
         if accounts.blank?
-          deny! "User has no valid accounts for this Product"
+          deny!(reason: "User has no valid accounts for this Product")
         elsif selected_account.present?
-          grant!
+          grant!(accounts: accounts)
```
jhanggi: Do we need to add the selected account to the verdict?

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -8,18 +8,21 @@ class SecureRoomsApi::ScansController < ApplicationController
...
+    verdict = SecureRooms::CheckAccess.new.authorize(
       @user,
       @card_reader,
-      accounts: accounts,
-      selected_account: selected_account,
+      requested_account_id: requested_account_id,
     )
 
-    render SecureRooms::ScanResponsePresenter.new(@user, @card_reader, access_verdict, accounts).response
+    SecureRooms::AccessManager.new(
```
jhanggi: I think they do each deserve to be their own object, but the usage of these three objects still feels a little funny to me. Maybe because the `AccessManager` is now acting more as a callback vs actually managing anything. Maybe it needs a better name? Maybe we hold off on seeing what else it does before we focus on the better name.

we could do an api like`SecureRooms::AccessManager.handle(verdict)`? I sometimes like using class methods like that to improve the API.

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -8,18 +8,21 @@ class SecureRoomsApi::ScansController < ApplicationController
...
+      verdict,
+    ).process
+
+    render SecureRooms::ScanResponsePresenter.new(
+      verdict,
+    ).response
```
jhanggi: On that same note, `SecureRooms::ScanResponsePresenter.present(verdict)` could be a nice api.

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
waggles: I wanted to be able to just do a `#find` but `accounts_for_product` returns a plain old list.. I'll at least pull it down though

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/account_selection_rule.rb
waggles: I don't think so but I'm not sure... in theory the grant just lets the user in without need for them to see any new info at this point. I'd leave it until we have a firm answer for this and then it should be a quick-ish change

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
waggles: yeah this is a bit of a dumb name at the moment haha

vendor/engines/secure_rooms/app/services/secure_rooms/access_manager.rb
```diff
@@ -0,0 +1,22 @@
+module SecureRooms
+
+  class AccessManager
+
+    def self.process(verdict)
```
jhanggi: What I was trying to suggest by improving the api is to leave everything that was there alone operating as primarily instance methods, but have a class method for a simpler interface. That gives you the advantage of being able to build out the instance methods without needing to pass around instance variables, but also have a cleaner interface. So something like:

```ruby
  def self.process(verdict)
    new(verdict).process
 end
 
 def process
   Event.create!(
     occurred_at: Time.current,
     card_reader: verdict.card_reader,
     user: verdict.user,
     # TODO: store a useful way of tracking the outcome
     outcome: verdict,
   )
 end
 ```

File: vendor/engines/secure_rooms/spec/services/secure_rooms/access_manager_spec.rb [hunk 10]
```diff
@@ -0,0 +1,22 @@
...
+    SecureRooms::AccessRules::Verdict.new(:deny, user, card_reader)
+  end
+
+  before { allow_any_instance_of(User).to receive(:accounts_for_product).and_return(accounts) }
+
+  describe "#process" do
+    it "creates an Event" do
+      expect {
+        described_class.process(verdict)
+      }.to change { SecureRooms::Event.count }.by(1)
```
jhanggi: We sometimes do `.to change(SecureRooms::Event, :count).by(1)`. Sometimes that saves a couple characters for dealing with line lengths.

I don't really care about the code climate complaint about `{..}` for multiline in this context.

jhanggi:
Looking pretty good. I think we're just about there.

vendor/engines/secure_rooms/app/services/secure_rooms/access_manager.rb
waggles: ok yeah that makes way more sense

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/verdict.rb
```diff
@@ -27,6 +32,16 @@ def has_result_code?(code)
...
       end
 
+      def translation_scope
+        "secure_rooms/access_rules.reasons"
+      end
+
+      private
+
+      def translate_reason(key)
+        return text(key) if key.present?
```
jhanggi: No need for the `return`. The other option would be to add `default: ""` to the `text` call, but I'm fine with the `if`

jhanggi:
I'd be fine if we do the grant reasons on a later PR.

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
```diff
@@ -23,19 +23,19 @@ def evaluate
...
-        Verdict.new(:pending, reason: reason)
+      def pending!(reason, options = {})
+        Verdict.new(:pending, reason, @user, @card_reader, options)
       end
 
-      def deny!(reason)
-        Verdict.new(:deny, reason: reason)
+      def deny!(reason, options = {})
+        Verdict.new(:deny, reason, @user, @card_reader, options)
       end
```
waggles: I kind of want to make this like a factory or something now but I feel I'm going too far

jhanggi: What are some of the options that we're still using? Is that something we could get rid of? I'm guessing not.

Doing a class hierarchy of `GrantVerdict` would save an argument.

Do we really need the `@user` and `@card_reader` as part of the verdict? Or is that something we could keep up at a higher level, like the controller?

waggles:
I made the default `Verdict` a `pass?` so now we just modify it before we return

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/base_rule.rb
```diff
@@ -12,30 +12,28 @@ def initialize(user, card_reader, params = {})
         @user = user
         @card_reader = card_reader
         @params = params
+        @verdict = Verdict.new(@user, @card_reader)
       end
 
       def call
-        evaluate || pass
+        evaluate
```
jhanggi: I'm kind of torn, I think I actually like the previous version where verdicts were more immutable, even if they had more constructor arguments. It's not so bad because the verdict only gets mutated in the same class it gets instantiated.

Thoughts?

waggles: Yeah I think I agree. should I roll back then?

PR #961, opened 2017-04-12
[#135084] fix users nav from reservations tab

meara:
- https://pm.tablexi.com/issues/135084

The "Users" link was disappearing from the left nav in the reservations tab on users (`/facilities/:facility_id/users/:user_id/reservations`) due to an ability issue. This PR keeps "Users" in this tab's nav, consistent with the other tabs.


jhanggi:
Looks good. Could you put a little more detailed description of the problem in the commit message when you merge it?

PR #962, opened 2017-04-12
[#134682] user price groups refactor

meara:
- https://pm.tablexi.com/issues/134682

Refactor default price groups for users. Groundwork for exposing this to be editable by admins.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -67,9 +67,8 @@ class UsersController < ApplicationController
     @user.password = generate_new_password
 
     if @user.save
-      # send email
-      Notifier.delay.new_user(user: @user, password: @user.password)
-      redirect_to facility_users_path(user: @user.id)
+      @user.create_default_price_group!
+      save_user_success
     else
       render(action: "new_external") && return
     end
@@ -84,6 +83,7 @@ class UsersController < ApplicationController
       flash[:error] = text("users.search.user_already_exists", username: @user.username)
       redirect_to facility_users_path
     elsif @user.save
+      @user.create_default_price_group!
       save_user_success
     else
       flash[:error] = text("create.error", message: @user.errors.full_messages.to_sentence)
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -214,7 +214,7 @@ class UsersController < ApplicationController
       add_role = html("create.add_role", link: facility_facility_user_map_user_path(current_facility, @user), inline: true)
       flash[:notice].safe_concat(add_role)
     end
-    Notifier.delay.new_user(user: @user)
+    Notifier.new_user(user: @user, password: @user.password).deliver_later
     redirect_to facility_users_path(user: @user.id)
   end
 
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -214,7 +214,7 @@ class UsersController < ApplicationController
       add_role = html("create.add_role", link: facility_facility_user_map_user_path(current_facility, @user), inline: true)
       flash[:notice].safe_concat(add_role)
     end
-    Notifier.delay.new_user(user: @user, password: nil)
+    Notifier.delay.new_user(user: @user)
     redirect_to facility_users_path(user: @user.id)
   end
 
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -12,7 +12,6 @@ class Notifier < ActionMailer::Base
   # Who created the account.  How to update.
   def new_user(args)
     @user = args[:user]
-    @password = args[:password]
     send_nucore_mail args[:user].email, text("views.notifier.new_user.subject")
   end
 
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -8,10 +8,10 @@ class Notifier < ActionMailer::Base
 
   default from: Settings.email.from, content_type: "multipart/alternative"
 
-  # Welcome user, login credentials.  CC to PI and Department Admin.
-  # Who created the account.  How to update.
-  def new_user(user:)
+  # Welcome user, login credentials
+  def new_user(user:, password:)
     @user = user
+    @password = password
     send_nucore_mail @user.email, text("views.notifier.new_user.subject")
   end
 
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -10,9 +10,9 @@ class Notifier < ActionMailer::Base
 
   # Welcome user, login credentials.  CC to PI and Department Admin.
   # Who created the account.  How to update.
-  def new_user(args)
-    @user = args[:user]
-    send_nucore_mail args[:user].email, text("views.notifier.new_user.subject")
+  def new_user(user:)
+    @user = user
+    send_nucore_mail @user.email, text("views.notifier.new_user.subject")
   end
 
   # Changes to the user affecting the PI or department will alert their
--- a/app/models/price_group.rb
+++ b/app/models/price_group.rb
@@ -20,9 +20,17 @@ class PriceGroup < ActiveRecord::Base
   scope :for_facility, -> (facility) { where(facility_id: [nil, facility.id]) }
   scope :globals, -> { where(facility_id: nil) }
 
-  scope :base, -> { globals.where(name: Settings.price_group.name.base) }
-  scope :external, -> { globals.where(name: Settings.price_group.name.external) }
-  scope :cancer_center, -> { globals.where(name: Settings.price_group.name.cancer_center) }
+  def self.base
+    globals.find_by(name: Settings.price_group.name.base)
+  end
+
+  def self.external
+    globals.find_by(name: Settings.price_group.name.external)
+  end
+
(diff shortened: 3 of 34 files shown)


spec/product_shared_examples.rb
```diff
@@ -1,50 +1,51 @@
...
-
-    @product = FactoryGirl.create(product_type,
-                                  facility: @facility,
-                                  facility_account: @facility_account)
-    @order = create(:order, account: account, created_by_user: @user, user: @user)
-    @order_detail = @order.order_details.create(attributes_for(:order_detail, account: account, product: @product, quantity: 1))
-
-    create(:account_price_group_member, account: @order.account, price_group: @price_group)
-    create(:account_price_group_member, account: @order.account, price_group: @price_group2)
+    UserPriceGroupMember.delete_all
```
meara: I'm having trouble deciding what I think makes most sense to do about this. On the one hand, it ain't broke. Perhaps we should just leave this be, unappealing as it is? @jhanggi, thoughts?

app/controllers/users_controller.rb
```diff
@@ -67,9 +67,8 @@ def create_external
     @user.password = generate_new_password
 
     if @user.save
-      # send email
-      Notifier.delay.new_user(user: @user, password: @user.password)
-      redirect_to facility_users_path(user: @user.id)
+      @user.create_default_price_group!
+      save_user_success
```
jhanggi: The existing `save_user_success` generates the mailer with`password: nil`. It might be both internal and external can use `password: @user.password` (I suspect netid users have a `nil` password field).

spec/product_shared_examples.rb
jhanggi: Yeah, I'm fine with leaving this as is. Not a huge fan of it, but if it's going to be too big of a hassle to clean up, then not worth it.

jhanggi:
I believe we're going to need a migration that adds the default price group to existing users.

db/migrate/20170421154535_create_default_price_groups_for_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDefaultPriceGroupsForUsers < ActiveRecord::Migration
+  def up
+    User.find_each do |user|
+      # does nothing if user_based_price_groups is not turned on
+      user.create_default_price_group!
+    end
+  end
+  def down
+    raise ActiveRecord::IrreversibleMigration
```
jhanggi: I think it would be safe to delete any user price group memberships that are part of the base or external. Might need to think a little more about that, though.

jhanggi: This whole feature feels like a fairly high-risk change, so I'd like to be confident we could roll back if necessary.

app/models/user.rb
```diff
@@ -216,4 +201,11 @@ def self.active
     where(deactivated_at: nil)
   end
 
+  def create_default_price_group!
+    return unless SettingsHelper.feature_on?(:user_based_price_groups)
+
+    price_group = email_user? ? PriceGroup.external : PriceGroup.base
+    price_group_members.create!(price_group: price_group)
```
jhanggi: Do we have any validations on uniqueness for price group membership? Not sure if it's worth doing. What would happen if we called this method multiple times? Would we end up with multiple memberships? If we prevent that, we can probably avoid the `uniq` scope above.

jhanggi:
I noticed that the demo seeds add UserPriceGroupMembers for the base rate for each of the users. We should probably change that to use the create default method (and make sure we don't end up with duplicates if we can avoid it). I checked NU, DC, and UIC and none of them have memberships for base/external, so we're good there.

meara:
@jhanggi We can certainly change the demo seed file, but there is a wrinkle. Right now all our demo users are email users--"ppi123@example.com", for example. They will therefore be added to external price group if we use `create_default_price_group!`, which will change the existing behavior of the seed file. Is that something we want to be doing?

meara:
Alternatively, we could change our demo users to have "NetID" looking usernames, so that `email_user?` returns true. That would affect `admin_editable?` and `password_updatable?` as well, though--not sure if that is a concern.

jhanggi:
We should try to keep the behavior the same, so I guess we should leave the migration alone, just as long as we're not going to end up with any duplicate memberships.

app/mailers/notifier.rb
```diff
@@ -12,7 +12,6 @@ class Notifier < ActionMailer::Base
   # Who created the account.  How to update.
   def new_user(args)
```
jhanggi: Not a big deal, but should we change the arguments to now take the `user` keyword and make it required?

app/models/price_group.rb
```diff
@@ -20,9 +20,17 @@ class PriceGroup < ActiveRecord::Base
   scope :for_facility, -> (facility) { where(facility_id: [nil, facility.id]) }
   scope :globals, -> { where(facility_id: nil) }
 
-  scope :base, -> { globals.where(name: Settings.price_group.name.base) }
-  scope :external, -> { globals.where(name: Settings.price_group.name.external) }
-  scope :cancer_center, -> { globals.where(name: Settings.price_group.name.cancer_center) }
+  def self.base
```
jhanggi: We should probably do a double check of the forks to make sure their specs aren't using the  `.first` style.

app/models/user.rb
```diff
@@ -1,17 +1,5 @@
 class User < ActiveRecord::Base
 
-  module Overridable
-
-    def price_groups
```
jhanggi: We should figure out a plan for how we're going to merge this into DC because it overrides this module/method. At least until we change their default behavior.

db/migrate/20170421154535_create_default_price_groups_for_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDefaultPriceGroupsForUsers < ActiveRecord::Migration
+  def up
+    User.find_each do |user|
+      # does nothing if user_based_price_groups is not turned on
+      user.create_default_price_group!
+    end
+  end
+  def down
```
jhanggi: Not sure why codeclimate isn't picking it up, but I would expect it to complain about adding a space between the methods. Also, the newlines at the top and bottom of the class.

spec/price_policies_controller_shared_examples.rb
```diff
@@ -10,7 +10,8 @@
     @facility_account = @authable.facility_accounts.create(FactoryGirl.attributes_for(:facility_account))
 
     # Delete the default price groups since they get in the way of testing
-    PriceGroup.all.each(&:delete)
+    UserPriceGroupMember.delete_all
```
jhanggi: So we still need this delete?

app/models/price_group.rb
meara: We'd see test failures on merge PRs, no? Is that something that makes sense to handle as part of a merge, since those changes would need to happen on the individual forks?

app/models/user.rb
meara: Ah okay--I would like to talk through this when you have some time.

db/migrate/20170421154535_create_default_price_groups_for_users.rb
meara: Looks like codeclimate ignores migrations, among other things.

https://github.com/tablexi/nucore-open/blob/master/.codeclimate.yml#L37

app/models/price_group.rb
jhanggi: Handing as part of a merge is fine. However, I'd like to have that branch ready to go (minus a rebase) in NU & DC before we merge this. I don't want problems from this in one of the forks on this to hold up releasing other features. So that means we should create a branch in the different schools, merge in this branch, and then see where we're at. Once we have that ready, we can do a merge from open, cherry-pick from our fork-specific branch, and issue that as a PR. Does that make sense?

spec/price_policies_controller_shared_examples.rb
meara: Evidently so--it looks like nil (unpersisted) price policies are getting created for the default price groups by `PricePolicyBuilder` (called from `PricePoliciesController#new`).

`PricePolicyBuilder` is very difficult to read and I suspect a lot of weird stuff is happening in there. Possibly we should look into refactoring it, but I think that's probably out of scope for this already-large-enough PR.

app/mailers/notifier.rb
```diff
@@ -10,10 +10,9 @@ class Notifier < ActionMailer::Base
 
   # Welcome user, login credentials.  CC to PI and Department Admin.
   # Who created the account.  How to update.
-  def new_user(args)
-    @user = args[:user]
-    @password = args[:password]
-    send_nucore_mail args[:user].email, text("views.notifier.new_user.subject")
+  def new_user(user:)
```
jhanggi: Just doing some testing. We do need to pass the password directly in here. When the user gets serialized for delayed_job, it doesn't serialize `password`, only the salt and `encrypted_password`. So when delayed_job comes to pick it up, `User#password` doesn't exist on the object, so it's not included with the email.

jhanggi:
I just did a bunch of testing, and all looks good except for the new user email issue.

PR #965, opened 2017-04-13
Clean up Roles module

jhanggi:
There was a lot of cruft in this module, so I cleaned up a bunch of it. I also moved it to modules/concerns, which is more where it belongs.

This is probably worth viewing through the individual commits as it tells a little more of the cleanups, especially prior to moving the file.

I vote to ignore the codeclimate duplication warning.


jhanggi:
I don't want to turn off the duplication warnings entirely. Sometimes they're useful. I'd rather see them and make the conscious choice to say, I'm ok with this duplication. I know rubocop has ways to declare a chunk of code to be ignored, but I'm not sure if codeclimate's duplication engine does. If there is, I'd be cool with tagging the duplicated code we've decided is fine.

PR #967, opened 2017-04-13
[#134318] Fix Project chosen width on Order Detail popup

jhanggi:
Sometimes, when the order detail modal pops up, chosen would not get the width
of the select box.

While a little kludgy, this adds a short timeout before loading
chosen to ensure that the browser has had time to set the box's width before
Chosen looks at it.

project problem

This also fixes a minor height issue on the search box where the box was too short.

Before:
screen shot 2017-04-13 at 6 30 48 pm

After:
screen shot 2017-04-13 at 6 31 41 pm


meara:
@jhanggi -- perhaps I'm misunderstanding how Chosen is supposed to work, but it looks like the width issue on the Project select is still there. Did you attach the incorrect screenshot?

jhanggi:
The big screenshot is what it looks like when it's broken.

jhanggi:
I'm going to hold off on this one until we see if #980 fixes the issue (I'll still want to pull in the style changes from this, though).

jhanggi:
Turns out the change from the other PR didn't solve the problem. This is now updated to use the new technique, though should be doing the same thing as the previous attempt (delaying for a few milliseconds).

PR #970, opened 2017-04-17
[#140938049] Add Occupancy

waggles:
* in progress addition of `Occupancy` and its handler

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170406162317) do
+ActiveRecord::Schema.define(version: 20170409125104) do
 
   create_table "account_users", force: :cascade do |t|
     t.integer  "account_id", limit: 4,  null: false
@@ -622,17 +622,37 @@ ActiveRecord::Schema.define(version: 20170406162317) do
   add_index "secure_rooms_card_readers", ["tablet_token"], name: "index_secure_rooms_card_readers_on_tablet_token", unique: true, using: :btree
 
   create_table "secure_rooms_events", force: :cascade do |t|
-    t.integer  "card_reader_id",  limit: 4
-    t.integer  "user_id",         limit: 4
+    t.integer  "card_reader_id",  limit: 4,   null: false
+    t.integer  "user_id",         limit: 4,   null: false
     t.datetime "occurred_at"
     t.string   "outcome",         limit: 255
     t.string   "outcome_details", limit: 255
     t.datetime "created_at",                  null: false
     t.datetime "updated_at",                  null: false
+    t.integer  "account_id",      limit: 4
   end
 
   add_index "secure_rooms_events", ["card_reader_id"], name: "index_secure_rooms_events_on_card_reader_id", using: :btree
   add_index "secure_rooms_events", ["user_id"], name: "index_secure_rooms_events_on_user_id", using: :btree
+
+  create_table "secure_rooms_occupancies", force: :cascade do |t|
+    t.integer  "product_id",     limit: 4, null: false
+    t.integer  "user_id",        limit: 4, null: false
+    t.integer  "account_id",     limit: 4
+    t.integer  "entry_event_id", limit: 4
+    t.datetime "entry_at"
+    t.integer  "exit_event_id",  limit: 4
+    t.datetime "exit_at"
+    t.datetime "orphan"
+    t.datetime "created_at",               null: false
+    t.datetime "updated_at",               null: false
+  end
+
+  add_index "secure_rooms_occupancies", ["account_id"], name: "index_secure_rooms_occupancies_on_account_id", using: :btree
+  add_index "secure_rooms_occupancies", ["entry_event_id"], name: "index_secure_rooms_occupancies_on_entry_event_id", using: :btree
+  add_index "secure_rooms_occupancies", ["exit_event_id"], name: "index_secure_rooms_occupancies_on_exit_event_id", using: :btree
+  add_index "secure_rooms_occupancies", ["product_id"], name: "index_secure_rooms_occupancies_on_product_id", using: :btree
+  add_index "secure_rooms_occupancies", ["user_id"], name: "index_secure_rooms_occupancies_on_user_id", using: :btree
 
   create_table "splits", force: :cascade do |t|
     t.integer "parent_split_account_id", limit: 4,                         null: false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -649,11 +649,11 @@ ActiveRecord::Schema.define(version: 20170409125104) do
     t.datetime "updated_at",               null: false
   end
 
-  add_index "secure_rooms_occupancies", ["account_id"], name: "fk_rails_75ebdba1ca", using: :btree
-  add_index "secure_rooms_occupancies", ["entry_event_id"], name: "fk_rails_eb85e7026f", using: :btree
-  add_index "secure_rooms_occupancies", ["exit_event_id"], name: "fk_rails_94eb785759", using: :btree
-  add_index "secure_rooms_occupancies", ["product_id"], name: "fk_rails_9195aec23b", using: :btree
-  add_index "secure_rooms_occupancies", ["user_id"], name: "fk_rails_d1c865d1b5", using: :btree
+  add_index "secure_rooms_occupancies", ["account_id"], name: "index_secure_rooms_occupancies_on_account_id", using: :btree
+  add_index "secure_rooms_occupancies", ["entry_event_id"], name: "index_secure_rooms_occupancies_on_entry_event_id", using: :btree
+  add_index "secure_rooms_occupancies", ["exit_event_id"], name: "index_secure_rooms_occupancies_on_exit_event_id", using: :btree
+  add_index "secure_rooms_occupancies", ["product_id"], name: "index_secure_rooms_occupancies_on_product_id", using: :btree
+  add_index "secure_rooms_occupancies", ["user_id"], name: "index_secure_rooms_occupancies_on_user_id", using: :btree
 
   create_table "splits", force: :cascade do |t|
     t.integer "parent_split_account_id", limit: 4,                         null: false
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -632,7 +632,7 @@ ActiveRecord::Schema.define(version: 20170409125104) do
     t.integer  "account_id",      limit: 4
   end
 
-  add_index "secure_rooms_events", ["account_id"], name: "fk_rails_a9da8518f4", using: :btree
+  add_index "secure_rooms_events", ["account_id"], name: "index_secure_rooms_events_on_account_id", using: :btree
   add_index "secure_rooms_events", ["card_reader_id"], name: "index_secure_rooms_events_on_card_reader_id", using: :btree
   add_index "secure_rooms_events", ["user_id"], name: "index_secure_rooms_events_on_user_id", using: :btree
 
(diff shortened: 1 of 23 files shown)


db/schema.rb
```diff
@@ -622,18 +622,38 @@
   add_index "secure_rooms_card_readers", ["tablet_token"], name: "index_secure_rooms_card_readers_on_tablet_token", unique: true, using: :btree
 
   create_table "secure_rooms_events", force: :cascade do |t|
-    t.integer  "card_reader_id",  limit: 4
-    t.integer  "user_id",         limit: 4
+    t.integer  "card_reader_id",  limit: 4,   null: false
+    t.integer  "user_id",         limit: 4,   null: false
```
jhanggi: I think we're going to want user_id to be nullable. It's probably fine like this for now, and we can change it back later if we need to.

db/schema.rb
```diff
@@ -622,18 +622,38 @@
...
 
+  create_table "secure_rooms_occupancies", force: :cascade do |t|
+    t.integer  "product_id",     limit: 4, null: false
+    t.integer  "user_id",        limit: 4, null: false
+    t.integer  "account_id",     limit: 4
+    t.integer  "entry_event_id", limit: 4
+    t.datetime "entry_at"
+    t.integer  "exit_event_id",  limit: 4
+    t.datetime "exit_at"
+    t.datetime "orphan"
```
jhanggi: Would this be better named `orphaned_at`? Making a method `orphan?` would probably be worth adding.

db/schema.rb
```diff
@@ -622,18 +622,38 @@
...
     t.datetime "created_at",                  null: false
     t.datetime "updated_at",                  null: false
+    t.integer  "account_id",      limit: 4
   end
 
   add_index "secure_rooms_events", ["card_reader_id"], name: "index_secure_rooms_events_on_card_reader_id", using: :btree
   add_index "secure_rooms_events", ["user_id"], name: "index_secure_rooms_events_on_user_id", using: :btree
 
+  create_table "secure_rooms_occupancies", force: :cascade do |t|
+    t.integer  "product_id",     limit: 4, null: false
```
jhanggi: Should we add foreign keys for a bunch of these columns?

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    belongs_to :secure_room, foreign_key: :product_id
+    belongs_to :user
+    belongs_to :account
+    belongs_to :entry_event, class_name: SecureRooms::Event
+    belongs_to :exit_event, class_name: SecureRooms::Event
+
+    validates :product_id, :user_id, presence: true
+
+    def mark_oprhaned!
```
jhanggi: We usually put all the scopes in the file before any instance methods.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -0,0 +1,32 @@
...
+    def self.valid
+      where.not(orphan: nil)
+    end
+
+    def self.current
+      valid.where(exit_event_id: nil).where.not(entry_event_id: nil)
+    end
+
+    def self.recent
+      # Either 24h or 10pax
```
jhanggi: I feel like this comment is misleading, or should maybe be tagged with `TODO`.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -0,0 +1,32 @@
...
+
+    def self.valid
+      where.not(orphan: nil)
+    end
+
+    def self.current
+      valid.where(exit_event_id: nil).where.not(entry_event_id: nil)
+    end
+
+    def self.recent
```
jhanggi: Would `recent_exits` be a more specific name? Although, leaving it as `recent` so we can swap out 24 hours for recent 10 makes sense. I'm good either way.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -0,0 +1,67 @@
...
+        event_attribute =
+          if entering?
+            { entry_event: event }
+          else
+            { exit_event: event }
+          end
+
+        if current_occupant? && exiting?
+          existing_occupancy.update!(event_attribute)
+        elsif new_occupant? && entering?
```
jhanggi: Would adding an `else` with a `TODO` be a good idea?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -0,0 +1,67 @@
...
+          new_occupancy.update!(event_attribute)
+        end
+
+        existing_occupancy || new_occupancy
+      end
+
+      def existing_occupancy
```
jhanggi: Is the existing or new logic something we could do with `find_or_initialize_by`?

vendor/engines/secure_rooms/spec/models/secure_rooms/occupancy_spec.rb
```diff
@@ -0,0 +1,6 @@
+require "rails_helper"
+
+RSpec.describe SecureRooms::Occupancy do
+  it { is_expected.to validate_presence_of :product_id }
```
jhanggi: Should we validate the presence of the actual product, rather than the id?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/event_handler.rb
```diff
@@ -0,0 +1,21 @@
+module SecureRooms
+
+  module AccessHandlers
+
+    class EventHandler
+
+      def self.process(verdict)
+        Event.create!(
```
jhanggi: Should this event be created with with an account if it's available?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -0,0 +1,67 @@
...
+        new(event).process
+      end
+
+      def initialize(event)
+        @event = event
+      end
+
+      def process
+        event_attribute =
+          if entering?
```
jhanggi: Could this be collapsed down into the main conditional below? Having these two sets of conditionals is a little confusing.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
waggles: oh dear apparently I can't spell either haha

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
waggles: it gets a little tricky because sometimes we want both the existing and the new (in one of the upcoming error cases) so I think we want them to be reference-able separately

jhanggi: Makes sense.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -0,0 +1,65 @@
...
+      end
+
+      def existing_occupancy
+        @existing_occupancy ||= Occupancy.current.find_by(
+          user: event.user,
+          secure_room: event.secure_room,
+        )
+      end
+
+      def new_occupancy
```
jhanggi: Can we add `private` to this class wherever it can go?

vendor/engines/secure_rooms/db/migrate/20170409125103_create_occupancies.rb
```diff
@@ -0,0 +1,27 @@
...
+      t.datetime :entry_at
+      t.references :exit_event
+      t.datetime :exit_at
+
+      t.datetime :orphaned_at
+
+      t.timestamps null: false
+    end
+
+    add_foreign_key :secure_rooms_occupancies, :products
```
jhanggi: This is an annoyance, but oracle doesn't automatically create indexes on foreign keys, so we'll also need to `add_index`. I forget, but I think you might be able to add `index: true` on the `references` as well.

Also, not a big deal (as long as this migration runs both up and down as is), but I believe you can add `foreign_key: true` to the `references` option and get this for free.

vendor/engines/secure_rooms/db/migrate/20170409125103_create_occupancies.rb
```diff
@@ -0,0 +1,23 @@
...
+      t.references :exit_event, index: true
+      t.datetime :exit_at
+
+      t.datetime :orphaned_at
+
+      t.timestamps null: false
+    end
+
+    add_foreign_key :secure_rooms_occupancies, :secure_rooms_events, column: :entry_event_id
+    add_foreign_key :secure_rooms_occupancies, :secure_rooms_events, column: :exit_event_id
```
waggles: I kept these separate since they use the extra `column: ` bit and I wasn't sure if that would work inline

vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
```diff
@@ -10,6 +10,7 @@ def initialize(ability)
 
     def extend(user, resource)
       ability.can :manage, CardReader if user.manager_of?(resource) || user.facility_senior_staff_of?(resource)
+      ability.can :index, Occupancy if user.operator_of?(resource)
```
waggles: I'm not sure what the permissions are--what is this supposed to be?

jhanggi: I think this makes sense for now.

vendor/engines/secure_rooms/app/views/secure_rooms/occupancies/index.html.haml
```diff
@@ -0,0 +1,24 @@
...
+= content_for :sidebar do
+  = render "admin/shared/sidenav_product", sidenav_tab: "secure_rooms"
+= content_for :tabnav do
+  = render "admin/shared/tabnav_product", secondary_tab: "occupancies"
+
+%h2= @product
+
+%p= text("views.admin.secure_rooms.occupancies.index.description")
+
+- if @occupancies.empty?
```
jhanggi: Not a big deal, especially as this is the more common pattern in NUcore, but I'm generally preferring these conditionals to be flipped so the positive case is first and the none case is at the bottom of the file.

vendor/engines/secure_rooms/db/migrate/20170409125103_create_occupancies.rb
jhanggi: Cool. Fine by me.

vendor/engines/secure_rooms/db/migrate/20170409125104_update_events.rb
```diff
@@ -0,0 +1,10 @@
+class UpdateEvents < ActiveRecord::Migration
+
+  def change
+    change_column :secure_rooms_events, :card_reader_id, :integer, null: false
+    change_column :secure_rooms_events, :user_id, :integer, null: false
+    add_column :secure_rooms_events, :account_id, :integer
+    add_foreign_key :secure_rooms_events, :accounts
```
jhanggi: I think we need the explicit index creation here as well.

jhanggi: I just ran this migration on oracle, and we definitely do need to add the index ourselves.

waggles:
@jhanggi updated

PR #971, opened 2017-04-17
[#140891941] Set up Secure Rooms Price Policies

jhanggi:
This PR ended up being bigger than I originally wanted, but a large portion of it is really moving things around.

I moved what was in `lib/price_policy_support` to `models/concerns/price_policies/quantity` and did some extractions out of InstrumentPricePolicy a usage-based concern that could be shared with SecureRoomPricePolicy.

Some of the validations have moved to better scope with their relevant policy types (and to avoid `allow_blank: true` where possible.

There are a lot of methods that have been rearranged for more intuitive ordering or to make methods private (there are a lot in InstrumentPricePolicyCalculations I would love to make private or rename, but the tests are really strict).

One of the illuminating insights this refactoring gave me was how Reservation-only, Usage, and Overage aren't really so different. They're all calculating a price based on a start and end time, the only difference is which values to use for start and end (reservation: reserve_start_at, reserve_end_at; usage: actual start/end; overage: reserve_start_at, max of reserve/actual end_at).

Actual pricing logic (i.e. what's in PricePolicyCalculations) will come in a separate PR.

Another thing I'd like to do, but need to figure out if it's ok is to get rid of `OldInstrumentPricePolicy` and its relevant columns (reservation_rate, overage_rate, etc). There are still some policies on prod that we were unable to convert to the newer version with those fields, so I'm not sure if we're safe to remove them. After that, I'd kind of like to see if it's possible to collapse `usage_rate`/`unit_cost` and `usage_subsidy`/`unit_subsidy`. The `usage` rates are really a cost per unit, where the unit is a minute. I think that would clean up some stuff.


app/views/instrument_price_policies/_adjustment_row.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  = text_field_tag "price_policy_#{price_group.id}[usage_subsidy]",
+    display_usage_subsidy(price_group, price_policy), size: 8, class: "usage_adjustment"
+
+%td
+  = t("price_policies.amount")
+  %br
+  %span.minimum_cost{ data: { usage_subsidy: price_policy.usage_subsidy } }
+  = hidden_price_policy_tag(price_group.id, price_policy.minimum_cost, :minimum_cost, params["price_policy_#{price_group.id}"])
+
+- if local_assigns[:cancellation] != false
```
jhanggi: It might be better for secure room price policies to be a completely different set of partials, but I went with this for now, since the only current difference between instrument and secure room policies is secure rooms don't have a cancelation fee. This was much quicker for now. We can revisit later if necessary.

app/views/instrument_price_policies/_table.html.haml
```diff
@@ -0,0 +1,58 @@
+- clazz = price_policies.first.class
```
waggles: it's a little odd to see `clazz` in a view, it makes me think there's about to be some metaprogramming or something going on

not really a big deal though

jhanggi: Yeah, it's really just to save some typing with the uses of `human_attribute_name` below.

I could go more with what I have in the next file, where there's `price_policy = price_policies.first` and then a bunch of `price_policy.class.human_attribute_name`. Either way, it would probably be worth being consistent between the two file.

PR #973, opened 2017-04-18
[#134502] Ability to see statements on cross facility accounts

jhanggi:
Billing..Payment Sources...[select an account]...Statements tab now will show statements in a cross facility context (i.e. as a billing administrator).

This also fixes the context for that tab so the correct sidebar appears depending on the `manage_payment_sources_with_users` setting.

In addition, it removes a couple of HTML versions of statement views which are no longer referenced anywhere.


spec/controllers/facility_accounts_controller_spec.rb
```diff
@@ -2,17 +2,19 @@
 require "controller_spec_helper"
 
 RSpec.describe FacilityAccountsController, feature_setting: { edit_accounts: true } do
-  let(:facility) { @authable }
+  let(:facility) { FactoryGirl.create(:facility) }
+  let(:account) { create_nufs_account_with_owner }
+  let(:admin) { @admin }
```
meara: Where is @admin defined?

jhanggi: Totally fair question. It gets created by `before(:all) { create_users }`. There's obviously further I could have gone with cleaning up the specs, but creating a few `let`s is where I stopped.

meara: Well, that is rather magical, but okay fair enough.

PR #975, opened 2017-04-19
[#80053] Remove legacy OldInstrumentPricePolicy

jhanggi:
Many moons ago (PR #84), we moved from a more complicated price policy that allowed things like different rates for overage time and reservation times, as well as charging in X-minute blocks rather than per-minute as we do now. 

The vast majority of the existing policies were compatible with the new model, so they were automatically converted. Some, especially in UIC's case, were non-convertible--mostly due to misconfiguration like having a usage or overage rate when the instrument was marked as "Reservation Only". These were left as is, and the facilities were expected to create new policies that adhere to the new model.

We left the old policy class around so we could reprice orders if necessary. All the orders that were using the old model are now in a "reconciled" state so they are no longer modifiable.

This PR has a migration that moves all the now-unused columns into a JSON hash in a new column and removes the old columns. This is so we still have the old information handy if we ever need it. I probably spent more time making a `down` migration rather than marking it as irreversible, but oh, well. 

It also removes a lot of models and tasks used for reporting on the pricing differences between the model, as well as the conversion.

I'm pretty sure there are conflicts with #971, but I figured this might provide additional context for that PR.


jhanggi:
Rebased off of latest master to clean up merge conflicts.

PR #976, opened 2017-04-19
[#135272] Responsive Search page

noelrappin:
https://pm.tablexi.com/issues/135272

Search results changed for mobile screens based on the pattern suggested in https://css-tricks.com/responsive-data-tables/.

I’m not completely sold on repeating the header data in the cells for the responsive version (the css-tricks had the responsive headers in the CSS, which I _really_ didn’t like), but I thought that an array that was used in both places was actually less readable.

### DIFF
--- a/app/assets/stylesheets/responsive.scss
+++ b/app/assets/stylesheets/responsive.scss
@@ -32,7 +32,7 @@
 
   // Responsive table behvavior inspired by
   // https://css-tricks.com/responsive-data-tables/
-  .responsive-table {
+  .js--responsive_table {
 
     table, thead, tbody, th, td, tr {
       display: block;
--- a/app/assets/stylesheets/responsive.scss
+++ b/app/assets/stylesheets/responsive.scss
@@ -50,16 +50,17 @@
       border: none;
       border-bottom: 1px solid #eee;
       position: relative;
-      padding-left: 50%;
+      padding-left: 35%;
 
       .responsive-header {
         display: block;
         position: absolute;
         top: 6px;
         left: 6px;
-        width: 45%;
+        width: 30%;
         padding-right: 10px;
         white-space: nowrap;
+        font-weight: bold;
       }
 
     }
--- a/app/assets/stylesheets/responsive.scss
+++ b/app/assets/stylesheets/responsive.scss
@@ -29,6 +29,47 @@
     padding: 10px 10px 0 10px;
     border-bottom: 0;
   }
+
+  // Responsive table behvavior inspired by
+  // https://css-tricks.com/responsive-data-tables/
+  .responsive-table {
+
+    table, thead, tbody, th, td, tr {
+      display: block;
+    }
+
+    thead tr {
+      position: absolute;
+      top: -9999px;
+      left: -9999px;
+    }
+
+    tr { border: 1px solid #ccc; }
+
+    td {
+      border: none;
+      border-bottom: 1px solid #eee;
+      position: relative;
+      padding-left: 50%;
+
+      .responsive-header {
+        display: block;
+        position: absolute;
+        top: 6px;
+        left: 6px;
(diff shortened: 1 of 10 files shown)


jhanggi:
I like the way this looks. In terms of styling, I think making the headers bold, and maybe less wide (give the values more space). The double-width border bothers me a little, but not overly so.

screen shot 2017-04-19 at 12 03 43 pm


I agree putting the headers in the CSS is not the way to go.

Would it be possible to use JS to copy the headers? If we start using this pattern elsewhere, I could see the duplication of the headers getting really annoying to have to maintain.

noelrappin:
It's certainly possible to do it in JavaScript, I can explore. Offhand, I suspect it's easier to have the headers be in an array in the partial that's used both places than to do it via JavaScript, but it would take some maintenance. I'm assuming the headers wouldn't change that often, though.

noelrappin:
I'll do those style requests.

jhanggi:
What does it look like as an array within the partial? Headers won't change that often, but if they do, I'd rather us not have to deal with thinking about the responsive side of it.

noelrappin:
With requested changes:

search_results_-_nucore

jhanggi:
This looks good, but I'm still not sure how I feel about the headers in the code. This almost feels like it might even be trickier to maintain down the line.

I spent about 10 trying to do it with JS. I found a [SO post](http://stackoverflow.com/questions/28320100/how-can-i-copy-the-corresponding-table-header-th-to-their-table-cell-td) and adapted it to support multiple tables per page.

```coffeescript
$ ->
  $(".responsive-table").each ->
    $th = $(this).find("thead th");
    $(this).find("tbody tr td").prepend ->
      $("<div>").addClass("responsive-header").text $th.eq($(this).index()).text()
```

What do you think?

noelrappin:
Kind of torn. The JS does mean we don't need to worry about maintaining duplication, but it does make the change kind of magical if you are working on the page. In general, I'm against adding JS to stuff that doesn't need it, but I do agree that losing the duplication has value.

So, I can argue both ways. Want me to do a PR with the JS so we can all see?

jhanggi:
Yeah, let's do a PR of the JS. Maybe use a class `.js--responsive-table` in addition to the styling class to make it clear JS is acting on it?

We can hold off on merging that one until we get to some of the other tables we want to make responsive and see how big of a pain it is.

noelrappin:
Okay, here's the JS version -- only additional thing we might consider is a global "responsive" variable so this code doesn't even get executed if the page isn't responsive. That might be a useful thing to have around going forward.

app/assets/stylesheets/responsive.scss
```diff
@@ -29,6 +29,48 @@
     padding: 10px 10px 0 10px;
     border-bottom: 0;
   }
+
+  // Responsive table behvavior inspired by
+  // https://css-tricks.com/responsive-data-tables/
+  .js--responsive_table {
```
jhanggi: I don't have a strong feeling on this, but would the current best practices say to use separate classes for the styles and the JS? (this project is an amalgam of best practices from various points in time).

jhanggi:
Cool. This looks pretty good. I like that we have a spec, too.

I agree a global variable is a good idea. We have a couple already being set in `shared/_head`. That might be a good place to set it.

app/assets/stylesheets/responsive.scss
noelrappin: Maybe. I can change, I don't have a strong feeling either.

noelrappin: Actually, I've just decided I had an opinion. Let's keep them as one style on the grounds that you don't want one without the other ever since the behavior depends on both.

jhanggi: I buy that.

PR #977, opened 2017-04-19
fix secure rooms migrations for oracle

meara:
Oracle has a 30-character limit on table names and indices. This updates the secure rooms schema naming to comply with those constraints.


vendor/engines/secure_rooms/db/migrate/20170406162317_create_events.rb
```diff
@@ -12,8 +12,13 @@ def change
       t.timestamps null: false
     end
 
-    add_index :secure_rooms_events, :user_id
-    add_index :secure_rooms_events, :card_reader_id
+    if NUCore::Database.oracle?
+      add_index :secure_rooms_events, :user_id, name: "index_rooms_events_on_user_id"
```
jhanggi: If we're creating custom names for an index, we should use that name on both oracle and mysql to keep them consistent.

jhanggi: Or, I guess they get updated in the new migration. So then it's fine.

db/migrate/20170419155310_rename_indices_for_oracle.rb
```diff
@@ -0,0 +1,10 @@
+class RenameIndicesForOracle < ActiveRecord::Migration
```
jhanggi: This migration should probably live in the secure_rooms engine.

db/migrate/20170419155310_rename_indices_for_oracle.rb
```diff
@@ -0,0 +1,10 @@
...
+  def change
+    unless NUCore::Database.oracle?
+      rename_index :secure_rooms_card_readers, "index_secure_rooms_card_readers_on_product_id", "index_card_readers_on_prod_id"
```
jhanggi: These are fine, but I sometimes like to go with `i_` as a prefix when you need to cut out a few letters.

db/migrate/20170419155310_rename_indices_for_oracle.rb
meara: Womp, you're right, I'll move it.

meara:
Actually, turns out when the migrations are run on oracle specifically, the index names are generated to fit within the character limit, so this should not be a problem.

PR #978, opened 2017-04-19
Log instead of Raise until error cases written

waggles:
@jhanggi @danhodos

### DIFF
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
@@ -27,7 +27,7 @@ module SecureRooms
           )
         else
           # TODO: Add error cases
-          raise NotImplementedError
+          Rails.logger.warn "Encountered Double-Scan Situation"
         end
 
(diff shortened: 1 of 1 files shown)


vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -27,7 +27,7 @@ def process
           )
         else
           # TODO: Add error cases
-          raise NotImplementedError
+          Rails.logger.warn "Encountered Double-Scan Situation"
```
jhanggi: Do we want to add any additional information in the log like which controller/user it was? This may be going away soon enough that we don't care.

danhodos: I imagine it's going away this week, so I vote to merge (selfishly to un-block me). 😸

waggles: yeah I figured I'm mainly unblocking and no one will actually read this message between now and actual implementation

PR #979, opened 2017-04-19
[#134005] Change column header on My Reservations

jhanggi:
NU requested "Reserved For" to be changed to read "Reservation
(Edit/Extend)”.

`reservations_table` is not used anywhere else, so I removed it from
`shared`

I have a question out on if they want it for both Upcoming and All tabs, so that may change based on their feedback.


jhanggi:
This is now ready for review.

PR #980, opened 2017-04-20
[#134547] Set fulfilled_at when completing order from modal

jhanggi:
This feature is to match the one where a facility staff can set the fulfilled_at date when adding to an existing order.

It sets the fulfilled_at to noon of the day you choose. This matches the behavior of the existing feature.

screen shot 2017-04-19 at 7 50 33 pm

The technique I setup here for dealing with binding events within the modal might also solve the Chosen issue in #967 (but I would still want to bring in the style change from that).


app/assets/javascripts/app/datepicker-data.coffee
```diff
@@ -1,9 +1,15 @@
...
 $ ->
-  # This will set up a datepicker based on data attributes
-  # Make sure you call `to_s` or `iso8601` on any dates you set in the views to
-  # ensure the proper format.
-  $picker = $(".datepicker__data")
-  $picker.datepicker(
-    minDate: new Date($picker.data("min-date")) # will be unbounded if not provided
-    maxDate: new Date($picker.data("max-date")) # will be unbounded if not provided
-  )
+  DatePickerData.activate()
```
meara: Why do we need to call `activate` effectively twice?

jhanggi: We're invoking it once here. The next line is just a reference to the function. That reference will get called whenever a modal gets loaded (after the ajax has finished) so that the datepicker gets bound to the new elements that were loaded on the page.

meara: Ah okay that makes sense, thanks.

PR #981, opened 2017-04-20
[#140891971] Add error scanning cases

waggles:
* Note: am cleaning up the `OccupancyHandler` spec, but only how it is organized, everything else is fair game for commentry

### DIFF
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
@@ -16,35 +16,21 @@ module SecureRooms
 
       def process
         if current_occupant? && exiting?
-          associate_exit_to(existing_occupancy)
+          existing_occupancy.associate_exit!(event)
         elsif new_occupant? && entering?
-          associate_entry_to(new_occupancy)
+          new_occupancy.associate_entry!(event)
         elsif new_occupant? && exiting?
           new_occupancy.mark_orphaned!
-          associate_exit_to(new_occupancy)
+          new_occupancy.associate_exit!(event)
         elsif current_occupant? && entering?
           existing_occupancy.mark_orphaned!
-          associate_entry_to(new_occupancy)
+          new_occupancy.associate_entry!(event)
+        else
+          raise NotImplementedError.new("Encountered unexpected scan context")
         end
       end
 
       private
-
-      def associate_entry_to(occupancy)
-        occupancy.update!(
-          entry_event: event,
-          entry_at: Time.current,
-        )
-        occupancy
-      end
-
-      def associate_exit_to(occupancy)
-        occupancy.update!(
-          exit_event: event,
-          exit_at: Time.current,
-        )
-        occupancy
-      end
 
       def existing_occupancy
         @existing_occupancy ||= Occupancy.current.find_by(
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
@@ -16,24 +16,35 @@ module SecureRooms
 
       def process
         if current_occupant? && exiting?
-          existing_occupancy.update!(
-            exit_event: event,
-            exit_at: Time.current,
-          )
+          associate_exit_to(existing_occupancy)
         elsif new_occupant? && entering?
-          new_occupancy.update!(
-            entry_event: event,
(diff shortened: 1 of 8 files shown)


vendor/engines/secure_rooms/app/views/secure_rooms/occupancies/index.html.haml
```diff
@@ -22,3 +22,21 @@
           %td= occupancy.entry_at
 - else
   %p.notice= text("views.admin.secure_rooms.occupancies.index.none")
+
+- if @problem_occupancies.present?
```
waggles: Am not sure if this is where this will end up, but this seems okay for now

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -16,25 +16,36 @@ def initialize(event)
...
-          # TODO: Add error cases
-          Rails.logger.warn "Encountered Double-Scan Situation"
+          associate_entry_to(new_occupancy)
+        elsif new_occupant? && exiting?
+          new_occupancy.mark_orphaned!
+          associate_exit_to(new_occupancy)
+        elsif current_occupant? && entering?
+          existing_occupancy.mark_orphaned!
+          associate_entry_to(new_occupancy)
         end
```
jhanggi: There shouldn't be any other potential conditions, but would it be worth adding an `else` that maybe raises an error indicating a bug? Might not be worth it, though.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -16,25 +16,36 @@ def initialize(event)
...
-
-        existing_occupancy || new_occupancy
       end
 
       private
 
+      def associate_entry_to(occupancy)
+        occupancy.update!(
```
jhanggi: Would it be worth adding `Occupancy#enter!(event)` that sets the event and marks the timestamp? Could even return `self` to clean these methods up (or even make these methods irrelevant).

vendor/engines/secure_rooms/app/views/secure_rooms/occupancies/index.html.haml
jhanggi: We should do these as separate tabs, like we do in reservations and orders. That means they should probably be different controller actions.

vendor/engines/secure_rooms/app/views/secure_rooms/occupancies/index.html.haml
```diff
@@ -22,3 +22,21 @@
...
+      %tr
+        %th= User.human_attribute_name(:username)
+        %th= SecureRooms::Occupancy.human_attribute_name(:entry_at)
+        %th= SecureRooms::Occupancy.human_attribute_name(:exit_at)
+        %th= SecureRooms::Occupancy.human_attribute_name(:orphaned_at)
+    %tbody
+      - @problem_occupancies.each do |occupancy|
+        %tr.secure_rooms_problem_occupancy{id: "secure_rooms_problem_occupancy_#{occupancy.id}"}
+          %td= occupancy.user.username
+          %td= occupancy.entry_at
```
jhanggi: How does this get formatted? I think we often use something like `human_datetime()` (a custom method of ours that's nil-safe).

vendor/engines/secure_rooms/app/views/secure_rooms/occupancies/index.html.haml
jhanggi: But I'm fine punting that to another PR.

waggles:
@jhanggi updated

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -16,21 +16,18 @@ def initialize(event)
...
+        elsif new_occupant? && exiting?
+          new_occupancy.mark_orphaned!
+          new_occupancy.associate_exit!(event)
+        elsif current_occupant? && entering?
+          existing_occupancy.mark_orphaned!
+          new_occupancy.associate_entry!(event)
         else
-          # TODO: Add error cases
-          Rails.logger.warn "Encountered Double-Scan Situation"
+          raise NotImplementedError.new("Encountered unexpected scan context")
```
jhanggi: I think we're using exploded style, so codeclimate wants `raise NotImplementedError, "Encountered unexpected scan context"`.

PR #982, opened 2017-04-20
[#140895293] Basic SecureRooms events API endpoint

jhanggi:
This is intentionally basic, only setting up the endpoint and logging,
in order to unblock other work. Future changes in implement the storage
of these events.


vendor/engines/secure_rooms/spec/controllers/secure_rooms_api/events_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
+
+RSpec.describe SecureRoomsApi::EventsController do
+  before do
```
danhodos: Might want to extract this out into a separate method at some point, but I think holding off until we have one more instance of it would be a-okay!

PR #985, opened 2017-04-21
[#144058475] Create order from Occupancy

waggles:
* what else do I need to know about order creation at this point?


vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -0,0 +1,30 @@
...
+          account: occupancy.account,
+          user: occupancy.user,
+          facility: occupancy.facility,
+          created_by_user: occupancy.user,
+        )
+
+        order.order_details.create!(
+          product: occupancy.secure_room,
+          created_by_user: occupancy.user,
+          quantity: 1,
```
jhanggi: You should validate that the `OrderDetail#order_status == OrderStatus.new_status` (or `inprocess`, I forget what we decided.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -0,0 +1,30 @@
+module SecureRooms
+
+  module AccessHandlers
+
+    class OrderHandler
+
+      def self.process(occupancy)
+        return unless occupancy.orderable?
+
+        order = Order.create!(
```
jhanggi: You should validate that the `state` is "purchased", `ordered_at` is set. I believe we'll need to call `order.validate_order!` and `order.purchase!` to do those transitions.

We should check `Order#cart_valid?` to see if there's anything else we need to test. I think we've already do most of the `has_valid_payment?` checks at the account selection level, so I'm not that worried about it (e.g. credit cards and POs are facility specific, so they shouldn't be displayed in the context of another facility on the kiosk). And that the `User#accounts_for_product` that we do in the account selection rule does indeed check that you have a valid price group when you select an account.

I think I'm probably ok if we raise an invalid state transition error all the way up in this case. It might be worth writing a spec that stubs out `Order#cart_valid?` to return false and see what happens. We might want to capture it and return something to the doors service (what does it do right now if it gets a 500 from us?)

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
jhanggi: Also that `OrderDetail#state == "new"` (or `inproces`). There is both a state and a status on order details. The states are new->inprocess->complete->reconciled (also canceled). The statuses are those, plus there are sub-statuses (e.g. "Pending Approval" is a sub-status of "New"). Order details have a `new` state from their creation, but don't get an `order_status` until they're purchased.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -34,6 +36,10 @@ def orphan?
       orphaned_at?
     end
 
+    def orderable?
```
jhanggi: I kind of feel like an occupancy shouldn't know or care if it's orderable. The OrderHandler should probably own that logic.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
jhanggi: As we add the extra method calls, it might be worth moving this over into instance methods.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -0,0 +1,30 @@
...
+          created_by_user: occupancy.user,
+        )
+
+        order.order_details.create!(
+          product: occupancy.secure_room,
+          created_by_user: occupancy.user,
+          quantity: 1,
+        )
+
+        order
```
jhanggi: We also need to attach the occupancy to the order detail. For reservations, we do this with a `has_one :reservation` on the order detail, and the reservation has an `order_detail_id` (nullable) column. I'm still unsure whether that was the right decision vs putting a nullable reservation_id on order details, but I think until it actually causes real problems we should stick with that pattern.

waggles: what's the preferred way to insert this association into the engine, re-opening the `OrderDetail` class?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
waggles: hmm I don't remember if we decided between `new`/`inprocess`-- does the latter maybe make more sense? I don't have full context of its meaning here

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
jhanggi: We usually create a module like `OrderDetailExtension` inside the engine, then in the engine's `to_prepare` block, add `OrderDetail.send :include, SecureRooms::OrderDetailExtension`. Take a look at the projects engine for a good example.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
jhanggi: I would say we should go with what reservations do when you start it, but haven't finished it. Looks like it puts it into "new" state with "New" order status.

vendor/engines/secure_rooms/lib/secure_rooms/order_detail_extension.rb
```diff
@@ -0,0 +1,13 @@
+module SecureRooms
```
jhanggi: We usually put these extensions in the app/models directory of the engine.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -5,6 +5,8 @@ class Occupancy < ActiveRecord::Base
     belongs_to :secure_room, foreign_key: :product_id
     belongs_to :user
     belongs_to :account
+    belongs_to :order
```
jhanggi: I don't think we need a direct relation to the `order`. If we need to get at the order, we can get at it from the `order_detail.order`.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+
+      def initialize(occupancy)
+        @occupancy = occupancy
+      end
+
+      def process
+        return unless occupancy.account_id?
+
+        create_order
```
jhanggi: We're doing a bunch of DB work here. I think it might be a good idea to wrap all this inside a transaction.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -5,11 +5,15 @@ class Occupancy < ActiveRecord::Base
...
     belongs_to :user
     belongs_to :account
+    belongs_to :order_detail
     belongs_to :entry_event, class_name: SecureRooms::Event
     belongs_to :exit_event, class_name: SecureRooms::Event
 
     validates :secure_room, :user, presence: true
 
+    delegate :facility, to: :secure_room
+    delegate :order, to: :order_detail
```
jhanggi: I'm not positive, but I think we'll need `allow_nil: true` on this to handle a non-order occupancies. And is this delegation even being used?

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -34,6 +38,10 @@ def orphan?
       orphaned_at?
     end
 
+    def complete?
```
jhanggi: Is this method being used?

jhanggi:
Looks pretty good. Just a couple questions about potentially unused code.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
waggles: oh you know I don't think it is, will remove

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
waggles: oops not in this branch, I'll pull it out so it goes in with context

jhanggi:
Looks good, but looks like we need to clean up a schema.rb conflict.

PR #986, opened 2017-04-22
Dynamic order detail time data

jhanggi:
Just wanted to get some feedback on this idea. I'm thinking of how to deal with order details that might have either a reservation or an occupancy. This will apply to several different stories, mainly in displays.

In the views, we are often relying on `Reservation#to_s` (defined in `Reservations::Rendering`), so when we get to occupancies, we'll want to implement a `to_s` method. We'll probably want to steal some of the code out of `Rendering` (it does smart things like `- ???` when there is no actual end, but it has started as well as using the reservation times when it hasn't started yet, but the actuals once it's there.

I have two commits here with two variations. I think they're worth checking out separately (view the first commit, then view the final version of the PR). I think the second idea is simpler, but is a little less flexible than the first. But we might not end up needing that flexibility.

I'm also very open to a better name than `time_data`.


waggles:
I think I actually find the first version's separate `TimeData` "simpler", it feels like it fits my expectations I guess

and errrr also not sure about the name...

meara:
I prefer the second variation, personally--it's kind of like a pseudo AR association, which feels more intuitive to me in terms of the actual relationship between these objects.

Thoughts on calling these relationships something like `TimedPurchase` or `TimeBasedPurchase`?

danhodos:
I basically don't know anything about NUcore, so take anything I say with a grain of salt. 😸 

It sounds like you're mainly trying to solve a view-level concern (i.e. how do I display order details, given that the thing being ordered may vary)? _If_ that's true, I'd probably recommend introducing a separate _domain model_ concept for the views, as opposed to continuing to build this logic into the _database models_ you're using for persistence.

A domain model object that takes the parameters you need for view rendering (often primitives like `String`, `Time`, etc.), and then either provides for a `from_instrument` class method (or defers that to a method on `Instrument` if you like) seems like it would keep your object responsibilities smaller (more cohesive) and hopefully reduce coupling between objects. The new domain model object (e.g. `TimeData`) then essentially acts as a common interface that the views know about, without it or the views having to know about the particularities of your other domain model objects, their interrelationships, etc.

Does that make sense? I may have beach brain, but hopefully it does at least a little.

jhanggi:
Closing in favor of #993

PR #988, opened 2017-04-24
Remove Reservation#duration_value/unit fields

jhanggi:
These fields are no longer used: we only ever work with minutes. The
larger goal is to remove the Reservations::DateSupport module entirely
with something more sane and more generic.

The UI fields always submit a minutes value for the duration.

It’s not entirely clear why the unit field existed (it came from the
pre “Initial Fork” commit, so git blame is useless here. Possibly,
there was a dropdown allowing you to set the duration as “X hours”.
Since that no longer exists, there is no reason for the added
complexity of supporting multiple units.


meara:
Also thanks for keeping these cleanup PRs separate from new functionality changes, super helpful. 😁

PR #990, opened 2017-04-25
[#144058895] Fulfill order

waggles:
* not sure what to do about price policies / prices yet, let's chat about that


vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -17,29 +17,33 @@ def initialize(occupancy)
...
       def create_order
+        ActiveRecord::Base.transaction do
+          create_order_and_detail
+          order.validate_order!
+          order.purchase!
+        end
+      end
+
+      def fulfill_order
+        order_detail.update_order_status! occupancy.user, OrderStatus.complete.first
```
waggles: while this updates the status to be complete, I'm not sure of a good context to reference to see what else is involved. any particular suggestions?

waggles:
* I fiddled with the factories and occupancy scopes bit to make them clearer while writing the specs
* specs using a fake price policy did blow up in a few ways, but seemed a bit too awkwardly unclear to add as pending for now

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -17,29 +17,33 @@ def initialize(occupancy)
...
         return unless occupancy.account_id?
 
-        ActiveRecord::Base.transaction do
-          create_order
-          create_order_detail
-
-          order.validate_order!
-          order.purchase!
-        end
+        create_order unless occupancy.order_detail_id?
```
waggles: if the `occupancy` here is `exit_at` only, we create the order for it, but it will fail the following condition and thus no fulfill attempt will be made--this should pave the way for either both-at-same-time and "scan in, scan in again and 2 orders are created"

jhanggi: I think this makes sense, I think these methods are just a little awkward. What if this were called `find_or_create_order` and push the predicate into that method?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -17,29 +17,33 @@ def initialize(occupancy)
...
         @order = Order.create!(
           account: occupancy.account,
           user: occupancy.user,
           facility: occupancy.facility,
           created_by_user: occupancy.user,
         )
-      end
-
-      def create_order_detail
         @order_detail = order.order_details.create!(
```
jhanggi: There's some almost circular logic here with your instance variables and methods (or at least it's very circuitous to me). I think it might be good to get rid of the memoized order/order_detail below and just pass around the instances. Maybe there's still a better way?

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/order_handler_spec.rb
```diff
@@ -58,6 +59,34 @@
...
+
+          it { is_expected.to be_complete }
+          it { is_expected.to be_fulfilled_at }
+
+          # TODO: only correct until the subsequent price specs are activated
+          it { is_expected.to be_problem }
+
+          xit "has price information" do
+            # TODO: expand once activated
+            expect(order_detail.price_policy).to be_present
```
jhanggi: I think we'll want to actually test for which policy it is, but this is fine for now.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -17,29 +17,42 @@ def initialize(occupancy)
...
 
-        ActiveRecord::Base.transaction do
-          create_order
-          create_order_detail
-
-          order.validate_order!
-          order.purchase!
-        end
+        find_or_create_order
+        fulfill_order if occupancy.complete?
```
jhanggi: It seems like it might be a good idea to leave the transaction here at this top level. Would we want to roll back any creations if the fulfillment fails? My brain is tired from traveling this morning, so I'm having trouble making arguments for or against.

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/order_handler_spec.rb
jhanggi: I just merged in #1001, which should have the calculations. I think if you rebase on master, you can probably un-pend this.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
waggles: I'm not fully sure of what's going on behind the scenes of `fulfill_order`, but it seems like we'd want to hang onto the order--if we just created the order and the fulfill fails, I think we still want the order around

jhanggi: Ok, makes sense.

jhanggi:
I think the only thing left to do here is to bring it the latest from master and update the specs for checking on price policy and pricing assignment.

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/order_handler_spec.rb
waggles: hmm still failing, but there's a buncha overlap in the current PRs, lemme look into what's happening

PR #992, opened 2017-04-25
[#135958] Fix error on item and service access list index

jhanggi:
Prior to #937, we were always setting the route for the `form_for` to
`facility_instrument_update_restrictions_path`, which would route, but
would be invalid. But services and items would never have schedule
groups so the submit would never appear, so we never had any problems.

In #937, we made it generic to support secure rooms, which also support
schedule groups.

A test around this would probably be a good idea, but I’d like to get it out there quickly.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -731,20 +731,6 @@ en:
       quantity: Quantity
 
   product_users:
-    index:
-      add_approved_user: Add Approved User
-      approval_explanation: |
-        This %{product_type} requires a user to be approved before they may
-        purchase the %{product_type}.
-      approved_users: Approved Users
-      confirm_removal: |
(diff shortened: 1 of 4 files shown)


config/locales/en.yml
```diff
@@ -731,20 +731,6 @@ en:
...
-      approved_users: Approved Users
-      confirm_removal: |
-        Are you sure you wish to remove this user from this %{product_type}?
-      no_users_approved: No users are currently approved.
-      th:
-        name: Name
-        user_name: User Name
-        email: Email
-      update: Update %{plural_label}
     new:
```
jhanggi: I did not move this locale because I know it is overridden in UIC/DC (they don't use netids in their searches).

PR #993, opened 2017-04-25
[#144059457] Display times on billing screens

jhanggi:
Working of what I started with in #986. I also have not addressed the specs yet.

I like @danhodos's comment about possibly thinking of some of this as presentation logic, but there is such a large reliance right now on `Reservation#to_s` as well as `order_detail.reservation`, that I chose to leave it there for now (although I did end up delegating it to a new presenter object), which also includes a new `TimeRange` object. That new class does mix some presentation (its `to_s` method) with business logic (`duration_mins`), and doesn't seem so complicated yet as to warrant splitting, but I could see it wanting to be split in the future.

There are places in business logic (see `OrderDetail#set_problem_order`) where we will need to switch to using something like `order_detail.time_data`.

Now that `Reservations::Rendering` is a more reasonable size, I'm thinking of folding it back into the main `Reservation` class.

One path I went down was using the `TimeData.for` mode of attack and have it return a SimpleDelegator around either reservation or occupancy, but `try` has problems with delegators, and there seem to be enough `try`s around that it might not be worth it. But something that may be worth revisiting.


app/models/order_detail.rb
```diff
@@ -843,10 +844,17 @@ def manual_fulfilled_at=(string)
...
 
+  def time_data
+    @time_data ||= product.time_data_for(self)
+  end
+
   private
 
   def has_completed_reservation?
-    !product.is_a?(Instrument) || (reservation && (reservation.canceled_at || reservation.actual_end_at || reservation.reserve_end_at < Time.zone.now))
+    return true unless product.timed?
```
jhanggi: This was a first attempt at using `time_data` in actual business logic. This will likely change in the future (one thing I'm considering is a NullObject for item/service time_data).

app/models/reports/export_raw.rb
```diff
@@ -69,9 +69,9 @@ def default_report_hash
         reservation_start_time: -> (od) { od.reservation.reserve_start_at if od.reservation },
         reservation_end_time: -> (od) { od.reservation.reserve_end_at if od.reservation },
         reservation_minutes: -> (od) { od.reservation.duration_mins if od.reservation },
-        actual_start_time: -> (od) { od.reservation.actual_start_at if od.reservation },
-        actual_end_time: -> (od) { od.reservation.actual_end_at if od.reservation },
-        actual_minutes: -> (od) { od.reservation.actual_duration_mins if od.reservation },
+        actual_start_time: -> (od) { od.time_data.actual_start_at if od.time_data },
```
jhanggi: I'm ignoring the rubocop warning here to maintain consistency with the rest of the file. Doing the rubocop cleanup should be its own PR/commit.

app/models/time_range.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def end_string
+    if start_at && end_at && start_at.day == end_at.day
+      l(end_at, format: :timeonly)
+    else
+      l(end_at)
+    end
+  end
+
+  def l(time, *options)
```
meara: I would lean towards this being called something like `localize` but I don't feel super strongly about it.

app/models/order_detail.rb
jhanggi: This is WIP, but here's what I'm doing with the null object: https://github.com/tablexi/nucore-open/commit/9d3cf661cd91df0632dc2173af3a90c2dd67e996 

It gets rid of the `timed?` method.

app/models/instrument.rb
```diff
@@ -44,6 +44,10 @@ def self.reservation_only
   # Instance methods
   # -------
 
+  def time_data_for(order_detail)
```
jhanggi: I don't really envision using it to display canceled. We're not planning to add the `canceled_at`, etc fields to Occupancy, so those fields will be reservation specific until we get the refactor of moving those fields onto OrderDetail.

Part of the idea is that this really returns more of a duck-type of the things that are shared between reservations and occupancy, things like `actual_start_at` (we will probably alias `entry_at` to that field to maintain the interface and are-there-problems-like-missing-times questions.

We'll keep reservation-specific things to `order_detail.reservation`.

app/models/time_range.rb
jhanggi: I felt ok with it because `l` is the normal way of using `I18n.localize` in views (it's an alias). I'll change it, though. I think it'll be a little clearer that way.

app/models/instrument.rb
meara: Ah okay, that makes much more sense, thank you. 😄

app/models/order_detail.rb
meara: 👍

PR #994, opened 2017-04-26
[#134671] force responsive nav background color

yoon:
[#134671](https://pm.tablexi.com/issues/134671)

This PR is being submitted with the awareness that the green header background color gets split in two when the hamburger menu opens. To prevent this would require a *lot* of messy overrides on the way Bootstrap styles the header / menu, in both the small and desktop versions. It may also serve as a subtle indicator as to where the hamburger menu ends when it opens.

screen shot 2017-04-26 at 10 57 24 am

### DIFF
--- a/app/assets/stylesheets/responsive.scss
+++ b/app/assets/stylesheets/responsive.scss
@@ -10,11 +10,11 @@ $navCollapseColor: #777;
   .visible-with-nav {
     display: none !important;
   }
-
   .nav-collapse {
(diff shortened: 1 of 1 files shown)


yoon:
@jhanggi thanks for that suggestion. we're getting into some pretty hairy css specificity for the colors, so I documented it for clarity.

jhanggi:
Just awaiting CI. There was a test failure that I'm pretty sure was unrelated.

PR #996, opened 2017-04-28
Refactor time_data to use Null Object

jhanggi:
I had to un-memoize time_data largely to deal with factories and specs where `@time_data` was getting called in the `set_problem_order` callback before the reservation was added. It should be quick enough that it won't really matter.

One thing this does is remove the delegation from Reservation#problem? to OrderDetail, which is only used in specs, and aliases `requires_but_missing_actuals?` to `problem?`. Occupancies will need to implement `problem?`.


app/models/order_detail.rb
```diff
@@ -845,16 +846,18 @@ def manual_fulfilled_at=(string)
   end
 
   def time_data
-    @time_data ||= product.time_data_for(self)
+    if product.respond_to?(:time_data_for)
```
jhanggi: This was the solution I came up with for dealing with items/services should be treated as null objects, but instruments require a reservation, so `product.time_data_for(self) || NullTimeData.new` was problematic in the cases where the instrument order doesn't have a reservation (mostly happens in the cart).

jhanggi: Not sure I love it yet: I might go back to more of the constant lookup. Still need to explore, but this works for now.

app/models/time_data/required_time_data.rb
```diff
@@ -0,0 +1,19 @@
...
+
+    def order_completable?
+      false
+    end
+
+    def blank?
+      true
+    end
+
+    def problem?
```
meara: This is a small thing, but it would be nice if the methods on `RequiredTimeData` were in the same order as in `NullTimeData`, just so it's easier to see the differences.

jhanggi: For sure, you're totally right.

PR #997, opened 2017-04-28
[#144318485] Refactor instrument price policy calculations

jhanggi:
Refactors the instrument price policy calculations so a good chunk of the logic is now in a separate class which I plan use for the secure rooms price policies calculations.

There's still more refactoring that could be done (I hate the interface for price policy calculations, but I didn't want to break too much). I'm not sure how much more I'll do on that side right now, but there's definitely specs to be moved or written.

### DIFF
--- a/spec/models/instrument_price_policy_calculations_spec.rb
+++ b/spec/models/instrument_price_policy_calculations_spec.rb
@@ -17,17 +17,6 @@ RSpec.describe InstrumentPricePolicyCalculations do
     policy.calculate_cost_and_subsidy_from_order_detail fake_order_detail
   end
 
-
-  # it "calculates cost based on the given duration and discount" do
-  #   policy.usage_rate = 5
-  #   expect(policy.calculate_cost(120, 0.15).round 2).to eq 1.5
-  # end
-
-  # it "calculates subsidy based on the given duration and discount" do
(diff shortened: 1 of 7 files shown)


jhanggi:
This is ready for a real review now.

spec/models/instrument_price_policy_calculations_spec.rb
```diff
@@ -16,18 +17,58 @@
     policy.calculate_cost_and_subsidy_from_order_detail fake_order_detail
   end
 
-  it "calculates cost based on the given duration and discount" do
-    policy.usage_rate = 5
-    expect(policy.calculate_cost(120, 0.15).round 2).to eq 1.5
+  describe "estimating cost and subsidy from an order detail" do
```
meara: These specs read a lot more like feature specs than unit tests, and to be honest it's hard for me to tell if they are testing all the right things. I gather most of the changes in this file are simple rearrangements, and it may not be worth the effort of re-writing these at this point, but thought I would flag it in case we end up doing more refactoring around price policies.

jhanggi: Yeah, these are kind of higher-level than unit tests, especially now that I've introduced the additional class. I ended up writing more unit-level tests for that, and left most of these tests around largely as regression protection.

PR #998, opened 2017-04-28
134682 price groups ui

meara:
See only the last commit ("admins can toggle..."). Previous commits are part of https://github.com/tablexi/nucore-open/pull/962.

screen shot 2017-04-28 at 1 55 54 pm
screen shot 2017-04-28 at 1 56 06 pm


app/controllers/users_controller.rb
```diff
@@ -166,6 +166,10 @@ def edit_user_params
     params.require(:user).permit(:email, :first_name, :last_name)
```
jhanggi: Since we're now allowing form submission for netid users, we should restrict the permitted params here in the controller.

db/migrate/20170421154535_create_default_price_groups_for_users.rb
```diff
@@ -0,0 +1,11 @@
+class CreateDefaultPriceGroupsForUsers < ActiveRecord::Migration
+
+  def up
+    User.find_each(&:create_default_price_group!)
```
jhanggi: We should probably remove any existing base/external memberships. I think it only happens in open, but after running the migrations, I have uses who have both internal and external groups.

meara:
The failing tests are coming from `new`, I realized--we don't want to expose this option for new user creation, right? I might like to pair a bit to tighten up this code.

app/models/user.rb
```diff
@@ -216,4 +203,28 @@ def self.active
...
   end
 
+  def internal?
+    price_groups.include?(PriceGroup.base)
+  end
+
+  def update_price_group!(params)
+    if params[:internal] == "true"
+      price_group_members.find_by(price_group: PriceGroup.external).try(:destroy)
+      price_group_members.create(price_group: PriceGroup.base)
```
jhanggi: I feel like this would be clearer as `find_or_create_by`. I was initially confused as to why it wan't creating duplicates every time I submitted the form.

app/controllers/users_controller.rb
meara: Do we need to conditionally restrict them? Not sure quite what the best way is to do this.

db/migrate/20170421154535_create_default_price_groups_for_users.rb
meara: If we have users who have custom price group memberships, do we want to preserve those? i.e. instead of removing existing memberships, if they have a base or default membership, we just do nothing?

app/views/users/_form.html.haml
```diff
@@ -1,9 +1,11 @@
 = simple_form_for [current_facility, @user] do |f|
   .form-inputs
-    = f.input :first_name
-    = f.input :last_name
-    = f.input :email
+    = f.input :first_name, readonly: @is_edit && !@user.admin_editable?
```
meara: I am fully aware that this is extremely gross. Can you suggest a better way of handling this?

app/views/users/_form.html.haml
```diff
@@ -1,9 +1,11 @@
-= simple_form_for [current_facility, @user] do |f|
+= simple_form_for [current_facility, user] do |f|
   .form-inputs
-    = f.input :first_name
-    = f.input :last_name
-    = f.input :email
+    = f.input :first_name, readonly: local_assigns[:edit] && !user.admin_editable?
```
meara: This is a lot of logic to have in a view, obviously. Does this make sense as a helper method, or is there a better way of doing this?

meara:
Closed in favor of #999.

PR #999, opened 2017-04-28
[#134682] UI for editing a user's base price group

meara:
https://pm.tablexi.com/issues/134682

- Depends on #999


app/views/users/edit.html.haml
```diff
@@ -8,8 +8,19 @@
...
-= render "form", button_text: text("shared.update")
+
+= simple_form_for [current_facility, @user] do |f|
+  .form-inputs
+    = f.input :first_name, readonly: !@user.admin_editable?
+    = f.input :last_name, readonly: !@user.admin_editable?
+    = f.input :email, readonly: !@user.admin_editable?
+    = f.input :username, readonly: true
+    - if current_user.administrator?
+      = f.input :internal?, label: 'Internal Pricing', as: :select, default: false
```
jhanggi: I18n.

app/views/users/show.html.haml
```diff
@@ -2,17 +2,18 @@
...
-  = render :partial => 'admin/shared/tabnav_users', :locals => {:secondary_tab => 'details'}
+  = render partial: 'admin/shared/tabnav_users', locals: { secondary_tab: 'details' }
 
 %h1= @user
 
 = readonly_form_for :user do |f|
   = f.input :full_name
   = f.input :email
-  = f.input :last_sign_in_at, :default_value => 'none'
+  = f.input :last_sign_in_at, default_value: 'none'
```
jhanggi: We should switch to double quotes when we can.

app/controllers/users_controller.rb
```diff
@@ -163,7 +163,11 @@ def update
   private
 
   def edit_user_params
-    params.require(:user).permit(:email, :first_name, :last_name)
+    params.require(:user).permit(:email, :first_name, :last_name) if @user.admin_editable?
+  end
+
+  def price_group_params
+    params.require(:user).permit(:internal)
```
jhanggi: If we're restricting the form element base on user role, we should probably restrict the allowed params as well.

PR #1000, opened 2017-04-28
[#134547] Restrict fulfilled at inputs

jhanggi:
Restrict to the beginning of the previous fiscal year up until today. If it is invalid, display an error.

I don't love the interface for failure when adding to an existing order in that it resets your inputs, but that's the current behavior for the failure case of trying to add an Instrument as complete.


app/models/valid_fulfilled_at_date.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  include ActiveModel::Validations
+  include DateHelper
+
+  validate :valid_format
+  validate :not_in_future, if: :to_time
+  validate :recent, if: :to_time
+
+  def self.min
+    # Beginning of previous fiscal year
```
meara: Okay, maybe I'm missing something, but why the previous fiscal year and not the current fiscal year? Wouldn't the previous fiscal year already be closed?

app/models/valid_fulfilled_at_date.rb
```diff
@@ -0,0 +1,56 @@
...
+    @string
+  end
+
+  # Returns nil if the date is invalid
+  def to_time
+    time = parse_usa_date(@string).try(:to_date)
+    time.beginning_of_day + 12.hours if time.present?
+  end
+  alias presence to_time
+  # This allows us to set `fulfilled_at = ValidFulfilledAtDate.new("XX/XX/XXXX")`
```
meara: I am a little confused by what's going on here. Is the "this" you're referring to for `presence`, or `in_time_zone`?

app/services/order_appender.rb
```diff
@@ -10,26 +10,25 @@ def initialize(original_order, user)
   end
 
   def add!(product, quantity, params)
-    products = product.is_a?(Bundle) ? product.products : [product]
+    OrderDetail.transaction do
```
meara: 👍

spec/controllers/facility_orders_controller_spec.rb
```diff
@@ -283,18 +284,18 @@
...
+                expect(flash[:error]).to include "cannot be in the future"
               end
             end
 
-            context "to a date before the start of this fiscal year" do
-              let(:fulfilled_at) { SettingsHelper.fiscal_year_beginning - 1.day }
+            context "to a date before the start of the previous fiscal year" do
+              let(:fulfilled_at) { SettingsHelper.fiscal_year_beginning - 1.year - 1.day }
 
               it_should_allow :director, "to add an item to existing order with fulfilled_at set to now" do
```
meara: Does this comment need updating? It doesn't seem to make sense with what the test is doing...

app/models/valid_fulfilled_at_date.rb
jhanggi: It's what the client asked for. They're mainly concerned about is the changeover to the new fiscal year as they're closing out the previous year. They want to be able to place orders into the previous fiscal year.

app/models/valid_fulfilled_at_date.rb
jhanggi: I can update the comment, but it's the `in_time_zone`. If you have that method, AR will treat it as a time, otherwise AR will throw it away as `nil`.

spec/controllers/facility_orders_controller_spec.rb
jhanggi: Probably, I need to figure out what the comment is actually trying to say...

jhanggi:
@meara updated.

PR #1001, opened 2017-04-29
[#144318485] Secure Room PricePolicy calculations

jhanggi:
I'm putting this into #997 so it's easier to see the diff. It's actually fairly trivial after that refactor.

I'm assuming that we'll test the actual assignment to an order detail as part of the scan out stories.

I'm also assuming that the specs in `PricePolicies::TimeBasedPriceCalculator` handle the majority of the edge and complicated cases, so I just did basic tests to make sure we're wired up properly.

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_room_price_policy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_room_price_policy.rb
@@ -2,4 +2,22 @@ class SecureRoomPricePolicy < PricePolicy
 
   include PricePolicies::Usage
 
+  def estimate_cost_and_subsidy_from_order_detail(order_detail)
+    calculate_cost_and_subsidy_from_order_detail(order_detail)
+  end
+
+  def calculate_cost_and_subsidy_from_order_detail(order_detail)
+    entry_at = order_detail.occupancy.entry_at
+    exit_at = order_detail.occupancy.exit_at
+    return unless entry_at && exit_at
+
+    calculator.calculate(entry_at, exit_at)
+  end
+
+  private
+
+  def calculator
+    PricePolicies::TimeBasedPriceCalculator.new(self)
+  end
+
 end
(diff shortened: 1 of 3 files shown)


vendor/engines/secure_rooms/app/models/secure_room_price_policy.rb
```diff
@@ -2,4 +2,22 @@ class SecureRoomPricePolicy < PricePolicy
 
   include PricePolicies::Usage
 
+  def estimate_cost_and_subsidy_from_order_detail(order_detail)
+    calculate_cost_and_subsidy_from_order_detail(order_detail)
+  end
+
+  def calculate_cost_and_subsidy_from_order_detail(order_detail)
+    entry_at = order_detail.occupancy.entry_at
```
jhanggi: I'm assuming at this point that `occupancy` exists. I'd like to keep it that way for now, but if it turns out to not be the case, we'll learn a little bit of how the code works, and we can drop in a `try`.

meara: What's the plan for determining whether or not we may need a `try`? I gather there's a lot of pieces still to come with SecureRooms and pricing--I just want to make sure we don't lose track of this and ship something to prod with a weird edge case error buried in the code.

jhanggi: I left this as the more pessimistic route of raising an error if `occupancy` is nil. That case _shouldn't_ happen, and if it does it most likely indicates a bug upstream somewhere. This is the kind of thing I'd like to catch early on if we can rather than not raising and returning `nil`s, which would be harder to track down.

If we do end up needing the `try`, we should at that point have found out why and can account for that, or at the least comment on what happened.

meara: 👍

waggles:
👍

PR #1002, opened 2017-05-01
[#135268] Responsive order page and order detail page

noelrappin:
https://pm.tablexi.com/issues/135268

This makes the order and order detail page responsive. Notes:

my_orders_-_nucore

order__38-56_-_nucore

The detail page had Quantity/Product as a two-column header, which I split because the colspan header was confusing the JavaScript creating the headers.


wtfluckey:
👍

PR #1004, opened 2017-05-01
[#135269] Responsive order receipt

noelrappin:
order_receipt_-_nucore

https://pm.tablexi.com/issues/135269

The only new thing here is hiding the existing header for the totals column so that I can add a responsive one — the JS responsive header won’t really work here, so I added it explicitly.

### DIFF
--- a/app/views/orders/receipt.html.haml
+++ b/app/views/orders/receipt.html.haml
@@ -53,7 +53,7 @@
       %td.hide-on-mobile{colspan: colspan}
         %strong= t('.total')
       %td.currency
-        .responsive-header Total
+        .responsive-header= t('.total')
         = number_to_currency(@order_details.sum { |d| d.total || 0 })
 
 = render :partial => '/price_display_footnote'
--- a/app/views/orders/receipt.html.haml
+++ b/app/views/orders/receipt.html.haml
@@ -20,7 +20,7 @@
   = f.input :user
   = f.input :created_by_user
 
-%table.table.table-striped.table-hover.old-table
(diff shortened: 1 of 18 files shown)


app/views/orders/receipt.html.haml
```diff
@@ -47,10 +47,13 @@
...
         %td.currency=order_detail.wrapped_total
+  %tfoot
     %tr
       - colspan = @order.has_subsidies? ? 5 : 4
-      %td{colspan: colspan}
+      %td.hide-on-mobile{colspan: colspan}
         %strong= t('.total')
-      %td.currency= number_to_currency(@order_details.sum { |d| d.total || 0 })
+      %td.currency
+        .responsive-header Total
```
jhanggi: Is it the colspan that's making this necessary? If it is necessary, can we use the locale?

noelrappin: You are correct, I should be using the locale.

This is necessary because the footer fields don't pull in the JavaScript to get the responsive headers added, and I didn't think the footers were likely to be consistent enough to be programmatically added. So it's explicit (here and on the cart).

But I'll change it to the locale.

noelrappin:
Ugh, git rebase put some extra changes here

jhanggi:
I'm not quite sure how those extra commits ended up in there, but if you `git rebase -i master`, you can simply remove the lines from the commits with PR numbers (those were the squash & merged PRs).

noelrappin:
I think this happens when I rebase from master into the PR

PR #1005, opened 2017-05-02
[#135265] Fix for responsive tables in reservations

noelrappin:
my_reservations_-_nucore

The obvious problem here is the overflow of the caption "Reservation (Edit/Extend)". There's no particularly easy way to get that to wrap (because of how the responsive div is set up). Is shortening the header to "Reservation" an option?

### DIFF
--- a/app/assets/javascripts/app/responsive_table.js.coffee
+++ b/app/assets/javascripts/app/responsive_table.js.coffee
@@ -20,6 +20,5 @@ class window.ResponsiveTable
     header = $(@table.find("thead th").eq(index))
     header.data("mobile-header") || header.text()
 
-
 $ ->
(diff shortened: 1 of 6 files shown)


noelrappin:
my_reservations_-_nucore

Allows for a short name to be specified as a mobile header.

jhanggi:
Looks like if the "actions" is empty, the row height gets messed up. I'd be fine if the row were  hidden in mobile view (if that makes it easier to deal with in HTML/JS instead of CSS).

screen shot 2017-05-03 at 1 49 22 pm

noelrappin:
I can either ensure that there's at least an &nbsp; in the cell (which is what I've done in other places) or drop the row in mobile if it's empty.

PR #1006, opened 2017-05-02
[#135267] Responsive Files

noelrappin:
my_files_-_nucore

This one is straightforward -- the only thing to note is that it will push wide right if the filename is long.


noelrappin:
Thanks @danhodos for the CSS suggestion...

PR #1007, opened 2017-05-02
[#144058895] Refactor problem descriptions

waggles:
Need some feedback here:
* this uses the idea that `time_data` is one of 4 objects which can be aware of their own problems.. however it's odd that `RequiredTimeData` and `Reservation` have the same thing going on... what situation causes `RequiredTimeData` to be used?
* do we want to move the i18n somewhere else? they are under a "shared" key which could be fine, not sure of convention here
* any reason to further move `:price_policy_missing if ...blank?` or is it cool in `OrderDetail`

### DIFF
--- a/app/helpers/facility_orders_helper.rb
+++ b/app/helpers/facility_orders_helper.rb
@@ -20,7 +20,7 @@ module FacilityOrdersHelper
     output = build_warnings(notices[:warnings])
     output += build_notices(notices[:notices])
 
-    output.join(" ").html_safe
+    safe_join(output)
   end
 
   def banner_date_label(object, field, label = nil)
--- a/app/helpers/facility_orders_helper.rb
+++ b/app/helpers/facility_orders_helper.rb
@@ -9,7 +9,7 @@ module FacilityOrdersHelper
     notices << "can_reconcile" if order_detail.can_reconcile_journaled?
     notices << "in_open_journal" if order_detail.in_open_journal?
 
-    warnings = [order_detail.problem_description]
+    warnings = Array(order_detail.problem_description)
 
     { warnings: warnings, notices: notices }
   end
--- a/app/helpers/facility_orders_helper.rb
+++ b/app/helpers/facility_orders_helper.rb
@@ -9,9 +9,7 @@ module FacilityOrdersHelper
     notices << "can_reconcile" if order_detail.can_reconcile_journaled?
     notices << "in_open_journal" if order_detail.in_open_journal?
 
-    warnings = []
-    warnings << "missing_actuals" if order_detail.complete? && order_detail.reservation.try(:requires_but_missing_actuals?)
-    warnings << "missing_price_policy" if order_detail.missing_price_policy? && !order_detail.reservation.try(:requires_but_missing_actuals?)
+    warnings = [order_detail.problem_description]
 
     { warnings: warnings, notices: notices }
   end
@@ -19,8 +17,8 @@ module FacilityOrdersHelper
   def order_detail_badges(order_detail)
     notices = order_detail_notices(order_detail)
 
-    output = build_badges(notices[:warnings], "label-important")
-    output += build_badges(notices[:notices], "label-info")
+    output = build_warnings(notices[:warnings])
+    output += build_notices(notices[:notices])
 
     output.join(" ").html_safe
   end
@@ -46,9 +44,15 @@ module FacilityOrdersHelper
 
   private
 
-  def build_badges(notices, label_class)
+  def build_notices(notices)
     notices.map do |notice|
-      content_tag(:span, t("order_details.notices.#{notice}.badge"), class: ["label", label_class])
+      content_tag(:span, t("order_details.notices.#{notice}.badge"), class: ["label", "label-info"])
+    end
+  end
+
+  def build_warnings(warnings)
+    warnings.map do |warning|
+      content_tag(:span, warning, class: ["label", "label-important"])
     end
   end
 
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -8,6 +8,7 @@ class OrderDetail < ActiveRecord::Base
   include OrderDetail::Accessorized
   include NUCore::Database::WhereIdsIn
   include NUCore::Database::ClobSafeDistinct
+  include TextHelpers::Translation
 
   versioned
 
@@ -30,7 +31,7 @@ class OrderDetail < ActiveRecord::Base
 
   before_save :set_problem_order
   def set_problem_order
-    self.problem = complete? && problem_description_key.present?
+    self.problem = complete? && problem_description.present?
     update_fulfilled_at_on_resolve if time_data.present?
     true # problem might be false; we need the callback chain to continue
   end
@@ -852,6 +853,17 @@ class OrderDetail < ActiveRecord::Base
     end
   end
 
+  def translation_scope
+    "activerecord.models.order_detail"
+  end
+
+  def problem_description
+    time_data_problem = time_data.problem_description
+    price_policy_problem = text(:price_policy_missing) if price_policy.blank?
+
+    time_data_problem || price_policy_problem
+  end
+
   private
 
   # Is there enough information to move an associated order to complete/problem?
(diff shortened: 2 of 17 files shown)


app/models/reservation.rb
```diff
@@ -282,6 +282,10 @@ def requires_but_missing_actuals?
   end
   alias problem? requires_but_missing_actuals?
```
jhanggi: If we're only ever using the key or checking its presence, should we get rid of `time_data#problem?`?

app/models/time_data/required_time_data.rb
```diff
@@ -10,6 +10,10 @@ def problem?
       true
     end
 
+    def problem_description_key
```
jhanggi: `RequiredTimeData` should only happen on instrument orders where the reservation doesn't exist. There are two scenarios I can think of: 1) a bug (this happened recently, we're cleaning up most of the data, but there will still be a few complete orders with no reservation) 2) when you're creating an order in your cart (part of a bundle, adding multiple reservations at once, etc), the order detail is created, but the reservation doesn't exist yet.

So in the first case, we would hit this because the order is complete, and we would want to consider it a problem.

In the second case, I don't think we ever hit the `problem?` method (we do hit  `order_completable?`), but it's still useful for staying parallel with the other classes.

app/views/shared/_problem_order_details.html.haml
```diff
@@ -28,9 +28,7 @@
         %td.centered= order_detail.quantity
         %td= order_detail_description(order_detail)
         %td= order_detail.account
-        - if order_detail.requires_but_missing_actuals?
-          %td.centered= "#{t ".actual_usage_missing"} #{link_to "Enter", manage_order_detail_path(order_detail), class: "manage-order-detail"}".html_safe
-        - else
-          %td.centered= "#{t ".price_policy_missing"} #{link_to "Assign", manage_order_detail_path(order_detail), class: "manage-order-detail"}".html_safe
+        - if order_detail.problem?
```
jhanggi: Do we need this conditional? I'm pretty sure all the orders that get rendered on this page are problems. I think we just had the check before to deal with deciding which message to display.

app/views/shared/_problem_order_details.html.haml
```diff
@@ -28,9 +28,7 @@
...
+          %td.centered= "#{t order_detail.problem_description_key} #{link_to "Update", manage_order_detail_path(order_detail), class: "manage-order-detail"}".html_safe
```
jhanggi: We should probably split this into multiple lines.
```
%td.centerered
  = text(order_detail.problem_description_key) # we should put it inside some model-level scope, not view level
  = link_to text("shared.update"), manage_order_detail_path(order_detail), class: "manage-order-detail"
```

app/models/order_detail.rb
```diff
@@ -926,4 +925,8 @@ def set_reconciled_at
     self.reconciled_at ||= Time.current
   end
 
+  def problem_description_key
+    time_data.problem_description_key || (:price_policy_missing if price_policy.blank?)
```
jhanggi: Something kind of bothers me about how this reads, but I'm not quite sure how to improve it.

Maybe something like this (unsure if this syntax actually works):
```
[
  time_data.problem_description_key,
  :price_policy_missing if price_policy.blank?,
].first
```

waggles:
so translation-wise... do we want to have res/occ/timedata have something like
```
def problem_description
  text(:missing_something)
end
```
and then pass that translated string down instead of the key?

jhanggi:
Hmm, that's a good question. That would allow us to use different translation scopes for different models. And could avoid some duplication in the views. Does it make testing any harder?

Also, we should look at `facility_orders/show` and dig down into the `order_detail_badges` helper. That should probably change to support this. I think that's the only other place we're showing problem descriptions.

jhanggi:
Should we add a few tests to make sure we're getting the right symbols/strings back?

waggles:
Hmm let's have a look at that `FacilityOrdersHelper` tomorrow perhaps, so I can pick up the context more quickly

config/locales/en.yml
```diff
@@ -1106,6 +1104,15 @@ en:
     recently_reviewed: Recently Reviewed
 
   activerecord:
+    models:
```
jhanggi: We should probably move these into `en.models.yml`. If we put these under `errors`, we could possibly use `self.errors.generate_message(:price_policy, :missing)`. That might be problematic for the null objects, though, and it might be overkill.

spec/models/reservation_spec.rb
```diff
@@ -1487,4 +1487,14 @@
...
     end
   end
+
+  describe "#problem_description" do
+    subject(:problem_description) { reservation.problem_description }
+
+    context "requires_but_missing_actuals?" do
+      before { expect(reservation).to receive(:requires_but_missing_actuals?).and_return(true) }
+
+      it { is_expected.to eq reservation.text(:actual_usage_missing) }
```
jhanggi: Might be a good idea to test the opposite case?

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -58,6 +60,18 @@ def actual_duration_mins
...
+    def problem_description
+      if entry_at.blank?
+        text(:missing_entry)
+      elsif exit_at.blank?
+        text(:missing_exit)
+      end
+    end
+
+    def translation_scope
+      "secure_rooms/occupancy"
```
jhanggi: If we're putting the other model messages under `activerecord.models`, we should probably do the same here.

app/helpers/facility_orders_helper.rb
```diff
@@ -9,18 +9,16 @@ def order_detail_notices(order_detail)
...
     { warnings: warnings, notices: notices }
   end
 
   def order_detail_badges(order_detail)
     notices = order_detail_notices(order_detail)
 
-    output = build_badges(notices[:warnings], "label-important")
-    output += build_badges(notices[:notices], "label-info")
+    output = build_warnings(notices[:warnings])
+    output += build_notices(notices[:notices])
```
jhanggi: Is this something we can use `safe_join` for and get rid of the extra `html_safe` marker?

jhanggi:
Just one minor comment, otherwise looks good.

waggles:
@jhanggi updated

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/order_handler_spec.rb
```diff
@@ -74,16 +74,23 @@
         describe "order_detail" do
           subject(:order_detail) { described_class.process(occupancy).order_details.first }
 
+          before do
+            SecureRoomPricePolicy.create(
```
jhanggi: Should we use the factory?

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/order_handler_spec.rb
```diff
@@ -74,16 +74,22 @@
         describe "order_detail" do
           subject(:order_detail) { described_class.process(occupancy).order_details.first }
 
+          let(:price_group) { PriceGroup.first }
```
waggles: I'm missing something conceptually here; I tried creating a `PriceGroup` in this let based on what attributes `PriceGroup.first` (which passes) has, but that didn't pass

app/helpers/facility_orders_helper.rb
```diff
@@ -9,20 +9,18 @@ def order_detail_notices(order_detail)
     notices << "can_reconcile" if order_detail.can_reconcile_journaled?
     notices << "in_open_journal" if order_detail.in_open_journal?
 
-    warnings = []
-    warnings << "missing_actuals" if order_detail.complete? && order_detail.reservation.try(:requires_but_missing_actuals?)
-    warnings << "missing_price_policy" if order_detail.missing_price_policy? && !order_detail.reservation.try(:requires_but_missing_actuals?)
+    warnings = [order_detail.problem_description]
```
jhanggi: Do we need a `compact` to make sure this is fine if there isn't a warning?

app/helpers/facility_orders_helper.rb
```diff
@@ -9,20 +9,18 @@ def order_detail_notices(order_detail)
     notices << "can_reconcile" if order_detail.can_reconcile_journaled?
     notices << "in_open_journal" if order_detail.in_open_journal?
 
-    warnings = []
-    warnings << "missing_actuals" if order_detail.complete? && order_detail.reservation.try(:requires_but_missing_actuals?)
-    warnings << "missing_price_policy" if order_detail.missing_price_policy? && !order_detail.reservation.try(:requires_but_missing_actuals?)
+    warnings = Array(order_detail.problem_description)
```
waggles: since it's just the one `nil` or a string I think this makes sense

PR #1008, opened 2017-05-02
[#144059361] move canceled_* to order detail

meara:
https://www.pivotaltracker.com/story/show/144059361


app/helpers/timeline_helper.rb
```diff
@@ -68,7 +68,7 @@ def public_timeline?
   end
 
   def display_status_class(reservation)
-    reservation.canceled_at ? "canceled" : reservation.order_detail.order_status.to_s.downcase
+    reservation.order_detail.canceled_at ? "canceled" : reservation.order_detail.order_status.to_s.downcase
```
jhanggi: Could this method take the order detail as its argument? Would that clean it up a bit?

app/models/order_detail.rb
```diff
@@ -220,7 +219,7 @@ def self.with_estimated_costs
 
   def in_review?
     # check in the database if self.id is in the scope
-    self.class.all_in_review.find_by_id(id) ? true : false
+    self.class.all_in_review.find_by(id: id) ? true : false
```
jhanggi: Could this use `self.class.all_in_review.exists?(id: id)` to get rid of the ternary?

app/models/order_detail.rb
```diff
@@ -288,9 +287,11 @@ def can_be_viewed_by?(user)
   scope :for_facilities, ->(facilities) { joins(:order).where("orders.facility_id in (?)", facilities) unless facilities.nil? || facilities.empty? }
   scope :for_products, ->(products) { where("order_details.product_id in (?)", products) unless products.blank? }
   scope :for_owners, lambda { |owners|
-    joins(:account)
-      .joins("INNER JOIN account_users on account_users.account_id = accounts.id and user_role = 'Owner'")
-      .where("account_users.user_id in (?)", owners) unless owners.blank?
+    unless owners.blank?
+      joins(:account)
+        .joins("INNER JOIN account_users on account_users.account_id = accounts.id and user_role = 'Owner'")
```
jhanggi: I like pulling the conditional up. Now that's it's at a block level, would it be clearer to use `if owners.any?`/`present?`?

Also, I feel like this could be cleaned up using something like `merge(AccountUser.owners)`, but I haven't taken the time to figure out exactly what that would be.

app/models/order_detail.rb
```diff
@@ -396,6 +397,10 @@ def self.ordered_or_reserved_in_range(start_date, end_date)
     end
   end
 
+  def is_canceled? # non-standard name to avoid conflict with AASM
+    canceled_at.present?
```
jhanggi: I really feel like we made a mistake a long time ago making canceled with charge go into "Complete" state, but that's a major refactoring and outside the scope of this work. But something to be aware of.

app/models/order_detail.rb
```diff
@@ -715,16 +720,18 @@ def in_dispute?
   end
 
   def disputed?
+    # only used in specs
     dispute_at.present? && !canceled?
   end
 
   def cancel_reservation(canceled_by, order_status: OrderStatus.canceled_status, admin: false, admin_with_cancel_fee: false)
+    # TODO: refactor more better
```
jhanggi: MOAR BETTER! Yes. But probably not now.

app/models/order_detail.rb
```diff
@@ -735,8 +742,8 @@ def cancel_reservation(canceled_by, order_status: OrderStatus.canceled_status, a
       end
     else
       return false unless res && res.can_cancel?
-      res.canceled_at = Time.zone.now # must set canceled_at after calling #can_cancel?
-      return false unless res.save
+      self.canceled_at = Time.zone.now # must set canceled_at after calling #can_cancel?
+      return false unless save
```
jhanggi: Could these two lines now be rolled up into something like `return false unless update(canceled_at: Time.current)`?

app/models/reports/export_raw.rb
```diff
@@ -33,55 +33,55 @@ def default_report_hash
       hash = {
         facility: :facility,
         order: :to_s,
-        ordered_at: -> (od) { od.order.ordered_at },
-        fulfilled_at: -> (od) { od.fulfilled_at },
-        order_status: -> (od) { od.order_status.name },
+        ordered_at: ->(od) { od.order.ordered_at },
```
jhanggi: I've been avoiding rubocopping this because of the annoyance of merge conflicts as I've been touching this quite a bit for rooms, but I think it's a worthwhile change. So we'll see who gets this in first!

app/models/reports/export_raw.rb
```diff
@@ -33,55 +33,55 @@ def default_report_hash
...
+        actual_cost: ->(od) { as_currency(od.actual_cost) },
+        actual_subsidy: ->(od) { as_currency(od.actual_subsidy) },
+        actual_total: ->(od) { as_currency(od.actual_total) },
+        reservation_start_time: ->(od) { od.reservation.reserve_start_at if od.reservation },
+        reservation_end_time: ->(od) { od.reservation.reserve_end_at if od.reservation },
+        reservation_minutes: ->(od) { od.reservation.duration_mins if od.reservation },
+        actual_start_time: ->(od) { od.time_data.actual_start_at if od.time_data.present? },
+        actual_end_time: ->(od) { od.time_data.actual_end_at if od.time_data.present? },
+        actual_minutes: ->(od) { od.time_data.actual_duration_mins if od.time_data.present? },
+        canceled_at: ->(od) { od.canceled_at },
```
jhanggi: This is totally fine, but fyi, if it's a method call directly on the order detail, you can just use the symbol `canceled_at: :canceled_at` rather than a proc.

app/models/reservation.rb
```diff
@@ -60,7 +71,7 @@ def self.joins_order
       .joins("LEFT JOIN orders ON orders.id = order_details.order_id")
   end
 
-  scope :not_canceled, -> { where(canceled_at: nil) }
+  scope :not_canceled, -> { where(order_details: { canceled_at: nil }) } # dubious
```
jhanggi: Indubitably. Might be interesting to see if we can get this scope out of reservation and into OrderDetail itself.

app/services/cancellation_fee_calculator.rb
```diff
@@ -12,10 +12,10 @@ def fee
 
   def calculated_fee
     unless defined?(@calculated_fee)
-      original_canceled_at = @reservation.canceled_at
-      @reservation.canceled_at = Time.current
+      original_canceled_at = @reservation.order_detail.canceled_at
```
jhanggi: Hmm, this is a spot where now I question the value of removing the delegate. Might be worth discussing.

app/services/instrument_offline_reservation_canceler.rb
```diff
@@ -18,7 +18,7 @@ def admin_user
 
   def cancel_reservation(reservation)
     reservation.order_detail.cancel_reservation(admin_user, admin: true)
-    reservation.update_attribute(:canceled_reason, "The instrument was offline")
+    reservation.order_detail.update_attributes(canceled_reason: "The instrument was offline")
```
jhanggi: I wonder if now `canceled_reason` should be an optional argument in `OrderDetail#cancel_reservation`.

config/locales/en.models.yml
```diff
@@ -302,9 +302,9 @@ en:
         actuals_string: Actual Usage
         actual_start_date: Actual Start
         actual_duration_mins: Actual Duration
-        canceled_by_user: Canceled By
-        canceled_by: Canceled By
-        canceled_at: Canceled At
+        canceled_by_user: Canceled By #
```
jhanggi: I'm guessing these `#` are meant as TODOs?

db/migrate/20170501201633_add_canceled_to_order_detail.rb
```diff
@@ -0,0 +1,19 @@
+class AddCanceledToOrderDetail < ActiveRecord::Migration
+  def up
+    add_column :order_details, :canceled_at, :datetime
+    add_column :order_details, :canceled_by, :integer
+    add_column :order_details, :canceled_reason, :string
+
+    Reservation.find_each do |reservation|
+      reservation.order_detail.update_attributes(canceled_at: reservation.canceled_at,
+                                                 canceled_by: reservation.canceled_by,
+                                                 canceled_reason: reservation.canceled_reason)
```
jhanggi: Should we add the remove columns to the reservations table? Totally cool leaving that be for now since it probably makes going up and down in the database easier and quicker while this is a WIP.

spec/controllers/facility_reservations_controller_spec.rb
```diff
@@ -188,7 +188,7 @@
 
     context "redirect on no edit" do
       before :each do
-        @reservation.update_attribute(:canceled_at, Time.zone.now)
+        @reservation.order_detail.update_attributes(canceled_at: Time.zone.now)
```
jhanggi: nbd, but I suspect we used `update_attribute` to be marginally quicker in the specs because we didn't care about validations. Does rubocop care about this? If so, does `update_column` also trigger the warning? I don't feel strongly about this.

Also, not caring too strongly, I believe we're leaning slightly in favour (all amatter of taste :)) of `Time.current` instead of `Time.zone.now`

spec/controllers/orders_controller_spec.rb
```diff
@@ -734,9 +734,9 @@ def setup_past_reservation
               expect(assigns[:order].order_details).to all(be_canceled)
             end
 
-            it "sets the canceled time on the reservation" do
-              assigns[:order].order_details.all? { |od| expect(od.reservation.canceled_at).not_to be_nil }
-              expect(reservation.reload.canceled_at).not_to be_nil
+            it "sets the canceled time on the order_detail" do
+              assigns[:order].order_details.all? { |od| expect(od.canceled_at).not_to be_nil }
+              expect(reservation.order_detail.reload.canceled_at).not_to be_nil
```
jhanggi: It's a pain to see in github, so I'm not looking too closely, but if we have direct access to `order_detail` as a variable in specs, I'd lean towards using that over `reservation.order_detail`.

spec/models/order_detail_spec.rb
```diff
@@ -208,7 +208,7 @@
     is_expected.not_to allow_value(nil).for(:product_id)
   end
 
-  it "should have a order" do
+  it "should have an order" do
```
jhanggi: Nice catch.

spec/services/accessories_child_updater_spec.rb
```diff
@@ -63,7 +63,7 @@
       end
 
       it "moves the child to canceled" do
-        expect(child_order_detail).to be_canceled
+        expect(child_order_detail.state).to eq "canceled"
```
jhanggi: I slightly lean towards the older syntax, but only slightly. Is this related to the `is_canceled?` method we've added?

meara: Oops, yeah, that was to avoid a name collision before I renamed to `is_canceled?`, forgot to revert that one.

spec/controllers/orders_controller_spec.rb
meara: Nope, no direct access, at least not for the order_detail that this reservation belongs to.

spec/controllers/facility_reservations_controller_spec.rb
meara: Rubocop flags on both `update_attribute` and `update_column`. I hesitate to change `Time.zone.now`, mostly because it's used multiple times throughout the file and I don't want to be inconsistent, but I also don't want to go on a huge style tangent in this PR.

db/migrate/20170501201633_add_canceled_to_order_detail.rb
meara: I didn't add the remove_columns into this migration because I thought perhaps we would prefer to try this on stage first and verify that the data migration went smoothly before dropping a bunch of data.  I'm not sure what the best strategy is for data migrations like this--might be good for the two of us to discuss.

app/services/instrument_offline_reservation_canceler.rb
meara: ...maybe, yeah. Will think about that.

app/services/cancellation_fee_calculator.rb
meara: Other than this being a bit clunky, what are the arguments you see in favor of the delegate?

app/models/reservation.rb
meara: Yeah, the way this ends up being used are generally in some pretty complicated queries that involve a lot of joins. Might be good to pair on finding a better solution.

app/models/order_detail.rb
meara: I was wondering what the purpose was of having `canceled` in the state machine and also separate `canceled_at` fields. Makes the logic fairly convoluted just to keep the data consistent.

spec/models/reservation_spec.rb
```diff
@@ -308,20 +311,30 @@
   end
 
   describe "#canceled?" do
+    # TODO: move to order_detail_spec
```
meara: I figure I'll leave these tests here, since Reservation#canceled? is technically still a method, and add some separately for `OrderDetail#is_canceled?`.

app/services/cancellation_fee_calculator.rb
jhanggi: Really just that this is now clunkly and getting close to law of demeter territory. Might be we don't worry about this for now. A better solution might be to have the cancellation fee calculator work on an OrderDetail rather than a reservation (I could see this coming up when we get to multiple signup).

db/migrate/20170501201633_add_canceled_to_order_detail.rb
jhanggi: That's probably a smart idea. This also reminds me we'll need to make sure to inform NU's BI team that these columns will be moving. It'll probably be confusing if they keep pointing at the old columns, but no new data is getting populated into them.

app/services/cancellation_fee_calculator.rb
meara: I would lean towards cleaning this up in a later refactoring, I think. I hesitate to add much more to this PR. Multiple signups might be a good opportunity.

meara:
@jhanggi I think this should be ready for a real review now. Also, it would be good to discuss what the migration/release plan is for this--maybe we'll want to have this in a release by itself, and have a second PR/release to drop the old DB columns once we've verified that the data migration went smoothly.

app/models/order_detail.rb
```diff
@@ -716,16 +722,16 @@ def in_dispute?
...
-    res = reservation
-    res.canceled_by = canceled_by.id
+  def cancel_reservation(canceled_by, canceled_reason: nil, order_status: OrderStatus.canceled_status, admin: false, admin_with_cancel_fee: false)
+    self.canceled_by = canceled_by.id
+    self.canceled_reason = canceled_reason
 
     if admin
-      res.canceled_at = Time.zone.now
-      return false unless res.save
+      return false unless update_attributes(canceled_at: Time.current)
```
jhanggi: Just curious where this might fail? I'm wondering if we can just use `assign_attributes` and then the save gets handled further down. I care a little less about the double save from a performance standpoint, but more from the versions table being easier to read (it's not something I have to look at too ofter, but when I do it's good to have it sane).

However, for this, I would lean towards ease of implementation and readability.

app/models/reservation.rb
```diff
@@ -36,6 +35,11 @@ class Reservation < ActiveRecord::Base
   delegate :lock_window, to: :product, prefix: true
   delegate :owner, to: :account, allow_nil: true
 
+  def canceled?
+    return nil unless order_detail
```
jhanggi: Should we return `false` here because it's a predicate method?

app/support/auto_canceler.rb
```diff
@@ -25,7 +25,7 @@ def cancelable_reservations
 
   def cancel_reservation(res)
     res.order_detail.cancel_reservation(admin, admin: true, admin_with_cancel_fee: true)
-    res.update_attribute :canceled_reason, "auto canceled by system"
+    res.order_detail.update_attributes(canceled_reason: "auto canceled by system")
```
jhanggi: Can we now do this in one line with the `canceled_reason` argument?

app/models/reservation.rb
meara: I went with returning `nil` because that duplicates the behavior of `delegate` with `allow_nil: true`, which is what I was originally doing before I had to write a custom method due to the name change. I'm fine with changing it to false, though.

spec/models/reservation_spec.rb
jhanggi: If we can't come up with something, it's fine, but it might be nice if we can think of a better name for our `is_canceled?` method. Something that encompass the fact it might mean canceled with fee (which is technically in complete state).

`manually_canceled`, `actual_canceled?`,`actual_cannibal_shia_lebeouf?`. I don't know. I got nothing.

jhanggi:
Yeah, I agree a separate release for this would be nice. We're just about ready for the NU release of the price group changes, so maybe we can get this in after that.

spec/models/reservation_spec.rb
meara: Um.... `has_canceled_at`? That's basically what this is checking for, as distinct from whatever the state machine says.

Precisely what the difference in meaning is between the presence of `canceled_at` and the order_detail state of canceled is less clear to me--I still find it very weird that these are independent events. The `canceled` state seems kind of not-very-meaningful, since canceled order details can also be in `complete` state.

app/models/order_detail.rb
meara: I'm not sure why this method is structured the way it is--I mostly didn't change it, just updated it to work with the new schema.  I thought about refactoring it more extensively, but I'm a little afraid of breaking it--I personally can't see why the save should fail, but I also don't see where the later save is that you're referring to...

meara:
@jhanggi updated. What are we thinking about the timeline for merging this?

PR #1009, opened 2017-05-03
[#140891973] Complete orphans

waggles:
* Adds flag and allows rules to record in verdict that a particular `:grant` does not require an order
* Adds `#order_completable?` logic for `Occupancy` and completes associated orders when an occupancy becomes orphaned

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
@@ -33,6 +33,7 @@ module SecureRooms
 
     def mark_orphaned!
       update!(orphaned_at: Time.current)
+      order_detail.update_order_status!(user, OrderStatus.complete.first) if order_detail.present?
     end
 
     def orphan?
@@ -53,6 +54,10 @@ module SecureRooms
         exit_at: Time.current,
       )
       self
+    end
+
+    def order_completable?
+      orphaned_at? || (entry_at? && exit_at?)
     end
 
     def actual_duration_mins
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
@@ -13,6 +13,7 @@ module SecureRooms
 
     delegate :facility, to: :secure_room
     delegate :to_s, to: :range
+    delegate :skip_order?, to: :entry_event, allow_nil: true
 
     def self.valid
       where(orphaned_at: nil)
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
@@ -32,6 +32,10 @@ module SecureRooms
 
     def mark_orphaned!
       update!(orphaned_at: Time.current)
+      attempt_order_complete!
+    end
+
+    def attempt_order_complete!
       order_detail.complete! if order_completable?
     end
 
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
@@ -13,7 +13,6 @@ module SecureRooms
 
     delegate :facility, to: :secure_room
     delegate :to_s, to: :range
-    delegate :skip_order?, to: :entry_event, allow_nil: true
 
     def self.valid
       where(orphaned_at: nil)
@@ -33,7 +32,7 @@ module SecureRooms
 
     def mark_orphaned!
       update!(orphaned_at: Time.current)
-      order_detail.update_order_status!(user, OrderStatus.complete.first) if order_detail.present?
+      order_detail.complete! if order_detail.present?
     end
 
     def orphan?
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
@@ -23,20 +23,24 @@ module SecureRooms
     end
 
     def self.current
-      valid.where(exit_event_id: nil).where.not(entry_event_id: nil)
+      valid.where(exit_at: nil).where.not(entry_at: nil)
     end
 
     def self.recent
-      valid.where.not(exit_event_id: nil).limit(10)
+      valid.where.not(exit_at: nil).limit(10)
     end
 
     def mark_orphaned!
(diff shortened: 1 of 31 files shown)


vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -32,6 +33,7 @@ def self.recent
 
     def mark_orphaned!
       update!(orphaned_at: Time.current)
+      order_detail.update_order_status!(user, OrderStatus.complete.first) if order_detail.present?
```
jhanggi: Can we use `OrderStatus.complete_status`? I'd like for us to eventually get to the point of being able to use `OrderStatus.complete` as the reference to the actual status, but that's a tedious change.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -15,14 +15,16 @@ def initialize(occupancy)
       end
 
       def process
-        return unless occupancy.account_id?
+        return if occupancy.skip_order?
```
jhanggi: I think you're trying to account for the case of swipe-out with no swipe in (because we'll have no order at that point), but I'm not super crazy about adding a database field, especially at the event level for this. Is there another way we can accomplish this?

waggles:
@jhanggi updated, just always creates an account for now, and instead just doesn't try to purchase without the account_id

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -15,10 +15,8 @@ def initialize(occupancy)
       end
 
       def process
-        return unless occupancy.account_id?
-
         find_or_create_order
-        fulfill_order if occupancy.complete?
+        occupancy.attempt_order_complete!
```
jhanggi: Part of me feels like what happens inside this method is more the responsibility of the order handler rather than the occupancy. I could be convinced otherwise, though, perhaps that it then has to know too much about the innards of occupancies. But on the other hand, as it is now, occupancy might know too much about orders. Thoughts?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -34,15 +32,13 @@ def find_or_create_order
...
-      end
-
       def create_order
         ActiveRecord::Base.transaction do
           create_order_and_detail
-          order.validate_order!
-          order.purchase!
+          if occupancy.account_id?
+            order.validate_order!
+            order.purchase!
```
jhanggi: I suspect this will get weird when we get to the scanning out without an in because you can't purchase an order without an id, but you also probably won't see unpurchased orders in the problems tab. We can deal with that later, though.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -32,14 +32,19 @@ def self.recent
...
+
+    def attempt_order_complete!
+      order_detail.complete! if order_completable?
     end
 
     def orphan?
       orphaned_at?
     end
 
     def complete?
```
jhanggi: Where else does `complete?` get used? It feels a little odd for an orphaned occupancy to be considered `complete?`.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
waggles: yeah.. this is a little odd since the code was sorta shared because of the orphaned case... perhaps if I move order completion out of `Occupancy` and just have a `#complete_order`-type method on both `OrderHandler` and `OccupancyHandler`?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
waggles: I'll add a comment so it's quick to find this spot when we get there

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
waggles: yeah, and looks like not used outside, so I'll just move the logic

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/occupancy_handler_spec.rb
```diff
@@ -30,15 +30,33 @@
     end
   end
 
+  shared_examples_for "a problem order" do
+    it "sets up a problem order" do
```
jhanggi: Maybe `it "results in a problem order"`? This almost makes it feel like you're testing the test setup.

vendor/engines/secure_rooms/spec/services/secure_rooms/access_handlers/occupancy_handler_spec.rb
```diff
@@ -30,15 +30,33 @@
...
+      expect(occupancy.order_detail).to be_complete
+      expect(occupancy.order_detail).to be_problem
+    end
+  end
+
   describe "#process" do
     context "current_occupant?" do
+      let(:order) do
```
jhanggi: Stylistically, I prefer `let`s to use `{ }`, even if it results in a slightly longer line. But that's also because I'm not a stickler for line length (well, until it gets 100+).

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -47,6 +47,13 @@ def new_occupancy
         )
       end
 
+      def complete_existing_occupancy!
+        existing_occupancy.mark_orphaned!
+        if existing_occupancy.order_completable?
+          existing_occupancy.order_detail.complete!
```
jhanggi: I forget, does this update the `order_status` as well as the state? We should probably include a spec for that.

jhanggi: Also, is there a way to move this order logic into the order handler? (Well, it looks like you have similar logic already. Can we dry it up a little?)

waggles: now that it uses the `OrderHandler`, that should be covered by those specs

PR #1010, opened 2017-05-03
[#135266] Responsive cart:

noelrappin:
A couple of changes from the existing cart. These changes apply to both the responsive and non-responsive versions. 

* [Remove] is not a separate column, I’ve now put it before the name of the product (and made it bold). Putting it after the name is tricky because the name and upload are part of the same partial, but I could move it to all the different partials if we wanted.

* There’s now a thicker bold line between bundles, but not between products that are part of the same bundle.

Here's the responsive bit:

cart_-_nucore

cart_-_nucore

And the non responsive version

cart_-_nucore

### DIFF
--- a/app/assets/stylesheets/responsive.scss
+++ b/app/assets/stylesheets/responsive.scss
@@ -1,6 +1,7 @@
 /* Bootstrap overrides go here. This is included after bootstrap. */
 $navCollapseBackgroundColor: #f7f7f7;
 $navCollapseColor: #777;
+$navbarActiveBorderColor: #aaa;
 
 /* 979px - 980px is where nav-collapse gets triggered */
 @media (max-width: 979px) {
@@ -71,7 +72,7 @@ $navCollapseColor: #777;
     tr { border: 1px solid #ccc; }
 
     .first-in-bundle {
-      border-top: 3px solid black;
+      border-top: 3px solid $navbarActiveBorderColor;
     }
 
     td {
--- a/app/assets/stylesheets/responsive.scss
+++ b/app/assets/stylesheets/responsive.scss
@@ -58,6 +58,10 @@ $navCollapseColor: #777;
       display: block;
     }
 
+    td.hide-on-mobile {
+      display: none;
+    }
+
     thead tr {
       position: absolute;
       top: -9999px;
@@ -65,6 +69,10 @@ $navCollapseColor: #777;
     }
 
     tr { border: 1px solid #ccc; }
+
+    .first-in-bundle {
+      border-top: 3px solid black;
+    }
 
     td {
       border: none;
@@ -100,4 +108,3 @@ $navCollapseColor: #777;
     display: none !important;
   }
 }
-
--- a/app/views/orders/_cart_row.html.haml
+++ b/app/views/orders/_cart_row.html.haml
@@ -1,8 +1,8 @@
 %tr{class: order_detail_iteration.first? ? "first-in-bundle" : ""}
   %td
     - if order_detail_iteration.first?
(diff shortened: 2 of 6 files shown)


jhanggi:
Can we make the thick lines a shade of gray? Something around `#999`, but we might have a dark gray in our palette already.

Also, here's my proposal:
Move `Remove` back into its own column, but also pull the bundle name, if there is one, into that column.
In non-responsive, it will get back some of the padding/indentation inside of bundles we had before. 
In responsive, that would give the remove and the bundle name its own row, hopefully making it a little more obvious it's a bundle.

noelrappin:
New non responsive (I used #aaa for the border)

cart_-_nucore

New responsive

cart_-_nucore

app/assets/stylesheets/responsive.scss
```diff
@@ -66,6 +71,10 @@ $navCollapseColor: #777;
 
     tr { border: 1px solid #ccc; }
 
+    .first-in-bundle {
```
jhanggi: Do we need this? I think the definition inside of tables.scss should cover this.

app/assets/stylesheets/responsive.scss
```diff
@@ -1,6 +1,7 @@
 /* Bootstrap overrides go here. This is included after bootstrap. */
 $navCollapseBackgroundColor: #f7f7f7;
 $navCollapseColor: #777;
+$navbarActiveBorderColor: #aaa;
```
jhanggi: Do we need this? It looks like this variable is already declared in `nucore-variables.scss` (although, I forget the load order, so this might be needed).

app/views/orders/_cart_row.html.haml
```diff
@@ -1,7 +1,10 @@
-%tr
-  - if order_detail_iteration.first?
-    %td.centered{ rowspan: order_details.count }= link_to text("shared.remove"), remove_order_path(@order, order_detail), method: :put
-
+%tr{class: order_detail_iteration.first? ? "first-in-bundle" : ""}
+  %td
+    - if order_detail_iteration.first?
+      %strong
+        [#{link_to text("shared.remove"), remove_order_path(@order, order_detail), method: :put}]
+        - if order_detail.bundle
```
jhanggi: I think with this here, I think we no longer need the bundle name tacked on before the product name. I think we can remove it from `product_description.html.haml`.

app/assets/stylesheets/responsive.scss
noelrappin: It's needed because of the load order

app/views/orders/_cart_row.html.haml
noelrappin: Ok, I wanted you to see this screenshot before I did that

app/assets/stylesheets/responsive.scss
jhanggi: Ok. We'll want to be sure to double check what happens in the schools because they override this variable (they do it in `nucore-overrides.scss`). It might be worth testing out here in open by setting something in that file and make sure it applies where it's supposed to and not where it's not.

app/views/orders/_cart_row.html.haml
jhanggi: Cool. I think it's redundant right now, so we can kill it.

jhanggi:
Could we finish this up? I'd like to get this out to DC soon.

jhanggi:
Added a minor tweak getting rid of the brackets around "Remove". That's the pattern we use elsewhere on the site, so I think it makes sense to stay consistent. I'll merge as soon as CI passes.

PR #1012, opened 2017-05-03
[#136246] Fix order detail sorting for accessories in oracle

jhanggi:
In Oracle, nulls are ordered differently than they are in MySQL. This
was causing accessories to appear before their parent in the order show
view.

| Order | Oracle         | MySQL       |
|------ | ------------ |------------|
| ASC   | NULLs last  | NULLs first |
| DESC | NULLs first | NULLs last |

Reference: http://www.sqlines.com/oracle-to-mysql/null_order_by

If we ever do Postgres support we might need to revisit this.

Problem:
screen shot 2017-05-03 at 5 57 08 pm

### DIFF
--- a/app/support/accessories/accessorizer.rb
+++ b/app/support/accessories/accessorizer.rb
@@ -110,6 +110,7 @@ class Accessories::Accessorizer
     attrs = @order_detail.attributes.slice("account_id", "created_by")
     attrs.merge(order: @order_detail.order,
                 product: accessory,
+                note: options[:note],
                 quantity: options[:quantity],
                 product_accessory: product_accessory(accessory),
                 state: "new")
--- a/lib/nucore.rb
+++ b/lib/nucore.rb
@@ -106,14 +106,15 @@ module NUCore
       def self.included(base)
         base.extend ClassMethods
       end
+
       module ClassMethods
 
         def order_by_asc_nulls_first(field)
-          NUCore::Database.oracle? ? order("#{field} asc nulls first") : order(field)
+          NUCore::Database.oracle? ? order("#{sanitize_sql(field)} asc nulls first") : order(field)
         end
 
         def order_by_desc_nulls_first(field)
-          NUCore::Database.oracle? ? order("#{field} desc nulls first") : order("-#{field}")
+          NUCore::Database.oracle? ? order("#{sanitize_sql(field)} desc nulls first") : order("-#{sanitize_sql(field)}")
         end
 
       end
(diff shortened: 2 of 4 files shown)


app/support/accessories/accessorizer.rb
```diff
@@ -110,6 +110,7 @@ def detail_attributes(accessory, options)
     attrs = @order_detail.attributes.slice("account_id", "created_by")
     attrs.merge(order: @order_detail.order,
                 product: accessory,
+                note: options[:note],
```
jhanggi: I decided to allow this so that it was easier to debug the tests. I don't think there's any harm in allowing a note field on an accessory order detail.

lib/nucore.rb
```diff
@@ -108,6 +108,10 @@ def self.included(base)
       end
       module ClassMethods
 
+        def order_by_asc_nulls_first(field)
+          NUCore::Database.oracle? ? order("#{field} asc nulls first") : order(field)
+        end
+
         def order_by_desc_nulls_first(field)
           NUCore::Database.oracle? ? order("#{field} desc nulls first") : order("-#{field}")
```
jhanggi: I decided not to change this even if the default in oracle for `desc` is nulls first because I'd rather be explicit about it.

noelrappin: So nulls first in either case? So the ascending order reversed is not the descending order? (Because nulls are always first). Or am I misreading this?

jhanggi: In both these cases we want nulls first. There are sometimes we want descending, i.e. reverse chronological, but put nulls first (we're doing this in the transaction search). Then in this case, we need nulls first because we want the parent-level order details at the top, but other than that we want normal ascending order.

The default for oracle and mysql are opposite. In oracle, `desc` has the nulls first by default. We could conceivably take it out here, but I wanted to make it clear what we're choosing to do, and not relying on the default (at least in oracle's case).

lib/nucore.rb
```diff
@@ -108,6 +108,10 @@ def self.included(base)
       end
       module ClassMethods
 
+        def order_by_asc_nulls_first(field)
+          NUCore::Database.oracle? ? order("#{field} asc nulls first") : order(field)
```
meara: CodeClimate is warning about SQL injection on this line. Is there some reason the `?` format won't work here? Otherwise, can we update this line and the corresponding line below?

jhanggi:
@meara updated

PR #1015, opened 2017-05-05
[#133813] Admins should be able to update user's username

jhanggi:
The javascript on the front end will make sure the username and email match for external users.


noelrappin:
👍

PR #1018, opened 2017-05-09
[#144059001] Add Admin Occupancies Tab

waggles:
* naming-wise... did we want this to be "Rooms" instead of "Occupancies"?

### DIFF
--- a/vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
+++ b/vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
@@ -1,0 +1,22 @@
+module SecureRooms
+
+  class FacilityOccupanciesController < ApplicationController
+
+    include TabCountHelper
+
+    admin_tab     :all
+    before_action :authenticate_user!
+    before_action :check_acting_as
+    before_action :init_current_facility
+
+    def initialize
+      super
+      @active_tab = "admin_occupancies"
+    end
+
+    def index
+    end
+
+  end
+
+end
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
@@ -11,7 +11,6 @@ module SecureRooms
     def extend(user, resource)
       ability.can :manage, CardReader if user.manager_of?(resource) || user.facility_senior_staff_of?(resource)
       ability.can :index, Occupancy if user.operator_of?(resource)
-      ability.can :index, Occupancy
     end
 
   end
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
@@ -11,6 +11,7 @@ module SecureRooms
     def extend(user, resource)
       ability.can :manage, CardReader if user.manager_of?(resource) || user.facility_senior_staff_of?(resource)
       ability.can :index, Occupancy if user.operator_of?(resource)
+      ability.can :index, Occupancy
     end
 
   end
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/order_detail_extension.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/order_detail_extension.rb
@@ -6,6 +6,9 @@ module SecureRooms
 
     included do
       has_one :occupancy, dependent: :destroy, inverse_of: :order_detail, class_name: SecureRooms::Occupancy
+
+      scope :non_occupancies, -> { joins(:product).where("products.type <> 'SecureRoom'") }
+      scope :occupancies, -> { joins(:product).where("products.type = 'SecureRoom'") }
     end
 
   end
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/order_detail_extension.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/order_detail_extension.rb
@@ -7,8 +7,7 @@ module SecureRooms
     included do
       has_one :occupancy, dependent: :destroy, inverse_of: :order_detail, class_name: SecureRooms::Occupancy
 
-      scope :non_occupancies, -> { joins(:product).where("products.type <> 'SecureRoom'") }
-      scope :occupancies, -> { joins(:product).where("products.type = 'SecureRoom'") }
+      scope :occupancies, -> { joins(:product).where(products: { type: SecureRoom }) }
     end
 
   end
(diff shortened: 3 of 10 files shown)


vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
```diff
@@ -11,6 +11,7 @@ def initialize(ability)
     def extend(user, resource)
       ability.can :manage, CardReader if user.manager_of?(resource) || user.facility_senior_staff_of?(resource)
       ability.can :index, Occupancy if user.operator_of?(resource)
+      ability.can :index, Occupancy
```
jhanggi: This seems like it's giving more power than is necessary. Why doesn't the previous line handle this?

vendor/engines/secure_rooms/app/models/secure_rooms/order_detail_extension.rb
```diff
@@ -6,6 +6,9 @@ module OrderDetailExtension
 
     included do
       has_one :occupancy, dependent: :destroy, inverse_of: :order_detail, class_name: SecureRooms::Occupancy
+
+      scope :non_occupancies, -> { joins(:product).where("products.type <> 'SecureRoom'") }
```
jhanggi: Are we using this scope anywhere? I actually got rid of the analogous `non_instruments` scope in #1013.

vendor/engines/secure_rooms/app/models/secure_rooms/order_detail_extension.rb
```diff
@@ -6,6 +6,9 @@ module OrderDetailExtension
...
+      scope :occupancies, -> { joins(:product).where("products.type = 'SecureRoom'") }
```
jhanggi: Matter of taste, but I tend to lean towards `joins(:product).where(products: { type: SecureRoom })` (or the string version of SecureRoom)

vendor/engines/secure_rooms/app/presenters/secure_rooms/link_collection_extension.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    extend ActiveSupport::Concern
+
+    included do
+      insert_index = tab_methods.index(:admin_products) || -1
+      tab_methods.insert(insert_index, :admin_occupancies)
+    end
+
+    def admin_occupancies
+      if single_facility? && SecureRoom.exists?
```
jhanggi: We should restrict this to there being a secure room in the current facility. We don't need to bother all the other cores with a "Rooms" tab they'll never use.

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -0,0 +1,5 @@
+= content_for :h1 do
+  =h current_facility
```
jhanggi: I know some of the old partials have the `h` from old Rails, but we should get rid of it where we can.

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -0,0 +1,5 @@
...
+
+= content_for :tabnav do
+  = render :partial => 'secure_rooms/shared/tabnav_occupancy'
```
jhanggi: We're trying to switch our standard to use 1.9 hash syntax, double quotes, and not using `partial` and `locals` in `render` where posssible. So `render "secure_rooms/shared/tabnav_occupancy"`

vendor/engines/secure_rooms/config/locales/en.yml
```diff
@@ -15,6 +15,9 @@ en:
       selection_needed: "Must select Account"
       rules_failed: "No rule generated a verdict"
 
+  pages:
+    admin_occupancies: Occupancies
```
jhanggi: I'm leaning towards "Rooms", but it's something we can always change later after client feedback.

vendor/engines/secure_rooms/app/models/secure_rooms/ability_extension.rb
waggles: hum, it does, this line was mis-added or something will remove

vendor/engines/secure_rooms/app/presenters/secure_rooms/link_collection_extension.rb
```diff
@@ -0,0 +1,23 @@
+module SecureRooms
+
+  module LinkCollectionExtension
+
+    extend ActiveSupport::Concern
+
+    included do
+      insert_index = tab_methods.index(:admin_products) || -1
```
jhanggi: Can we move this one more to the left so it's between "Reservations" and "Billing". Conceptually it fits closer with Orders and Reservations, so I think it should live there.

vendor/engines/secure_rooms/app/presenters/secure_rooms/link_collection_extension.rb
```diff
@@ -0,0 +1,23 @@
...
+      tab_methods.insert(insert_index, :admin_occupancies)
+    end
+
+    def admin_occupancies
+      if single_facility? && SecureRoom.for_facility(facility).exists?
```
jhanggi: Can we also add a `not_archived` scope on here? I think if you deactivate a room, we should let this go away. I wouldn't use `Product.active` because it also excludes hidden, but not-archived instruments. There's definitely some awkwardness to the names there.

waggles: for some reason we wrote the story to show the tab "archived or not" but I don't remember why
https://www.pivotaltracker.com/story/show/144059001

do we need to update story?

jhanggi: Hmm, let's leave it like this for now and we can always change it later. I think it'll be useful to see what ends up in the subtabs.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
```diff
@@ -0,0 +1,22 @@
+module SecureRooms
+
+  class FacilityOccupanciesController < ApplicationController
+
+    include TabCountHelper
```
jhanggi: Are we using this helper anywhere yet? If you know for sure already we're going to be using it, I'm fine leaving it in here.

waggles: ah, not in this PR, that bit in tabnav_occupancy is where it will be needed immediately following

PR #1019, opened 2017-05-09
Foreign keys and generating ERD diagrams

noelrappin:
This PR 

* Adds the rails-erd gem which allows us to create database relationship diagrams

* Adds a bunch of foreign keys and indexes

* Fixes some tests that break with foreign keys

My research of all the things that look like foreign keys but don’t have foreign keys defined is at:

https://docs.google.com/a/tablexi.com/spreadsheets/d/1fbVPc9qr1iqirQLvJNTSflpyOCb-Fuitr-4FetE3ZEU/edit?usp=sharing

In creating foreign keys, I did not use all the possible ones, I tried to pick the ones that were in join tables, or involved users, or seemed to be significant.

I attempted to run the migration against recent production of the UIC and Dartmouth databases, the migration file is commented with the lines that cause problems in those databases.


app/models/statement_row.rb
```diff
@@ -6,7 +6,15 @@ class StatementRow < ActiveRecord::Base
...
-  after_destroy { @parent_statement.destroy if @parent_statement.statement_rows.reload.empty? }
+  
+  after_destroy do
+    if @parent_statement.statement_rows.reload.empty?
+      @parent_statement.order_details.each do |order_detail|
+        order_detail.update_attributes(statement: nil)
+      end
+      @parent_statement.destroy
+    end
+  end
```
noelrappin: This is necessary because without it deleting the parent statement causes a foreign key violation in the order detail. I think it's a timing issue, but this seemed to be the minimal way to mitigate.

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
...
+  def change
+    add_index :notifications, [:subject_id, :subject_type]
+    add_index :order_details, :project_id
+    add_index :order_statuses, :parent_id
+    add_index :orders, :merge_with_order_id
+    add_index :price_group_members, :account_id
+    add_index :products, :initial_order_status_id
+
+    # The dartmouth database violates this one
+    add_foreign_key :account_users, :users
```
jhanggi: I fixed up Dartmouth's production here. Anything with user_id `114,115,337,168` you can safely delete. That'll probably be quicker than me giving you a fresh snapshot.

noelrappin:
Looks like CI found some different test failures, too.

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
...
+    add_foreign_key :price_group_members, :accounts
+
+    # This is in a migration, the Dartmouth database has it, the UIC
+    # one does not
+    #add_foreign_key :price_group_members, :price_groups
+
+    add_foreign_key :price_group_members, :users
+
+    # The Dartmouth database fails this one
+    add_foreign_key :product_users, :users
```
jhanggi: I cleaned up dartmouth.

Both this and account_users are fine on NU.

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    add_foreign_key :order_details, :users, column: :assigned_user_id
+    add_foreign_key :order_details, :journals
+    add_foreign_key :order_details, :order_statuses
+    add_foreign_key :order_details, :statements
+    add_foreign_key :orders, :order_imports
+    add_foreign_key :orders, :orders, column: :merge_with_order_id
+
+    #The UIC and Dartmouth databases violate this
+    add_foreign_key :orders, :users
```
jhanggi: DC#677.

jhanggi: NU is fine.

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
...
+    add_foreign_key :statements, :accounts
+
+    # The UIC db has one row that violates this
+    add_foreign_key :user_roles, :facilities
+
+    add_foreign_key :user_roles, :users
+
+    # this is already null: false in the dartmouth database
+    # it also fails a lot of tests 
+    # change_column :journal_rows, :order_detail_id, :integer, null: false
```
jhanggi: So this is `null: true` in open? Where does it break things?

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
...
+    add_foreign_key :orders, :orders, column: :merge_with_order_id
+
+    #The UIC and Dartmouth databases violate this
+    add_foreign_key :orders, :users
+
+    add_foreign_key :price_group_members, :accounts
+
+    # This is in a migration, the Dartmouth database has it, the UIC
+    # one does not
+    #add_foreign_key :price_group_members, :price_groups
```
jhanggi: I see this foreign key when I look at the different schemas as well as my snapshots of UIC, open, and NU, it's just a different name in some of them `sys_c008583`. Maybe we can can drop the constraint and re-add it got get it consistent.

db/migrate/20170508150135_add_foreign_keys.rb
noelrappin: It very much doesn't like Journal#create_journal_rows, which calls JournalRowBuilder, where I think the `add_journal_rows_from_product_recharges` which calls `product_to_journal_row` creates a journal row without an `order_detail_id`.

noelrappin: sorry, it's not `null: false` in Dartmouth, this constraint should just be deleted

jhanggi: Can we add the foreign keys from `journal_rows` to `journals` and `order_details` (and leave `order_detail_id` nullable (I believe that's as it should be because there can be journal rows that don't have order_details)?

app/models/statement_row.rb
jhanggi: I'm note quite sure if we care about callbacks/validations, but we could possibly get away with `@parent_statement.order_details.update_all(statement: nil)`. If we care about `updated_at`, we could include that in there, too.

Another option I don't know if it would work is to add `dependent: ::nullify` on the Statement `has_many :order_details`.

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
...
+    add_foreign_key :product_users, :users
+    add_foreign_key :statement_rows, :order_details
+
+    #The UIC database violates this
+    add_foreign_key :statement_rows, :statements
+
+    add_foreign_key :statements, :accounts
+
+    # The UIC db has one row that violates this
+    add_foreign_key :user_roles, :facilities
```
jhanggi: Looks like that was Chris from way back in the day. I'll clean that up.

db/schema.rb
```diff
@@ -713,6 +715,7 @@
     t.string  "role",        limit: 255, null: false
   end
 
+  add_index "user_roles", ["facility_id"], name: "fk_rails_dca27403dd", using: :btree
```
jhanggi: I think I mentioned this before, but Oracle is kind of annoying and won't automatically generate indexes for foreign keys, so we need to manually add them as part of the migration.

jhanggi:
I think this is just about ready to go. Just needs some comment cleanup and maybe some explicit indexes on foreign keys (because silly, oracle).

jhanggi:
And looks like we now have a merge conflict on schema.

jhanggi:
```

**Comment by jhanggi** [comment]
On CI:
```
You have 1 pending migration:
  20170508150135 AddForeignKeys
Run `rake db:migrate` to update your database then try again.
```

File: db/migrate/20170508150135_add_foreign_keys.rb [hunk 9]
```diff
@@ -0,0 +1,49 @@
...
+    add_foreign_key :journal_rows, :journals
+    add_foreign_key :journal_rows, :order_details
+    add_foreign_key :order_details, :users, column: :assigned_user_id
+    add_foreign_key :order_details, :journals
+    add_foreign_key :order_details, :order_statuses
+    add_foreign_key :order_details, :statements
+    add_foreign_key :orders, :order_imports
+    add_foreign_key :orders, :orders, column: :merge_with_order_id
+
+    #The UIC and Dartmouth databases violate this
```
jhanggi: Can we remove these comments?

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,49 @@
+class AddForeignKeys < ActiveRecord::Migration
+  def change
+    add_index :notifications, [:subject_id, :subject_type]
+    add_index :order_details, :project_id
+    add_index :order_statuses, :parent_id
+    add_index :orders, :merge_with_order_id
+    add_index :price_group_members, :account_id
+    add_index :products, :initial_order_status_id
```
jhanggi: Am I correct in assuming that any other foreign keys we're adding already had indexes?

noelrappin: Yes

noelrappin:
@jhanggi This should be ready to go again

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,31 @@
+class AddForeignKeys < ActiveRecord::Migration
+  def change
+    add_index :notifications, [:subject_id, :subject_type]
+    add_index :order_details, :project_id
```
jhanggi: Adding this index should live in the `projects` engine.

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,31 @@
...
+    add_index :order_statuses, :parent_id
+    add_index :orders, :merge_with_order_id
```
jhanggi: This index already exists in NU. There were a bunch of indexes that we didn't have in oracle because of it's weirdness around foreign keys, so we added them. Seems like we might have added too many.

https://github.com/tablexi/nucore-nu/commit/e3a67c1a4ee478473bee3a5ccd9182c5ffbe8e35

There's also `:price_group_members, :account_id` already.

jhanggi: Is there a way we can make the `add_index` conditional on the index not already being there? Or maybe have it `rescue` and move on?

db/migrate/20170508150135_add_foreign_keys.rb
```diff
@@ -0,0 +1,31 @@
...
+    add_index :price_group_members, :account_id
+    add_index :products, :initial_order_status_id
+
+    add_foreign_key :account_users, :users
+    add_foreign_key :journal_rows, :accounts
+    add_foreign_key :journal_rows, :journals
+    add_foreign_key :journal_rows, :order_details
```
jhanggi: This one had a few problems in NU. I believe this should do the trick. Sound reasonable? They're order details that no longer exist, they're all very old.

```
update journal_rows
set order_detail_id = null
where id in(
select journal_rows.id
from journal_rows
left join order_details
on order_details.id = journal_rows.order_detail_id
where order_details.id is null
and order_detail_id is not null);
```

jhanggi:
That's it. Everything else ran just fine.

db/migrate/20170508150135_add_foreign_keys.rb
noelrappin: It does look like `index_exists?` is a thing, I'll try that

db/migrate/20170508150135_add_foreign_keys.rb
noelrappin: I _think_ mysql complains that you can't specify journal rows as the update table and the from table, don't know about Oracle. Might have to do it in a two-line script.

jhanggi: I'm happy to run it manually on NU. I ran it locally and everything appeared fine to me; I just want to make sure the script looks right to you.

noelrappin: seems right

noelrappin:
Okay, minor tweaks

jhanggi:
Tested against NU, and it looks good.

jhanggi:
@noelrappin Looks like another schema conflict. Let's get that cleaned up and then merge this in.

noelrappin:
@jhanggi Cleaned up, can I merge?

db/migrate/20170508150135_add_foreign_keys.rb
jhanggi: Fun times. This migration is failing on UIC. And yes, you were right about mysql complaining. What did you end up doing to get your migration working?

jhanggi: For reference, this worked.
```
update journal_rows
set order_detail_id = null
where id in(
select distinct journal_rows.id
from (SELECT * FROM journal_rows) AS jrs
left join order_details
on order_details.id = jrs.order_detail_id
where order_details.id is null
and order_detail_id is not null);
```

PR #1020, opened 2017-05-10
[#144059055] Occupancies in progress

waggles:
* There's a buncha copypasta here from facility_reservations
* Also it includes the "problem" tab but without switching over to a new partial on it yet
* Note: writing specs with the right fixture composition when orders are involved is such a pain


vendor/engines/secure_rooms/spec/features/view_facility_occupancies_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+
+      before do
+        SecureRooms::AccessHandlers::OrderHandler.process(problem_occupancy)
+      end
+
+      it "shows the order details" do
+        visit show_problems_facility_occupancies_path(facility)
+
+        expect(current_path).to eq(show_problems_facility_occupancies_path(facility))
+        # TODO: Update specs after page is updated
```
waggles: so I assume we can add the entry/exit by extending the `ProblemOrderDetailsHelper` and  renaming the related `show_reservation_start_at`?

app/models/facility.rb
```diff
@@ -129,12 +129,12 @@ def problem_reservation_order_details
     complete_problem_order_details.reservations
   end
 
-  private
```
waggles: not sure why this was private before... any ideas?

jhanggi: Looks like it was only being used internally by `problem_non_reservation_order_details` and `problem_reservation_order_details` so there was no reason to make it public.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
```diff
@@ -2,19 +2,73 @@ module SecureRooms
...
     before_action :authenticate_user!
     before_action :check_acting_as
     before_action :init_current_facility
+    before_action :sanitize_sort_params, only: :index
+
+    load_and_authorize_resource class: Occupancy
+
+    SORTING_CLAUSES = {
+      "entry_at" => "secure_rooms_occupancies.entry_at",
+      "user_name" => "users.last_name",
```
jhanggi: Should this be `user.last_name, users.first_name` so it sorts when people have the same last name. Might need to do something like make this an array so you can apply the direction to each one to come out with `users.last_name desc, users.first_name desc`.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
```diff
@@ -2,19 +2,73 @@ module SecureRooms
...
+          :product,
+          :account,
+          :occupancy,
+        )
+    end
+
+    def problem_order_details
+      current_facility
+        .complete_problem_order_details
+        .joins(:occupancy)
```
jhanggi: Does this join do what we need to filter to only those with occupancies?

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
```diff
@@ -2,19 +2,73 @@ module SecureRooms
...
+        .merge(SecureRooms::Occupancy.order(entry_at: :desc))
+    end
+
+    def sanitize_sort_params
+      sort_clause = SORTING_CLAUSES[sort_column]
+      @order_by_clause = [sort_clause, sort_direction].join(" ")
+    end
+
+    def sort_column
```
jhanggi: Does this need to be its own separate method? Could we use `params.fetch(:sort, "entry_at")` to have a default?

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
```diff
@@ -2,19 +2,73 @@ module SecureRooms
...
+      params[:sort] || "entry_at"
+    end
+
+    def sort_direction
+      (params[:dir] || "") =~ /asc/i ? "asc" : "desc"
```
jhanggi: `params[:dir].to_s`?

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -3,3 +3,37 @@
...
+          %th= sortable 'entry_at', SecureRooms::Occupancy.human_attribute_name(:entry_at)
+          %th= sortable 'user_name', User.model_name.human
+          %th.order-note= sortable 'product_name', SecureRoom.model_name.human
+          %th= sortable 'payment_source', Account.model_name.human
+      %tbody
+        - @order_details.each do |od|
+          - order = od.order
+          - occ = od.occupancy
+          %tr
+            %td.centered= check_box_tag("order_detail_ids[]", od.id, false, {:class => 'toggle', :id => nil })
```
jhanggi: 1.9 hash syntax, please.

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -3,3 +3,37 @@
...
+            %td.centered= link_to od.order_id, facility_order_path(current_facility, order)
+            %td.centered= link_to od.id, manage_order_detail_path(od), :class => 'manage-order-detail'
+            %td= human_datetime(occ.entry_at)
+            %td=h od.user.nil? ? '' : od.user.full_name
```
jhanggi: Maybe just `od.user.try(:full_name)`. Not quite sure why there wouldn't be a user, though. Also, get rid of the `h`.

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
waggles: is there a cop for this one? I keep missing them and that would help me spot

jhanggi: We have haml-lint set up which should catch this, codeclimate doesn't have an engine for it, so you need to remember to run it manually.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
waggles: this and `sort_column` are required for us to use the `sortable` helper in the views, but I feel like this should just be a default in the helper so I might move it

and then also yeah I'll update ofc

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
waggles: I'll double check what it would take to factor out the need for this to be defined

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
waggles: yeah, and apparently just `.occupancies` wasn't enough due to sorting

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
waggles: oh actually I was gonna ask... would `username` be sufficient here? not sure when we care about `username` vs `email` vs `full_name` for ID'ing a user

neither first/last name currently have indexes so I am not sure if this indicates they typically aren't expected to be sorted, or just that the indexes are missing

app/models/facility.rb
waggles: ok cool makes sense

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
jhanggi: Let's stick with what the other pages are doing. I think there are usually few enough orders on these views that the indexing doesn't really matter.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
waggles: okay I don't want to put stuff in `ApplicationController` but I do still want to use `sortable` in views (it handles directions) so I think I just prefer to leave these as methods... what do you think?

waggles:
updated

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -3,3 +3,37 @@
...
+            %td.centered= link_to od.id, manage_order_detail_path(od), class: "manage-order-detail"
+            %td= human_datetime(occ.entry_at)
+            %td= od.user.nil? ? "" : od.user.try(:full_name)
+            = render partial: "shared/order_detail_cell", locals: { od: od, show_occupancy: false }
+            %td= od.account
+
+      = render "table_controls"
+
+  = will_paginate(@order_details)
+  = render "/price_display_footnote", locals: { admin: true }
```
jhanggi: I don't think we need the `locals` key here.

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/show_problems.html.haml
```diff
@@ -0,0 +1,13 @@
...
+
+= content_for :tabnav do
+  = render partial: "secure_rooms/shared/tabnav_occupancy",
+    locals: { secondary_tab: "problem" }
+
+- if @order_details.empty?
+  %p.notice= t(".no_problem_occupancies")
+- else
+  %p.notice= t(".instructions")
+  = render partial: "shared/problem_order_details",
```
jhanggi: Get rid of keys here. And can we flip the conditional?

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/show_problems.html.haml
```diff
@@ -0,0 +1,13 @@
...
+    locals: { assign_price_policies_path: assign_price_policies_to_problem_orders_facility_occupancies_path }
```
jhanggi: I'm curious if we can do something like use `[:assign_price_policies_to_problem_orders, current_facility, Occupancy]` for the path and use something else to fill in the last item so we can get away without needing to pass anything in to `shared/problem_order_details`. Not a big deal, though.

vendor/engines/secure_rooms/app/views/secure_rooms/shared/_tabnav_occupancy.html.haml
```diff
@@ -1,2 +1,4 @@
 %ul.nav.nav-tabs
-  = # TODO: Add tabs
+  = tab "In Progress", facility_occupancies_path
```
jhanggi: I18n.

vendor/engines/secure_rooms/app/views/secure_rooms/shared/_tabnav_occupancy.html.haml
```diff
@@ -1,2 +1,4 @@
...
+  - if current_ability.can?(:show_problems, SecureRooms::Occupancy)
+    = display_tab "Problem Occupancies", show_problems_facility_occupancies_path, action: :show_problems
```
jhanggi: I18n.

vendor/engines/secure_rooms/config/routes.rb
```diff
@@ -15,7 +15,13 @@
       patch :"card_number/update", to: "secure_rooms/card_numbers#update"
     end
 
-    resources :occupancies, controller: "secure_rooms/facility_occupancies", only: :index
+    resources :occupancies, controller: "secure_rooms/facility_occupancies", only: :index do
+      collection do
+        post "assign_price_policies_to_problem_orders"
```
jhanggi: I'm still interested in DRYing this up, but I won't consider a show stopper for now, especially if there are going to be more routes coming up.

vendor/engines/secure_rooms/spec/features/view_facility_occupancies_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+      expect(page).to have_content('There are no "In Process" occupancies.')
+    end
+  end
+
+  context "with occupancy" do
+    let(:account) { create :account, :with_account_owner, owner: facility_staff }
+
+    before do
+      # Allow purchase of order
+      allow_any_instance_of(Order).to receive(:cart_valid?).and_return(true)
```
jhanggi: My new PR #1024 has some new factories that should allow for avoiding these kinds of stubbing. We can clean them up later, though. Especially in feature specs, I shy away from these kind of stubs.

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/show_problems.html.haml
waggles: I looked into this; I don't know of a way to get to `Occupancy`/`Reservation` for room/instrument but then to `Order` for everything else without adding logic--is there one?

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -3,3 +3,37 @@
...
+- else
+  %p.notice There are no "In Process" occupancies.
```
jhanggi: Can we I18n this?

jhanggi:
Found a couple issues with this. Let's discuss them when we meet in a few minutes.

waggles:
@jhanggi I think I addressed what we went through, lmk if I missed anything

vendor/engines/secure_rooms/spec/features/view_facility_occupancies_spec.rb
jhanggi: This is merged in now.

jhanggi:
Still getting this:

```
Showing /Users/jhanggi/Documents/workspace/nucore-open/app/views/shared/_problem_order_details.html.haml where line #32 raised:

translation missing: en.order_details.notices.missing_exit.badge
```

PR #1021, opened 2017-05-10
[#144058895] Remove translation of problem description

waggles:
I did a grep for other calls to `order_detail_notices` and it looks like this is the only other spot

This issue was introduced through in #1007

### DIFF
--- a/app/helpers/facility_orders_helper.rb
+++ b/app/helpers/facility_orders_helper.rb
@@ -9,10 +9,7 @@ module FacilityOrdersHelper
     notices << "can_reconcile" if order_detail.can_reconcile_journaled?
     notices << "in_open_journal" if order_detail.in_open_journal?
 
-    # TODO: Share translation key between badges and notices
-    warnings = []
-    warnings << "missing_actuals" if order_detail.complete? && order_detail.reservation.try(:requires_but_missing_actuals?)
-    warnings << "missing_price_policy" if order_detail.missing_price_policy? && !order_detail.reservation.try(:requires_but_missing_actuals?)
+    warnings = Array(order_detail.problem_description_key)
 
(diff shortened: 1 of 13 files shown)


jhanggi:
The old messages were longer than the badge version. We need to maintain the longer messages in this view.

"This order's reservation does not have an actual time. Please ensure that actual times are set and there is a price policy for the date this order was fulfilled."
"This order does not have a price policy assigned. Please ensure that there is a price policy for the date this order was fulfilled. Saving will attempt to assign a price policy."

waggles:
hmm interesting, do we need to do something like:
```
def problem_key
  :actual_usage_missing if requires_but_missing_actuals?
end
def problem_description_badge
  text(problem_description_key) if problem_description_key
end
```
in the models and then translate the longer text in the controllers?

jhanggi:
I'm almost leaning now towards reverting back and only having the `problem_key` on the model and doing the rest of the translations in the views/controllers/helpers. Since we're showing some sort of message, especially now that we have multiple, it seems more like view/presenter logic than model logic.

What do you think?

waggles:
Yeah, that's a good point--it should just be the keys until it's in a presentation layer

waggles:
note: above rolls back the broken part, so in a pinch could be releasable... but further refactor is coming next

waggles:
updated, I believe everything is now displaying while still using the `problem_description_key` to detect problems in the order_detail

waggles:
This issue was introduced through [#144058895] in #1007

jhanggi:
Looks like `_problem_order_details.html.haml` didn't get switched back.

waggles:
I also pulled out the translation scopes, I don't think they are used anymore

PR #1023, opened 2017-05-10
[#144059203] Edit Occupancy Times in Modal

jhanggi:
This adds support for modifying an occupancies entry and exit time. I used what was there for JournalCutoffDates as my jumping off point (after a few bad initial attempts) since that was my favorite method we've come up with so far for handling date inputs from forms, so that's why there are changes in that section (I tried to isolate that to the first commit)

Will come in future PRs:
* Implement support for duration
* Better handling against problem occupancies
* Refactor Reservations to use this (this might not happen just yet because it's a significant chunk of work, but I _really_ want to kill off Reservations::DateSupport)

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
@@ -2,7 +2,6 @@ module SecureRooms
 
   class Occupancy < ActiveRecord::Base
 
-    include TextHelpers::Translation
     include DateTimeInput::Model
 
     date_time_inputable :entry_at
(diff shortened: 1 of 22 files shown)


jhanggi:
This is now ready for a full look. A few things as mentioned in the description will come in later PRs.

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -2,6 +2,12 @@ module SecureRooms
 
   class Occupancy < ActiveRecord::Base
 
+    include TextHelpers::Translation
```
waggles: do we use this?

jhanggi: Probably not. Might have been a merge thing from when we were using it for problem descriptions. I'll remove it.

PR #1026, opened 2017-05-14
[#145319377] Avoid processing denied scans

waggles:
Adds guards against processing through the chain unless the event was "successful"

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/event.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/event.rb
@@ -11,6 +11,10 @@ module SecureRooms
 
     delegate :direction, :secure_room, :ingress?, :egress?, to: :card_reader
 
+    def successful?
+      outcome == "grant"
+    end
+
   end
 
 end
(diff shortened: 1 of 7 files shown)


vendor/engines/secure_rooms/app/models/secure_rooms/event.rb
```diff
@@ -11,6 +11,10 @@ class Event < ActiveRecord::Base
 
     delegate :direction, :secure_room, :ingress?, :egress?, to: :card_reader
 
+    def successful?
```
jhanggi: Would it be worth pulling something into the `AccessManager` that breaks out of the loop if a result is not successful? And have all the handlers implement a success predicate? That might be overkill, though.

And just from a nitpicky perspective, I think `success?` is a little more idiomatic.

waggles: hmm I do like that better and it should be a quick update

waggles:
I ended up going with a compromise--since the handler instances aren't kept around, that would have to change to ask them `#success?` in the `AccessManager`. I left in the guard clause for `event.success?` but also added in skipping of handlers if `nil` is returned.

PR #1027, opened 2017-05-15
[#144059233] Duration as quantity

waggles:
does the use of duration for quantity appear elsewhere (at least, within the scope of this story)?


jhanggi:
I think we're good here. If we find any more spots we can drop in a story for them.

PR #1028, opened 2017-05-15
[#145319401] Return 404s in JSON

waggles:
* Adds a `rescue_from` to avoid 404'ing in html
* doesn't yet address whether or not sending an error message would be helpful - need to chat with @danhodos about it

### DIFF
--- a/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
+++ b/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
@@ -2,6 +2,8 @@ module SecureRoomsApi
 
   class ScansController < SecureRoomsApi::ApiController
 
+    rescue_from ActiveRecord::RecordNotFound, with: :render_not_found
+
     before_action :load_user_and_reader
 
     def scan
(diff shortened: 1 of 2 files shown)


danhodos:
@waggles - for a 404 we'll [report to Rollbar right now with the details](https://github.com/tablexi/nucore-rooms/blob/develop/apps/doors/lib/doors/nu_core.ex#L45-L63) of whatever you send in the response. So if you want to add the `Exception#message`, that might be pretty good (assuming it'll be something like "Couldn't find User with 'indala_number'=12345" or something).

waggles:
updated with message!

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -26,6 +28,10 @@ def load_user_and_reader
       )
     end
 
+    def render_not_found(exception)
+      render json: { error: exception.message }, status: 404
```
jhanggi: What's an example of the message that comes out of this?

waggles: `"Couldn't find SecureRooms::CardReader"`

waggles: yeah looks like it's not readily available and we'd have to dig out the right id since we're finding these by different-named attributes like `#card_number`

PR #1029, opened 2017-05-15
Add a warning when the Sanger sequencing is printed

noelrappin:
Per https://www.tjvantoll.com/2012/06/15/detecting-print-requests-with-javascript/ adds a warning to the Sanger order form page, requesting that users continue their submission if they print.

If the user does print, they get an alert warning them to continue, and the message on the page changes.

### DIFF
--- a/vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/print_warning.js.coffee
+++ b/vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/print_warning.js.coffee
@@ -17,8 +17,12 @@ class @PrintWarning
   mediaQuery: -> window.matchMedia("print")
 
   onPrintAction: ->
-    $(".js--print-warning").text("Even if you have printed the order, this order is not submitted. Please click Save Submission and move to the next step.")
-    alert "This order is not submitted. Please click Save Submission and move to the next step before printing your sample sheet."
+    $(".js--print-warning").text(
+      "Even if you have printed the order, this order is not submitted.
+      Please click Save Submission and move to the next step.")
+    alert("This order is not submitted.
+      Please click Save Submission and move to the next step before
+      printing your sample sheet.")
 
 $ ->
   new PrintWarning().initListener()
--- a/vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/print_warning.js.coffee
+++ b/vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/print_warning.js.coffee
@@ -17,9 +17,6 @@ class @PrintWarning
   mediaQuery: -> window.matchMedia("print")
 
   onPrintAction: ->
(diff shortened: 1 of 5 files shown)


vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/print_warning.js.coffee
```diff
@@ -0,0 +1,24 @@
...
+    @mediaQuery().addListener (query) =>
+      return unless query.matches
+      @onPrintAction()
+
+  mediaQuery: -> window.matchMedia("print")
+
+  onPrintAction: ->
+    $(".js--print-warning").text("Even if you have printed the order, this order is not submitted. Please click Save Submission and move to the next step.")
+    alert "This order is not submitted. Please click Save Submission and move to the next step before printing your sample sheet."
+
```
noelrappin: Is there a way we localize JS?

vendor/engines/sanger_sequencing/spec/javascripts/sanger_sequencing/print_warning_spec.js.coffee
```diff
@@ -0,0 +1,19 @@
...
+    warning.initListener()
+    expect(warning.listen).not.toHaveBeenCalled()
+
+  it "warns when the right elements are in place", ->
+    warning = new PrintWarning
+    spyOn(warning, "listen")
+    setFixtures($("<h2>").addClass("js--print-warning"))
+    expect(warning.shouldWarn()).toBeTruthy()
+    warning.initListener()
+    expect(warning.listen).toHaveBeenCalled()
```
noelrappin: Couldn't find an easy way to test triggering the listener, but there's no logic there that isn't already covered by the browser code, presumably.

vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/print_warning.js.coffee
```diff
@@ -0,0 +1,28 @@
...
+  mediaQuery: -> window.matchMedia("print")
+
+  onPrintAction: ->
+    $(".js--print-warning").text(
+      "Even if you have printed the order, this order is not submitted.
+      Please click Save Submission and move to the next step.")
+    alert("This order is not submitted.
+      Please click Save Submission and move to the next step before
+      printing your sample sheet.")
```
noelrappin: Not sure if or how we localize JS?

jhanggi:
It's only a little extra complexity, but I'm not sure changing the message on the page adds all that much. I would vote for just leaving that message the same and keeping the code simpler.

I like the rest of it, though.

We don't usually do too much with localizing JS. Where possible, we'll sometimes put the text in a data attribute or somewhere in the HTML so the JS doesn't need to have the strings.

noelrappin:
No problem to take it out, it was a suggestion in the standup or IPM, but it's just one line of code.

PR #1031, opened 2017-05-15
[#136734] Drop reservation canceled_* columns

meara:
Depends on #1008, DO NOT MERGE until that has been deployed/verified on stage.

Will need to be rebased on current master after #1008 is merged in, which should resolve the current test failures.


meara:
@jhanggi are we ready to merge this in before the next stage release, or do we want to release bounceback fixes etc first?

jhanggi:
We've created a new story to track this, so I added the ticket number. We can do a rebase once we're ready to merge this in, which will be after the next production release.

PR #1032, opened 2017-05-16
[#145286873] Order detail modal has duration for occupancies

jhanggi:
Behavior should be identical to placing a new reservation:

* If you adjust the start time, duration stays the same, but end time updates to be start time plus duration
* If you adjust the duration, the end time adjusts accordingly
* If you adjust the end time, duration adjusts accordingly
* If you try to put the end time before the start time, then the end time reverts to its previous value

Additional notes:

* Setting the end time before the start time should not trigger a price update


vendor/engines/secure_rooms/app/views/order_management/order_details/secure_rooms/_occupancy.html.haml
```diff
@@ -1,4 +1,26 @@
-.datetime-block.js--pricingUpdate
+.datetime-block.js--timeFieldAdjustor
   = f.simple_fields_for :occupancy_attributes, f.object.time_data do |of|
     = of.input :entry_at, as: :time_dropdown
+
+    = of.input :actual_duration_mins, input_html: { class: "timeinput" }
+
     = of.input :exit_at, as: :time_dropdown
+
+:coffeescript
```
waggles: does this have to be inline in the haml?

jhanggi: If we weren't in an engine, this would be a little easier to extract out of being inline. If we end up ever adding more JS to the engine, then it might be worth getting it in via a view hook, but that just didn't seem worth it for this, especially as this is a configuration tying directly to the form elements that get rendered in the same partial.

app/views/order_management/order_details/_warnings.html.haml
```diff
@@ -2,7 +2,8 @@
 - if notices[:warnings].any?
   .alert.alert-error
     - notices[:warnings].each do |warning|
-      = t("order_details.notices.#{warning}.alert")
+      -# TODO: Should be getting fixed in another PR
```
meara: Is there a reason it's not being fixed in this PR?

spec/javascripts/app/reservation_time_field_adjustor_spec.coffee
```diff
@@ -0,0 +1,105 @@
+#= require jquery
+
+describe "ReservationTimeFieldAdjustor", ->
```
meara: Thanks for writing all these specs. 👍

app/views/order_management/order_details/_warnings.html.haml
jhanggi: That other PR already existed. Now with a merge/rebase off master, this isn't needed anymore, so I'm removing it.

PR #1039, opened 2017-05-17

waggles:
Notes:
* `AutoExpireReservation` uses `Time.zone.now - 12.hours` for the window, but I feel like I remember saying we might just run this job overnight since rooms would typically be closed? If not and we do want the 12 hour window, should that be moved into a real constant and/or setting then?
* I'm not sure how whenever/`schedule.rb` work within the engine, and figured it's easier to chat about it if we're already familiar with that.

### DIFF
--- a/app/controllers/admin/services_controller.rb
+++ b/app/controllers/admin/services_controller.rb
@@ -9,6 +9,16 @@ module Admin
       render nothing: true
     end
 
+    def process_five_minute_tasks
+      workers = [AutoExpireReservation, EndReservationOnly, AutoLogout]
+
+      workers.each do |worker|
+        worker.new.perform
+      end
+
+      render nothing: true
+    end
+
   end
 
 end
--- a/app/controllers/admin/services_controller.rb
+++ b/app/controllers/admin/services_controller.rb
@@ -2,7 +2,7 @@ module Admin
 
   class ServicesController < ApplicationController
 
-    skip_before_action :verify_authenticity_token, only: :cancel_reservations_for_offline_instruments
+    skip_before_action :verify_authenticity_token
 
     def cancel_reservations_for_offline_instruments
       InstrumentOfflineReservationCanceler.new.cancel!
@@ -10,13 +10,14 @@ module Admin
     end
 
     def process_five_minute_tasks
-      workers = [AutoExpireReservation, EndReservationOnly, AutoLogout]
-
-      workers.each do |worker|
+      self.class.five_minute_tasks.each do |worker|
         worker.new.perform
       end
-
       render nothing: true
+    end
+
+    def self.five_minute_tasks
+      @five_minute_tasks ||= [AutoExpireReservation, EndReservationOnly, AutoLogout]
     end
 
   end
--- a/vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
+++ b/vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
@@ -13,7 +13,13 @@ module SecureRooms
     private
 
     def occupancy_order_details
-      OrderDetail.joins(:occupancy).merge(SecureRooms::Occupancy.current)
+      OrderDetail.joins(:occupancy).merge(long_running_occupancies)
+    end
+
+    def long_running_occupancies
+      SecureRooms::Occupancy.current.where(
+        SecureRooms::Occupancy.arel_table[:entry_at].lt(Time.current - 12.hours),
+      )
     end
 
     def orphan_occupancy(od)
--- a/vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
+++ b/vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
@@ -17,9 +17,7 @@ module SecureRooms
     end
 
     def long_running_occupancies
-      SecureRooms::Occupancy.current.where(
-        SecureRooms::Occupancy.arel_table[:entry_at].lt(Time.current - 12.hours),
-      )
+      SecureRooms::Occupancy.current.where("entry_at < ?", 12.hours.ago)
     end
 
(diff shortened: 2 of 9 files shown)


jhanggi:
Let's stick with 12 hours for now. I think it'll make sense being congruent with instruments. Changing that should be pretty easy later. If auto-logout time was a field on Product, I'd say use that, but turns out it's a field on Relay.

I need to think a bit on how to hook this in with cron.

waggles:
@jhanggi updated, uses the `Admin::ServicesController` to trigger all 5-min interval tasks

vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    private
+
+    def occupancy_order_details
+      OrderDetail.joins(:occupancy).merge(long_running_occupancies)
+    end
+
+    def long_running_occupancies
+      SecureRooms::Occupancy.current.where(
+        SecureRooms::Occupancy.arel_table[:entry_at].lt(Time.current - 12.hours),
```
jhanggi: It's not a big deal, but I've started leaning against `arel` for these kinds of things. Abstractly, I definitely prefer it, but `where("entry_at < ?", 12.hours.ago)` is so much easier to read. I could go either way, though.

app/controllers/admin/services_controller.rb
```diff
@@ -2,13 +2,24 @@ module Admin
...
 
-    skip_before_action :verify_authenticity_token, only: :cancel_reservations_for_offline_instruments
+    skip_before_action :verify_authenticity_token
 
     def cancel_reservations_for_offline_instruments
       InstrumentOfflineReservationCanceler.new.cancel!
       render nothing: true
     end
 
+    def process_five_minute_tasks
```
jhanggi: I'm not in love with this name, especially as the timing is controlled outside of this controller. Not sure I have a better idea, though.

app/controllers/admin/services_controller.rb
```diff
@@ -2,13 +2,24 @@ module Admin
...
+      self.class.five_minute_tasks.each do |worker|
+        worker.new.perform
+      end
+      render nothing: true
+    end
+
+    def self.five_minute_tasks
+      @five_minute_tasks ||= [AutoExpireReservation, EndReservationOnly, AutoLogout]
```
jhanggi: Do each of these tasks have their own error handling? I'd like to avoid the case where one of the first jobs crashes and prevents the rest from running.

vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
waggles: yeahh this line is 🤢  if you're cool with the readable version I'd go for it too

app/controllers/admin/services_controller.rb
waggles: yeah, not qure sure what other trait binds these tasks together, hmm

vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
jhanggi: Go for it.

app/controllers/admin/services_controller.rb
waggles: Each rescues and rolls-back its own transaction block, but not via shared code so it's not necessarily guaranteed in the future, do you think that's sufficient?

jhanggi: I think it's probably a good idea to rescue, trigger an exception notification, then continue on with the rest of them.

jhanggi: On a separate note, I did a quick test and these jobs are all fast enough I'm not worried about one slowing the others down. So no need to do any delayed jobs or anything.

waggles: is that what `ActiveSupport::Notifications.instrument("background_error"` triggers?

jhanggi: Yeah, it does. Looking at it, would we end up with double exceptions? We don't need that.

waggles: I think at most one per task, which I think would match previous behavior? but my brain is a bit slow today

except `AutoLogout` which is just stderr for some reason

jhanggi: Options: 

1) We change Autologout to match and have it trigger the background notification.
2) Don't catch anything in the individual jobs and instead catch at the controller level.

I need to think a little bit, but I think I lean towards 2. That isolates the error handling into one place. The rollbacks will still happen because an exception is still getting raised; it'll get re-raised by the transaction after rolling back.

waggles: whoa wait doesn't `.transaction` catch and not-release the `ActiveRecord::Rollback`?
http://api.rubyonrails.org/v4.2/classes/ActiveRecord/Rollback.html

jhanggi: Right, but if we don't `rescue` at all, then any exception will hit the `.transaction`, trigger the rollback, and then get re-raised.

waggles: I think we are okay then, since each method called by the transactions has an inline rescue of StandardErrors, reports that, and then raises AR::Rollback, right? If we want a better handling of anything beyond that, I think I'd prefer to just take advantage of booting them into the job queue and leveraging that error handling rather than writing our own stuff.

jhanggi: That makes sense because we're raising the exception on a per-order level and we want to make sure we continue to run the job on the rest of the orders. Let's stick with what we have, but can we update `Autologout` to trigger an error notification like the other jobs?

waggles: updated~!

PR #1043, opened 2017-05-18
[#134974] separate table for order details with expired payment

meara:
- https://pm.tablexi.com/issues/134974 

screen shot 2017-05-18 at 2 43 23 pm

### DIFF
--- a/app/views/facility_journals/_problem_transactions.html.haml
+++ b/app/views/facility_journals/_problem_transactions.html.haml
@@ -1,0 +1,68 @@
+.row
+  .span12
+    %p.alert.alert-error= t("facility_journals.new.expired_payment")
+.row
+  .span12
+    = will_paginate(order_details) if order_details.respond_to? :total_pages
+    %table.table.table-striped.table-hover.dense.old-table
+      %thead
+        %tr
+          %th= Order.model_name.human
+          %th= OrderDetail.model_name.human
+          %th.nowrap= OrderDetail.human_attribute_name(@date_range_field)
+          - if @extra_date_column
+            %th.nowrap= OrderDetail.human_attribute_name(@extra_date_column_header || @extra_date_column)
+          - if current_facility.blank? || cross_facility?
+            %th= Facility.model_name.human
+          %th.nowrap= [Account.model_name.human, Account.human_attribute_name("expires_at")].join(" ")
+          %th.order-note= OrderDetail.human_attribute_name("description")
+          %th.nowrap= OrderDetail.human_attribute_name("user")
+          - unless @account
+            %th= Account.model_name.human
+            %th= Account.human_attribute_name("owner")
+          %th.currency= OrderDetail.human_attribute_name("cost")
+          %th= OrderDetail.human_attribute_name("order_status")
+          - if local_assigns[:show_statements]
+            %th= Statement.model_name.human
+      %tbody
+        - order_details.each do |order_detail|
+          %tr{ class: row_class(order_detail) }
+            - if @order_detail_link
+              %td.nowrap= link_to @order_detail_link[:text], @order_detail_link[:proc].call(order_detail) if @order_detail_link[:display?].call(order_detail)
+            - if backend?
+              %td.nowrap= link_to order_detail.order.id, facility_order_path(order_detail.order.facility, order_detail.order)
+              %td.nowrap= link_to order_detail.id, manage_order_detail_path(order_detail), :class => 'manage-order-detail'
+            - else
+              %td.nowrap= link_to order_detail.order.id, order_path(order_detail.order)
+              %td.nowrap= link_to order_detail.id, order_order_detail_path(order_detail.order, order_detail)
+            %td= order_detail.send(:"#{@date_range_field}").try(:strftime, "%m/%d/%Y")
+            - if @extra_date_column and order_detail.send(@extra_date_column)
+              %td= order_detail.send(@extra_date_column).strftime("%m/%d/%Y")
+            - if current_facility.blank? || cross_facility?
+
+              %td= order_detail.order.facility
+            %td= order_detail.account.expires_at.try(:strftime, "%m/%d/%Y")
+            %td.user-order-detail.order-note
+              .order-detail-description
+                = order_detail_description(order_detail)
+                - order_detail.extend PriceDisplayment
+                = "(#{order_detail.wrapped_quantity})".html_safe unless order_detail.problem?
+                - if order_detail.time_data.present?
+                  %br
+                  %em= order_detail.time_data
+              - if order_detail.note.present?
+                .order-detail-extra.order-detail-note
+                  = render "shared/order_detail_note", order_detail: order_detail
+
+
+            %td= order_detail.order.user.full_name
+            - unless @account
+              %td= order_detail.account
+              %td= order_detail.account.owner_user
+            %td.currency
+              - order_detail.send(:extend, PriceDisplayment)
+              = order_detail.wrapped_total
+            %td.nowrap= order_detail.order_status
+            - if local_assigns[:show_statements]
+              %td= link_to "Download", account_statement_path(@account, order_detail.statement_id, format: :pdf) if order_detail.statement
+    = will_paginate(order_details) if order_details.respond_to? :total_pages
--- a/app/views/facility_journals/new.html.haml
+++ b/app/views/facility_journals/new.html.haml
@@ -25,6 +25,8 @@
 = content_for :top_block do
   = render :partial => 'shared/transactions/top', :locals => { :tab => 'new_journal' }
 
+- if @problem_order_details.any?
+  = render 'problem_transactions', order_details: @problem_order_details
 - if @order_details.any?
   = content_for :action_instructions do
     - if has_pending_journals?
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -143,7 +143,9 @@ class FacilityJournalsController < ApplicationController
   end
 
   def set_default_variables
-    @order_details = @order_details.need_journal
+    need_journal = @order_details.need_journal
+    @order_details = need_journal.valid_account
+    @problem_order_details = need_journal.expired_account
 
     set_pending_journals
     set_earliest_journal_date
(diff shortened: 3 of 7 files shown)


app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
+.row
+  .span12
+    %p.alert.alert-error= t("facility_journals.new.expired_payment")
+.row
```
meara: I borrowed a ton of this from `shared/transactions/_table_inside`--not sure what else, if anything, we might want to strip out.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+  .span12
+    = will_paginate(order_details) if order_details.respond_to? :total_pages
```
jhanggi: We can probably take this pagination out.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+    %table.table.table-striped.table-hover.dense.old-table
+      %thead
+        %tr
+          %th= Order.model_name.human
+          %th= OrderDetail.model_name.human
+          %th.nowrap= OrderDetail.human_attribute_name(@date_range_field)
+          - if @extra_date_column
```
jhanggi: I would go through and try to take out conditionals wherever possible.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+            %th= Facility.model_name.human
+          %th.nowrap= [Account.model_name.human, Account.human_attribute_name("expires_at")].join(" ")
+          %th.order-note= OrderDetail.human_attribute_name("description")
+          %th.nowrap= OrderDetail.human_attribute_name("user")
+          - unless @account
+            %th= Account.model_name.human
+            %th= Account.human_attribute_name("owner")
+          %th.currency= OrderDetail.human_attribute_name("cost")
+          %th= OrderDetail.human_attribute_name("order_status")
+          - if local_assigns[:show_statements]
```
jhanggi: I don't think this will ever be true in this scenario. Should probably take it out.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+          %tr{ class: row_class(order_detail) }
+            - if @order_detail_link
+              %td.nowrap= link_to @order_detail_link[:text], @order_detail_link[:proc].call(order_detail) if @order_detail_link[:display?].call(order_detail)
+            - if backend?
+              %td.nowrap= link_to order_detail.order.id, facility_order_path(order_detail.order.facility, order_detail.order)
+              %td.nowrap= link_to order_detail.id, manage_order_detail_path(order_detail), class: 'manage-order-detail'
+            - else
+              %td.nowrap= link_to order_detail.order.id, order_path(order_detail.order)
+              %td.nowrap= link_to order_detail.id, order_order_detail_path(order_detail.order, order_detail)
+            %td= order_detail.send(:"#{@date_range_field}").try(:strftime, "%m/%d/%Y")
```
jhanggi: I know this was copied, but we should probably switch to `public_send`. And can we use `format_usa_date`? It's a null-safe version of `I18n.l(date.to_date, format: :usa)`? It'll help our dates stay more consistent.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+            - if @extra_date_column && order_detail.send(@extra_date_column)
+              %td= order_detail.send(@extra_date_column).strftime("%m/%d/%Y")
+            - if current_facility.blank? || cross_facility?
```
jhanggi: I think the cross facility checks are probably the only conditions we might want to keep.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+                .order-detail-extra.order-detail-note
+                  = render "shared/order_detail_note", order_detail: order_detail
+
+
+            %td= order_detail.order.user.full_name
+            - unless @account
+              %td= order_detail.account
+              %td= order_detail.account.owner_user
+            %td.currency
+              - order_detail.send(:extend, PriceDisplayment)
```
jhanggi: Seems like we're calling this multiple times on the same order detail.

app/views/facility_journals/new.html.haml
```diff
@@ -25,6 +25,8 @@
 = content_for :top_block do
   = render :partial => 'shared/transactions/top', :locals => { :tab => 'new_journal' }
 
+- if @problem_order_details.any?
```
jhanggi: This might not be an issue, but it might be worth calling `to_a` before the call to `any?`. We're going to be iterating anyways. That would avoid an extra `count` query. I'd only worry about that if the query is slow.

app/views/facility_journals/new.html.haml
```diff
@@ -25,6 +25,8 @@
...
+  = render 'problem_transactions', order_details: @problem_order_details
```
jhanggi: Use double quotes.

jhanggi:
Looks pretty good. If we're not going to share the table partial, we should strip it down to the minimum needed.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,68 @@
...
+          - unless @account
+            %th= Account.model_name.human
+            %th= Account.human_attribute_name("owner")
+          %th.currency= OrderDetail.human_attribute_name("cost")
+          %th= OrderDetail.human_attribute_name("order_status")
+          - if local_assigns[:show_statements]
+            %th= Statement.model_name.human
+      %tbody
+        - order_details.each do |order_detail|
+          %tr{ class: row_class(order_detail) }
```
noelrappin: Maybe pull the row into a partial with render collection?

app/views/facility_journals/_problem_transactions.html.haml
noelrappin: Possibly a decorator object that returns the cells or the table row that might be easier to refactor?

app/views/facility_journals/new.html.haml
noelrappin: I'd be surprised if this was a signifiant performance issue.

meara:
Feature spec to come.

meara:
@jhanggi updated.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,54 @@
...
+      %thead
+        %tr
+          %th= Order.model_name.human
+          %th= OrderDetail.model_name.human
+          %th.nowrap= OrderDetail.human_attribute_name(@date_range_field)
+          - if current_facility.blank? || cross_facility?
+            %th= Facility.model_name.human
+          %th.order-note= OrderDetail.human_attribute_name("description")
+          %th.nowrap= OrderDetail.human_attribute_name("user")
+          - unless @account
```
jhanggi: I don't think we need to worry about this conditional. I think in this scenario it will always want to render.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,54 @@
...
+            %th= Account.model_name.human
+            %th.nowrap= Account.human_attribute_name("expires_at")
+            %th= Account.human_attribute_name("owner")
+          %th.currency= OrderDetail.human_attribute_name("cost")
+          %th= OrderDetail.human_attribute_name("order_status")
+      %tbody
+        - order_details.each do |order_detail|
+          %tr{ class: row_class(order_detail) }
+            - if @order_detail_link
```
jhanggi: Is there ever a case in this flow where this conditional is necessary?

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,54 @@
...
+              %td.nowrap= link_to @order_detail_link[:text], @order_detail_link[:proc].call(order_detail) if @order_detail_link[:display?].call(order_detail)
+            - if backend?
```
jhanggi: I think we're always in `backed?` in this usage.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -0,0 +1,54 @@
...
+                - order_detail.extend(PriceDisplayment)
+                = "(#{order_detail.wrapped_quantity})".html_safe unless order_detail.problem?
+                - if order_detail.time_data.present?
+                  %br
+                  %em= order_detail.time_data
+              - if order_detail.note.present?
+                .order-detail-extra.order-detail-note
+                  = render "shared/order_detail_note", order_detail: order_detail
+            %td= order_detail.order.user.full_name
+            - unless @account
```
jhanggi: I think `@account` is always blank in this flow.

spec/features/creating_a_journal_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    login_as admin
+  end
+
+  describe "new journal page" do
+    before do
+      visit new_facility_journal_path(facility)
+    end
+
+    it 'has only journalable order details' do
```
jhanggi: double quotes here and elsewhere in the file.

spec/features/creating_a_journal_spec.rb
```diff
@@ -0,0 +1,55 @@
...
+      expect(page).to have_content("Select the orders that you wish to journal.")
+      expect(page).to have_content(reviewed_order_detail.id)
+      expect(page).not_to have_content(unreviewed_order_detail.id)
+    end
+
+    it 'has problem order details' do
```
jhanggi: Can we call this something else like "non-journalable"? "Problem order details" is a separate concept.

spec/features/creating_a_journal_spec.rb
jhanggi: This seems a little odd to call it "only journalable" when the page also has the un-journalable orders.

app/views/facility_journals/_problem_transactions.html.haml
meara: Seems so, I'm hitting an error testing locally without it.

meara:
@jhanggi updated.

PR #1044, opened 2017-05-18
[#144059135] Fix route when mass assigning occupancy price policies

jhanggi:
To be completely honest, I’m not sure why the `current_facility` is
necessary here, but not for the other problem controllers.


vendor/engines/secure_rooms/spec/features/view_facility_occupancies_spec.rb
```diff
@@ -64,10 +64,47 @@
 
         expect(current_path).to eq(show_problems_facility_occupancies_path(facility))
         # TODO: Update specs after page is updated
```
jhanggi: @waggles what was the intention behind this TODO?

PR #1047, opened 2017-05-20
[#144059233] [#145287173] Better handling of problem occupancies in modal

jhanggi:
This got bigger than I wanted it to be for a single PR, but all these pieces ended up being kind of linked together. It's largely a bunch of little changes to make the UI of editing problem occupancies better.

* Duration now starts out blank when editing a problem occupancy or reservation (before, reservations defaulted to an end time of now resulting in _very long_ times).
* Occupancy time dropdowns will be blank if the time is missing
* Show ??? for both reservation and occupancy actuals when missing start/end on FacilityOrders#show ([#144059233])
* Better error flash messaging when editing in the popup for both reservations and occupancies
  * "cannot be blank" when missing start/end
  * "must be at least 1 minute" if you put in zero duration
* Do not show inline error messages in popup so we can avoid float problems (must be at least one minute error was causing float problems)
* More intuitive JS for editing problem occupancies
  * When end time missing, it start off blank
    * Entering a duration will populate the end at
    * Entering the different pieces of the end time does nothing until you've filled them all, then it populates duration
    * If you populate the end time so it's earlier than the start time, it'll reset the start time to match the time you just put in
  * When start time is missing
    * Entering a duration will subtract that time from the end to populate the start
    * Entering the pieces of the start time does nothing until you've filled them all, then populates duration
      * If you populate the start time so it's later than the end time, it'll reset the end time to match the time you just put in

I would definitely appreciate some usability testing locally to make sure I caught all the weird scenarios (I don't know what would happen if both start and end are missing, but that should never happen for now).


app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
```diff
@@ -16,6 +17,9 @@ class window.DateTimeSelectionWidgetGroup
 
     @change()
 
+  valid: =>
+    @getDateTime() && !isNaN(@getDateTime().getTime())
```
waggles: is there an alternative syntax to `&& !isNaN` here? sigh javascript

app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
```diff
@@ -78,12 +89,24 @@ class window.ReservationTimeFieldAdjustor
...
+      # end time to X minutes after the start time
+      endTime = @reserveStart.getDateTime().addMinutes(duration)
+      @reserveEnd.setDateTime(endTime)
+    else
+      # If there was not a duration already, calculate from start/end
+      if @calculateDuration() >= 0
+        @durationField().val(@calculateDuration())
+        @durationField().trigger("change")
+      else
+        # When the input start time was after than the end time, set the end time to the beginning
```
waggles: cue The End Is The Beginning Is The End 🎶

vendor/engines/secure_rooms/config/locales/en.yml
```diff
@@ -55,6 +55,9 @@ en:
     attributes:
       user:
         card_number: Card Number
+      order_detail/occupancy:
+        exit_at: Time of Exit
+        actual_duration_mins: Duration
```
waggles: these are also below under `secure_rooms/occupancy`, do they need to be under`order_detail/occupancy` as well for some lookup thing?

app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
jhanggi: Not that I know of. That's why I isolated this to its own function.

vendor/engines/secure_rooms/config/locales/en.yml
jhanggi: It's for nested attributes error messages. Not sure why it couldn't pick up the normal version, but the messages were like "Occupancy exit at cannot be blank".

app/assets/javascripts/app/reservation_time_field_adjustor.js.coffee
jhanggi: Aren't you a little young to remember that song? Of course, I had repressed memories of that movie.

waggles:
poking around there are a few pretty minor things that I don't really care about tbh (like the adjustment select turns yellow and fades when I change things) and nothing I think is worth holding up the PR

jhanggi:
That yellow means it's recalculating the price. As long as it still has pricing in there, then it's fine.

PR #1048, opened 2017-05-22
[#144058835] Assign account when entry missing

waggles:
* Exits access handling chain if associated user cannot purchase the room, thus facility operators no longer generate orders.
* If user can purchase but order has no associated account, grabs first account from their list of `accounts_for_product`
* Error messaging now begins by stating the account must be double-checked.

### DIFF
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
@@ -15,6 +15,8 @@ module SecureRooms
       end
 
       def process
+        return unless user_can_purchase_secure_room?
+
         find_or_create_order
         complete_order
 
@@ -38,14 +40,21 @@ module SecureRooms
 
       def create_order
         ActiveRecord::Base.transaction do
+          assign_account unless occupancy.account_id?
+
           create_order_and_detail
-          # TODO: This avoids an exception but results in un-purchased order on
-          # problem occupancy, which may not appear in the problem-order scope
-          if occupancy.account_id?
-            order.validate_order!
-            order.purchase!
-          end
+          order.validate_order!
+          order.purchase!
         end
+      end
+
+      def assign_account
+        accounts = occupancy.user.accounts_for_product(occupancy.secure_room)
+        occupancy.update(account: accounts.first)
+      end
+
+      def user_can_purchase_secure_room?
+        occupancy.user.accounts_for_product(occupancy.secure_room).present?
       end
 
       def create_order_and_detail
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
@@ -15,6 +15,7 @@ module SecureRooms
       end
 
       def process
+        # TODO: [#145957283] Ensure facility operators with accounts do not create orders on exit
         return unless user_can_purchase_secure_room?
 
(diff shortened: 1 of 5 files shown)


vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -15,6 +15,8 @@ def initialize(occupancy)
       end
 
       def process
+        return unless user_can_purchase_secure_room?
```
waggles: This covers occupancies that we don't want orders for, provided someone who can enter for free can't also purchase the room. Otherwise, are there any error cases that we specifically want to watch out for?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -38,16 +40,23 @@ def complete_order
...
-            order.purchase!
-          end
+          order.validate_order!
+          order.purchase!
         end
       end
 
+      def assign_account
+        accounts = occupancy.user.accounts_for_product(occupancy.secure_room)
+        occupancy.update(account: accounts.first)
```
waggles: I'm not sure if this first account will _always_ work.. if not, we could at least cycle through the `accounts_for_product` until `validate_order!` succeeds

jhanggi: It should be fine. `accounts_for_product` should do enough validation that `validate_order!` should work.

vendor/engines/secure_rooms/config/locales/en.yml
```diff
@@ -21,8 +21,8 @@ en:
   order_details:
     notices:
       missing_entry:
-        badge: Missing Entry
-        alert: This order's occupancy does not have an entry time. Please ensure that all times are set and there is a price policy for the date this order was fulfilled.
+        badge: Inferred Account
```
jhanggi: I think I'd be fine leaving the badge the same. Could be convinced otherwise, though.

waggles: ya this is something I'd leave up to user feedback, but as a default for now I think I'll switch to "Missing Entry" + the expanded message

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
jhanggi: We should definitely test for the case of a facility operator who has accounts that could purchase. There's a very high probability of that.

For them, exiting the room shouldn't create a problem order.

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
jhanggi: Do we need  a check here for if there are no accounts?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -38,16 +40,23 @@ def complete_order
 
       def create_order
         ActiveRecord::Base.transaction do
+          assign_account unless occupancy.account_id?
```
jhanggi: Can we add a TODO until we get to the operator scan-out issue?

waggles: done, with pivotal number now that the story exists 👍

PR #1049, opened 2017-05-22
[DNM] [WIP] [#145788259] Add new order/cart related scope thing

waggles:
#### Note I'm just tossing this together for context of discussing this change, it's not a real PR

* What is currently `non_reservations` is in theory conceptually changing since occupancies are now a similar concept. This adds a scope that in theory could filter to "cart items" (which totally might be the wrong concept, let's discuss)
* each place we use `non_reservations` is highlighted with a `TODO` so we can scan through 'em all


waggles:
I'm gonna close this for now and reopen it when it's closer to play time

PR #1050, opened 2017-05-22
[#140077839] Prevent admins and non-approved users from exiting a room

jhanggi:
Re-order rules so that global admins or people not on the access list
are unable to exit a room (just like they are unable to enter a room).

### DIFF
--- a/vendor/engines/secure_rooms/spec/services/secure_rooms/check_access_spec.rb
+++ b/vendor/engines/secure_rooms/spec/services/secure_rooms/check_access_spec.rb
@@ -1,17 +1,119 @@
 require "rails_helper"
 
 RSpec.describe SecureRooms::CheckAccess, type: :service do
-  subject(:check_access) { described_class.new(rules) }
-
-  let(:card_reader) { build :card_reader }
-  let(:card_user) { build :user }
+  let(:secure_room) { create(:secure_room, :with_schedule_rule, :with_base_price) }
 
   context "with a single deny rule" do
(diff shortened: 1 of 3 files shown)


vendor/engines/secure_rooms/spec/services/secure_rooms/check_access_spec.rb
```diff
@@ -1,17 +1,118 @@
...
+
+          it "is denied exiting as well" do
+            expect(verdict).to be_denied
+            expect(verdict.reason).to include("access list")
+          end
+        end
+      end
+
+      describe "the user is on the access list" do
+        before { secure_room.product_users.create!(user: card_user, approved_by: 0) }
```
waggles: while we're here, would `approved_by: 0` make more sense as a `trait` instead?

jhanggi: This isn't a factory, and we don't actually have a factory at all for ProductUsers.

waggles: ah ya I mean more generally, should we use something more semantic, possibly higher up inside a factory

might not be worth it for this change, but generally getting the right seeds for tests has been a pain point for me because they can feel kinda random like this one

jhanggi: I can get behind that idea, but yeah, I don't think it's worth it for right now.

PR #1051, opened 2017-05-22
[#144058951] Complete orders for entry only rooms

jhanggi:
If there is an entry-only room (no exit scanner), then go ahead and complete the order as is.

Once the occupancy is created, it can be given a duration and exit at. I tried to see about making the form only have the single time for entry-only rooms, but dealing with updating the exit time to match the entry time whenever you change it proved to be beyond of what I consider the expected effort.

### DIFF
--- a/vendor/engines/secure_rooms/spec/features/scans_controller_spec.rb
+++ b/vendor/engines/secure_rooms/spec/features/scans_controller_spec.rb
@@ -23,10 +23,10 @@ RSpec.describe "Scanning in and out", type: :controller do
     describe "scanning in" do
       before do
         post :scan,
-               card_number: user.card_number,
-               reader_identifier: in_reader.card_reader_number,
-               controller_identifier: in_reader.control_device_number,
-               account_identifier: account.id
+             card_number: user.card_number,
+             reader_identifier: in_reader.card_reader_number,
+             controller_identifier: in_reader.control_device_number,
+             account_identifier: account.id
       end
 
       it "creates a new order in purchased state" do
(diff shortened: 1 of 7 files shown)


vendor/engines/secure_rooms/spec/features/scans_controller_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "rails_helper"
+
+RSpec.describe "Scanning in and out", type: :controller do
```
jhanggi: I wasn't quite sure what to do with this. My goal was to have a full-stack happy path, but it's hard to do API requests in feature specs. And this seemed somewhat different than the other ScansController controller specs. I'm open to other ideas.

waggles: while it's not necessarily correct I usually expect anything under `spec/features` to use capybara type stuff, so I'd just keep it under `spec/controllers` for that. since it's in a `...Api` named module I'd probably look first in there for specs anyway

it's not too big a deal to me though

jhanggi: There's already a `scans_controller_spec` there and this is testing a very different layer than what's currently in there. Would calling it something like `scans_controller_feature_spec.rb` be too weird? Or maybe it's better just to put it in that same file?

meara: I feel like if it's acting as a feature spec, it should be in `spec/features`, even if it's using controller tooling rather than capybara. Functionally, this is a feature spec.

This seems like an important spec to have, but if I saw this is `spec/controllers` I'd be pretty confused, because this doesn't follow a controller spec structure.

What if you just added a comment in this file explaining that this is intended as a feature spec, but you are using controller tooling because it's hard to test APIs with capybara?

jhanggi: That sounds like a good plan to me.

jhanggi: Updated.

PR #1052, opened 2017-05-22
[#144059265] More generic handling on importing unsupported product types

jhanggi:
Only allow import of Items and Services by default. This will give a
useful message when trying to import SecureRooms or instruments telling
you the particular product type is not supported for imports.

There are plenty of opportunities to refactor here, but I this was originally considered a "chore", so I opted to just make a minimal change.

### DIFF
--- a/app/support/order_row_importer.rb
+++ b/app/support/order_row_importer.rb
@@ -24,6 +24,8 @@ class OrderRowImporter
     :fulfillment_date,
   ].map { |k| HEADERS[k] }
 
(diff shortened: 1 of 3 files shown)


meara:
It occurred to me after I approved this--should there be a spec for SecureRooms that import is not allowed? Does this already exist?

jhanggi:
Spec added. I remembered why I didn't originally: It was kind of annoying to set up.

PR #1056, opened 2017-05-23
[#135657] journal expired accounts csv

meara:
- Depends on #1043, will need to be rebased once that's merged in.
- https://pm.tablexi.com/issues/135657

### DIFF
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -32,6 +32,7 @@ class FacilityJournalsController < ApplicationController
 
     respond_to do |format|
       format.csv do
+        # used for "Export as CSV" link for order details with expired accounts
         @order_details = @order_details.expired_account
       end
       format.any {}
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -29,6 +29,13 @@ class FacilityJournalsController < ApplicationController
   def new_with_search
     set_default_variables
     @layout = "two_column_head"
+
+    respond_to do |format|
+      format.csv do
+        @order_details = @order_details.expired_account
+      end
+      format.any {}
+    end
   end
 
   # PUT /facilities/journals/:id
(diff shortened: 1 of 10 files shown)


lib/csv_helper.rb
```diff
@@ -1,15 +1,6 @@
...
-        when "1.8.7"
-          require "faster_csv" unless defined?(FasterCSV)
-
-          FasterCSV
-        else
-          require "csv" unless defined?(CSV)
-
-          CSV
-    end
+  require "csv"
```
jhanggi: 🎉  I fully support finally getting rid of this. Do we need the explicit `require "csv"` here?

app/controllers/facility_journals_controller.rb
```diff
@@ -29,6 +29,13 @@ def index
   def new_with_search
     set_default_variables
     @layout = "two_column_head"
+
+    respond_to do |format|
+      format.csv do
+        @order_details = @order_details.expired_account
```
jhanggi: I feel like a comment here might be helpful since we are doing something a little funky.

app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -3,6 +3,9 @@
     %p.alert.alert-error= t("facility_journals.new.expired_payment")
 .row
   .span12
+    %p= link_to t("reports.account_transactions.export"), url_for(params.merge(format: :csv)), class: 'js--exportSearchResults pull-right', data: { form: '.search_form' }
```
jhanggi: Double quotes.

jhanggi: And after that, can we remove this from the haml-lint todo file?

meara: Still gonna flag on @date_range_field (instance variable in a partial), unless we want to break the search functionality.

meara:
@jhanggi updated.

PR #1059, opened 2017-05-24
[#144059161] Remove Orders/Reservations disputed tab

jhanggi:
Users are using the facility disputed tab under Billing now.

> I think we're OK with killing the "Disputed" tabs under "Orders" and
"Reservations". The volume of disputes has gone way down since BA's now
update many of the Payment Sources themselves. As you note the "Notices"
tab leads to the combined Orders/Reservations location in Billing, so I
think the utility of the standalone "Orders" and "Reservations" Disputed
tabs has gone way down.


jhanggi:
@meara good catch on the HAML files. Updated.

UIC definitely does not because they don't have a review period. If DC does, we can easily refer them over to the disputed tab under "Billing", which includes both orders and reservations.

PR #1061, opened 2017-05-24

waggles:
NOTE: This and #1060 cannot both go in, as it's based off of the commit that would be reverted

* This is the other way to get to a passing build, but requires a bunch of spec setup that I'm not confident about without dissecting a bit more. It's like one of those indiana-jones style tomb puzzles to get to the right situation needed to run the spec.


waggles:
also closing this to reduce the clutter, reopening after https://www.pivotaltracker.com/story/show/145957283
since they will conflict

PR #1062, opened 2017-05-24
[#145490555] Rename Control Device Number to MAC address

jhanggi:
I considered actually renaming the column, but eventually decided against it. I also considered adding an `alias_attribute :control_device_number, :mac_address`. It seemed to make things clearer and more obvious when looking at very specific places (e.g. inside the factory or validations), but got confusing when you stepped back and saw both in the same picture.

### DIFF
--- a/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
+++ b/vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
@@ -22,9 +22,10 @@ module SecureRoomsApi
 
     def load_user_and_reader
       @user = User.find_by!(card_number: params[:card_number])
+
       @card_reader = SecureRooms::CardReader.find_by!(
         card_reader_number: params[:reader_identifier],
-        control_device_number: params[:controller_identifier],
+        control_device_number: params[:controller_identifier].upcase,
       )
     end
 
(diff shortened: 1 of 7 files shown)


vendor/engines/secure_rooms/app/controllers/secure_rooms_api/scans_controller.rb
```diff
@@ -22,9 +22,10 @@ def scan
 
     def load_user_and_reader
       @user = User.find_by!(card_number: params[:card_number])
+
       @card_reader = SecureRooms::CardReader.find_by!(
         card_reader_number: params[:reader_identifier],
-        control_device_number: params[:controller_identifier],
+        control_device_number: params[:controller_identifier].upcase,
```
jhanggi: This actually won't really do much in MySQL because it's search is case-insensitive, but will affect Oracle. The test around this will pass on mysql with our without this `upcase`, but it should fail without it on Oracle.

vendor/engines/secure_rooms/spec/factories/card_reader.rb
```diff
@@ -2,7 +2,8 @@
   factory :card_reader, class: SecureRooms::CardReader do
     secure_room
     sequence(:card_reader_number) { |n| "card_reader_#{n}" }
-    sequence(:control_device_number) { |n| "control_device_#{n}" }
+    # Generate a sequential MAC address
+    sequence(:control_device_number) { |n| format("%012X", n).scan(/.{2}/).join(":") }
```
waggles: fancy

PR #1063, opened 2017-05-24
[#145957283] Skip order for exempt users

waggles:
* runs `CheckAccess` again, since that will leave `accounts` blank if the user was granted from an earlier rule--then, if `accounts` is populated, the user would have had to select a valid account
* cleans up the `OrderHandler` specs with better fixtures available

### DIFF
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
@@ -53,6 +53,11 @@ module SecureRooms
         occupancy.update(account: accounts.first)
       end
 
+      # Whether or not an occupant must pay for their time is determined by
+      # CheckAccess at scan-in time. In the event we don't have access to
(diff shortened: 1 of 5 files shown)


vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/order_handler.rb
```diff
@@ -54,8 +53,9 @@ def assign_account
         occupancy.update(account: accounts.first)
       end
 
-      def user_can_purchase_secure_room?
-        occupancy.user.accounts_for_product(occupancy.secure_room).present?
+      def user_exempt_from_purchase?
+        in_reader = occupancy.secure_room.card_readers.ingress.first
```
jhanggi: Can you add some commentary on why this works? I'm pretty sure it's fine, but it's definitely not intuitive.

waggles: kk will do

waggles:
@jhanggi updated

PR #1065, opened 2017-05-24

waggles:
* the spec now relies on fixtures!!

Original PR: #1039

### DIFF
--- a/vendor/engines/secure_rooms/lib/tasks/restricted_rooms_tasks.rake
+++ b/vendor/engines/secure_rooms/lib/tasks/restricted_rooms_tasks.rake
@@ -1,4 +1,8 @@
-# desc "Explaining what the task does"
-# task :secure_rooms do
-#   # Task goes here
-# end
+namespace :secure_rooms do
+  namespace :order_details do
+    desc "mark order_details with long-running occupancies as orphaned"
+    task orphan_occupancies: :environment do
+      SecureRooms::AutoOrphanOccupancy.new.perform
+    end
+  end
+end
--- a/vendor/engines/secure_rooms/spec/support/auto_orphan_occupancy_spec.rb
+++ b/vendor/engines/secure_rooms/spec/support/auto_orphan_occupancy_spec.rb
@@ -39,7 +39,7 @@ RSpec.describe SecureRooms::AutoOrphanOccupancy, :time_travel do
 
       it "sets the order_detail problem status" do
         action.perform
-        expect(order_detail.reload.problem).to be true
+        expect(order_detail.reload).to be_problem
       end
 
       it "does not assign pricing" do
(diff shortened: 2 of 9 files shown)


vendor/engines/secure_rooms/lib/tasks/restricted_rooms_tasks.rake
```diff
@@ -1,4 +1,8 @@
-# desc "Explaining what the task does"
-# task :secure_rooms do
-#   # Task goes here
-# end
+namespace :secure_rooms do
+  namespace :order_details do
+    desc "mark order_details with long-running occupancies as orphaned"
+    task orphan_occupancies: :environment do
```
jhanggi: Do we need this task anymore? I'm fine if we leave it, just wondering.

waggles: not directly, but I figured it might still be useful to have

jhanggi:
Can you rebase this off master just so we can make sure everything works as expected before we merge it in?

And now that that is in, can we add a spec here that an operator who gets auto-orphaned does not get a problem order created?

waggles:
updated.. the new spec resulted in an update to the service object too

vendor/engines/secure_rooms/spec/support/auto_orphan_occupancy_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+          secure_room: secure_room,
+        )
+      end
+
+      it "orphans the occupancy" do
+        expect { action.perform }.to change { occupancy.reload.orphan? }.from(false).to(true)
+      end
+
+      it "sends the occupancy through the order handler" do
+        expect(SecureRooms::AccessHandlers::OrderHandler).to receive(:process).with(occupancy)
```
jhanggi: Seems like mocking this out doesn't really ensure that the order doesn't get created because this would short-circuit the logic we want to test. Right?

vendor/engines/secure_rooms/spec/support/auto_orphan_occupancy_spec.rb
```diff
@@ -0,0 +1,128 @@
...
+        action.perform
+      end
+
+      it "sets the order_detail problem status" do
+        action.perform
+        expect(order_detail.reload.problem).to be true
```
jhanggi: Could be `expect(order_detail.reload).to be_problem`, but I'm fine either way.

vendor/engines/secure_rooms/spec/support/auto_orphan_occupancy_spec.rb
jhanggi: Oh, n/m. This is just checking in this particular test. The next test checks more of the actual behavior. Is this test really necessary, then?

waggles: yeah this is kinda testing implementation but with this and the follow spec, I wanted to make sure that it both attempted to create an order by calling it, and then didn't create one

vendor/engines/secure_rooms/spec/support/auto_orphan_occupancy_spec.rb
waggles: might as well for consistency

PR #1068, opened 2017-05-25
[#145788259] Exclude occupancies from facility orders view

jhanggi:
This creates a new `NewInprocessController` concern that is used for the sorting and loading of the `index` view. It also updates the `non_reservation` scope to be `plain_orders` (I couldn't come up with a better name) that filters by the product type.

PR #1070, opened 2017-05-25
[#145558575] Include problem occupancies in notices

jhanggi:
Extends from #1068

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -2,9 +2,6 @@ class Product < ActiveRecord::Base
 
   include TextHelpers::Translation
 
-  cattr_accessor(:types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
-  cattr_accessor(:mergeable_types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
-
   belongs_to :facility
   belongs_to :initial_order_status, class_name: "OrderStatus"
   belongs_to :facility_account
@@ -17,6 +14,9 @@ class Product < ActiveRecord::Base
   has_many   :accessories, through: :product_accessories, class_name: "Product"
   has_many   :price_policies
   has_many   :training_requests, dependent: :destroy
+
+  cattr_accessor(:types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
+  cattr_accessor(:mergeable_types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
 
   # Allow us to use `product.hidden?`
   alias_attribute :hidden, :is_hidden
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -2,6 +2,9 @@ class Product < ActiveRecord::Base
 
   include TextHelpers::Translation
 
+  cattr_accessor(:types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
+  cattr_accessor(:mergeable_types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
+
   belongs_to :facility
   belongs_to :initial_order_status, class_name: "OrderStatus"
   belongs_to :facility_account
@@ -55,14 +58,6 @@ class Product < ActiveRecord::Base
   scope :archived, -> { where(is_archived: true) }
   scope :not_archived, -> { where(is_archived: false) }
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
-
-  def self.types
-    @product_types ||= [Instrument, Item, Service, Bundle]
-  end
-
-  def self.mergeable_types
-    @mergeable_types ||= [Instrument, Item, Service, Bundle]
-  end
 
   # Products that can be used as accessories
   def self.accessorizable
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -15,9 +15,6 @@ class Product < ActiveRecord::Base
   has_many   :price_policies
   has_many   :training_requests, dependent: :destroy
 
-  cattr_accessor(:types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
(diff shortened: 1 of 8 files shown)


app/models/product.rb
```diff
@@ -15,6 +15,9 @@ class Product < ActiveRecord::Base
   has_many   :price_policies
   has_many   :training_requests, dependent: :destroy
 
+  cattr_accessor(:types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
+  cattr_accessor(:mergeable_types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
```
waggles: `types` is such a generic name that adding it to the class_variable list feels a bit scary, especially since we already have a class hierarchy under `Product` that will result in `Item.types`, etc appearing. Should we change to `product_types` at least? Or even stick with the class instance variables we had previously since this isn't saving that many lines?

app/models/order_detail.rb
```diff
@@ -260,8 +255,10 @@ def can_be_viewed_by?(user)
       .where.not(statement_id: nil)
   }
 
-  scope :non_reservations, -> { joins(:product).where("products.type <> 'Instrument'") }
-  scope :reservations, -> { joins(:product).where("products.type = 'Instrument'") }
+  # Supports single type or an array of product types
+  scope :for_product_type, ->(product_type) { joins(:product).where(products: { type: product_type }) }
+  scope :plain_orders, -> { for_product_type(["Item", "Service"]) }
```
waggles: I certainly don't have an answer but is there any more descriptive category for these two than "plain"?

app/controllers/facility_reservations_controller.rb
```diff
@@ -221,32 +206,27 @@ def update_prices
...
-                    )
-                    .where("reservations.id IS NOT NULL")
-                    .order(order_by_clause)
-  end
-
   def problem_order_details
     current_facility
       .problem_reservation_order_details
-      .joins(:reservation)
+      .includes(:reservation)
```
waggles: if we do `includes` don't we need `references(:reservation)` too?

app/controllers/facility_reservations_controller.rb
```diff
@@ -221,32 +206,27 @@ def update_prices
...
       .order("reservations.reserve_start_at DESC")
```
waggles: because of this, but I could be glazing over something
https://apidock.com/rails/ActiveRecord/QueryMethods/references

meara: P sure you need `includes` if you have `references`, but not the other way around. I've definitely used `includes` without it before.

app/models/product.rb
meara: I'm with Waggles on not loving the `types` naming--if we changed the API to `product_types` would that break a bunch of existing code?

app/models/product.rb
```diff
@@ -15,6 +15,9 @@ class Product < ActiveRecord::Base
   has_many   :price_policies
   has_many   :training_requests, dependent: :destroy
 
+  cattr_accessor(:types) { ConfigurableArray.new([Instrument, Item, Service, Bundle]) }
```
meara: I don't entirely follow why we need a special `ConfigurableArray` class rather than using a vanilla array. Is the order so important that we can't just use `.unshift` or `<<`?

app/controllers/facility_reservations_controller.rb
jhanggi: I think if we were doing a `where(reservations: {})` we would need the `references`. But we're just doing this to avoid an N+1 here, so I don't think it's necessary. It works, so I'm guessing it's not.

app/models/product.rb
jhanggi: the idea of `ConfigurableArray` is to wrap up some somewhat obtuse insertion logic where you want to insert something before or after a certain element.

The ordering of Secure Rooms being before bundles is not critical, but I think it makes much more logical sense.

jhanggi: I could be convinced to go back to the old way where we did the `index` and `insert(index, xxx)` within the engines.

meara: So, if we think we are going to end up refactoring the existing inserts to use ConfigurableArray, or if we foresee reusing it in more places in the future, I'm okay with keeping it. It's certainly more readable than `insert(index, ...)`.

I mostly just felt like adding an entire class for just one or two use cases was a bit overkill when Ruby already has an array class, but if we think we'll end up reusing this I'm fine with it.

app/controllers/facility_reservations_controller.rb
meara: Also yay on avoiding N+1's, good we are doing these refactors.

app/models/product.rb
jhanggi: The old hierarchy also exposes `Item.types`, so switching to `cattr_accessor` doesn't really change that behavior. In addition to typing, I do think it makes it a little easier to recognize as a setting.

I'm not opposed to moving towards `product_types`, but I think that's a big enough change to be outside the scope of this PR.

meara: Fair enough.

app/models/product.rb
jhanggi: Yeah, I didn't replace everywhere this could be used, but there are definitely a few more spots where this could be used. And it's really just a decorator, so it's still using the normal Array class. I contemplated monkey-patching Array, but that felt weird.

jhanggi:
Rebased off master for a cleaner diff.

app/models/product.rb
jhanggi: Hmm, there was something weird with this if I disabled the secure_rooms engine. I was getting a circular reference what autoloading Instrument. Switching back to the class methods fixed it.

jhanggi:
Updated. The `get_` codeclimate warning would involve changing the whole hierarchy since it's a template method being used by the base class.

PR #1069, opened 2017-05-25
[#145964569] Tokenized public dashboard

waggles:
Depends on #1067 , see https://github.com/tablexi/nucore-open/compare/dashboard-145965563...public-dashboard-145964569
Phase 1 of https://www.pivotaltracker.com/story/show/145964569

Does include:
* Adds new layout and route to be used by tablet
* Add token to identify `SecureRoom`s

Does not include yet:
* Auto-refresh/show last refreshed at
* Specifically accurate styling of the partial

PR #1072, opened 2017-05-26
[#145964569] Refresh dashboard

waggles:
Depends on #1069  , see https://github.com/tablexi/nucore-open/compare/public-dashboard-145964569...refresh-dashboard-145964569
Phase 2 of https://www.pivotaltracker.com/story/show/145964569

Adds auto-refresh/show last refreshed at


waggles:
also updated~

vendor/engines/secure_rooms/app/controllers/secure_rooms/occupancies_controller.rb
```diff
@@ -0,0 +1,25 @@
+module SecureRooms
+
+  class OccupanciesController < ApplicationController
+
+    layout "plain"
+
+    before_action :load_occupancies
+
+    def index
+      @refresh_url = refresh_facility_secure_room_occupancies_url(current_facility, @secure_room.dashboard_token)
```
jhanggi: Is there a way to collapse these two actions into one? Do something like `render ..., layout: !request.xhr?` so it has no layout when being fetched with ajax.

vendor/engines/secure_rooms/app/controllers/secure_rooms/occupancies_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+    end
+
+    def refresh
+      render partial: "secure_rooms/shared/secure_rooms_dashboard", locals: { room: @secure_room }
+    end
+
+    private
+
+    def load_occupancies
```
jhanggi: Would `load_secure_room` be a little closer to what this is actually doing?

vendor/engines/secure_rooms/app/controllers/secure_rooms/occupancies_controller.rb
waggles: hmm I would have to move the javascript either into the layout or into assets somehow... I'm not sure which setup I prefer

vendor/engines/secure_rooms/app/controllers/secure_rooms/occupancies_controller.rb
waggles: yeah good point, updating

PR #1075, opened 2017-05-26
[#145119365] Display count on New/IP tab

waggles:
* Similar to orders/reservations, this should be a `display_tab`


waggles:
I approve of this gif 👍

PR #1077, opened 2017-05-26
[#146173013] Configure res/occ timeout in settings

waggles:
Since the setting is now required.. should we add a default directly into the code? Like:
```
Settings.reservations.timeout_period || 12.hours
```

### DIFF
--- a/app/support/auto_expire_reservation.rb
+++ b/app/support/auto_expire_reservation.rb
@@ -15,7 +15,7 @@ class AutoExpireReservation
   end
 
   def earliest_allowed_time
-    Time.zone.now - Settings.reservations.timeout_period
+    Settings.reservations.timeout_period.seconds.ago
   end
 
   def purchased_active_order_details
--- a/app/support/auto_expire_reservation.rb
+++ b/app/support/auto_expire_reservation.rb
@@ -14,11 +14,15 @@ class AutoExpireReservation
     purchased_active_order_details
   end
 
+  def earliest_allowed_time
+    Time.zone.now - Settings.reservations.timeout_period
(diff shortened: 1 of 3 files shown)


app/support/auto_expire_reservation.rb
```diff
@@ -14,11 +14,15 @@ def order_details
     purchased_active_order_details
   end
 
+  def earliest_allowed_time
+    Time.zone.now - Settings.reservations.timeout_period
```
jhanggi: I would go with `Settings.reservations.timeout_period.ago`.

app/support/auto_expire_reservation.rb
```diff
@@ -14,11 +14,15 @@ def order_details
     purchased_active_order_details
   end
 
+  def earliest_allowed_time
+    Settings.reservations.timeout_period.seconds.ago
```
jhanggi: I think we can get away with just `Settings.reservations.timeout_period.ago` since the settings already have the `.hours`.

Surprisingly, 
```
12.hours.seconds
=> 43200 seconds
```

waggles: I thought so, but I was getting errors about `Fixnum#ago` not existing. I could dig further

jhanggi: Hmm, interesting. Looks like ERB will convert end up doing a `to_i` or `to_s` so `12.hours` gets rendered into the YAML as some primitive value. So `.seconds.ago` is fine. For the review period, we use `Time.zone.now + Settings.billing.review_period`, but I would lean towards `ago` going forward.

PR #1078, opened 2017-05-26
[#140895293] Add alarm events table

waggles:
There are a couple of uncertainties I left hanging until we can clarify:
* I named these "alerts" since scans are "events", so that now clashes with the name of the api controller.
* I'm not sure what the actual data types are, I'm assuming some might be Integers? We can just do strings though, probably.

Here's the example data I got from Dan for reference:
```
Event{
  additional_data: nil,
  class_code: "20",
  event_code: "24",
  event_description: "root password is still the default value and should be changed",
  mac_address: "00:06:8E:41:17:21",
  message_id: "001213",
  message_time: "03:25:26 GMT 01/22/2000",
  message_type: "7",
  priority: "200",
  task_code: "7",
  task_description: "monitor task"
}
```

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170526170933) do
+ActiveRecord::Schema.define(version: 20170531183710) do
 
   create_table "account_users", force: :cascade do |t|
     t.integer  "account_id", limit: 4,  null: false
@@ -596,7 +596,7 @@ ActiveRecord::Schema.define(version: 20170526170933) do
     t.boolean "on_sat",                                                            null: false
   end
 
-  add_index "schedule_rules", ["product_id"], name: "fk_rails_bd1c8c4ecb", using: :btree
+  add_index "schedule_rules", ["product_id"], name: "index_schedule_rules_on_product_id", using: :btree
 
   create_table "schedules", force: :cascade do |t|
     t.string   "name",        limit: 255
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -479,8 +479,10 @@ ActiveRecord::Schema.define(version: 20170526170933) do
     t.text     "training_request_contacts", limit: 65535
     t.boolean  "note_available_to_users",                 default: false, null: false
     t.integer  "cutoff_hours",              limit: 4,     default: 0,     null: false
+    t.string   "dashboard_token",           limit: 255
   end
 
+  add_index "products", ["dashboard_token"], name: "index_products_on_dashboard_token", using: :btree
   add_index "products", ["facility_account_id"], name: "fk_facility_accounts", using: :btree
   add_index "products", ["facility_id"], name: "fk_rails_0c9fa1afbe", using: :btree
   add_index "products", ["schedule_id"], name: "i_instruments_schedule_id", using: :btree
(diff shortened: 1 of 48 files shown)


vendor/engines/secure_rooms/app/controllers/secure_rooms_api/events_controller.rb
```diff
@@ -3,10 +3,28 @@ module SecureRoomsApi
...
+    def alert_params
+      params.permit(
+        :additional_data,
+        :class_code,
+        :event_code,
+        :event_description,
+        :mac_address,
+        :message_id,
+        :message_time,
+        :message_type,
```
jhanggi: Can we add a test for datetime parsing? Particularly that month/day are parsed USA-style rather than rest-of-the-world style?

jhanggi:
This looks pretty good as a starting point. Would it be a good idea to have a `raw` column that is the raw request (or JSON serialized version of the request)? That way if we do get fields that don't belong in one of those columns, we still capture it?

waggles:
I think it's a good idea to store the raw, especially for now so we can have it around as we learn more about what the incoming data is like

waggles:
a couple things here:
* @danhodos Is the `message_time` value a raw value? or do you have some control over parsing it before it gets sent along?
* @jhanggi we (probably, depending on above) have to both parse the time and clean-up/store the raw_json--where do you prefer this logic goes? If it remains to be request-parsing stuff, it feels like it could stay in the controller, but if it's more than that we could consider a service object. opine?

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/events_controller.rb
```diff
@@ -2,11 +2,37 @@ module SecureRoomsApi
 
   class EventsController < SecureRoomsApi::ApiController
 
+    before_action :parse_time
+
     def create
-      Rails.logger.info "SecureRoomsApi Event: #{params}"
+      SecureRooms::Alert.create(alert_params)
```
jhanggi: We could possibly do this as something like (I haven't actually verified this is syntactically correct):
```
SecureRooms::Alert.create(
  alert_params.merge(
    message_time: parse_time(params[:message_time)),
    raw: JSON.parse(request.body),
  )
)
```

I do it occasionally, but I generally prefer to avoid modifying the `params` hash.

jhanggi: And we should probably do some sort of if/else. I think I'd want to know if this fails. At least a rollbar error, even if we return an OK to the client.

waggles: would just using a `create!` and letting it end up in rollbar that way be sufficient?
@danhodos do you care about the response from this controller (if it's an error)?

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/events_controller.rb
```diff
@@ -3,10 +3,40 @@ module SecureRoomsApi
   class EventsController < SecureRoomsApi::ApiController
 
     def create
-      Rails.logger.info "SecureRoomsApi Event: #{params}"
+      # TODO: Discuss how to handle errors
```
jhanggi: I think this is probably fine for now. None of these fields are validated, so there _shouldn't_ be any problems. I think we can remove the TODO.

db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170525143813) do
+ActiveRecord::Schema.define(version: 20170526170933) do
```
jhanggi: I suspect this is where the conflict is.

PR #1080, opened 2017-05-30
only migrate where canceled is not nil, skip validations

meara:


db/migrate/20170501201633_add_canceled_to_order_detail.rb
```diff
@@ -4,10 +4,11 @@ def up
     add_column :order_details, :canceled_by, :integer
     add_column :order_details, :canceled_reason, :string
 
-    Reservation.where.not(order_detail_id: nil).find_each do |reservation|
-      reservation.order_detail.update_attributes(canceled_at: reservation.canceled_at,
+    Reservation.where.not(order_detail_id: nil, canceled_at: nil).find_each do |reservation|
+      reservation.order_detail.assign_attributes(canceled_at: reservation.canceled_at,
```
jhanggi: The save needs to `order_detail`.

I'm looking to see if this has different behavior than `update_columns`.

jhanggi: Looks like the `save(validate: false)` still runs callbacks. `update_columns` is definitely faster.

Also, adding `includes(:order_detail)` into the query is making this much faster as well.

I just ran it with the includes and `update_columns` and it definitely looks faster (I'm seeing UPDATEs fly by much faster).

Usually I don't care too much about the performance of migrations, but once they go over a couple minutes I start to get worried. If we were to want to really boost this, we'd do it in a single `UPDATE` clause, but that's not as obvious what's going on.

noelrappin: Not sure how big a deal this is, but maybe define a dummy `Reservation` class at the top of the file (as in http://railsguides.net/change-data-in-migrations-like-a-boss/) -- I normally prefer data changes to be outside migrations, but I understand that might be logistically complicated.

meara: I ended up putting the data change in the migration because it seemed important to make sure that the data change was definitely run before the old `reservation.canceled_*` columns were dropped. Putting it in a migration ensured that it would be impossible to miss migrating the data, where with a rake task, one could conceivably inadvertently skip running it before dropping the old columns and end up with data loss.

Coordinating running a rake task across four or five staging environments, three production environments, and everyone's dev environment seemed a little error-prone to me. Idk if it's worth changing from a migration at this point, but curious what your preferred alternative might be, Noel.

jhanggi: I'm inclined to agree with Meara in this case. I would consider this change to be an integral part of the migration, and needs to happen on every row across every environment. If you're updating more specific rows or specific environments, then a rake task is better.

PR #1084, opened 2017-05-31
[#136226] prevent deletion of order form for purchased orders

meara:
- https://pm.tablexi.com/issues/136226

### DIFF
--- a/app/controllers/order_detail_stored_files_controller.rb
+++ b/app/controllers/order_detail_stored_files_controller.rb
@@ -39,7 +39,7 @@ class OrderDetailStoredFilesController < ApplicationController
     @file.created_by = session_user.id ## this is correct, session_user instead of acting_user
 
     if @file.save
-      flash[:notice] = "Order File uploaded successfully" # TODO: I18n
+      flash[:notice] = text('upload_order_file.success')
 
       if @order_detail.order.to_be_merged?
         @order_detail.merge! # trigger the OrderDetailObserver callbacks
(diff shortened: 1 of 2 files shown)


app/controllers/order_detail_stored_files_controller.rb
```diff
@@ -48,19 +48,23 @@ def upload_order_file
...
       render :order_file
     end
   end
 
   # GET /orders/:order_id/order_details/:order_detail_id/remove_order_file
   def remove_order_file
-    if @order_detail.stored_files.template_result.all?(&:destroy)
-      flash[:notice] = "The uploaded Order File has been deleted successfully" # TODO: I18n
+    if @order_detail.new?
+      if @order_detail.stored_files.template_result.all?(&:destroy)
```
jhanggi: Will `@order_detail.stored_files.template_result.destroy_all` do the same thing? I forget what it returns.

app/controllers/order_detail_stored_files_controller.rb
```diff
@@ -48,19 +48,23 @@ def upload_order_file
...
+        flash[:notice] = text('remove_order_file.success')
```
jhanggi: Use double quotes.

meara: Huh it is weird that rubocop didn't flag on that...

app/controllers/order_detail_stored_files_controller.rb
meara: It returns the collection of objects that it destroys... idk if that is quite what we we want for that boolean.

https://apidock.com/rails/ActiveRecord/Relation/destroy_all

jhanggi: I think it's fine to leave as is. Although, technically `destroy` also returns the single object that was just destroyed.

I think it's probably fine to leave as is. I can't think of where this wouldn't work. And really there's only ever one of these files at a time.

app/controllers/order_detail_stored_files_controller.rb
meara: Okay, I am genuinely mystified by this. I can't find anything in the rubocop config that explains why it should not have flagged on this.

PR #1085, opened 2017-05-31
[#146399533] Require complete? to view description

waggles:
This was using the wrong method, originally the logic was `#missing_price_policy?` which includes `complete?` like so:
```
def missing_price_policy?
  complete? && price_policy.nil?
end
```


jhanggi:
Good change. Can we add a spec around this?

waggles:
it is done

jhanggi:
I probably would have stuck with a model spec, but this works, too.

PR #1086, opened 2017-05-31
[#134979] add invoice number and charge_for fields to export raw

meara:
- https://pm.tablexi.com/issues/134979

### DIFF
--- a/app/models/reports/export_raw.rb
+++ b/app/models/reports/export_raw.rb
@@ -84,6 +84,8 @@ module Reports
         journal_date: ->(od) { od.journal.journal_date if od.journal },
         reconciled_note: :reconciled_note,
         reconciled_at: :reconciled_at,
+        invoice_number: ->(od) { od.statement.try(:invoice_number) },
+        charge_for: ->(od) { od.price_policy.try(:charge_for) },
       }
       if SettingsHelper.has_review_period?
         hash


app/models/reports/export_raw.rb
```diff
@@ -84,6 +84,8 @@ def default_report_hash
         journal_date: ->(od) { od.journal.journal_date if od.journal },
         reconciled_note: :reconciled_note,
         reconciled_at: :reconciled_at,
+        invoice_number: ->(od) { od.statement.try(:invoice_number) },
+        charge_for: ->(od) { od.price_policy.try(:charge_for) },
```
meara: @jhanggi -- this field is nullable. It's a little unclear to me from the ticket--are they okay with this being null? I'm not sure it would ever be set to "quantity", for instance...

jhanggi: This might be slightly more complicated than we initially hoped.

We might need to talk through this with the client. Here are the scenarios I can think of:

* Price policy is present, but does not respond to charge_for because it's an Item or ServicePricePolicy.
  * In that case, we'll want to display "Quantity".
* Price policy is blank because the order hasn't been completed
  * Maybe we should display the current price policy's charge_for. Or do we leave it blank?
* Price policy is blank because it's a problem & missing actuals
  * We would probably want to do the same as we do for being uncomplete
* Price policy is blank because it's a problem & missing price policy
  * Probably leave it blank in this case.

jhanggi: We should probably adjust the I18n to have the header be "Current Charge Mode". And make sure our translations display the right thing or "Overage", "Usage", "Reservation".

meara:
Closed in favor of https://github.com/tablexi/nucore-open/pull/1104

PR #1094, opened 2017-06-01
[#136391] hide statements and orders tabs for account managers

meara:
- https://pm.tablexi.com/issues/136391

### DIFF
--- a/app/views/admin/shared/_tabnav_payment_method.html.haml
+++ b/app/views/admin/shared/_tabnav_payment_method.html.haml
@@ -1,8 +1,8 @@
 %ul.nav.nav-tabs
   = tab t(".details"), facility_account_path(current_facility, @account), (secondary_tab == "details")
   -# TODO: use a feature flag here instead
-  - if @account.statements.any? && !current_user.account_manager?
+  - if @account.statements.any? && current_ability.can?(:show, @account.statements.first)
     = tab t(".statements"), facility_account_statements_path(current_facility, @account), (secondary_tab == "statements")
   = tab t(".members"), facility_account_members_path(current_facility, @account), (secondary_tab == "members")
-  - unless current_user.account_manager?
-    = tab t(".orders"), facility_account_orders_path(current_facility, @account), (secondary_tab == "orders")
+  - if current_ability.can?(:show, @account.orders.first)
+    = tab t(".orders"), facility_account_orders_path(current_facility, @accountZ), (secondary_tab == "orders")
--- a/app/views/admin/shared/_tabnav_payment_method.html.haml
+++ b/app/views/admin/shared/_tabnav_payment_method.html.haml
@@ -1,7 +1,8 @@
 %ul.nav.nav-tabs
   = tab t(".details"), facility_account_path(current_facility, @account), (secondary_tab == "details")
   -# TODO: use a feature flag here instead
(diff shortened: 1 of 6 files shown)


app/views/admin/shared/_tabnav_payment_method.html.haml
```diff
@@ -1,7 +1,8 @@
 %ul.nav.nav-tabs
   = tab t(".details"), facility_account_path(current_facility, @account), (secondary_tab == "details")
   -# TODO: use a feature flag here instead
-  - if @account.statements.any?
+  - if @account.statements.any? && !current_user.account_manager?
     = tab t(".statements"), facility_account_statements_path(current_facility, @account), (secondary_tab == "statements")
   = tab t(".members"), facility_account_members_path(current_facility, @account), (secondary_tab == "members")
-  = tab t(".orders"), facility_account_orders_path(current_facility, @account), (secondary_tab == "orders")
+  - unless current_user.account_manager?
```
meara: There is probably a more sensible thing to switch on, but I'm not sure quite what it would be.

jhanggi: Is there an ability we can do this on? `current_ability.can? :orders, @account`? You could probably figure it out by seeing what ability is checked by the controller this links to.

spec/lib/ability_spec.rb
```diff
@@ -66,7 +66,7 @@
       let(:facility) { Facility.cross_facility }
 
       it { is_expected.to be_allowed_to(:manage_accounts, facility) }
-      it { is_expected.to be_allowed_to(:manage, Account) }
+      it { is_expected.to be_allowed_to([:read, :edit], Account) }
```
meara: @jhanggi, whenever you have a sec--this spec is still failing and I can't figure out why. What am I missing?

jhanggi: The `be_allowed_to` matcher only accepts a single ability, not an array.

We have `it_is_allowed_to([:read, :edit], Account)`, which essentially expands to
```
it { is_expected.to be_allowed_to(:read, Account) }
it { is_expected.to be_allowed_to(:edit, Account) }
```

meara: Ah thanks. Updated.

meara:
@jhanggi I added suspend/unsuspend account abilities to fix a failing split accounts spec--are there any other abilities we should double check that might not have test coverage?

meara:
I added `:update` too because I wasn't sure if `:edit` was sufficient--seemed like update was the more comprehensive ability from my reading of the docs.

spec/lib/ability_spec.rb
```diff
@@ -66,7 +66,7 @@
       let(:facility) { Facility.cross_facility }
 
       it { is_expected.to be_allowed_to(:manage_accounts, facility) }
-      it_is_allowed_to([:read, :edit, :suspend, :unsuspend], Account)
```
jhanggi: If you include `:edit, :new` in here, does it still pass?

meara: Yup, looks like `:create` covers `:new` as well, and `:update` covers `:edit`. Updated the ability file accordingly.

jhanggi:
👍  pending CI

PR #1097, opened 2017-06-06
Rake task for reverting engine migrations

jhanggi:
Example: `rake engines:db:migrate_down[projects]`

Conveniently (and honestly a little surprisingly), this will happily run even if the engine is disabled in the Gemfile. 

If the engine is enabled, then following this up with a `db:migrate` should leave the schema in a similar state (column ordering might be different).


meara:
I like this solution--can you add a line in a readme documenting the use of this?

PR #1101, opened 2017-06-06
[#136529] Responsive sanger receipt

jhanggi:
This includes #1100 and #1099.

The main change here is updating the responsive table JS to handle nested tables.

This does some fixing, but I'm wondering if there's a way we can get the height of the row to fit the header if it spans multiple lines. I wasn't having very much luck. I'd rather not use the data attribute with a shorter name, but we could fall back to that if there's no other way.

screen shot 2017-06-06 at 6 01 00 pm

### DIFF
--- a/app/assets/javascripts/app/responsive_table.js.coffee
+++ b/app/assets/javascripts/app/responsive_table.js.coffee
@@ -20,7 +20,10 @@ class window.ResponsiveTable
     @table.find("tbody tr").each (index, row) => @add_header_to_row($(row))
 
   add_header_to_row: ($row) ->
-    $row.find("> td").prepend (index) => @responsive_header($row, index)
+    # Only get the immediate child cells of the row. Without the `>`, it could
+    # also find cells in nested tables.
+    cells = $row.find("> td")
+    cells.prepend (index) => @responsive_header($row, index)
 
   responsive_header: ($row, index) ->
     $("<div>").addClass("responsive-header").text(@text_for_header($row, index))
--- a/app/assets/javascripts/app/responsive_table.js.coffee
+++ b/app/assets/javascripts/app/responsive_table.js.coffee
@@ -28,7 +28,7 @@ class window.ResponsiveTable
   responsive_header: ($row, index) ->
     $("<div>").addClass("responsive-header").text(@text_for_header($row, index))
 
-  text_for_header: ($row, index) =>
+  text_for_header: ($row, index) ->
     header = $($row.closest("table").find("thead th").eq(index))
     header.data("mobile-header") || header.text()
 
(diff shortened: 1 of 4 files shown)


noelrappin:
I couldn't figure out how to do that with the header either if I recall correctly.

spec/javascripts/responsive_tables_spec.coffee
```diff
@@ -32,3 +32,29 @@ describe "Responsive table support", ->
...
+      expect($outerTableHeaders.eq(3).text()).toEqual("Facility")
+
+    it "sets the inner table headers", ->
+      $innerTableHeaders = $("#inner-table .responsive-header")
+      expect($innerTableHeaders.size()).toEqual(4)
+      expect($innerTableHeaders.eq(0).text()).toEqual("Inner Header 1")
+      expect($innerTableHeaders.eq(1).text()).toEqual("Inner Header 2")
+      expect($innerTableHeaders.eq(2).text()).toEqual("Inner Header 1")
+      expect($innerTableHeaders.eq(3).text()).toEqual("Inner Header 2")
+
```
noelrappin: Thanks for adding the test.

app/assets/javascripts/app/responsive_table.js.coffee
```diff
@@ -20,13 +20,13 @@ class window.ResponsiveTable
     @table.find("tbody tr").each (index, row) => @add_header_to_row($(row))
 
   add_header_to_row: ($row) ->
-    $row.find("td").prepend (index) => @responsive_header(index)
+    $row.find("> td").prepend (index) => @responsive_header($row, index)
```
noelrappin: In this case, I might consider either pulling `$row.find("> td")` into a method with a meaningful name or a comment explaining what `>` does here because I think it's a little opaque otherwise. (Possibly just me not remembering all the jQuery sigils....)

jhanggi: It's from CSS, but yeah, I can add a comment. I think having a comment will also help clarify _why_ it's there in addition to what it's doing.

jhanggi:
Updated. I was able to use `float: left` on the header and that seemed to work better than `position: absolute`. Might be worth some additional looking around at other responsive tables, but everything I've looked at is fine.

screen shot 2017-06-07 at 1 05 51 pm
screen shot 2017-06-07 at 1 05 46 pm

app/assets/javascripts/app/responsive_table.js.coffee
```diff
@@ -20,13 +20,16 @@ class window.ResponsiveTable
     @table.find("tbody tr").each (index, row) => @add_header_to_row($(row))
 
   add_header_to_row: ($row) ->
-    $row.find("td").prepend (index) => @responsive_header(index)
+    # Only get the immediate child cells of the row. Without the `>`, it could
+    # also find cells in nested tables.
+    cells = $row.find("> td")
```
noelrappin: Thank you, this is great.

PR #1103, opened 2017-06-07
[#134974] new journal page account validation

meara:
- https://pm.tablexi.com/issues/134974

### DIFF
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -32,12 +32,12 @@ class FacilityJournalsController < ApplicationController
 
     @after_search_hook = lambda do
       @valid_order_details, @invalid_order_details = ValidatorFactory.partition_valid_order_details(@order_details.unexpired_account)
-      @unjournalable_order_details = @invalid_order_details + @order_details.expired_account
+      @invalid_order_details += @order_details.expired_account
 
       respond_to do |format|
         format.csv do
           # used for "Export as CSV" link for order details with expired accounts
-          @order_details = @unjournalable_order_details
+          @order_details = @invalid_order_details
         end
         format.any {}
       end
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -30,12 +30,17 @@ class FacilityJournalsController < ApplicationController
     set_default_variables
     @layout = "two_column_head"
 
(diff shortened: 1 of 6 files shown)


app/controllers/facility_journals_controller.rb
```diff
@@ -30,12 +30,17 @@ def new_with_search
     set_default_variables
     @layout = "two_column_head"
 
-    respond_to do |format|
-      format.csv do
-        # used for "Export as CSV" link for order details with expired accounts
-        @order_details = @order_details.expired_account
+    @after_search_hook = lambda do
+      @valid_order_details, @invalid_order_details = ValidatorFactory.partition_valid_order_details(@order_details.unexpired_account)
+      @invalid_order_details += @order_details.expired_account
```
jhanggi: Is there anything we need to do around sorting these together since it'll be two sets appended to each other?

lib/transaction_search.rb
```diff
@@ -144,7 +145,8 @@ def handle_csv_search
   end
 
   def email_csv_export
-    AccountTransactionReportMailer.delay.csv_report_email(to_email, @order_details.pluck(:id), @date_range_field)
+    order_detail_ids = @order_details.respond_to?(:pluck) ? @order_details.pluck(:id) : @order_details.map(&:id)
```
noelrappin: Is this allowing it to be called with an association or an array? Maybe `@order_details.to_a.map(&:id)`?

jhanggi: We went with this to give us the performance/memory boost if it's a relation, but then still work if it's already an array.

Part of me wanted to monkey patch `Array` to support `pluck`, but that seemed like a bad idea.

app/controllers/facility_journals_controller.rb
meara: We can sort them, but it looks like the basic scope (`.need_journal`) doesn't do any ordering so I'm not sure what would make sense to order by.  There can't be duplicates, since they're disjoint sets.

lib/transaction_search.rb
noelrappin: I'm not suggesting that you do that, but I see the appeal.

meara:
Hey @noelrappin, @jhanggi, are there changes needed, or is this good to go?

noelrappin:
Good to go as far as I'm concerned

PR #1104, opened 2017-06-07
[#134979] Add charge mode and invoice number to export raw

jhanggi:
Continues on from #1086. 

As part of this, I set up a new custom matcher to clean up how we're testing export raw. I haven't updated the SecureRooms export raw spec yet. I wanted to make sure people liked this change first.

### DIFF
--- a/app/models/charge_mode.rb
+++ b/app/models/charge_mode.rb
@@ -1,0 +1,12 @@
+class ChargeMode
+
+  def self.for_order_detail(order_detail)
+    if order_detail.price_policy
+      order_detail.price_policy.charge_for
+    else
+      current_policy = order_detail.product.current_price_policies.first
+      current_policy.try(:charge_for)
+    end
+  end
+
+end
--- a/spec/models/charge_mode_spec.rb
+++ b/spec/models/charge_mode_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+
+RSpec.describe ChargeMode do
+
+  describe ".for_order_detail" do
+    subject(:charge_for) { described_class.for_order_detail(order_detail) }
+
+    describe "it has an assigned price policy" do
+      let(:order_detail) { build(:order_detail, price_policy: price_policy) }
+
+      describe "for an item" do
+        let(:price_policy) { build(:item_price_policy) }
+        it { is_expected.to eq("quantity") }
+      end
+
+      describe "for an instrument" do
+        let(:price_policy) { build(:instrument_price_policy, charge_for: "overage") }
+        it { is_expected.to eq("overage") }
+      end
+    end
+
+    describe "with no assigned price policy and no current policy for the product" do
+      let(:item) { build_stubbed(:item) }
+      let(:order_detail) { build(:order_detail, price_policy: nil, product: item) }
+
+      it { is_expected.to be_blank }
+    end
+
+    describe "with no assigned price policy but there is a current policy" do
+      let(:instrument) { create(:setup_instrument) }
+      let(:order_detail) { build_stubbed(:order_detail, product: instrument) }
+
+      it { is_expected.to eq("reservation") }
+
+      describe "and the policy is overage" do
+        before { instrument.price_policies.update_all(charge_for: "overage") }
+
+        it { is_expected.to eq("overage") }
+      end
+    end
+
+  end
+
+end
--- a/spec/models/reports/export_raw_spec.rb
+++ b/spec/models/reports/export_raw_spec.rb
@@ -39,20 +39,22 @@ RSpec.describe Reports::ExportRaw do
       expect(report.to_csv.split("\n").length).to eq(2)
     end
 
-    it "populates the report", :aggregate_failures do
-      expect(report).to have_column(Facility.model_name.human).with_value("My Facility (MF)")
-      expect(report).to have_column("Order").with_value(order_detail.to_s)
-      expect(report).to have_column("Ordered By").with_value(user.username)
-      expect(report).to have_column("First Name").with_value(user.first_name)
-      expect(report).to have_column("Last Name").with_value(user.last_name)
-      expect(report).to have_column("Quantity").with_value("3")
-      expect(report).to have_column("Estimated Cost").with_value("$39.99")
-      expect(report).to have_column("Estimated Subsidy").with_value("$29.99")
-      expect(report).to have_column("Estimated Total").with_value("$10.00")
-      expect(report).to have_column("Actual Cost").with_value("$19.99")
-      expect(report).to have_column("Actual Subsidy").with_value("$9.99")
-      expect(report).to have_column("Actual Total").with_value("$10.00")
-      expect(report).to have_column("Charge For").with_value("Quantity")
+    it "populates the report" do
+      expect(report).to have_column_values(
+        Facility.model_name.human => "My Facility (MF)",
+        "Order" => order_detail.to_s,
+        "Ordered By" => user.username,
+        "First Name" => user.first_name,
+        "Last Name" => user.last_name,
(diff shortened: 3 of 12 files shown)


app/models/charge_mode.rb
```diff
@@ -0,0 +1,12 @@
...
+  def self.for_order_detail(order_detail)
+    if order_detail.price_policy
+      order_detail.price_policy.charge_for
+    else
+      current_policy = order_detail.product.current_price_policies.first
+      current_policy.try(:charge_for)
+    end
+  end
+
+end
```
noelrappin: Why is this a separate class and not an instance method of `OrderDetail`?

spec/support/matchers/csv_column_matcher.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  failure_message do
+    if @actual_values
+      %(Expected report to have value(s) "#{@expected_values.join(', ')}" in column #{column_header}, but got "#{@actual_values.join(', ')}")
+    else
+      %(Report did not have column with header "#{column_header}")
+    end
+  end
+end
```
noelrappin: I like this!

spec/models/reports/export_raw_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+        )
+      end
+    end
+
+    it "exports correct number of line items" do
+      expect(report.to_csv.split("\n").length).to eq(2)
+    end
+
+    it "populates the report", :aggregate_failures do
+      is_expected.to have_column(Facility.model_name.human).with_value("My Facility (MF)")
```
noelrappin: I have a mild preference for `expect(report).to` rather than using subject and `is_expected.to` because it is more explicit, but it's not a big deal

spec/models/reports/export_raw_spec.rb
```diff
@@ -0,0 +1,105 @@
...
+      is_expected.to have_column("Quantity").with_value("3")
+      is_expected.to have_column("Estimated Cost").with_value("$39.99")
+      is_expected.to have_column("Estimated Subsidy").with_value("$29.99")
+      is_expected.to have_column("Estimated Total").with_value("$10.00")
+      is_expected.to have_column("Actual Cost").with_value("$19.99")
+      is_expected.to have_column("Actual Subsidy").with_value("$9.99")
+      is_expected.to have_column("Actual Total").with_value("$10.00")
+      is_expected.to have_column("Charge For").with_value("Quantity")
+    end
+
```
noelrappin: idle speculation, is this slow because it's running the CSV report for each matcher, and if so, might the matcher be more like `is_expected.to have_columns("Order" => order_detail.to_s, "Ordered By" => user.username, ...` -- might be harder to read and not worth it, just wondering.

spec/models/reports/export_raw_spec.rb
jhanggi: I like it when you can do `it { is_expected.to`, but I think in this case since we have a description, it probably makes sense to switch to `expect(report).to`

spec/models/reports/export_raw_spec.rb
jhanggi: Yeah, it's slow because of the setup of orders and what not. That's an interesting idea. I think it could be made to read decently. I'll play around with that.

app/models/charge_mode.rb
jhanggi: Fair question. The initial reason was simply that `OrderDetail` is giant enough as is. Later reasons revolve around the fact that it's not really the order detail's responsibility. If there's a price policy assigned, then it could delegate easily enough, but when there's not one assigned, it almost becomes the product's responsibility. Since who was responsible based on the state, I decided to give it its own class.

If we had a `NullPricePolicy`, I could see some of the logic going there, but adding that in feels pretty dangerous at this point: there are a lot of checking on the presence of `price_policy`. If only you could make a null object falsy...

spec/models/charge_mode_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+RSpec.describe ChargeMode do
+
+  describe ".for_order_detail" do
+    subject(:charge_for) { described_class.for_order_detail(order_detail) }
+
+    describe "it has an assigned price policy" do
+      let(:order_detail) { build(:order_detail, price_policy: price_policy) }
+
+      describe "for an item" do
```
meara: How do price policies work with bundles? Do they have one for the bundle as a whole, or is it somehow delegated to the products within the bundle?

spec/support/matchers/csv_column_matcher.rb
meara: Agree, this is cool.

meara:
Also, it appears that codeclimate wants arrays of words to be notated with %w--not sure I like that, tbh.

spec/models/charge_mode_spec.rb
jhanggi: An order detail never has a bundle as its product. If you buy a bundle of say "Item 1" and "Service 2", you end up with two order details, one for each product.

jhanggi:
Updated to include testing the tester.

app/models/charge_mode.rb
noelrappin: In that case, I have no problem making this kind of a service object for order details, but maybe I'd prefer an class with an instance method, not a class method. (Maybe even a SimpleDelegator).

spec/support/matchers/csv_column_matcher.rb
```diff
@@ -0,0 +1,56 @@
...
+# expect(report).to have_column("Quantity").with_values("1", "2")
+# expect(report).to have_column("Quantity").with_values(["1", "2"])
+RSpec::Matchers.define :have_column do |column_header|
+  match do |report|
+    lines = report.to_csv.split("\n")
+    headers = lines.shift.split(",")
+    @column_index = headers.index(column_header)
+    break false unless @column_index
+    break true unless @expected_values
+    @actual_values = lines.map { |line| line.split(",")[@column_index].to_s }
```
ahorner: I'd use Ruby's built-in `CSV` handling for this. Naïvely using `#split` is somewhat fraught, because your cells can technically contain internal commas, so long as they're wrapped in quotes.

The implementation using `CSV` also reads a lot better to me:

```ruby
csv = CSV.parse(report.to_csv)
break false unless csv.headers.include?(column_header)
break true unless @expected_values

@actual_values = csv.map { |row| row[column_header] }
@actual_values == @expected_values
```

jhanggi:
@meara I'm thinking of adding
```
Style/WordArray:
  MinSize: 3
```
or even a bigger size. This will only trigger the cop for arrays of at least 3 items. I do think once the array gets larger, it makes sense to switch to `%w`. What do you think?

meara:
@jhanggi I feel like `MinSize: 5` and I'd be on board.

spec/support/matchers/csv_column_matcher.rb
jhanggi: Thanks, that's definitely nicer. 

Looks like we need `headers: true` as an option in the `CSV.parse` call.

jhanggi:
Updated. Making the matcher composable might be overkill, but it was definitely a fun and informative experiment into the inner workings of rspec.

spec/models/charge_mode_spec.rb
meara: 👍

spec/support/matchers/csv_column_matcher.rb
```diff
@@ -0,0 +1,65 @@
+require "csv"
+require "pry"
+# Examples:
```
meara: I appreciate you including examples. 😃

PR #1105, opened 2017-06-08
Update to TXI's latest rubocop-txi.yml

jhanggi:


.rubocop-txi.yml
```diff
@@ -4,7 +4,6 @@ AllCops:
     - ".bundle/**/*" # Auto-generated
     - "bin/**/*"     # Auto-generated
 
-  TargetRubyVersion: 2.2
```
jhanggi: Rubocop will use `.ruby-version` if it's present.

.rubocop.yml
```diff
@@ -1,22 +1,31 @@
...
+# Project-specific overrides go here, but try to conform to TXI standards as
+# specified in .rubocop-txi.yml when possible.
 
+# We prefer this to a one-liner:
+# def methodname
+# end
 Style/EmptyMethod:
   EnforcedStyle: expanded
-Style/HashSyntax:
-  EnforcedStyle: ruby19
```
jhanggi: This seems to be the default now, so the explicit setting it isn't necessary anymore.

.rubocop-txi.yml
jhanggi: Well, rubocop should do that, but code climate doesn't seem to look at it.

PR #1107, opened 2017-06-08
Prevent spurious test failures when factory sequence adds a character

jhanggi:
Before, we would occasionally get test failures because `"9100-7777777" < "999-7777777"`
when we were dealing with sorting. This will allow us many more before it rolls over
to add an extra character.

NU overrides this factory to do something similar, but abides by its custom
validation rules.

Example of failure, when the test is checking ordering.
```
expected: "nufs account aaaaabml / 91000-7777777"
     got: "nufs account aaaaabmk / 9999-7777777"
```


noelrappin:
:+1

PR #1109, opened 2017-06-09
[#126401] Update user info on LDAP login

jhanggi:
When a user logs in via LDAP, we want to update the user's first name, last name, and email address.

This also isolates the LDAP functionality into an engine so schools that don't use it, don't need to load the code or include the dependencies.


vendor/engines/ldap_authentication/lib/ldap_authentication/user_converter.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def to_user
+      ::User.new(attributes)
+    end
+
+    def attributes
+      CONVERTABLE_ATTRIBUTES.each_with_object({}) do |field, output|
+        output[field] = @user_entry.public_send(field)
+      end
+    end
```
noelrappin: explicitly make `output` the last value at the end of this for readability?

noelrappin: I also wonder if this might make more sense in the `UserEntry` class as a `to_h`, which might let you do the meta programming inside the module.

jhanggi: Are you suggesting?
```
CONVERTABLE_ATTRIBUTES.each_with_object({}) do |field, output|
  output[field] = @user_entry.public_send(field)
  output
end
```

I'll play with moving this into `UserEntry`.

vendor/engines/ldap_authentication/README.md
```diff
@@ -0,0 +1 @@
+# LdapAuthentication
```
meara: Can we link to the howto_ldap and provide some context on what this is used for? (Should the LDAP how-to be in this engine and not loose in open? I don't have strong opinions but that might make sense with encapsulation.)

vendor/engines/ldap_authentication/lib/ldap_authentication/user_extension.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    class_methods do
+      # Overrides default from devise_ldap_authenticatable
+      def find_for_ldap_authentication(attributes = {})
+        resource = super
+        resource unless resource.authenticated_locally?
+      end
+    end
+
+    # Overrides the default no-op from devise_ldap_authenticatable
```
meara: Thank you for all these helpful comments. 👍

vendor/engines/ldap_authentication/lib/ldap_authentication/user_entry.rb
```diff
@@ -0,0 +1,48 @@
+module LdapAuthentication
+
+  class UserEntry
```
meara: I like this.

vendor/engines/ldap_authentication/lib/ldap_authentication/user_converter.rb
noelrappin: Nevermind, I was confused a little. The result of `each_with_object` is the last expression, that's what's returned.

jhanggi: I started to play around with a `UserEntry#to_h`, but then `UserConverter` became almost nothing, and it felt weird to move the `::User` information into `UserEntry`. I prefer `UserEntry` just being the LDAP information and then the `UserConverter` to know about which fields from LDAP it can bring over into `::User`.

jhanggi:
Updated. I had to make some changes to how the engine decides to turn itself on and off so it can work with or without a config file in test mode. I also updated the documentation.

PR #1110, opened 2017-06-10
Move matcher_specs directory

jhanggi:
RSpec is loading everything in the support folder, including these specs. That
was causing them to always run, even if you are just trying to run a single file's
specs.

This moves it to `spec/matcher_specs` instead of `spec/support/matcher_specs`.


noelrappin:
+1 once the tests pass

meara:
This change looks completely reasonable but circle is having some kind of weird problem bundling... If we can get the tests passing, 👍

PR #1111, opened 2017-06-12
[#136306] Ability to make notes field required for product

jhanggi:
This handles the cart flow. There's a little funkiness currently with reservations.

I'm hooking into the additional validations functionality for NU Research safety since this is something we only want to check on purchase. However, if this fails, you're taken to the cart view. Did we establish this was ok behavior for research safety?

The flow for reservation purchasing does a lot of redirects. It starts from Reservations#create which does some validation checks (like the account has been set and the reservation isn't invalid) and then redirects to Orders#purchase which does the actual purchasing (this is where the research safety validation runs).

I can pull some of the validations into the Reservations#create and have it be a separate list of validators because we probably don't want the research safety check to hit twice.

I'm going to do the configurable hint/placeholder in a separate PR.


meara:
@jhanggi there is this note about instruments in cart: https://pm.tablexi.com/issues/134985#note-16

For research safety, if we fail validations on just one reservation, we are taken back to the instrument view, however: https://github.com/tablexi/nucore-nu/pull/239

meara:
Oh here's what we did for the redirect on reservations: https://github.com/tablexi/nucore-open/pull/947/files#diff-8394f3a652a395b55691e9fa1176eb28R285

jhanggi:
Thanks for that info. Looks like what I'm seeing has to do with some of the refactoring I did where I made the validators not raise an error and instead return false. I feel like this note field should be a softer error (I'd prefer it to re-render the form) than we're using research safety. I'll continue to play with it.

app/models/products/user_note_mode.rb
```diff
@@ -0,0 +1,49 @@
+module Products
+
+  class UserNoteMode
```
meara: I don't entirely grasp the raison d'être of this object. It's sort of a presenter and sort of a validator, and it's weird to me that it's in products and not in lib.  Can you explain why you made this choice?

jhanggi: It sort of came out of an experimentation with "whole value objects" (if you have a ruby tapas subscription: https://www.rubytapas.com/2016/04/14/episode-401-whole-value/), and I ended up liking this as it encapsulated the concept of a "mode" pretty well, and is a lot more versatile than . Otherwise a lot of this code would be tacked on inside of Product.

It's not really doing any validation here, just telling you what the valid modes are (I could maybe come up with with better names for the two class methods). And realy `valid_values` is kind of a private method: you should use `values` for getting a list of valid `UserNoteMode` instead of strings.

The only real presenter-y part of this is `to_label`, which is a way of getting a localized version of the value. To me, it's similar to `human_attribute_name` or `model_name`. Calling it `to_label` is to give it an API that simple_form will use before it tries `to_s`.

As far as why it's in models and not lib is to me this is application/domain logic, not really library code (I would consider library code to be things that could conceivably be pulled into a separate gem, used in an other project and still make sense). To be fair, we do have some things in `lib` that probably don't belong there (e.g. `AffiliateAccount` would probably live better in `app/models/concerns`).

Does that all make sense?

meara: This is an interesting pattern.  It still feels a little weird to me--I gotta watch that tapas ep--but the goal of keeping bloat out of Product is a worthy one.

`.valid_values` still feels a little odd though. If your intent is for it to be essentially private, I might expect that to be a constant. Can you add a comment about your intent for `.values` to be the public-facing method?

app/models/products/user_note_mode.rb
```diff
@@ -0,0 +1,49 @@
...
+      valid_values.map { |value| new(value) }
+    end
+
+    def self.valid_values
+      %w(hidden optional required).freeze
+    end
+
+    def initialize(raw_value)
+      @raw_value = raw_value
+      freeze
```
noelrappin: Possibly check against valid values here? That wouldn't seem unusual to me in a value object

app/validators/note_presence_on_purchase_validator.rb
```diff
@@ -0,0 +1,22 @@
...
+    @order = order
+  end
+
+  def valid?
+    # Use select instead of none? to make sure we loop over everything
+    invalid_orders = @order.order_details.select do |od|
+      if od.product.user_notes_field_mode.required? && od.note.blank?
+        od.errors.add(:note, :blank) # truthy return
+      end
+    end
```
noelrappin: I think this is right, but the object boundaries do feel a little odd to me. I think because the inside of the block feels conceptually like an instance method of order detail (feature envy). But I completely understand not wanting to put more code in order detail, so maybe it's a separate validator?

app/models/products/user_note_mode.rb
meara: I did do a bit of a double-take that we weren't doing that check here, although it makes sense that we would have the validation on product.

jhanggi: I wanted to do that, but then we get an exception rather than a validation error when assigning within the Product. If anyone has a good suggestion for how best to balance that, I'm open.

meara: What order does rails call these in? If we have both the validation on product and throw an exception within the UserNoteMode, will we get the validation error or the hard exception first? If the validation error happens first, is there any downside to having a fallback check within the UserNoteMode object?

app/services/order_purchase_validator.rb
```diff
@@ -0,0 +1,28 @@
+class OrderPurchaseValidator
+
+  cattr_accessor(:additional_validations) { [AllDetailsOnOrderValidator.build(NotePresenceValidator)] }
```
meara: Can we add some comments on this class and the AllDetailsOnOrder validator clarifying that one validates the order and the other validates all order details belonging to that order against the provided order detail validator(s)? Looking at this here, I have some context from your commit messages, but I think for later viewers comments would help a lot, especially around this call.

config/locales/en.yml
```diff
@@ -511,6 +511,12 @@ en:
     new:
       mark_offline: Take instrument offline
 
+  products:
+    user_notes_field_mode:
+      hidden: "No"
```
meara: Is there a reason this is quoted when the others aren't?

spec/controllers/orders_controller_spec.rb
```diff
@@ -279,7 +279,7 @@ def deliver
 
       it "updates the quantity", :aggregate_failures do
         @order_detail.reload
-        expect(flash[:notice]).to include("Quantities have changed")
+        expect(flash[:error]).to include("Quantities have changed")
```
meara: i18n?

app/models/products/user_note_mode.rb
jhanggi: It will initialize the object first. Then when you call `valid?` or `save` on the model, that will check if the inclusion list `include?` this object. I have an idea that I may play with later.

config/locales/en.yml
jhanggi: `No` is considered `false` in YAML. So it needs the quotes to make sure it's treated as a string.

app/controllers/orders_controller.rb
```diff
@@ -326,7 +314,7 @@ def index
   # Group into chunks by the group id. If group_id is nil, then it should get its
   # own chunk.
   def grouped_order_details
-    @order.order_details.order(:group_id).slice_when do |before, after|
+    @order.order_details.sort_by(&:group_id).slice_when do |before, after|
```
jhanggi: Oops. You are correct. Did not mean to commit this. This was related to a failed experiment I was trying.

jhanggi: Never mind, this is necessary. I forgot why I did this. You are correct that it takes it out of AR and into memory. But we are doing that anyways by `slice_when`. When done this way, `@order.order_details` will be memoized on `@order`. With the `order`, it returns a fresh set of objects (even though they have the same values). We run the validations on `@order.order_details`, but render the objects returned from `grouped_order_detail`. I want to render the same objects that get the validations (and therefore have `errors` on them) in order to display the validation errors in line and in red.

I'll add a comment explaining that since this is definitely non-obvious, and even I forgot why I made the change.

app/models/products/user_note_mode.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    attr_reader :raw_value
+
+    # An array of UserNoteModes representing the possible modes
+    # Use this in validations and form collections.
+    def self.all
+      VALID_MODES.map { |value| new(value) }
+    end
+
+    def self.from_string(mode_string)
```
jhanggi: Another option would be `def self.[](mode_string)`, which would use the semantics of `UserNoteMode["required"]`, which is fairly idiomatic, though rare. There's also the ["conversion function"](http://www.virtuouscode.com/2012/05/07/a-ruby-conversion-idiom/) `UserNoteMode("required")`, but I find the setup for that annoying, and I would probably only want to use that if we were supporting converting from multiple formats (e.g. string and integer).

I may still change the name of this method to be able to take a `UserNoteMode` instance to make it idempotent, but I may wait on that until I write some specs.

```
if mode.is_a?(UserNoteMode)
  mode
elsif VALID_MODES.include?(mode)
  new(mode)
else
  InvalidUserNoteMode.new(mode_string)
end
```

jhanggi:
Updated. I still need to clean up the schema conflict and write some specs, but I wanted to make sure the primary constructs are in place before @meara goes on vacation and you're on board with what I'm doing.

I also wouldn't mind another 👀 from @noelrappin since you're probably more familiar with some of the patterns I'm taking for inspiration.

For retroactive context, this is the road I was considering, but had not yet explored, when I created the [`ChargeMode`](https://github.com/tablexi/nucore-open/blob/master/app/models/charge_mode.rb) class in #1104 and why I was thinking of it more as a model class and less of a presenter.

app/models/products/user_note_mode.rb
noelrappin: I like the idempotent version, though I'm not sure that `UserNoteMode.from_string("thing")` is much better than putting the logic in the constructor and having it be `UserNoteMode.new("thing"). I might be inclined to this bit of weirdness:

```
VALID_MODES.each do |mode_string|
  define_singleton_method mode_string.to_sym do
    new(mode_string)
  end
end
```

Which gives you `UserNoteMode.hidden` and the like, but which might be too confusing to use.

app/models/products/user_note_mode.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    def ==(other)
+      case other
+      when self.class
+        raw_value == other.raw_value
+      when String
+        raw_value == other
+      else
+        false
+      end
```
noelrappin: Maybe just reduce this to `self.to_s == other.to_s`?

app/models/products/user_note_mode.rb
noelrappin: And which doesn't let you call it from an arbitrary string, so maybe it's not useful at all...

using `UserNoteMode["required"]` is also fine.

meara: I don't mind `from_string`, although the `[]` option would be okay too. Either way we go, though, I'd like it if we added a comment that this is the preferred option for the public API over `UserNoteMode.new`.

I do think this is a nice solution to the hard/soft error problem.

app/validators/all_details_on_order_validator.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  # This class should only be initialized through the `.build` method
+  def initialize(order)
+    raise "Should not be initialized directly. Use the `.build` factory method instead" unless order_detail_validator_class
+    @order = order
+  end
+
+  def valid?
+    # Use select instead of none? to make sure we loop over everything
```
meara: This comment does not make sense to me--select isn't being used? I don't actually think we need a comment here, the comments up top are sufficiently helpful.

config/locales/en.yml
meara: 👍

app/models/products/user_note_mode.rb
jhanggi: The main reason for moving it to a class method/factory rather than the initializer it to allow it to return an instance of something other than `UserNoteMode` (e.g. `InvalidUserNoteMode.`, or in an extreme version of this where each mode is actually a separate class the API doesn't need to change). 

I think I'm probably going to end up using `[]` if y'all are cool with it.

app/validators/all_details_on_order_validator.rb
jhanggi: You're right. I switched from `select` to `reject`. The reason I added the comment was to explain why we don't compact this method to just ` @order.order_details.none? do |od| ...`. That, similar `any?`, will short circuit if it finds something in the list that matches.

Do you think the comment is useful if I update it to reflect the code?

meara: Ah I see what you're saying. I would then amend the comment to include why we want to make sure to loop over all the validators, i.e. we want to collect all the errors.

jhanggi:
This should be ready for a final review.

app/controllers/orders_controller.rb
```diff
@@ -326,7 +314,10 @@ def index
   # Group into chunks by the group id. If group_id is nil, then it should get its
   # own chunk.
   def grouped_order_details
-    @order.order_details.order(:group_id).slice_when do |before, after|
+    # Do not do any additional scopes on `@order.order_details` to ensure the list
+    # of OrderDetail objects is memoized. Validations are run on `@order.order_details`
+    # and we want to be able to render validation errors inline.
+    @order.order_details.sort_by(&:group_id).slice_when do |before, after|
       before.group_id.nil? || before.group_id != after.group_id
```
noelrappin: I wonder as I look at this again if the comment indicates that the code wants to be factored a little differently, maybe pulling `@order.order_details` into a separate load method so as to force a memoized version?

app/models/products/user_note_mode.rb
```diff
@@ -0,0 +1,74 @@
...
+
+    # Prefer the use of the factory `UserNoteMode[raw_value]` over directly
+    # initializing this class with `new` as it will not hard error on invalid
+    # inputs.
+    def initialize(raw_value)
+      @raw_value = raw_value
+      raise ArgumentError, "Invalid value: #{raw_value}" unless VALID_MODES.include?(raw_value)
+      freeze
+    end
+
```
noelrappin: The comment doesn't match the code? It does hard error on invalid inputs. Also maybe

```
def self.valid?(mode)
  VALID_MODES.include?(mode)
end
```

app/validators/all_details_on_order_validator.rb
```diff
@@ -0,0 +1,39 @@
...
+    Class.new(AllDetailsOnOrderValidator) do
+      self.order_detail_validator_class = order_detail_validator_class
+    end
+  end
+
+  # This class should only be initialized through the `.build` method
+  def initialize(order)
+    raise "Should not be initialized directly. Use the `.build` factory method instead" unless order_detail_validator_class
+    @order = order
+  end
```
noelrappin: I think I'm not completely clear as to why the `order_detail_validator_class` is a class attribute and not an instance attribute?

app/validators/all_details_on_order_validator.rb
noelrappin: I'm with Meara here.

app/models/products/user_note_mode.rb
jhanggi: Took me a second to get what you were saying. I read (because I wrote it) with the factory being the antecedent of "it". Clearly, it's unclear. I'll try to reword it.

app/controllers/orders_controller.rb
jhanggi: The issue has more to do with the API of the validators. They take an Order and drop errors on the association's entries. We could change that API to take an array of orderdetails, but that means we'd need to make additional changes downstream in NU. Also, the validator in NU (for research safety) needs access to the order (for `ordered_on_behalf_of?`.

Not quite sure how to make it better.

app/validators/all_details_on_order_validator.rb
jhanggi: The way these validators get used is there's an array of classes, and we loop over each class and call `clazz.new(order).valid?`. `AllDetailsOnOrderValidator` is essentially an abstract class. The `build` method is used to create a new subclass of `AllDetailsOnOrderValidator` that has a reference to order detail validator class so it can use it internally later. This class doesn't get instantiated for a while.

I originally wanted to add a `def order_detail_validator_class` inside the new subclass, but I couldn't get it to reference the passed in `order_detail_validator_class`.

The other option I considered was varying levels of generic vs specific. One step along the way was something like this. That seemed something where generalization into a superclass felt right (though maybe premature).

```ruby
class AllDetailsNotePresenceValidator
  ...
  def valid?
    # Loop over everything so that all order details get `errors` applied to them
    invalid_orders = @order.order_details.reject do |od|
      NotePresenceValidator.new(od).valid?
    end

    invalid_orders.none?
  end
```

Does that make sense?

app/controllers/orders_controller.rb
noelrappin: Okay, leave it as-is

app/validators/all_details_on_order_validator.rb
noelrappin: Yeah, I've got what it's doing now. I think there's a non-metaprongramming version that doesn't create a new subclass, but I'm not sure it's worth digging for.

app/services/order_purchaser.rb
```diff
@@ -3,32 +3,55 @@ class OrderPurchaser
   attr_reader :acting_as, :order, :order_in_past, :params, :user
   attr_accessor :backdate_to
 
-  cattr_accessor(:additional_validations) { [] }
+  # TODO: remove this once all child instances have been updated
+  def self.additional_validations
+    warn "Use OrderPurchaseValidator.additional_validations instead of OrderPurchaser at #{caller(1..1)}"
```
meara: This is a very helpful warning, I like this.

db/migrate/20170609214859_optional_required_notes_field.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def up
+    add_column :products, :user_notes_field_mode, :string, null: false, default: "hidden"
+
+    Product.find_each do |product|
+      mode = product.note_available_to_users? ? "optional" : "hidden"
+      product.update_column(:user_notes_field_mode, mode)
+    end
+
+    remove_column :products, :note_available_to_users
```
meara: Separate migrations for the drop column might be overkill in this case, but do we want this in a transaction block? So it'll rollback/not drop the column if the data migration somehow fails?

jhanggi: In this case, I'm using `update_column`, so the only way it should fail would be something catastrophic (i.e. db connection being lost). And if it does fail, we're going to end up in a weird state.

I don't think you can put DDL statements into a transaction on mysql. One thing I miss about postgres is that it automatically wraps the migrations in a transaction, so if anything fails it can roll back back schema changes.

meara: Oh mysql. Okay. I'm fine with this then.

PR #1113, opened 2017-06-13
[#136649] Users unable to tap facilities under "Manage Facilities" on phones

noelrappin:
https://pm.tablexi.com/issues/136649

See https://stackoverflow.com/questions/17178606/bootstrap-v2-dropdown-navigation-not-working-on-mobile-browsers for a description, but this is a bootstrap bug, and short of manipulating the bootsrap source, this seems to be the fix.

### DIFF
--- a/app/assets/javascripts/app/responsive_dropdown.js.coffee
+++ b/app/assets/javascripts/app/responsive_dropdown.js.coffee
@@ -5,6 +5,6 @@
 #
 
 $ ->
-  $("li.dropdown a").click (e) ->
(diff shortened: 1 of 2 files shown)


jhanggi:
There's some coffee linting to be done, but looks good. I noticed the dropdown has some ugly hover styles, but we can address that in the other ticket.

jhanggi:
Also, we'll want to fast-track this into dartmouth prod since it's currently causing problems for some of their users.

PR #1114, opened 2017-06-14
[#136662] Bundles showing "no current price policies"

jhanggi:
screen shot 2017-06-13 at 1 45 51 pm

Show "Product missing current price policies" if one of the bundled products is missing current policies. Do not show anything if all the products have active policies. Show "No current price policies" on the bundle's product list.

screen shot 2017-06-13 at 7 41 07 pm
screen shot 2017-06-13 at 7 40 57 pm

Addresses NU's #134312

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -1,16 +1,10 @@
 module ProductsHelper
 
   def price_policy_errors(product)
-    @facility_price_groups ||= current_facility.price_groups.count
-    price_policy_count = product.price_policies.current.count
-    error_msg = ""
-    if price_policy_count == 0
+    if product.is_a?(Bundle)
+      error_msg = t("price_policies.errors.missing_for_bundle") if product.products_missing_price_policies.any?
+    elsif product.current_price_policies.none?
       error_msg = t("price_policies.errors.none_exist")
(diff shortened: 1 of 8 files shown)


app/helpers/products_helper.rb
```diff
@@ -1,16 +1,10 @@
 module ProductsHelper
 
   def price_policy_errors(product)
-    @facility_price_groups ||= current_facility.price_groups.count
-    price_policy_count = product.price_policies.current.count
-    error_msg = ""
-    if price_policy_count == 0
+    if product.is_a?(Bundle)
```
noelrappin: I don't know if it'd be clearer, but the `is_a?` makes me wonder if you might be better off implementing `missing_price_policy?` (or `has_price_policy?`) on `Product` and `Bundle`, though you'd still have to switch on the error message I suppose.

jhanggi: I struggled a bit with that. I initially considered overriding `Bundles#price_policies` with `has_many :price_policies, through: :products`, but was scared that would have unintended side effects. Ultimately, I was worried about future confusion, and knew that I would want to switch here anyways to display different messages.

meara: I'm fine with this as is.

PR #1118, opened 2017-06-14
[#134990] facility dash redirect

meara:
- Refactor logic for the "default" facility admin homepage path to allow it to be extended/overridden by school-specific engines
- https://pm.tablexi.com/issues/134990
- Accompanies https://github.com/tablexi/nucore-uic/pull/130

### DIFF
--- a/lib/ability.rb
+++ b/lib/ability.rb
@@ -148,7 +148,7 @@ class Ability
 
         can :index, User if controller.is_a?(FacilityUserReservationsController)
 
-        can [:list, :show], Facility
+        can [:list, :dashboard, :show], Facility
         can :act_as, Facility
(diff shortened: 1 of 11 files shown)


meara:
@jhanggi updated

jhanggi:
Well, looks like we still have a couple test failures looking for old behavior.

jhanggi:
👍

lib/ability.rb
```diff
@@ -148,7 +148,7 @@ def initialize(user, resource, controller = nil)
 
         can :index, User if controller.is_a?(FacilityUserReservationsController)
 
-        can [:list, :show], Facility
+        can [:list, :dashboard, :show], Facility
```
meara: @jhanggi, I just wanted to double check--I added this to the ability file to resolve the sanger spec failures. This permission is correctly located, right? It shouldn't be overreaching anywhere?

jhanggi: Just curious, without this, what happens if you're a lowly facility staff and you try to get in to the manage facility? I would expect it to fail. Maybe we should add something in a feature spec?

PR #1119, opened 2017-06-14
[#136649] Another attempt to fix bootstrap dropdowns on touchscreen

jhanggi:
Previous attempts in #1117 and #1113.

We should test this out in android and older versions of iOS. So far I have only tested on iOS 10.


app/assets/stylesheets/nucore.scss
```diff
@@ -354,3 +354,12 @@ textarea.wide {
...
 }
+
+/* Bootstrap adds an element with this class when you expand a dropdown and your
+ * browser supports the ontouchstart event. It claims to be necessary "if mobile
+ * we we use a backdrop because click events don't delegate". Instead, this
+ * is taking the tap and preventing the actual link to receive the click.
+ */
+.dropdown-backdrop {
+  display: none;
+}
```
noelrappin: Do you have a reference?

jhanggi: I found this by using the web inspector with the simulator. This explains why they do it, so maybe this is not right solution: https://v4-alpha.getbootstrap.com/components/dropdowns/#usage

jhanggi: I also found this: https://stackoverflow.com/questions/17579750/drop-down-menu-not-working-on-mobile-devices

Initial testing, using `position: static;` seems to work as well. Both do seem to close the dropdown if you tap elsewhere on the page.

jhanggi:
Tested in iOS 9.3 as well as 10.3. Anyone have an android device or simulator installed to test there?

noelrappin:
+1 if testing is good.

PR #1123, opened 2017-06-16
[#136702] Prevent invalid quantities from cart into Sanger form

jhanggi:
In the cart, if the user changes the quantity field to be blank or negative and then clicks the "Complete Order Form" we end up passing a blank string or negative number in the quantity query parameter to the Sanger form. This was then triggering an `ArgumentError: quantity must be positive` in `Submission#create_samples!`.

I also tested with decimal numbers. Those were being passed as decimals, but the form was handling that by dropping the fractional part.

This now handles it at both the cart level and the create_samples level. The cart will disable the "Complete Order Form" button if the quantity is not a valid value. If an invalid value does slip through, then the Sanger form will default to having one sample.

### DIFF
--- a/app/assets/javascripts/app/cart.coffee
+++ b/app/assets/javascripts/app/cart.coffee
@@ -4,16 +4,28 @@ class window.CartQuantityReplacer
   toString: ->
     @originalPath.replace(/\bquantity=\d+(&|$)/, "quantity=#{@newQuantity}$1")
 
+class window.Cart
+  constructor: (cartSelector) ->
+    @$cart = $(cartSelector)
+    @$cart.find("[data-quantity-field]").each @setupQuantityListeners
+
+  setupQuantityListeners: (_i, link) ->
+    $link = $(link)
+    $quantityField = $("#" + $(link).data("quantity-field"))
+    originalHref = $link.attr("href")
+
+    $quantityField.change (e) ->
+      newQuantity = parseInt($quantityField.val())
+
+      if newQuantity > 0
+        $link
+          .removeClass("disabled")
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/app/cart.coffee
```diff
@@ -4,16 +4,28 @@ class window.CartQuantityReplacer
...
+      if newQuantity > 0
+        $link
+          .removeClass("disabled")
+          .attr("href", new CartQuantityReplacer(originalHref, newQuantity))
 
+      else
+        $link
+          .addClass("disabled")
+          .removeAttr("href")
+
```
noelrappin: Probably this isn't better, but:

`$link.toggleClass("disabled", newQuantity > 0)` prevents weird state problems, but you'd still have to deal with the attribute...

jhanggi: I get there could be something adding/removing the class outside of this, but how would using `toggleClass` be different from how this behaves? Or is there some funky behavior if you `addClass` multiple times?

noelrappin: In theory, there's no difference, I feel like jQuery occasionally behaves weirdly if you add/remove a bunch. But that might just be superstition.

jhanggi: I'm not sure. I feel like this is clearer than two things switching on the same condition. I'd be a little more comfortable using `toggleClass` if I could also address the attribute in a parallel structure with something like `.attr("href", quantity > 0 ? href : "")`, but an `<a>` without an `href` is different than an `<a href="">`.

PR #1126, opened 2017-06-19
[#136682] Lost tab when on Bulk Email or Payment Sources tab

jhanggi:
When under the main "Users" tab, if you navigate to the Bulk Email tab or Payment Sources, the "Users" tab in the sidebar goes away. That's because it is checking for `:index, User` on the ability, but the admin is only granted `:administer`.

screen shot 2017-06-19 at 3 12 04 pm

See for reference:
* Changed from using `:administer` to `:index` in `_sidenav_users.html.haml` allowing Business Admins to view the tab in cross facility mode. https://github.com/tablexi/nucore-open/commit/2b73d4fb90ff0ef27f192964ac1fb5c0d18ac99f 
* Added the line around `FacilityUserReservationsController` that I am removing here #961

### DIFF
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -91,7 +91,7 @@ class NavTab::LinkCollection
   end
 
   def admin_users
-    if single_facility? && ability.can?(:index, User)
+    if single_facility? && ability.can?(:administer, User)
       NavTab::Link.new(tab: :admin_users, url: facility_users_path(facility))
     end
   end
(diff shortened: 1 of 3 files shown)


jhanggi:
Well, looks like I might not be able to get rid of the `administer` role after all. There's the `UsersController#index`, which is the main listing of customers with the ability to search (that's the one we're dealing with here), but there's also `FacilityUsersController`, which is the administrative controller for dealing with staff and roles within the facility. They both check `can? :index, User`. I need to rethink this.

jhanggi:
Yeah, it's weird. Especially where we add some specific rules based on what controller you're in. That gets confused when you're in the same section of the site, but a different controller, but we want to conditionally display a link based on if you have access to that resource.

PR #1128, opened 2017-06-19
[#136739] Configurable user note field label

jhanggi:
Builds on #1111

* Adds an optional configurable label that will be displayed on the Note field to the user.
* Will only be displayed if the notes field is set to optional or required
* This also adds the required asterisk if the notes field is required
* If the label is set as blank, then it will fall back to "Note"

I still need to add a few tests, but this should generally be ready for a look.

screen shot 2017-06-19 at 6 21 52 pm

screen shot 2017-06-19 at 6 24 29 pm

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -289,16 +289,16 @@ ActiveRecord::Schema.define(version: 20170619213358) do
   add_index "order_details", ["account_id"], name: "fk_od_accounts", using: :btree
   add_index "order_details", ["assigned_user_id"], name: "index_order_details_on_assigned_user_id", using: :btree
   add_index "order_details", ["bundle_product_id"], name: "fk_bundle_prod_id", using: :btree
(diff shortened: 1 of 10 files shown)


db/schema.rb
```diff
@@ -11,7 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170609214859) do
+ActiveRecord::Schema.define(version: 20170619213358) do
```
noelrappin: what's with all the name change in the schema file?

jhanggi: I'll clean up the DB renames. The open schema should be cleaned up so it matches a db:migrate from scratch, but I often run off of an older database. I should probably figure out if I can copy all my data over to a fresh schema...

Can you also take another look over #1111 as this depends on that.

noelrappin:
👍

jhanggi:
Rebased off of master, so could use a quick 👁 👁

PR #1131, opened 2017-06-20
Disable UnnecessaryStringOutput haml linter

jhanggi:
The linter prefers `%p Note: #{order_detail.note}` to `%p= "Note:
#{order_detail.note}”`. We usually go with the latter where the string
interpolation is explicit.

I'm open to debate on this. This can probably wait until Meara is back so she can weigh in on it.

### DIFF
--- a/.haml-lint.yml
+++ b/.haml-lint.yml
@@ -3,3 +3,6 @@ inherits_from: .haml-lint_todo.yml
 linters:
   LineLength:
     enabled: false
+
+  UnnecessaryStringOutput:
+    enabled: false


noelrappin:
I kind of prefer the former, but don't really have strong feelings about it either way.

meara:
I would probably go with the explicit string interpolation option, but I don't have strong feelings about it.

With style things, I tend to bias towards the more obvious formatting, just because I think it helps make codebases friendlier for newer devs.

jhanggi:
I prefer it with `=` and the quotes mainly because my editor syntax highlights it as a string. But that's just me.

noelrappin:
I'm fine with either

PR #1132, opened 2017-06-21
[#135577] Update RSpec metadata settings.

noelrappin:
Per https://github.com/tablexi/nucore-dartmouth/pull/104 some tests were
dependent on the `user_based_price_groups` setting, while I was there,
cleaned up some settings changes to use metadata

### DIFF
--- a/spec/models/product_spec.rb
+++ b/spec/models/product_spec.rb
@@ -209,7 +209,7 @@ RSpec.describe Product do
       end
     end
 
-    context "can_purchase?", feature_setting: { user_based_price_groups: true } do
+    context "can_purchase?" do
       class TestPricePolicy < PricePolicy
 
         def rate_field
@@ -222,11 +222,7 @@ RSpec.describe Product do
         @product = TestProduct.create!(facility: @facility, name: "Test Product", url_name: "test")
         @price_group = FactoryGirl.create(:price_group, facility: @facility)
         @price_group2 = FactoryGirl.create(:price_group, facility: @facility)
-        @user = FactoryGirl.create(:user)
-        FactoryGirl.create(:user_price_group_member, user: @user, price_group: @price_group)
-        @user.reload
-
-        @user_price_group_ids = @user.price_groups.map(&:id)
+        @user_price_group_ids = [@price_group]
       end
 
       it "should not be purchasable if it is archived" do
@@ -339,6 +335,7 @@ RSpec.describe Product do
                                                         can_purchase: true)
         expect(@product).not_to be_can_purchase(@user_price_group_ids)
       end
+
       it "should be purchasable if there are no current policies, but two future policies, one of which is purchasable and one is not" do
         expect(@product.current_price_policies).to be_empty
         @price_policy_pg1 = TestPricePolicy.create!(price_group: @price_group,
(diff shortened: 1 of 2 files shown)


spec/models/product_spec.rb
```diff
@@ -232,7 +209,7 @@ def requires_account?
       end
     end
 
-    context "can_purchase?" do
+    context "can_purchase?", feature_setting: { user_based_price_groups: true } do
```
jhanggi: I have a thought on this one since most of the tests in this context still pass even if this is set to `false`.

The `can_purchase?` method doesn't really care about a user's price groups, it just cares about an array of price group ids. What if we ditched the `@user_price_group_ids` variable altogether and use what its value would be if the feature were enabled, `[@price_group.id]`? There are a few places where we create new `PriceGroupMember`s. We don't really need to create those, we could just add its price group to the array.

Thoughts?

noelrappin: Probably something for @meara to weigh in on, she was the person who touched this last.

spec/models/product_spec.rb
```diff
@@ -245,12 +222,9 @@ def rate_field
         @product = TestProduct.create!(facility: @facility, name: "Test Product", url_name: "test")
         @price_group = FactoryGirl.create(:price_group, facility: @facility)
         @price_group2 = FactoryGirl.create(:price_group, facility: @facility)
-        @user = FactoryGirl.create(:user)
-        FactoryGirl.create(:user_price_group_member, user: @user, price_group: @price_group)
-        @user.reload
-
-        @user_price_group_ids = @user.price_groups.map(&:id)
+        @user_price_group_ids = [@price_group]
```
meara: These should just be `@price_group_ids` now, right? Did we already make that change and it just didn't get pushed up?

jhanggi: I don't think we changed the name of the variable, although it might make sense to do so. If we are thinking of doing that, I would vote for getting rid of the instance variable altogether and either use a local variable or just put it inline (`expect(@product).to be_can_purchase([@price_group])`. That would put the expectations a little closer to where they're used. I don't feel too strongly about it, as I'd like to get this wrapped up soon.

spec/models/user_spec.rb
```diff
@@ -40,34 +40,38 @@
...
-  it "belongs to price groups of accounts" do
-    cc = create(:nufs_account, account_users_attributes: account_users_attributes_hash(user: user))
-    pg = FactoryGirl.create(:price_group, facility: facility)
-    AccountPriceGroupMember.create(account: cc, price_group: pg)
-    expect(user.account_price_groups.include?(pg)).to be true
-  end
+    it "is a member of any explicitly mapped price groups" do
+      pg = FactoryGirl.create(:price_group, facility: facility)
+      UserPriceGroupMember.create(user: user, price_group: pg)
+      expect(user.price_groups.include?(pg)).to eq(true)
```
jhanggi: I don't have a strong feeling which way we go, but everywhere else we're using `be true`. We should probably be consistent. Or maybe better change all these to `expect(user.price_groups).to include(pg)`. But again, I'm more interested in getting this wrapped up than spending a ton of time on cleanup.

spec/models/product_spec.rb
meara: I would kinda prefer not to have a variable called @user_price_group_ids since these in fact have nothing to do with users now. For future readability's sake.

noelrappin:
I just changed the variable name, commit now?

jhanggi:
Feel free to merge.

meara:
👍

PR #1133, opened 2017-06-21
[#125709] Show post-complete badges for account admins

jhanggi:
Adds blue status badges to a couple views under "My Payment Sources" so that account managers and PIs can see the status beyond "complete".

* "In Review"
* "Open Journal"
* "Can Reconcile"
* "In Open Journal"

These badges were already being used on the facility's Order#show view. They are now available on the "Transaction History" and OrderDetail#show views.
transaction history
order detail show

Since the transaction history table partial is shared with the billing tab, these badges are now available there as well.
billing tab

In order to DRY some things up, this changes the problem column of the problem orders tabs into badges.

problem orders

I currently have some questions out to the client confirming this behavior is all correct and acceptable.

### DIFF
--- a/app/presenters/order_detail_notice_presenter.rb
+++ b/app/presenters/order_detail_notice_presenter.rb
@@ -1,0 +1,61 @@
+class OrderDetailNoticePresenter < DelegateClass(OrderDetail)
+
+  include ActionView::Helpers::OutputSafetyHelper
+
+  def statuses
+    [].tap do |statuses|
+      statuses << Notice.new(:in_review) if in_review?
+      statuses << Notice.new(:in_dispute) if in_dispute?
+      statuses << Notice.new(:can_reconcile) if can_reconcile_journaled?
+      statuses << Notice.new(:in_open_journal) if in_open_journal?
+    end
+  end
+
+  def warnings
+    [].tap do |warnings|
+      warnings << Notice.new(problem_description_key, :warning) if problem?
+    end
+  end
+
+  def notices
+    statuses + warnings
+  end
+
+  def badges_to_html
+    safe_join(notices.map(&:badge_to_html))
+  end
+
+  # Not meant to be used outside the presenter class
+  class Notice
+
+    include ActionView::Helpers::TagHelper
+
+    def initialize(key, severity = :status)
+      @key = key
+      @severity = severity
+    end
+
+    def badge_text
+      I18n.t("order_details.notices.#{@key}.badge")
+    end
+
+    def alert_text
+      I18n.t("order_details.notices.#{@key}.alert")
+    end
+
+    def badge_to_html
+      content_tag(:span, badge_text, class: ["label", label_class])
+    end
+
+    private
+
+    def label_class
+      {
+        status: "label-info",
+        warning: "label-important",
+      }.fetch(@severity)
+    end
+
+  end
+
+end
--- a/app/presenters/order_detail_notice_presenter.rb
+++ b/app/presenters/order_detail_notice_presenter.rb
@@ -22,8 +22,11 @@ class OrderDetailNoticePresenter < DelegateClass(OrderDetail)
     statuses + warnings
   end
 
-  def badges_to_html
-    safe_join(notices.map(&:badge_to_html))
+  # Filter to only status or warnings by passing `only: :status` or `only: :warning`
+  def badges_to_html(only: [:status, :warning])
+    filtered = notices.select { |notice| Array(only).include?(notice.severity) }
+
+    safe_join(filtered.map(&:badge_to_html))
   end
 
   def alerts_to_html
@@ -48,6 +51,8 @@ class OrderDetailNoticePresenter < DelegateClass(OrderDetail)
   class Notice
 
     include ActionView::Helpers::TagHelper
+
+    attr_reader :severity
 
     def initialize(key, severity = :status)
       @key = key
--- a/app/presenters/order_detail_notice_presenter.rb
+++ b/app/presenters/order_detail_notice_presenter.rb
@@ -1,5 +1,6 @@
 class OrderDetailNoticePresenter < DelegateClass(OrderDetail)
 
+  include ActionView::Helpers::TagHelper
   include ActionView::Helpers::OutputSafetyHelper
 
   def statuses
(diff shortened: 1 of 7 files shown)


app/presenters/order_detail_notice_presenter.rb
```diff
@@ -0,0 +1,85 @@
+class OrderDetailNoticePresenter < DelegateClass(OrderDetail)
+
```
noelrappin: Is there a functional reason why `DelegateClass` is better than `SimpleDelegator`?

app/presenters/order_detail_notice_presenter.rb
```diff
@@ -0,0 +1,85 @@
...
+  def statuses
+    [].tap do |statuses|
+      statuses << Notice.new(:in_review) if in_review?
+      statuses << Notice.new(:in_dispute) if in_dispute?
+      statuses << Notice.new(:can_reconcile) if can_reconcile_journaled?
+      statuses << Notice.new(:in_open_journal) if in_open_journal?
+    end
+  end
+
```
noelrappin: I think that I find `tap` hard to read relative to creating a local and returning the local. 

Hmm... what about.. (I may have the splats wrong here...)
```
def build_a_notice(criterion, *args)
  Notice.new(*args) if public_send(criterion)
end

def build_notices(*notice_args)
  notice_args.map { |criterion, args| build_a_notice(criterion, args) }.compact
end

def statuses 
  build_notices([:in_review?, :in_review], [:in_dispute?, :in_dispute], [:can_reconcile_journaled?, :can_reconcile], [:in_open_journal?, :in_open_journal])
end

def warnings 
  build_notices(:problem, problem_description_key, :warning)
end
```

The build methods could be class methods of `Notice`, too...

noelrappin: Too dry?

noelrappin: I really like the use of the presenter, though.

app/presenters/order_detail_notice_presenter.rb
jhanggi: None that are super compelling. Slight performance improvement, and better introspection, `OrderDetailNoticePresenter.instance_methods` includes the OrderDetail methods when it's `DelegateClass`, but doesn't know about them if it's `SimpleDelegator`.

I'd be happy to switch if you think there's a good reason to prefer `SimpleDelegator`.

app/presenters/order_detail_notice_presenter.rb
jhanggi: I feel like that's almost too clever. I find that pretty hard to reason through, especially with the splats. Maybe not as elegant, I think the current way makes it pretty obvious what's happening without having to dive down the rabbit hole of the helper methods.

I personally like the `tap` version. The local variable version looks like a wall. Line breaks definitely make it better, but I kind of prefer the indentation. I'd be happy to switch if you definitely prefer it.

```
    statuses = []
    statuses << Notice.new(:in_review) if in_review?
    statuses << Notice.new(:in_dispute) if in_dispute?
    statuses << Notice.new(:can_reconcile) if can_reconcile_journaled?
    statuses << Notice.new(:in_open_journal) if in_open_journal?
    statuses
```

app/presenters/order_detail_notice_presenter.rb
noelrappin: Nah, it's fine as is...

app/presenters/order_detail_notice_presenter.rb
noelrappin: I mildly prefer the local variable version to the tap version, but if you don't have a problem reading the tap version, don't change it on my account.

app/presenters/order_detail_notice_presenter.rb
meara: I'm not super familiar with how/why DelegateClass (or SimpleDelegator, for that matter) would be used. We're essentially adding additional function onto the order detail object with OrderDetailNoticePresenter as a sort of wrapper class? What is the advantage of doing this delegation rather than having an entirely separate presenter class?

app/presenters/order_detail_notice_presenter.rb
meara: I kind of hate `tap`, to be honest. (In general, not just here--which is a good case to use it if  ever there was one.)  It's just more cognitive effort to figure out exactly what is happening/what is being returned in this method with `tap` vs the local variable version, which makes it super obvious what the return value on the method is.

`tap` isn't a deal-breaker for me though.

app/presenters/order_detail_notice_presenter.rb
```diff
@@ -0,0 +1,85 @@
...
+  def warnings
+    [].tap do |warnings|
+      warnings << Notice.new(problem_description_key, :warning) if problem?
```
meara: Couldn't this just be

```ruby
[Notice.new(problem_description_key, :warning)] if problem?
```

There's only ever going to be one `Notice` object in the array, right?

I think the `tap` weirds me out here and above because my brain subconsciously expects us to be iterating over something, because the pattern looks too much like an `each` or a `map` kind of operation.

app/presenters/order_detail_notice_presenter.rb
```diff
@@ -0,0 +1,85 @@
...
+    end
+  end
+
+  def notices
+    statuses + warnings
+  end
+
+  # Filter to only status or warnings by passing `only: :status` or `only: :warning`
+  def badges_to_html(only: [:status, :warning])
```
meara: I like this, very elegant.

The `filtered` line isn't going to be a performance hit when we want both anyway, though?

app/presenters/order_detail_notice_presenter.rb
jhanggi: The main advantage from my perspective is brevity. We can use method calls with an implicit `self` within the class. So we can just call `in_review?` rather than `order_detail.in_review?`.

app/presenters/order_detail_notice_presenter.rb
jhanggi: I can change it. I think it's just something that I've gotten used to, so it doesn't take more effort, but I can see how it does if you're not used to it, and I think that's a fair reason to change it.

app/presenters/order_detail_notice_presenter.rb
jhanggi: Yeah, there's only ever going to be one here. The reason I did it this way was to keep the shape of the `warnings` and `notices` the same. I think it's easier to see they're doing similar logic when they're parallel.

app/presenters/order_detail_notice_presenter.rb
jhanggi: I don't think there'll be a significant performance hit. I'd actually be more worried about doing extra logic when we only want one (we do all the logic, but filter it), but these checks are all pretty snappy.

app/presenters/order_detail_notice_presenter.rb
meara: Ah okay, makes sense.

PR #1134, opened 2017-06-22
[#136792] Consider 2-digit years as an invalid format

jhanggi:
Force to always use 4 digit years. Before, “3/4/17” was being treated
as  the date “0003-4-17” and coming through as valid causing some interesting database values.

Since the method returns nil for invalid dates, this could still cause
unexpected behavior. For example, you can backdate an order, but if the
input is blank or invalid, it defaults to `Time.current`. This could
result in setting the ordered_at to now rather than what the user
expected to see.

On the front-end side, this highlights the form element in red if the date format is invalid
(only happens if the user manually adjusts the date w/o using the widget).

I considered doing something like disabling the form, but that started
to get weird with also wanting to disable the submit buttons when
thinking about multiple things possibly enabling/disabling.

I’m hoping just highlighting the problem will dissuade users from
hitting submit.


app/helpers/date_helper.rb
```diff
@@ -14,11 +16,22 @@ def parse_usa_import_date(date_string)
   end
 
   def parse_usa_date(date, time_string = nil)
-    date_string = (date =~ /\d{1,2}\/\d{1,2}\/\d{4}/ ? Date.strptime($&, "%m/%d/%Y") : date).to_s
-    date_string += " #{time_string}" if time_string
+    date = date.to_s.strip
+
+    # TODO: Many tests pass either a Date, a Time, or an YYYY-MM-DD formatted
```
jhanggi: There were a lot of ways I tried engineering this ranging from nested conditionals to case statements, you name it. I liked this the best because there's only one line that's "wrong" and coerces incorrect inputs into what we want. If we clean up the specs, then just the one line needs to go.

app/assets/javascripts/app/datepicker-validate.coffee
```diff
@@ -0,0 +1,27 @@
+$ ->
+  # Parse to a date if it's a string, otherwise just return the original argument.
+  # This will handle "5/6/2016", unix timestamps, and `Date`s.
+  parseDate = (dateOrString, format) ->
+    if typeof dateOrString is "string"
+      $.datepicker.parseDate(format, dateOrString)
+    else
+      dateOrString
```
noelrappin: What if the input isn't a date either? Or is that not likely to be an issue?

app/assets/javascripts/app/datepicker-validate.coffee
```diff
@@ -0,0 +1,27 @@
...
+      date = parseDate($input.val(), format)
+
+      if maxDate && date > maxDate
+        throw new Error("Beyond maxDate")
+      if minDate && date < minDate
+        throw new Error("Before minDate")
+
+      $input.closest(".control-group").removeClass("error")
+    catch e
+      $input.closest(".control-group").addClass("error")
```
noelrappin: I would pull some or all of this into its own method or class, but then I'm generally kind of an outlier in JS code structure (I normally like event handlers to call regular functions or methods).

app/helpers/date_helper.rb
```diff
@@ -14,11 +16,22 @@ def parse_usa_import_date(date_string)
   end
 
   def parse_usa_date(date, time_string = nil)
-    date_string = (date =~ /\d{1,2}\/\d{1,2}\/\d{4}/ ? Date.strptime($&, "%m/%d/%Y") : date).to_s
-    date_string += " #{time_string}" if time_string
+    date = date.to_s.strip
```
noelrappin: Maybe comment with expected input format?

app/views/orders/_edit_date.html.haml
```diff
@@ -1,18 +1,27 @@
...
+    = f.input :order_status_id,
+      collection: @order_statuses,
+      label_method: :name_with_level,
+      selected:  params[:order_status_id] || @order_statuses.first.try(:id),
+      label: OrderDetail.human_attribute_name(:order_status)
 
-    = f.input :send_notification, :as => :boolean, :input_html => { :value => 1 }, :default => params[:send_notification], :label => false, :inline_label => true
-
-  :javascript
-    $(function() { $('.datepicker').datepicker({maxDate: '#{format_usa_date(Time.zone.now)}'}); });
```
noelrappin: Sorry, this is probably a silly question, but I see that the date picker class changed, is `datepicker__data` converted to an actual date picker someplace else?

spec/helpers/date_helper_spec.rb
```diff
@@ -41,32 +41,56 @@
...
+      end
+
+      it "returns nil  with invalid time" do
+        expect(parse_usa_date("5/10/2012", "34:23 FM")).to be_nil
+      end
+
+      it "returns nil for valid time, but nil input" do
+        expect(parse_usa_date(nil, "3:17 PM")).to be_nil
+      end
+    end
```
noelrappin: Should there be a specific test for the `date =~ /\A\d{4}-\d{2}-\d{2}/` condition or am I missing it?

app/assets/javascripts/app/datepicker-validate.coffee
noelrappin: Is this worth a spec?

jhanggi: Yeah, that's probably a good call.

app/helpers/date_helper.rb
jhanggi: Good call.

app/views/orders/_edit_date.html.haml
jhanggi: Our use of jquery-ui's datepicker is a little weird. We only call `$(".datepicker").datepicker()` on the views that need it. I believe we did this so we could control `maxDate` and `minDate` on a per-page level.

A while back, I decided that using data attributes would be a good solution so we could call `.datepicker()` from the main `application.js`. I went with `.datepicker__data` to avoid conflicts and to imply that it's meant to be used with data attributes. Not sure it's the perfect convention, but it works.

See `app/assets/javascripts/app/datepicker-data.coffee`. We load the whole `javascripts/app` tree into `application.js`.

spec/helpers/date_helper_spec.rb
jhanggi: Yeah, that's probably a good idea.

app/assets/javascripts/app/datepicker-data.coffee
```diff
@@ -3,12 +3,16 @@ class window.DatePickerData
...
     # Make sure you call `to_s` or `iso8601` on any dates you set in the views
     # to ensure the proper format.
-    $picker = $(".datepicker__data")
+    $pickers = $(".datepicker__data")
 
-    $picker.datepicker(
-      minDate: new Date($picker.data("min-date")) # will be unbounded if not provided
-      maxDate: new Date($picker.data("max-date")) # will be unbounded if not provided
-    )
+    $pickers.each (i, picker) ->
```
jhanggi: This change was to allow supporting multiple datepickers on a page with different min/max values.

app/assets/javascripts/app/datepicker-data.coffee
```diff
@@ -3,12 +3,16 @@ class window.DatePickerData
...
+      $picker = $(picker)
+      $picker.datepicker(
+        minDate: new Date($picker.data("min-date")) # will be unbounded if not provided
+        maxDate: new Date($picker.data("max-date")) # will be unbounded if not provided
+      )
+
+    new DatePickerValidate($pickers).activate()
```
jhanggi: This felt like a reasonable initialization process. I could possibly be convinced to squash these two classes together.

noelrappin: Seems fine to me

PR #1136, opened 2017-06-23
[#98769] Add journal reference to order detail popup

jhanggi:
screen shot 2017-06-23 at 3 43 57 pm

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -74,6 +74,8 @@ class OrderDetail < ActiveRecord::Base
   end
 
   delegate :journal_date, to: :journal, allow_nil: true
+  delegate :reference, to: :journal, prefix: true, allow_nil: true
+
   def statement_date
     statement.try(:created_at)
   end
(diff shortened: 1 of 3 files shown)


app/models/order_detail.rb
```diff
@@ -74,6 +74,8 @@ def current_journal_rows
   end
 
   delegate :journal_date, to: :journal, allow_nil: true
+  delegate :reference, to: :journal, prefix: true, allow_nil: true
+
   def statement_date
```
jhanggi: I considered trying to move some of these methods around since instance methods seemed mixed up in `delegates`, but in really these are essentially delegations where we want to use a different name.

jhanggi:
Code climate is complaining about the number of lines in OrderDetail.

noelrappin:
Yeah, the length of order detail is out of scope here.

meara:
Oh, a thought--`reference` gets populated across all three schools, right?  Idk if this would look a bit weird if the field was nil.

jhanggi:
If the field is blank/nil, then the label doesn't show up. That's handled in the `banner_label` helper.

meara:
Oh cool, thanks for clarifying.

PR #1137, opened 2017-06-23
[#136008] Use facility abbreviation in bulk email subject

jhanggi:
The default will now be `[NUcore EF]` instead of `[NUcore Example Facility]` as the subject prefix because many facilities have very long names.

I am passing both the name and abbreviation to the locale so a fork can easily
override it and revert back if they want.

### DIFF
--- a/vendor/engines/bulk_email/app/services/bulk_email/content_generator.rb
+++ b/vendor/engines/bulk_email/app/services/bulk_email/content_generator.rb
@@ -2,6 +2,8 @@ module BulkEmail
 
   class ContentGenerator
 
+    include TextHelpers::Translation
+
     DEFAULT_RECIPIENT_NAME = "Firstname Lastname".freeze
 
     attr_reader :facility, :subject_product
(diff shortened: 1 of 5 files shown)


vendor/engines/bulk_email/app/services/bulk_email/content_generator.rb
```diff
@@ -25,17 +27,21 @@ def wrap_text(text, recipient_name = nil)
...
     end
 
     def signoff
       if facility.single_facility?
-        I18n.t("bulk_email.body.signoff", facility_name: facility.name)
+        text("body.signoff", facility_name: facility.name)
       end
     end
 
+    def translation_scope
```
meara: Why do we need this?

jhanggi: The `text_helpers` gem uses this. It gets auto-added to controllers and views, so if you use the helpers in other classes, you need to define this scope.

jhanggi: And the reason I switched to using the `text` helper over regular `I18n` was to get the `!app_name!` interpolation feature.

meara: Ah gotcha. Thanks.

PR #1138, opened 2017-06-23
Avoid constantly looking up `User#user_roles`

jhanggi:
While these methods were on UserRole itself, the results of `User#user_roles`
were not being memoized. It would hit the cache, but on some pages this would still
slow things down noticeably.

### DIFF
--- a/app/models/user_role.rb
+++ b/app/models/user_role.rb
@@ -11,6 +11,16 @@ class UserRole < ActiveRecord::Base
   FACILITY_STAFF = "Facility Staff".freeze
   FACILITY_SENIOR_STAFF = "Facility Senior Staff".freeze
 
+  module AssociationExtension
+    def operator?(facility)
+      any? { |ur| ur.facility == facility && ur.in?(facility_roles) }
+    end
+
+    def manager?(facility)
+      any? { |ur| ur.facility == facility && ur.in?(facility_management_roles) }
+    end
+  end
+
   def self.account_manager
     [ACCOUNT_MANAGER]
   end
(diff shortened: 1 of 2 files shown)


app/models/user_role.rb
```diff
@@ -11,6 +11,18 @@ class UserRole < ActiveRecord::Base
...
+  module AssociationExtension
+
+    def operator?(facility)
+      any? { |ur| ur.facility == facility && ur.in?(facility_roles) }
+    end
+
+    def manager?(facility)
+      any? { |ur| ur.facility == facility && ur.in?(facility_management_roles) }
+    end
+
```
noelrappin: I'm sure this works, and there's no real need to throw more time into it...

But, it's not my favorite Rails construct and I wonder if the problem is the `any?` and you could alternately compose scopes:

Feel perfectly free to ignore as the ravings of a madman.

```
def self.in_facility(facility)
  where(facility: facility) 
end

def self.one_of(roles)
  where(role: roles)
end

def self.operator?(facility)
  in_facility(facility).one_of(facility_roles)
end

def self.manager?(facility)
  in_facility(facility).one_of(facility_managment_roles)
```

jhanggi: What I'm seeing is there sometimes lots of alls to `manager?` and it's does a lot of `SELECT * FROM user_roles where user_id = ?`s. Most of them is hitting the cache, but it still comes up as a noticeable performance difference to not hit the cache (it is possible it's just the logging, so it wouldn't change anything in production, but it slows down in development, where I notice it).

I think what you're proposing would still generate a lot of SQL calls (hopefully mostly cached). The extensions works because the `user_roles` gets memoized after it's first call.

meara: Can you explain how this change fixes it? I understand the problem but I don't entirely follow the solution.

jhanggi: I'm not 100% what rails is doing under the hood, but I believe every time you call `user.operator_of?`, it's treating `user.user_roles.any?` as a new query, so it doesn't memoize `user.user_roles`. Moving this into an extension seems to handle the memoization. http://guides.rubyonrails.org/association_basics.html#association-extensions

meara: Oh rails, y u so magical. Thanks for the ref.

PR #1139, opened 2017-06-26
[#136823] Instrument util report errors on problem reservation

jhanggi:
When a problem reservation is missing its end time this report errors. The error
manifests when a problem reservation's fulfilled_at falls into the "Fulfilled between"
range.

```
nil can't be coerced into Fixnum
  app/models/reports/instrument_utilization_report.rb:56:in `+'
```


meara:
What do you mean by the "Fulfilled between" range?

jhanggi:
"Fulfilled between" is the label for the date range fields on the report page.
screen shot 2017-06-26 at 1 23 26 pm

PR #1142, opened 2017-06-27
drop roles table if exists

meara:
NU and open do not have the roles table--UIC and Dartmouth do. None of the schools actually use this table. This migration will drop the table if it exists, in order to get the schools' schemas in line with each other.

### DIFF
--- a/db/migrate/20170627182248_drop_roles_if_exists.rb
+++ b/db/migrate/20170627182248_drop_roles_if_exists.rb
@@ -1,9 +1,8 @@
 class DropRolesIfExists < ActiveRecord::Migration
-  def change
-    if table_exists?(:roles)
-      drop_table(:roles) do |t|
-        t.string "name", limit: 255
-      end
-    end
+  def up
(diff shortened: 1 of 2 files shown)


db/migrate/20170627182248_drop_roles_if_exists.rb
```diff
@@ -0,0 +1,9 @@
+class DropRolesIfExists < ActiveRecord::Migration
+  def change
+    if table_exists?(:roles)
+      drop_table(:roles) do |t|
```
jhanggi: I haven't used this syntax before. Is this to make it rollback-able? How does that interact with the `table_exists?` method?

This may be one where we want the rollback to be a no-op. It's not a big deal.

meara: I tested the rollback both on my local UIC and open databases--it effectively is a no-op (it doesn't restore the table, but also doesn't error).

This syntax doesn't make it super clear that that's what's happening, though--I can do separate up/down migrations to make it more obvious.

PR #1144, opened 2017-06-28
[#136521] Minor change to order detail so that it plays better with transaction search.

noelrappin:
See https://github.com/tablexi/nucore-dartmouth/pull/114

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -289,8 +289,8 @@ class OrderDetail < ActiveRecord::Base
   scope :for_owners, lambda { |owners|
     return if owners.blank?
     joins(account: :account_users)
-      .where(account_users: { user_role: AccountUser::ACCOUNT_OWNER })
-      .where("account_users.user_id in (?)", owners)
+      .where(account_users: { user_role: AccountUser::ACCOUNT_OWNER,
+                              user_id: owners })
   }
   scope :for_order_statuses, ->(statuses) { where("order_details.order_status_id in (?)", statuses) unless statuses.nil? || statuses.empty? }
 
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -287,11 +287,10 @@ class OrderDetail < ActiveRecord::Base
   scope :for_facilities, ->(facilities) { joins(:order).where("orders.facility_id in (?)", facilities) unless facilities.nil? || facilities.empty? }
   scope :for_products, ->(products) { where("order_details.product_id in (?)", products) unless products.blank? }
   scope :for_owners, lambda { |owners|
-    if owners.present?
-      joins(:account)
-        .joins("INNER JOIN account_users on account_users.account_id = accounts.id and user_role = 'Owner'")
-        .where("account_users.user_id in (?)", owners)
-    end
+    return if owners.blank?
+    joins(account: :account_users)
+      .where(account_users: { user_role: AccountUser::ACCOUNT_OWNER })
+      .where("account_users.user_id in (?)", owners)
   }
   scope :for_order_statuses, ->(statuses) { where("order_details.order_status_id in (?)", statuses) unless statuses.nil? || statuses.empty? }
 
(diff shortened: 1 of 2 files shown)


app/models/order_detail.rb
```diff
@@ -287,11 +287,10 @@ def can_be_viewed_by?(user)
...
   scope :for_owners, lambda { |owners|
-    if owners.present?
-      joins(:account)
-        .joins("INNER JOIN account_users on account_users.account_id = accounts.id and user_role = 'Owner'")
-        .where("account_users.user_id in (?)", owners)
-    end
+    return if owners.blank?
+    joins(account: :account_users)
+      .where(account_users: { user_role: AccountUser::ACCOUNT_OWNER })
+      .where("account_users.user_id in (?)", owners)
```
jhanggi: Did we want to change this to `where(account_users: { user_id: owners })`?

jhanggi: Oh, and then we could actually combine the two `.where(account_users: { user_role: AccountUser::ACCOUNT_OWNER, user_id: owners })`

spec/transaction_search_shared_examples.rb
```diff
@@ -11,8 +11,9 @@
       let(:params) { super().merge(account_owners: [3, 4]) }
 
       it "applies the parameter" do
-        expect(assigns(:order_details).where_values)
-          .to be_include("account_users.user_id in ('3','4')")
+        where_values = assigns(:order_details).where_values.reject { |v| v.is_a?(String) }
+        expect(where_values.map(&:to_sql))
+          .to include("`account_users`.`user_id` IN (3, 4)")
       end
```
jhanggi: This should probably be tested against oracle before we merge it. I don't think it quotes tables/columns the same way. I'm not really sure why we tested this way. Is there a better test we could do?

jhanggi: Yep, oracle doesn't like this. It was working before before because we were setting a direct string so it was consistent across databases. I'm really not sure what the right fix is for this. Any ideas?

jhanggi: Here's the failure message:

`expected ["\"ORDER_DETAILS\".\"ACCOUNT_ID\" IN (SELECT \"ACCOUNTS\".\"ID\" FROM \"ACCOUNTS\" INNER JOIN \"ACCO...TE\" = :a1", "\"ACCOUNT_USERS\".\"USER_ROLE\" = 'Owner'", "\"ACCOUNT_USERS\".\"USER_ID\" IN (3, 4)"] to include "`account_users`.`user_id` IN (3, 4)"`

noelrappin: Well, we could have it match a Regex or match either string (both of which seem kind of brittle. Or we could go up a level and search for the Arel object directly.

jhanggi: Checking the Arel level makes sense if it's not too much effort. I like that better than the regex or checking for one of two strings. I think the idea was to avoid needing to set up a bunch of orders in order to run the queries and check the results.

spec/transaction_search_shared_examples.rb
```diff
@@ -11,8 +11,15 @@
...
-        expect(assigns(:order_details).where_values)
-          .to be_include("account_users.user_id in ('3','4')")
+        in_nodes = assigns(:order_details)
+                   .where_values.select { |v| v.is_a?(Arel::Nodes::In) }
+        expect(in_nodes).to satisfy { |nodes|
+          nodes.any? do |node|
+            node.left.name == :user_id &&
+              node.left.relation.name.casecmp("account_users").zero? &&
+              node.right.map(&:val) == ["3", "4"]
+          end
```
noelrappin: This is a rubocop thing to add `casecmp`, there's a clearer `casecmp?` method, but I think that's Ruby 2.4.

meara:
+1 on the spec being complicated, but this looks good to me.

PR #1146, opened 2017-06-29
[#136900] User details being shown on calendar

jhanggi:
There is a `show_details` “Show reservation details (name, email, etc)
on schedule” option on instruments that should only show users’ name
and email if the option is on.

However, what was happening is we were sending the string `”false”`
into `Reservations::CalendarPresenter`, which is `present?`, so we were
rending the details if if we didn’t want to.

Even if that was fixed, using a query parameter like this isn’t very
secure. Anyone could conceivably change the query parameter to “true”.
This changes the logic so that if you request details (which happens in
most of the facility-side views), it also checks to make sure that
either the `show_details` option is on or that you have facility-level
permissions.

I also did a little bit of JS cleanup getting rid of the `withDetails`
javascript global since it’s only used in the one case of a user
placing a reservation to disable the tooltips.

There's definitely more that could be done on refactoring #index, but this seemed like enough for getting the bug fixed.

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -54,8 +54,8 @@ class ReservationsController < ApplicationController
     respond_to do |format|
       as_calendar_object_options = { start_date: @start_at.beginning_of_day, with_details: @show_details }
       format.js do
-        render json: @reservations.map { |r| r.as_calendar_object(as_calendar_object_options) } +
-                     Array(@unavailable).map { |r| r.as_calendar_object(as_calendar_object_options) }
+        render json: Reservation.as_calendar_objects(@reservations, as_calendar_object_options) +
+                     ScheduleRule.as_calendar_objects(@unavailable, as_calendar_object_options)
       end
     end
   end
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -23,45 +23,39 @@ class ReservationsController < ApplicationController
 
   # GET /facilities/1/instruments/1/reservations.js?_=1279579838269&start=1279429200&end=1280034000
   def index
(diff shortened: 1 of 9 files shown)


app/controllers/reservations_controller.rb
```diff
@@ -23,45 +23,39 @@ def public_timeline
...
-                   end
+    @show_details = params[:with_details] == "true" && (@instrument.show_details? || can?(:administer, Reservation))
 
     respond_to do |format|
-      as_calendar_object_options = { start_date: @start_date, with_details: params[:with_details] }
+      as_calendar_object_options = { start_date: @start_at.beginning_of_day, with_details: @show_details }
       format.js do
-        render json: @reservations.map { |r| r.as_calendar_object(as_calendar_object_options) }.flatten +
-                     @unavailable.map { |r| r.as_calendar_object(as_calendar_object_options) }.flatten
+        render json: @reservations.map { |r| r.as_calendar_object(as_calendar_object_options) } +
```
meara: Just noticed you took `.flatten` out for both `@reservations` and `@unavailable`, but only wrapped one of them with `Array()`.  Possibly it's not necessary for `@reservations`, I'm not sure what form the data is in here, but I just wanted to double check in case this was a typo.

jhanggi: The `Array` is there because `@unavailable` may be nil. The `flatten` is not a typo, but definitely a mistake. Turns out that Reservation's `as_calendar_object` behaves differently than `ScheduleRule`'s. Reservation's always returns a single hash, ScheduleRule returns an array of hashes (this is because a single schedule, say weekdays from 9-5, needs 5 calendar items, one for each Mon-Fri). Let me think about how to improve this.

PR #1147, opened 2017-06-30
rm ruby 2.2 polyfill

meara:
Noticed in the course of upgrading ruby versions, https://pm.tablexi.com/issues/132541


meara:
@jhanggi did you have comments on this?

PR #1148, opened 2017-07-03
update rails to 4.2.9

meara:
- https://pm.tablexi.com/issues/136933


vendor/engines/ldap_authentication/ldap_authentication.gemspec
```diff
@@ -16,6 +16,6 @@ Gem::Specification.new do |s|
   s.files = Dir["{app,config,db,lib}/**/*", "README.md"]
   s.test_files = Dir["spec/**/*"]
 
-  s.add_dependency "rails", "~> 4.2.7.1"
+  s.add_dependency "rails", "~> 4.2.9"
```
jhanggi: I'm wondering if we should make the engine rails dependencies looser. So something like `~> 4.2`, or even `>= 4.2`. I just ran into this on DC while merging open. Or should we get rid of the dependency altogether?

meara: Was there a specific reason we included the rails dependency on this engine? Most of the other engines don't include rails as a dependency, even though they use it...

jhanggi: My guess is that it's part of the default generated file. I think it's probably safe to get rid of, but we can do that down the line.

PR #1149, opened 2017-07-03
[#136941] Bundle admin issues

jhanggi:
* Bundled products tab loses the tabs
* Fix "Docs" tab for bundles
* Remove "notes" field options for bundles

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -102,7 +102,7 @@ Nucore::Application.routes.draw do
     resources :bundles do
       get :manage, on: :member
       resources :bundle_products, controller: "bundle_products", except: [:show]
-      resources :file_uploads, path: "files", only: [:new, :create, :destroy]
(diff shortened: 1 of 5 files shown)


config/routes.rb
```diff
@@ -102,7 +102,7 @@
     resources :bundles do
       get :manage, on: :member
       resources :bundle_products, controller: "bundle_products", except: [:show]
-      resources :file_uploads, path: "files", only: [:new, :create, :destroy]
+      resources :file_uploads, path: "files", only: [:index, :create, :destroy]
```
meara: Why are we removing the `:new` action here? Was it not being used?

jhanggi: Correct, it's not being used. The upload form is on same view as the index list.

jhanggi: For reference, here's where the mistake was introduced: https://github.com/tablexi/nucore-open/commit/2a6483fbbc0eb0423e8d29479fe5e246adcbcf58

PR #1150, opened 2017-07-03
[#132541, #136148] update ruby to 2.4

meara:
- https://pm.tablexi.com/issues/136148
- https://pm.tablexi.com/issues/132541


.rubocop.yml
```diff
@@ -4,7 +4,6 @@ inherit_from:
 
 AllCops:
   DisplayCopNames: true
-  TargetRubyVersion: 2.2
   Exclude:
```
noelrappin: Do we want to explicitly target `2.4` here?

circle.yml
```diff
@@ -1,10 +1,10 @@
 machine:
-  ruby:
-    version: 2.2.4
   timezone: "America/Chicago"
```
noelrappin: Similarly, I think we want to explicitly keep the version here.

meara: CircleCI will use the version from `.ruby-version`: https://circleci.com/docs/1.0/configuration/#ruby-version

.rubocop.yml
meara: So my understanding is that Rubocop and Circle should both read from the `.ruby-version` file, now that we're including it.

http://rubocop.readthedocs.io/en/latest/configuration/#setting-the-target-ruby-version

jhanggi:
This all looks reasonable to me. If we were to change the `ruby-version` back to `2.2`, does everything continue to work? I hesitate a little bit to merge this just yet as this will require devops tasks like installing a new version of ruby on the servers.  I think we can ask @mlineen how we want to handle this.

meara:
@jhanggi I dropped the ruby version back down to 2.2.4--assuming the build passes, would it make sense to merge this (once we reach consensus on pinning the version for rubocop/circle), and increment the ruby version in a separate PR once we're ready for the devops steps? Or should we wait and do it all together?

Gemfile
```diff
@@ -70,6 +72,7 @@ gem "sanger_sequencing", "~> 0.0.1", path: "vendor/engines/sanger_sequencing"
 gem "secure_rooms", path: "vendor/engines/secure_rooms"
 gem "split_accounts", "~> 0.0.1", path: "vendor/engines/split_accounts"
 gem "synaccess_connect", "0.2.2", github: "tablexi/synaccess"
+gem "net-telnet"
```
jhanggi: I remember this was an issue, but where does it get loaded? Was it a bundler thing? I feel like this would only get required if we used `Settings.relays.connect_module` as `NetBooter::Telnet`, which has been years since we used that option.

meara: Yeah, it caused an issue bundling unless Telnet was included, because of synaccess if I recall correctly. If we don't want to include this here, we could try updating synaccess, but I don't know what other projects use it/what the downstream consequences would be there.

meara:
This will work with Ruby at 2.2, if necessary, but I dropped that commit while rebasing, since the plan is to update Ruby on all the servers before merging this.

Gemfile
jhanggi: Let's not worry about that for now. I may look at the synaccess gem at some point, because it would be nice to be able to get rid of that dependency.

PR #1151, opened 2017-07-03
[#136943] Fix error when item/service and a bundle are in cart

jhanggi:
`comparison of Fixnum with nil failed`

Non-bundled order details have a nil group_id, so we can’t directly
compare a nil and a number.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -316,7 +316,8 @@ class OrdersController < ApplicationController
     # Do not do any additional scopes on `@order.order_details` to ensure the list
     # of OrderDetail objects is memoized. Validations are run on `@order.order_details`
     # and we want to be able to render validation errors inline.
-    @order.order_details.sort_by(&:group_id).slice_when do |before, after|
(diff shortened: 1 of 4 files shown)


app/controllers/orders_controller.rb
```diff
@@ -316,7 +316,8 @@ def grouped_order_details
     # Do not do any additional scopes on `@order.order_details` to ensure the list
     # of OrderDetail objects is memoized. Validations are run on `@order.order_details`
     # and we want to be able to render validation errors inline.
-    @order.order_details.sort_by(&:group_id).slice_when do |before, after|
+    sorted_order_details = @order.order_details.sort_by { |od| od.group_id.to_i }
```
noelrappin: Is it worth taking care of this at the model level, either with a `safe_group_id` method, or by making the group_id have a default?

jhanggi: Actually, I think it might make sense to pull the whole thing up into a model method on `Order`. 'm not sure giving it a default value makes sense. In this value, I do think `null` is more meaningful than `0`.  I like the `safe_group_id` idea.

PR #1152, opened 2017-07-03
[#98881] Auto-Log Out after Begin Reservation

noelrappin:
NUCore should automatically log out users (in customer mode only) after they click "Begin Reservation" from "My Reservations". Identical functionality as what happens when users click "End Reservation", including the 60 second delay and possibility of remaining logged in. T

The existing usage didn’t work from the order page, but it does now. I don’t think this presenter is used any place else, so this shouldn’t trigger anyplace else.

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -270,12 +270,13 @@ class ReservationsController < ApplicationController
       ReservationInstrumentSwitcher.new(@reservation).switch_off!
       flash[:notice] = "The instrument has been deactivated successfully"
     end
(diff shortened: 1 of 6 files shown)


meara:
Given that the existing usage wasn't working, should there be a feature spec for this? (Both existing and added functionality.)

noelrappin:
The existing usage wasn't broken, (I don't think), it was just not integrated in all the pages. My thought was that a feature test wasn't worth the effort, but I can add one if you'd like.

meara:
Oh, I get what you meant now. Yeah, a feature spec might be a bit overkill--I think the presenter spec suffices.

PR #1155, opened 2017-07-05
Fix Date to Time conversion so it uses the app's timezone

jhanggi:
`to_time` uses the system's timezone, not the app's.

This was broken by changes made in https://github.com/tablexi/nucore-open/pull/1134

Circle is set to central time as is the application, so that's why this was fine until we merged to DC, which has circle set to central and the app set to eastern. I'm going to open another PR for trying to deal with that.

### DIFF
--- a/app/helpers/date_helper.rb
+++ b/app/helpers/date_helper.rb
@@ -26,13 +26,11 @@ module DateHelper
     date = format_usa_date(Date.parse(date)) if date =~ /\A\d{4}-\d{2}-\d{2}/
 
     return unless usa_formatted_date?(date)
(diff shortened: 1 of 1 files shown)


meara:
Hey, do we need to worry about this `to_time` reference as well?

https://github.com/tablexi/nucore-open/blob/master/lib/date_time_input/model.rb#L12

jhanggi:
I'll look at that other reference later. I'm not sure if we need to worry about it.

jhanggi:
That instance of `to_time` is actually fine because it's not on an actual date object, it's a `DateTimeInput::FormData`, which takes in the hash from the form and runs it through `parse_usa_date`.

PR #1158, opened 2017-07-05
[#132894] creating iCal from reservation

noelrappin:
Before I put this in the UI, wanted to make sure of the basics.

* It’s in a presenter because the Reservations class seemed long enough, and this code would only be used in one context.

* Maybe add a class method on the Presenter as a shortcut

* Wanted to confirm that the name and the descriptions were okay (or if we want them in i18n?)

* Also wanted to confirm that time zone conversion from the Reservation object isn’t needed?


app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,19 @@
...
+  def initialize(reservation)
+    super(reservation)
+  end
+
+  def as_ical
+    cal = Icalendar::Calendar.new
+    cal.event do |e|
+      e.dtstart = Icalendar::Values::DateTime.new(reserve_start_at)
+      e.dtend = Icalendar::Values::DateTime.new(reserve_end_at)
+      e.summary = "Reservation for #{product.name}"
```
jhanggi: We should probably include the facility name in there. Maybe `"#{facility.abbreviation}: #{product.name}"`? I'm not sure we need "Reservation for". I could be convinced otherwise, though.

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,19 @@
...
+      e.description = "Reservation for #{product.name}"
```
jhanggi: I think we should flesh out the description (I'm assuming summary is sort of the "subject" and the desciption is the "body"/"notes"). We should include the full facility name and the order number. If we can do HTML in there, I would link to the order receipt. Or is there a URL field?

Maybe add the facility name as the location?

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,19 @@
+class ReservationCalendar < SimpleDelegator
+
+  def initialize(reservation)
+    super(reservation)
+  end
+
+  def as_ical
+    cal = Icalendar::Calendar.new
+    cal.event do |e|
+      e.dtstart = Icalendar::Values::DateTime.new(reserve_start_at)
```
jhanggi: I'm not quite sure what you mean by time zone conversion. I would guess that as long as the timezone gets rendered out as part of the dtstart/dtend, then it should be fine. I'm not quite sure how icalendar handles timezones.

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,19 @@
+class ReservationCalendar < SimpleDelegator
+
+  def initialize(reservation)
+    super(reservation)
```
jhanggi: If we're not doing anything special in the constructor, I think we don't need it.

app/presenters/reservation_calendar.rb
noelrappin: My assumption is that we don't need to worry about dates being in different timezones, so I don't need to worry about conversion?

app/presenters/reservation_calendar.rb
noelrappin: I'll defer to your context, this is basically just to show the structure.

app/presenters/reservation_calendar.rb
noelrappin: I'll check the rest of the fields in the spec.

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,31 @@
+class ReservationCalendar < SimpleDelegator
+
+  attr_accessor :hostname
+
+  def self.to_calendar(reservation)
+    ReservationCalendar.new(reservation).as_ical
```
jhanggi: I'd just go with `new(reservation).as_ical`.

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,31 @@
...
+    @hostname = hostname
+  end
+
+  def as_ical
+    cal = Icalendar::Calendar.new
+    cal.event do |e|
+      e.dtstart = Icalendar::Values::DateTime.new(reserve_start_at)
+      e.dtend = Icalendar::Values::DateTime.new(reserve_end_at)
+      e.summary = "#{facility.abbreviation}: #{product.name}"
+      e.description = "#{facility.name} reservation for #{product.name}. Order number #{order.id}"
```
jhanggi: I would probably go with I18n-ing this now. I don't really care if we pass more interpolations than the locale ends up needing, either.

Also, we should use `OrderDetail#to_s`/`OrderDetail#order_number`. That'll give us the full XXX-XXX number including the detail id.

spec/presenters/reservation_calendar_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    it "generates an ical file" do
+      expect(ical.events.size).to eq(1)
+      event = ical.events.first
+      expect(event.dtstart).to eq(reservation.reserve_start_at)
+      expect(event.dtend).to eq(reservation.reserve_end_at)
+      expect(event.summary).to eq("FA: Instrument")
+      expect(event.description).to eq("Facility reservation for Instrument. Order number #{order.id}")
+      expect(event.location).to eq("Facility")
+      expect(event.url.to_s).to eq("http://localhost/orders/1001/order_details/1002/reservations/1003")
```
jhanggi: Where are you stubbing the IDs so this works?

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,31 @@
...
+      e.location = facility.name
+      e.url = Icalendar::Values::Uri.new(
+        Rails.application.routes.url_helpers.order_order_detail_reservation_url(
```
jhanggi: I would link to the `/orders/XXX/order_details/XXX` route rather than the reservation. It has a little more information and can easily link into the reservations as well as the receipt. (Also, I notice there's no way to get from the reservation view back to the order detail view, which is not ideal, but that's how it is right now).

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,31 @@
...
+
+  attr_accessor :hostname
+
+  def self.to_calendar(reservation)
+    ReservationCalendar.new(reservation).as_ical
+  end
+
+  def initialize(reservation, hostname)
+    super(reservation)
+    @hostname = hostname
```
jhanggi: I don't necessarily have a problem with this, but I feel like there's got to be a better way to get the hostname out of the request or something.

app/presenters/reservation_calendar.rb
noelrappin: Yeah, that's better.

app/presenters/reservation_calendar.rb
noelrappin: You can get the host name from the request, but this presenter won't have the request, I assumed that the controller that was creating this object would pass the hostname in.

app/presenters/reservation_calendar.rb
noelrappin: Ok

spec/presenters/reservation_calendar_spec.rb
noelrappin: Ack. I'm not, so it's fragile, it needs to look like the description line and interpolate them in.

spec/presenters/reservation_calendar_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it "generates an ical file" do
+      expect(ical.events.size).to eq(1)
+      event = ical.events.first
+      expect(event.dtstart).to eq(reservation.reserve_start_at)
+      expect(event.dtend).to eq(reservation.reserve_end_at)
+      expect(event.summary).to eq("FA: Instrument")
+      expect(event.description).to eq("Facility reservation for Instrument. Order number #{order_detail.order_number}.")
+      expect(event.location).to eq("Facility")
+      expect(event.url.to_s).to eq("http://localhost/orders/#{order.id}/order_details/#{order_detail.id}")
```
jhanggi: Are we going to need to deal with configuring`https`, port numbers (for development)?

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def as_ical
+    cal = Icalendar::Calendar.new
+    cal.event do |e|
+      e.dtstart = Icalendar::Values::DateTime.new(reserve_start_at)
+      e.dtend = Icalendar::Values::DateTime.new(reserve_end_at)
+      e.summary = I18n.t(
+        "ical.summary", facility: facility.abbreviation, product: product.name)
+      e.description = I18n.t(
```
jhanggi: We might not be at the point yet, but it's looking like it might be worth extracting some private methods soon.

spec/presenters/reservation_calendar_spec.rb
noelrappin: probably, but I didn't want to overbuild at this point

noelrappin:
Want me to merge this and do the connections in a separate PR, or keep going here?

jhanggi:
I could go either way. I feel like I'd rather have a user-facing feature before we do the merge, but don't feel too strongly.

meara:
Is this still a WIP?

noelrappin:
Maybe? There are still other places that the iCal connection should be linked, and I was going to look into a webcal link that would go directly to google calendar

meara:
Okay--I don't have any notes on what's here at present, so I'll wait to take another look whenever you're ready for a final review.

noelrappin:
@jhanggi Up for controller code review...

noelrappin:
For the record, I was able to get Rails to create a `webcal` link, but I'm not sure it'll be super useful (my calendar program wants to treat it as a subscription source...)

app/controllers/reservations_controller.rb
```diff
@@ -138,6 +138,17 @@ def show
     @reservation  = Reservation.find(params[:id])
 
     raise ActiveRecord::RecordNotFound if @reservation != @order_detail.reservation
+
+    respond_to do |format|
+      format.html
+
+      format.ics do
+        calendar = ReservationCalendar.new(@reservation, request.host)
+        send_data(calendar.to_ical,
```
jhanggi: I think this is mostly fine, though we may want to consider abstracting it up if we add this to other controllers. Look at what we did for CSVs in `config/initializers/csv_renderer.rb`. I'd be fine if you say, let's deal with that when we add it to more places.

app/controllers/reservations_controller.rb
```diff
@@ -138,6 +138,17 @@ def show
...
+                  type: 'text/calendar', disposition: 'attachment',
+                  filename: "reservation_cal_#{@reservation.id}.ics")
```
jhanggi: I would go with the order number (XXX-XXX) in the filename. The reservation ID is never really shown anywhere except URLs so has little semantic value.

app/controllers/reservations_controller.rb
```diff
@@ -138,6 +138,17 @@ def show
     @reservation  = Reservation.find(params[:id])
 
     raise ActiveRecord::RecordNotFound if @reservation != @order_detail.reservation
+
+    respond_to do |format|
+      format.html
+
+      format.ics do
+        calendar = ReservationCalendar.new(@reservation, request.host)
```
jhanggi: Still want to hold off on dealing with HTTPS and ports?

app/controllers/reservations_controller.rb
```diff
@@ -138,6 +138,17 @@ def show
...
+        send_data(calendar.to_ical,
+                  type: 'text/calendar', disposition: 'attachment',
```
jhanggi: Use double quotes.

app/views/orders/receipt.html.haml
```diff
@@ -41,6 +41,10 @@
           .order-detail-description= order_detail_description(order_detail)
           - if res
             .order-detail-extra= res
+            .order-detail-extra
+              = link_to(fa_icon("calendar"),
+                order_order_detail_reservation_path(@order, order_detail, res, format: :ics),
+                title: "Calendar file download")
```
jhanggi: I18n the tooltip.

app/controllers/reservations_controller.rb
noelrappin: Yes, I think the `request.host` actually might handle this (my developer one came in as `localhost:3000`.

app/controllers/reservations_controller.rb
noelrappin: Yeah, that's what I'd say for now

app/controllers/reservations_controller.rb
noelrappin: But the order might have multiple reservations, no?

jhanggi: An order might have multiple reservations, but an order detail can only have one reservation. We use `OrderDetail#to_s` or `OrderDetail#order_number` (they're aliases) in quite a few places.

app/controllers/reservations_controller.rb
jhanggi: I wouldn't mind verifying for HTTPS before we merge since we'll be dealing with it as soon as we put it on NU stage. Maybe use ngrok to test that out?

noelrappin: Using ngrok, it picks up the port, but not the Https, checking...

noelrappin: Okay, now fixed.

noelrappin:
@jhanggi this adds controller logic for the facility. In both cases the URL inside the ical itself is of the form: `http://localhost:3000/orders/37/order_details/48`. Is that okay?

jhanggi:
That's going to be a problem because facility staff doesn't have access to `/orders/37/order_details/48`. They need to access it through their facility. And we should probably send them directly to the order view (since this controller action redirects them anyways).

noelrappin:
So, send admins to `/orders/37`, and facility staff to `/facility/whatever/orders/37`? Is there a way to send everybody to the same URL?

app/presenters/reservation_calendar.rb
```diff
@@ -1,15 +1,14 @@
 class ReservationCalendar < SimpleDelegator
 
-  attr_accessor :hostname, :ical, :protocol
+  attr_accessor :url, :ical
 
-  def self.to_calendar(reservation, hostname, protocol)
-    new(reservation, hostname, protocol).to_ical
+  def self.to_calendar(reservation, url = nil)
```
jhanggi: Not a big deal as nothing is using this yet, but this seems like a good candidate for a keyword arg.

app/controllers/reservations_controller.rb
```diff
@@ -138,6 +138,16 @@ def show
     @reservation  = Reservation.find(params[:id])
 
     raise ActiveRecord::RecordNotFound if @reservation != @order_detail.reservation
+
+    respond_to do |format|
+      format.html
+
+      format.ics do
+        calendar = ReservationCalendar.new(@reservation)
+        send_data(calendar.to_ical, type: "text/calendar",
```
jhanggi: Part of me still wants to abstract this, but I'd be fine waiting until we have 3 places. 

And not a big deal, but the two controllers we have it in are formatted slightly differently.

app/controllers/facility_order_details_controller.rb
```diff
@@ -14,9 +14,20 @@ def initialize
...
+        # This is deprecated in favor of OrderManagement::OrderDetailsController, but
+        # we want to avoid 404s
+        redirect_to facility_order_path(current_facility, @order)
+      end
+
+      format.ics do
+        calendar = ReservationCalendar.new(@order_detail.reservation)
+        send_data(calendar.to_ical,
+                  type: "text/calendar", disposition: "attachment",
+                  filename: "reservation_cal_#{@order_detail.order_number}.ics")
```
jhanggi: Can we use `calendar.filename` here?

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def generate_ical
+    @ical = Icalendar::Calendar.new
+    ical.event do |e|
+      e.dtstart = Icalendar::Values::DateTime.new(reserve_start_at)
+      e.dtend = Icalendar::Values::DateTime.new(reserve_end_at)
+      e.summary = summary
+      e.description = description
+      e.location = facility.name
+      e.url = url if url.present?
```
jhanggi: Will the controller be responsible for wrapping the URL in whatever wrapper Icalendar needs? (we can deal with that when we need to, I suppose)

app/mailers/notifier.rb
```diff
@@ -76,6 +77,15 @@ def order_detail_status_change(order_detail, old_status, new_status, to)
 
   private
 
+  def attach_ical(order)
+    order.order_details.map(&:reservation).compact.each do |reservation|
+      calendar = ReservationCalendar.new(reservation)
+      attachments[calendar.filename] = {
```
jhanggi: Have you been able to test this to see if gmail picks it up? We can always test from the open stage if you haven't.

app/presenters/reservation_calendar.rb
noelrappin: Actually, I was thinking I'd take it out, since nobody is suing it.

app/presenters/reservation_calendar.rb
noelrappin: No, the wrapping should be here, I'll put it back

app/controllers/facility_order_details_controller.rb
noelrappin: Yes

app/mailers/notifier.rb
noelrappin: I tested that letter_opener received it as an attachment correctly.

app/controllers/facility_order_details_controller.rb
```diff
@@ -14,9 +14,20 @@ def initialize
...
-    redirect_to facility_order_path(current_facility, @order)
+    respond_to do |format|
+      format.html do
+        # This is deprecated in favor of OrderManagement::OrderDetailsController, but
+        # we want to avoid 404s
+        redirect_to facility_order_path(current_facility, @order)
+      end
+
+      format.ics do
+        calendar = ReservationCalendar.new(@order_detail.reservation)
```
jhanggi: I don't think we need to worry about someone hacking URLs and accessing a non-instrument order detail, but something to think about.

app/mailers/notifier.rb
jhanggi: What I was really saying is does gmail recognize it as an event invitation? If we can't test this until it gets to stage, that's fine, we should be able to verify from open stage.

app/presenters/reservation_calendar.rb
```diff
@@ -0,0 +1,50 @@
+class ReservationCalendar < SimpleDelegator
+
+  attr_accessor :url, :ical
+
+  def self.to_calendar(reservation, url = nil)
+    new(reservation, url).to_ical
+  end
+
+  def initialize(reservation, url = nil)
```
jhanggi: I thought we were getting rid of this since it wasn't being used. It's fine, but I might consider using a keyword arg for this.

app/controllers/facility_order_details_controller.rb
noelrappin: Right now, it'd error if the reservation is `nil`, I suspect that's okay at the moment

app/presenters/reservation_calendar.rb
noelrappin: Kept the URL to keep the wrapper logic. But making it a keyword is easy enough, and I think it's fine to remove the class method

PR #1159, opened 2017-07-06
[#136975] Order import errors appearing in "Notes" field

jhanggi:
If the uploaded import file does not have a notes column (the current template
does have this column, but older versions of the template might not), then any
import errors in the error report appear in the "Notes" column as opposed to the
"Errors" column.

There are definitely opportunities for refactoring here, but didn't seem worth the effort for this bug fix.

### DIFF
--- a/app/support/order_row_importer.rb
+++ b/app/support/order_row_importer.rb
@@ -70,10 +70,14 @@ class OrderRowImporter
     @order_key ||= [user_field, chart_string_field, order_date_field]
   end
 
-  def row_with_errors # TODO: refactor
-    new_row = @row.dup
+  def row_with_errors
+    # Start with a hash of HEADERS keys with nil values to ensure optional columns
+    # are included in the report even if they are not in the uploaded CSV.
+    new_row = Hash[HEADERS.values.zip []]
+    new_row.merge!(@row)
     new_row[HEADERS[:errors]] = errors.join(", ")
-    new_row
+
+    CSV::Row.new(new_row.keys, new_row.values)
   end
 
   def user
--- a/app/support/order_row_importer.rb
+++ b/app/support/order_row_importer.rb
@@ -73,7 +73,7 @@ class OrderRowImporter
   def row_with_errors
     # Start with a hash of HEADERS keys with nil values to ensure optional columns
     # are included in the report even if they are not in the uploaded CSV.
-    new_row = Hash[HEADERS.values.zip []]
+    new_row = HEADERS.values.each_with_object({}) { |header, hash| hash[header] = nil }
     new_row.merge!(@row)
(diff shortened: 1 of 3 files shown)


app/support/order_row_importer.rb
```diff
@@ -70,10 +70,14 @@ def order_key
     @order_key ||= [user_field, chart_string_field, order_date_field]
   end
 
-  def row_with_errors # TODO: refactor
-    new_row = @row.dup
+  def row_with_errors
+    # Start with a hash of HEADERS keys with nil values to ensure optional columns
+    # are included in the report even if they are not in the uploaded CSV.
```
noelrappin: This is super minor, but it confused me for a second -- it's a hash of keys with empty array values, not nil values, right?

jhanggi: No, it's actually nil values. If you zip and one is longer than the other you get nils.

`HEADERS.values.zip []` => [["Header 1", nil], ["Header 2", nil], ...]`, then `Hash[]` => `{ "Header 1" => nil, "Header 2" => nil, ... }`. I'm open to better ways to get to this, or describe it. Ultimately, what I need is to make sure `new_row` has all the expected columns. Before, it might not have had "Notes", which is why the "Errors" were ending up in that column sometimes.

noelrappin: Right. Okay, I see that now. Might be a clearer way to create that empty hash. Would this work?

```
Headers.values.each_with_object({}) { |value, hash| hash[value] = nil }
```

noelrappin: Once upon a time I wrote a Hash extension that would do this cleanly called like `Hash#key_map` or something. 

Or...

```
HEADERS.invert.transform_values { |v| nil }
```

jhanggi: Thanks. I think those are both clearer. I think I like the `each_with_object` the best. 

The `transform_values` method seems pretty nice for some cases, though (looked it up, it appears to be a Rails 4.2 extension and then is added to ruby in 2.4).

meara: I like the `each_with_object` too.

jhanggi: Updated.

spec/app_support/order_row_importer_spec.rb
```diff
@@ -56,7 +55,6 @@
   let(:quantity) { "column4" }
   let(:order_date) { "column5" }
   let(:fulfillment_date) { "column6" }
-  let(:errors) { "column7" }
   let(:notes) { "column8" }
```
meara: Notes shouldn't still be column 8, right? I realize this is a string and doesn't really matter, but just for clarity's sake...

PR #1160, opened 2017-07-07
[#136924] Ready for Journal/Statement badges

jhanggi:
* Ready for Journal
* Ready for Statement
* Awaiting Payment (statemented, but not reconciled)
* Get rid of `disputed?` method, which is only used in a couple specs
* A couple refactors to use local methods vs hitting the database again
to prevent N+1 queries

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -201,7 +201,7 @@ class Account < ActiveRecord::Base
 
   def can_reconcile?(order_detail)
     if self.class.using_journal?
-      order_detail.journal.try(:successful?) || OrderDetail.need_journal.exists?(order_detail.id)
+      order_detail.journal.try(:successful?) || order_detail.ready_for_journal?
     elsif self.class.using_statements?
       order_detail.statement_id.present?
     else
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -211,17 +211,6 @@ class OrderDetail < ActiveRecord::Base
     where("estimated_cost IS NOT NULL")
   end
 
-  def in_review?
-    # check in the database if self.id is in the scope
-    self.class.in_review.exists?(id: id)
-    # this would work without hitting the database again, but duplicates the functionality of the scope
-    # state == 'complete' and !reviewed_at.nil? and reviewed_at > Time.zone.now and (dispute_at.nil? or !dispute_resolved_at.nil?)
-  end
-
-  def reviewed?
-    reviewed_at.present? && !in_review? && !in_dispute?
-  end
-
   def can_be_viewed_by?(user)
     order.user_id == user.id || account.owner_user.id == user.id || account.business_admins.any? { |au| au.user_id == user.id }
   end
@@ -527,10 +516,6 @@ class OrderDetail < ActiveRecord::Base
     end
   end
 
-  def can_dispute?
-    in_review?
-  end
-
   def customer_account_changeable?
     journal_id.blank? && statement_id.blank? && !canceled?
   end
@@ -711,15 +696,6 @@ class OrderDetail < ActiveRecord::Base
     price_policy.nil? && estimated_cost && estimated_subsidy && actual_cost.nil? && actual_subsidy.nil?
   end
 
-  def in_dispute?
-    dispute_at && dispute_resolved_at.nil? && !canceled?
-  end
-
-  def disputed?
-    # only used in specs
-    dispute_at.present? && !canceled?
-  end
-
   def cancel_reservation(canceled_by, canceled_reason: nil, order_status: OrderStatus.canceled_status, admin: false, admin_with_cancel_fee: false)
     self.canceled_by = canceled_by.id
     self.canceled_reason = canceled_reason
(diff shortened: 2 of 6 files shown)


jhanggi:
Not sure if the block length codeclimate warning in the spec is worth addressing.

noelrappin:
I'd shut all the length cops off, honestly

app/models/order_detail.rb
```diff
@@ -771,6 +747,34 @@ def problem_order?
     problem
   end
 
+  def can_dispute?
+    in_review? && dispute_at.blank?
+  end
```
noelrappin: is the extra boolean clause here to cover a potential bug?

app/models/order_detail.rb
```diff
@@ -771,6 +747,34 @@ def problem_order?
...
+    dispute_at.present? && dispute_resolved_at.nil? && !canceled?
+  end
+
+  def in_review?
+    reviewed_at.try(:future?) && !in_dispute?
+  end
+
+  def reviewed?
+    reviewed_at.try(:past?) && !in_dispute?
+  end
```
noelrappin: Trying to understand this... and the `past?` bit means we don't also need to check `in_review?`, right?

It's also just weird that `present?` and `past?` mean totally different things.

app/models/order_detail.rb
jhanggi: Not quite following the question. This needs to be a little more restrictive than `in_review? && !in_dispute?`. An order is no longer `in_dispute?` once `dispute_resolved` is set or the order is canceled, but it does need to be `!can_dispute?`.

app/models/order_detail.rb
jhanggi: An order cannot be `in_review?` and `reviewed?` at the same time. It seemed clearer to me to make them independent and parallel as opposed to calling each other and then also needing additional redundant calls.

I think `past?` might have been confusing on its own, but when paired with `future?`, I thought it made it more obvious than a `< Time.current` and ` > Time.current` And what about `🎁 ?`?

noelrappin: Yeah, independent makes sense, I was just trying to make sure I got it right in my head. `past?` and `future?` are the right calls, I just got tangled comparing the old with the new..

app/models/order_detail.rb
noelrappin: I was comparing it to the old `can_dispute?` method, but I guess this is just to make all of them parallel and independent?

jhanggi: My goal was to make some of them independent so you ideally don't have to go more than one method down (and also make sure I don't have any infinite loops), but also to make them at least somewhat intention revealing, even if it means some predicates are redundant through other method calls.

I just pushed up another spec with a scenario that fails without the `dispute_at.blank?`.

app/models/account.rb
```diff
@@ -201,7 +201,7 @@ def validate_against_product(product, user)
 
   def can_reconcile?(order_detail)
     if self.class.using_journal?
-      order_detail.journal.try(:successful?) || OrderDetail.need_journal.exists?(order_detail.id)
+      order_detail.journal.try(:successful?) || order_detail.ready_for_journal?
```
meara: I like this refactor a lot, `ready_for_journal?` is so much clearer.

app/models/order_detail.rb
```diff
@@ -771,6 +747,34 @@ def problem_order?
...
+
+  def ready_for_statement?
+    reviewed? && statement_id.blank? && Account.config.using_statements?(account.type)
+  end
+
+  def ready_for_journal?
+    reviewed? && journal_id.blank? && Account.config.using_journal?(account.type)
+  end
+
+  def awaiting_payment?
```
meara: This has meaning only for statements? Not for journals as well?

spec/controllers/order_details_controller_spec.rb
```diff
@@ -29,13 +29,17 @@
       context "with a blank dispute_reason" do
         let(:dispute_reason) { "" }
 
-        it { expect(order_detail.reload).not_to be_disputed }
+        it "does not set set the dispute at" do
```
meara: Double sets.

spec/controllers/order_details_controller_spec.rb
```diff
@@ -29,13 +29,17 @@
...
+          expect(order_detail.reload.dispute_at).to be_blank
+        end
       end
 
       context "with a dispute_reason" do
         let(:dispute_reason) { "Too expensive" }
 
-        it { expect(order_detail.reload).to be_disputed }
+        it "caputures the dispute at time" do
```
meara: Typo

app/models/account.rb
jhanggi: And it saves an N+1!

app/models/order_detail.rb
jhanggi: Correct, while it has slightly different underlying logic, this method is analogous to `in_open_journal?`

meara: Ah gotcha.

app/models/account.rb
meara: 👍

PR #1163, opened 2017-07-10
[#137029] update simple_form

meara:
- https://pm.tablexi.com/issues/137029

Fix deprecation warnings from Ruby 2.4.


noelrappin:
I also mildly lean toward using `~>` in projects that are basically locked down, I don't in projects that are still in initial development.

meara:
Updated.

Gemfile
```diff
@@ -43,7 +43,7 @@ gem "jquery-rails"
 gem "jquery-ui-rails", "~> 6.0.1"
 gem "vuejs-rails"
 gem "clockpunch", "~> 0.1.12"
-gem "simple_form", "~> 3.3.1"
+gem "simple_form", "~>3.5.0"
```
jhanggi: minor: looks like we usually put a space between the `~>` and the number.

PR #1164, opened 2017-07-11
Ruby/rails post update cleanup

meara:
Some syntax cleanup following up https://github.com/tablexi/nucore-open/pull/1148.

### DIFF
--- a/app/models/order_status.rb
+++ b/app/models/order_status.rb
@@ -16,30 +16,30 @@ class OrderStatus < ActiveRecord::Base
   end
 
   scope :for_facility, ->(facility) { where(facility_id: [nil, facility.id]).order(:lft) }
-  scope :new_os, -> { where(name: "New").limit(1) }
-  scope :inprocess, -> { where(name: "In Process").limit(1) }
-  scope :canceled, -> { where(name: "Canceled").limit(1) }
-  scope :complete, -> { where(name: "Complete").limit(1) }
-  scope :reconciled, -> { where(name: "Reconciled").limit(1) }
+  scope :new_os, -> { find_by(name: "New") }
+  scope :inprocess, -> { find_by(name: "In Process") }
+  scope :canceled, -> { find_by(name: "Canceled") }
+  scope :complete, -> { find_by(name: "Complete") }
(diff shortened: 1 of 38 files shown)


app/models/order_status.rb
```diff
@@ -16,34 +16,29 @@ class OrderStatus < ActiveRecord::Base
...
-  scope :canceled, -> { where(name: "Canceled").limit(1) }
-  scope :complete, -> { where(name: "Complete").limit(1) }
-  scope :reconciled, -> { where(name: "Reconciled").limit(1) }
 
   def self.new_status
-    new_os.first
+    find_by(name: "New")
   end
 
   def self.complete_status
```
jhanggi: I like this. Though, now I almost want to go with `OrderStatus.complete` (without the somewhat redundant `_status`). Not completely necessary, though. Maybe leave both with one the alias for the other while we migrate?

meara: To clarify--do you mean that you want all the `_status` methods to lose the status, basically? So it would become `OrderStatus.complete`, `OrderStatus.new`, etc?

I'm down to make that change. Do we have a lot of fork-specific code that we are concerned about migrating? Might be better to do a two stage migration if we want to make that change, in order to avoid possible collisions with the scopes that I eliminated in this PR--I'd rather get a method missing error than have some possible weird side effects.

jhanggi: Yes, that's what I'm proposing (although `new` might need to stay `new_status` to avoid conflicting with the initialize method).

I'm not sure how much fork specific code we have to worry about. We could do something like this. I'm not sure we need to worry too much about the scopes going away, anywhere that used them should be calling `first` on it, which would error on an instance of `OrderStatus`.

```
def self.complete
  find_by(name: "Complete")
end

def self.complete_status
  warn "Deprecated complete_status called. Use complete instead. At #{caller(1..1)}"
  complete
end

`

meara:
Updated.

meara:
@jhanggi UIC and Dartmouth forks build cleanly. NU had a couple outdated references in engines but I've got the commits to fix them in https://github.com/tablexi/nucore-nu/pull/290.

Good to merge this?

PR #1168, opened 2017-07-12
[#133776] offline reservations display on calendar/timeline

meara:
- https://pm.tablexi.com/issues/133776

- Show the offline blocks on the facility timeline view (I feel like this could be pretty useful). If we do this we should make sure auto-canceled reservations are in front of the offline block
- Show the offline blocks in the calendar on the place reservation view. Show when ordering on behalf of.
- Show the offline blocks in the calendar where you take the instrument online and offline

screen shot 2017-07-12 at 5 31 29 pm

screen shot 2017-07-12 at 5 45 50 pm

screen shot 2017-07-12 at 5 46 22 pm

PR #1170, opened 2017-07-14
[#136142] Add Instrument Name to calendar tooltip

jhanggi:
> This is specifically useful for shared calendars.

There was a lot of copy/pasted code between `reservations.js` and `instruments.js` for setting up the calendar. This consolidates it into one single class, and then is initialized through those two files.

This will conflict with #1168 in the calendar presenter, but should be straightforward to reconcile.

New reservation:
screen shot 2017-07-14 at 5 57 08 pm

Public calendar (with details disabled):
screen shot 2017-07-14 at 5 57 58 pm

### DIFF
--- a/app/assets/javascripts/app/calendar.js.coffee
+++ b/app/assets/javascripts/app/calendar.js.coffee
@@ -8,8 +8,9 @@ class window.FullCalendarConfig
     events: events_path
     loading: (isLoading, view) =>
       @toggleOverlay(isLoading)
-      @toggleNextPrev(view) if !isLoading
+
     eventAfterRender: @buildTooltip
+    eventAfterAllRender: @toggleNextPrev
 
   toggleOverlay: (isLoading) ->
     if isLoading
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/app/calendar.js.coffee
```diff
@@ -0,0 +1,73 @@
+class window.FullCalendarConfig
+  constructor: (@$element) ->
+
+  init: (options) ->
+    @$element.fullCalendar($.extend(@options(), options))
+
```
noelrappin: Why not just pass the options to the constructor?

app/assets/javascripts/app/calendar.js.coffee
```diff
@@ -0,0 +1,73 @@
...
+      $(".fc-button-next").toggleClass("fc-state-disabled", endDate > window.maxDate)
+
+  buildTooltip: (event, element) ->
+    tooltip = [
+      $.fullCalendar.formatDate(event.start, "h:mmTT"),
+      $.fullCalendar.formatDate(event.end,   "h:mmTT")
+    ].join("&ndash;") + "<br/>"
+
+    # Default for our tooltip is to show.
+    if $("#calendar").data("show-tooltip") != false
```
noelrappin: why `!= false` instead of `if !`

app/assets/javascripts/app/calendar.js.coffee
```diff
@@ -0,0 +1,73 @@
...
+      $("#overlay").addClass("on").removeClass("off")
+    else
+      $("#overlay").addClass("off").removeClass("on")
+
+  toggleNextPrev: (view) ->
+    # window.minDate/maxDate are strings formatted like 20170714
+    try
+      startDate = $.fullCalendar.formatDate(view.start, "yyyyMMdd")
+      endDate   = $.fullCalendar.formatDate(view.end, "yyyyMMdd")
+
```
noelrappin: possibly pull the format into a method so as not to duplicate the format string?

app/assets/javascripts/app/calendar.js.coffee
jhanggi: I forget exactly, but I suspect it it has to do with truthiness in js. There are three cases for the value of the data: `true`, `false`, `undefined`. We want the condition to run for `true` and `undefined`. We only want to not hit it when the value is explicitly `false`.

app/assets/javascripts/app/calendar.js.coffee
jhanggi: I didn't really change this piece of the code, other than to add a comment about why it's this way, but pulling it into a method is probably a good idea.

jhanggi:
Updated.

coffeelint.json
```diff
@@ -65,7 +65,7 @@
     "console": true
   },
   "no_empty_functions": {
-    "level": "error"
+    "level": "ignore"
```
meara: 👍

jhanggi:
Forgot to update the usages after I changed the constructor arguments. Because JS is so forgiving, it still loaded just fine, it just ignored options in `instruments.js`.

PR #1171, opened 2017-07-17
[#136735] First pass at logging events.

noelrappin:
Okay, second try at architecture.

There's a model for `LogEvents`

There’s a class method of `LogEvent` called by the relevant controller methods (did I catch them all)

I kept the `created_by` in `User`, but I can take it out if you don't want it.


db/migrate/20170713183443_create_log_events.rb
```diff
@@ -0,0 +1,11 @@
+class CreateLogEvents < ActiveRecord::Migration
+  def change
+    create_table :log_events do |t|
+      t.references :loggable, polymorphic: true
+      t.string :event_type
+      t.references :user, index: true, foreign_key: false
```
noelrappin: Foreign key is false because none of the `created_by` or `updated_by` columns are foreign keys. Also, I want to minimize the chance this save causes an error

app/models/log_event.rb
```diff
@@ -0,0 +1,6 @@
+class LogEvent < ActiveRecord::Base
+
+  belongs_to :user
```
meara: If  `user_id` isn't a foreign key, why does LogEvent belong_to user here? I don't quite follow how you're thinking about the relationship between LogEvent and user.

noelrappin: The `user_id` still references a user in the ideal case, even if we aren't sure enough that it's always a real user enough to force the database to enforce it.

db/migrate/20170713183443_create_log_events.rb
jhanggi: I think I'd rather see this be a foreign key, but still nullable.

db/migrate/20170713183443_create_log_events.rb
```diff
@@ -0,0 +1,11 @@
+class CreateLogEvents < ActiveRecord::Migration
+  def change
+    create_table :log_events do |t|
+      t.references :loggable, polymorphic: true
+      t.string :event_type
+      t.references :user, index: true, foreign_key: true, nil: false
```
jhanggi: Did you mean `null: false`?

spec/services/account_builder_spec.rb
```diff
@@ -211,6 +211,10 @@
                                        })
     end
 
+    before(:example) do
+      allow(account).to receive(:destroyed?).and_return(false)
```
jhanggi: What is this for?

app/models/log_event.rb
```diff
@@ -0,0 +1,11 @@
...
+  belongs_to :loggable, polymorphic: true
+
+  def self.log(loggable, event_type, user)
+    LogEvent.create(
```
jhanggi: Can we drop the the `LogEvent` here and just call `create`?

app/controllers/account_users_controller.rb
```diff
@@ -47,6 +48,7 @@ def destroy
     @account_user.deleted_by = session_user.id
 
     if @account_user.save
```
jhanggi: Not the fault of the new code, but the new code sheds light on the existing code. Might be for another time, but the `if save` right next to a logging `delete` seems counter-intuitive. I wonder if it would be better to create a custom `destroy` method on `AccountUser`.

app/controllers/users_controller.rb
```diff
@@ -65,6 +65,8 @@ def create
   def create_external
     @user = User.new(params[:user])
     @user.password = generate_new_password
+    @user.created_by = current_user.try(:id)
```
jhanggi: I'm not a fan of our existing relationships where `created_by` is an ID, rather than a `belongs_to`. I would lean towards changing the column to `created_by_id` and adding a `belongs_to`.

db/migrate/20170713183443_create_log_events.rb
```diff
@@ -0,0 +1,11 @@
+class CreateLogEvents < ActiveRecord::Migration
+  def change
+    create_table :log_events do |t|
+      t.references :loggable, polymorphic: true
```
jhanggi: We should probably add an index on `loggable_type, loggable_id`.

spec/services/account_builder_spec.rb
noelrappin: Oh, actually left over from when I had the code in account builder. Disregard.

app/controllers/account_users_controller.rb
noelrappin: I think I'd probably call the method `soft_delete`, but I don't really have a strong feeling about it.

app/controllers/users_controller.rb
noelrappin: I think I'd maybe leave this right now for consistency and then change all of them together?

That said, I don't really need the `created_by` anymore, we can just ditch it.

jhanggi: I'm fine ditching it.

If you do keep it, a few of our classes have `belongs_to :created_by_user, class_name: "User", foreign_key: :created_by` (which I think is awkward, but I think we did that to avoid changing any of the column names, which we don't like doing often because NU has a business analysis tool that sits on top of the DB and changing column names sometimes affects their queries.)

db/migrate/20170713183443_create_log_events.rb
```diff
@@ -0,0 +1,14 @@
...
```
jhanggi: Could you include the index with `index: true` here? I thought that it worked for polymorphic references as well.

app/controllers/users_controller.rb
meara: @noelrappin looks like you dropped the DB column but missed this reference, making your tests sad.

app/controllers/users_controller.rb
```diff
@@ -83,6 +85,8 @@ def create_internal
       flash[:error] = text("users.search.user_already_exists", username: @user.username)
       redirect_to facility_users_path
     elsif @user.save
+      @user.update(created_by: current_user)
```
meara: This should go too probably

spec/controllers/users_controller_spec.rb
```diff
@@ -208,6 +208,7 @@
             it_should_allow_operators_only :redirect do
               expect(assigns(:user)).to be_kind_of User
               expect(assigns(:user)).to be_persisted
+              expect(assigns(:user).created_by).to be_truthy
```
meara: Needs to go.

PR #1172, opened 2017-07-19
[#137184] fix fulfilled_at bug

meara:
- https://pm.tablexi.com/issues/137184

The fulfilled_at value was being overwritten to nil when updating problem orders for reservation-only instruments.

How to reproduce:
1) Place a reservation in the past for another user on a reservation only instrument that is missing a price policy for that date (this is not a problem for timer instruments)
2) The reservation should appear in "Problem Reservations", and it will have a fulfillment date at this point.
3) Create a price policy that covers the fulfillment date
4) In problem reservations, open up the order detail modal and click save
5) The order will no longer be in problem reservations, but if you open up the modal for that reservation, the fulfillment date is gone.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -942,7 +942,7 @@ class OrderDetail < ActiveRecord::Base
 
   def update_fulfilled_at_on_resolve
     if problem_changed? && !problem_order?
-      self.fulfilled_at = time_data.actual_end_at
+      self.fulfilled_at = time_data.actual_end_at if time_data.actual_end_at
     end
   end
 
--- a/spec/models/order_detail_spec.rb
+++ b/spec/models/order_detail_spec.rb
@@ -409,13 +409,14 @@ RSpec.describe OrderDetail do
 
     context "with instruments" do
       let(:instruments) do
-        create_list(:instrument, 4, facility_account: facility_account, facility: facility)
+        create_list(:instrument, 5, facility_account: facility_account, facility: facility)
       end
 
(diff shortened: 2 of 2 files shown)


app/models/order_detail.rb
```diff
@@ -942,7 +942,7 @@ def clear_statement
 
   def update_fulfilled_at_on_resolve
     if problem_changed? && !problem_order?
-      self.fulfilled_at = time_data.actual_end_at
+      self.fulfilled_at = time_data.actual_end_at if time_data.actual_end_at
```
jhanggi: Would doing `self.fulfilled_at ||= time_data.actual_end_at` make sense? There's two scenarios to think about:
1) Resolving missing actual time (I believe that's what this is ultimately trying to handle)
2) Resolving missing price policies (it probably shouldn't change much in that case)

meara: So this is a callback that happens anytime we save an order detail, not just for problem orders. Will we never have a case where `time_data.actual_end_at` changes, but the previous value is not nil?

```
before_save :set_problem_order
  def set_problem_order
    self.problem = problem_description_key.present?
    update_fulfilled_at_on_resolve if time_data.present?
    true # problem might be false; we need the callback chain to continue
  end
```

jhanggi: There's probably a weird edge case where `actual_end_at` might change, so maybe `||=` doesn't make sense. I think what you have sounds reasonable. Hopefully when you're writing specs you don't run into anything funky.

PR #1173, opened 2017-07-20
[#137096] Replaces Vestal Versions with Paper Trail

noelrappin:
There’s not much to this

* Swap gems
* Swap database tables
* Swap config 
* Add the whodunnit config
* Swap in OrderDetail
* Change some syntax in tests


db/migrate/20170719210832_move_versions_table.rb
```diff
@@ -0,0 +1,5 @@
+class MoveVersionsTable < ActiveRecord::Migration
+  def change
+    rename_table :versions, :vestal_versions
```
jhanggi: Can you confirm that the oracle adapter support this? This is the kind of thing where I'd be afraid it doesn't.

db/migrate/20170720161356_create_versions.rb
```diff
@@ -0,0 +1,36 @@
...
+class CreateVersions < ActiveRecord::Migration
+
+  # The largest text column available in all supported RDBMS is
+  # 1024^3 - 1 bytes, roughly one gibibyte.  We specify a size
+  # so that MySQL will use `longtext` instead of `text`.  Otherwise,
+  # when serializing very large objects, `text` might not be big enough.
+  TEXT_BYTES = 1_073_741_823
+
+  def change
+    create_table :versions, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_general_ci" do |t|
```
jhanggi: This is mysql specific. Can we update it to be more db-agnostic?

db/migrate/20170719210832_move_versions_table.rb
jhanggi: We should probably also test this against the version of mariadb DC is on and mysql UIC is on in terms of automatically renaming indexes and foreign keys. That bit us recently.

db/migrate/20170720161356_create_versions.rb
noelrappin: Hmm... This is what the gem generates when it sees that your database is MySQL. I doubt there's a default oracle configuration, so we'll have to figure out what changes to make.

noelrappin:
If I understand how these things work, I guess you could have vestal in the order_detail class, but with it set not to save any more data. But... they both want to use the `versions` table, and I suspect there are other name collisions if they are both on. I wouldn't recommend it. Maybe we can write a small script or model around the vestal table to make it easier to get at the data?

db/migrate/20170720161356_create_versions.rb
jhanggi: I think we can probably just ignore the `options` altogether.

jhanggi:
I think I'd be fine if we had a basic model with a `has_many` on the order detail. Just as long as I can get a pretty view of the changes. I often use `pp od.versions.order(:id).pluck(:created_at, :modifications)` (the `order` because oracle doesn't default to id ordering). If I can do something similar, I'd be good.

db/migrate/20170720161356_create_versions.rb
noelrappin: Would the oracle adapter just ignore these, I wonder?

noelrappin: Probably not, I suppose

db/migrate/20170719210832_move_versions_table.rb
noelrappin: Per the gem source, oracle does rename the table and indexes. (https://github.com/rsim/oracle-enhanced/blob/c2caf15919e3d34129287958be4f2301a595b972/lib/active_record/connection_adapters/oracle_enhanced/schema_statements.rb). 

The other... I guess we need to test, as far as I can see it should work.

db/migrate/20170720161356_create_versions.rb
noelrappin: Looking at the source, it seems like the Oracle adapter will try and drop the options in directly, so we probably should just remove it.

app/models/vestal_version.rb
```diff
@@ -0,0 +1,11 @@
+# This is a _temporary_ class to make up for the fact that the
+# vestal versions gem is no longer in the project. It's here to
+# allow access to the vestal data.
+# once that data is no longer needed, this class can be removed
+# along with the has_many clause in OrderDetail
+
+class VestalVersion < ActiveRecord::Base
+
+  belongs_to :versioned, polymorphic: true
```
jhanggi: Can we add `serialize :modifications`? That's what converts the nasty stored text into something readable.

noelrappin: wish, command, etc...

PR #1174, opened 2017-07-20
[#137190] note error message

meara:
- https://pm.tablexi.com/issues/137190

If you try to change a note of an in-progress reservation, you receive the error message "Actual duration may not be blank" but it still lets the note get changed even though you have to use cancel to get out of the screen.

In process Reservation
Click on the Order Number
Change note
Click Save
See error message
Click Cancel
Note is changed


meara:
@noelrappin If the `editing_time_data` flag is true, there is an extra validation that runs on reservation#save.  That validation wasn't necessary for the use case where we were seeing the error--we moved it to a more specific location so the validation would only run when needed.

https://github.com/tablexi/nucore-open/blob/master/app/support/reservations/validations.rb#L26

PR #1177, opened 2017-07-21
[#137189] feature flag ready for journal badges/alerts

meara:
- https://pm.tablexi.com/issues/137189

UIC doesn't use internal billing currently, and would like "ready for journal" badges and notices hidden to avoid confusing facility directors.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -125,7 +125,7 @@ feature:
   set_statement_search_start_date_on: false
   send_statement_emails_on: true
   responsive_on: true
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -125,6 +125,7 @@ feature:
   set_statement_search_start_date_on: false
   send_statement_emails_on: true
   responsive_on: true
+  ready_for_journal_on: true
```
meara: This will need to be switched to false when we pull these changes into UIC.

PR #1178, opened 2017-07-21
[#132893] time based services

meara:
- https://pm.tablexi.com/issues/132893

New product type, time-based services.

- When setting up price policies, set pricing by hours, similar to instruments. Also have subsidy.
- Charge by the minute
- Use existing time-based input (hour:minute) on front-end and editing in popup.
- Only keeping track of duration, no start/end times


meara:
This is getting sufficiently massive that it could probably benefit from some feedback as I continue working on it. Specs and handling product accessory joins still to come.

app/views/facilities/show.html.haml
```diff
@@ -2,15 +2,17 @@
   = current_facility
 
 - if acting_as? || session_user.try(:operator_of?, current_facility)
-  - instruments = current_facility.instruments.active_plus_hidden
-  - items       = current_facility.items.active_plus_hidden
-  - services    = current_facility.services.active_plus_hidden
-  - bundles     = current_facility.bundles.active_plus_hidden
+  - instruments    = current_facility.instruments.active_plus_hidden
```
jhanggi: If we're going to update the style/format of this code, I would lean towards getting rid of all the extraneous space before the `=`. It'll make them ragged, but it'll make diffs cleaner.

app/views/instrument_price_policies/_amount_row.html.haml
```diff
@@ -4,13 +4,14 @@
...
     display_usage_rate(price_group, price_policy), size: 8, class: "#{price_group.master_internal? ? 'master_usage_cost' : ''} usage_rate"
 
-%td
-  = t("price_policies.amount")
-  %br
-  = text_field_tag "price_policy_#{price_group.id}[minimum_cost]",
-    params["price_policy_#{price_group.id}"] ? params["price_policy_#{price_group.id}"][:minimum_cost] : number_to_currency(price_policy.minimum_cost, unit: "", delimiter: ""),
-    size: 8,
-    class: price_group.master_internal? ? "master_minimum_cost" : ""
+- if local_assigns[:minimum_cost] != false
```
jhanggi: I wonder if at this point, we should pull this out of `instrument_price_policies` into a more shared location?

app/views/instrument_price_policies/_table.html.haml
```diff
@@ -6,7 +6,8 @@
       %th.actions{ rowspan: 2 }
       %th{ rowspan: 2 }= "#{PriceGroup.model_name.human} (#{price_policy.class.human_attribute_name(:type)})"
       %th.currency{ colspan: 2 }= price_policy.class.human_attribute_name(:hourly_usage_rate)
-      %th.currency{ rowspan: 2 }= price_policy.class.human_attribute_name(:minimum_cost)
+      - if local_assigns[:cancellation] != false
```
jhanggi: Should this be testing for `local_assigns[:minimum_cost]`?

app/views/orders/_timed_service_desc.html.haml
```diff
@@ -0,0 +1 @@
+= render :partial => 'orders/product_description', :locals => { :order_detail => order_detail }
```
jhanggi: `render "orders/product_description", order_detail: order_detail`

app/views/timed_service_price_policies/_adjustment_row.html.haml
```diff
@@ -0,0 +1 @@
+= render "instrument_price_policies/adjustment_row", price_group: price_group, price_policy: price_policy, cancellation: false, minimum_cost: false
```
jhanggi: If we do move stuff out of `instrument_price_policies`, does it make sense for the default to now be `false` and you pass in `true` if you want these extra features?

app/views/instrument_price_policies/_price_policy_fields.html.haml
```diff
@@ -5,7 +5,7 @@
 
 - price_policy = @price_policies.first
 = f.input :charge_for,
-  collection: charge_for_options(price_policy.product),
+  collection: local_assigns[:collection] ? collection : charge_for_options(price_policy.product),
```
jhanggi: I would lean towards calling the option something like `charge_for_collection`, and if it's not there, don't show the dropdown at all.

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -28,13 +28,13 @@ def calculate_discount(start_at, end_at)
 
     private
 
-    def cost_and_subsidy(duration_mins, discount)
-      costs = { cost: duration_mins * usage_rate * discount }
+    def cost_and_subsidy(duration_mins, discount_multiplier)
```
jhanggi: I like this naming change.

app/models/timed_service_price_policy.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    calculate_for_time(duration)
+  end
+
+  private
+
+  def calculate_for_time(duration)
+    start_at = Time.current.beginning_of_day
+    end_at = start_at + duration.minutes
+    PricePolicies::TimeBasedPriceCalculator.new(self).calculate(start_at, end_at)
```
jhanggi: Something about this feels awkward to me. We're making times that don't really have a meaning. I would probably lean toward doing a straight `{ cost: usage_rate * quantity, usage_subsidy * quantity }`. That also avoids an awkward check for `schedule_rules` in the calculator.

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -14,8 +14,8 @@ def initialize(price_policy)
     def calculate(start_at, end_at)
       return if start_at > end_at
       duration_mins = TimeRange.new(start_at, end_at).duration_mins
-      discount = calculate_discount(start_at, end_at)
-      cost_and_subsidy(duration_mins, discount)
+      discount_multiplier = product.try(:schedule_rules) ? calculate_discount(start_at, end_at) : 1
```
jhanggi: If we are going to keep this around, I would lean towards using `respond_to?` rather than `try` as it conveys the intention a little better.

app/views/facilities/show.html.haml
```diff
@@ -2,15 +2,17 @@
...
+  - items          = current_facility.items.active_plus_hidden
+  - services       = current_facility.services.active_plus_hidden
+  - timed_services = current_facility.timed_services.active_plus_hidden
+  - bundles        = current_facility.bundles.active_plus_hidden
```
jhanggi: Should this also have the `reject` on it like the else clause? Not that worried about it since it wasn't there before.

app/views/instrument_price_policies/_amount_row.html.haml
meara: Ah, that's an idea. Yeah, probably.

app/views/facilities/show.html.haml
meara: I feel like it's possible that facility managers would find having the old bundles visible to be useful--I don't know if they use that for anything currently but wouldn't want to break it if they do.

jhanggi:
I'm about to sit down and review the code, and probably do some testing. It might be a good idea to make a branch in NU (and maybe the other schools) with this just to make sure the tests are all passing.

Also, would it be possible to update the description to be a more user-focused explanation of the feature, maybe with a couple screenshots? I'd like something that I can reference when I put the release notes out for DC/UIC who don't already have context on the feature.

config/locales/views/admin/en.products.yml
```diff
@@ -34,6 +34,10 @@ en:
         index:
           description: Services are provided by the !facility_downcase! on behalf of the user. Forms for collection of additional information may be attached to the service.  Results files may be returned to the user.
 
+      timed_services:
+        index:
+          description: Timed services are provided by the !facility_downcase! on behalf of the user; billing is based on time used.
```
jhanggi: Maybe add, "charge by the minute" or something similar?

app/models/timed_service_price_policy.rb
```diff
@@ -0,0 +1,29 @@
+class TimedServicePricePolicy < PricePolicy
+
+  include PricePolicies::Usage
+
+  CHARGE_FOR = { usage: "usage" }.freeze
```
jhanggi: Can we get rid of this and the associated dropdown? Since it would only be one element in dropdown, there's no real need for it to be there.

app/models/product.rb
```diff
@@ -60,16 +60,16 @@ class Product < ActiveRecord::Base
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
 
   def self.types
-    @types ||= [Instrument, Item, Service, Bundle]
+    @types ||= [Instrument, Item, Service, TimedService, Bundle]
```
jhanggi: Part of me wants to DRY up these lists, but I'm not quite sure the best way to do it. We could make a constant `DEFAULT_TYPES = [..]` and then this and mergeable_types could do `@types ||= DEFAULT_TYPES.dup`. I'm also ok on punting on that.

app/presenters/order_detail_presenter.rb
```diff
@@ -3,6 +3,7 @@ class OrderDetailPresenter < SimpleDelegator
   include ActionView::Helpers::NumberHelper
   include DateHelper
   include Rails.application.routes.url_helpers
+  include PriceDisplayment
```
jhanggi: We don't need to do it now, but now that we have this, I'd kind of like to go through and get rid of some of the `.send(:extend, PriceDisplayment)` and replace them with the presenter.

app/views/facilities/show.html.haml
```diff
@@ -24,12 +26,14 @@
   = form_for @order_form, url: add_order_path(acting_user.cart(session_user)), html: {method: :put } do |f|
     = render partial: 'product_list', locals: {products: instruments, f: f, title_extra: daily_view_link }
     = render partial: 'product_list', locals: {products: services, f: f}
+    = render partial: 'product_list', locals: {products: timed_services, f: f}
```
jhanggi: I initially noticed the `partial`/`local` and single quotes we want to get rid of, but let's save the linting here for another time in order to avoid obfuscating the diff.

app/views/timed_service_price_policies/_adjustment_row.html.haml
```diff
@@ -0,0 +1 @@
+= render "time_based_price_policies/adjustment_row", price_group: price_group, price_policy: price_policy, cancellation: false, minimum_cost: false
```
jhanggi: If the default for `cancellation` and `minimum` cost are essentially `false`, do we need to include them here? I would probably accept the argument that this is more explicit.

app/views/timed_service_price_policies/_price_policy_fields.html.haml
```diff
@@ -0,0 +1 @@
+= render "time_based_price_policies/price_policy_fields", f: f, cancellation: false, minimum_cost: false, charge_for_collection: TimedServicePricePolicy::CHARGE_FOR.map { |k, v| [k.to_s.titleize, v] }
```
jhanggi: Here's where I would like to get rid of the CHARGE_FOR.

app/services/price_policies/time_based_price_calculator.rb
noelrappin: 👍

app/models/order_detail.rb
```diff
@@ -250,7 +250,7 @@ def can_be_viewed_by?(user)
 
   # Supports single type or an array of product types
   scope :for_product_type, ->(product_type) { joins(:product).where(products: { type: product_type }) }
-  scope :untimed_orders, -> { for_product_type(["Item", "Service"]) }
+  scope :item_and_service_orders, -> { for_product_type(["Item", "Service", "TimedService"]) }
```
waggles: is there a categorization here that could explain a bit more about why these are grouped together? "untimed" gave a little hint as to what the purpose of the group was

app/views/timed_service_price_policies/_adjustment_row.html.haml
waggles: I support changing this and the weird `!= false` logic

app/models/order_detail.rb
jhanggi: It has to do with where the orders end up in the UI. Do they go under "My Orders" or "My Reservations" for the user who purchased them? Do they go under "Orders", "Reservations", or "Occupancies" on the admin side? Essentially, Reservations and Occupancies are treated somewhat specially, and then items and services are lumped together as the rest. `untimed_orders` sort of called this out better, but now that we have timed services that need to be in "the rest", it made at name not quite correct. This grouping is basically every product type that isn't special.

app/views/facility_orders/index.html.haml
```diff
@@ -28,20 +28,18 @@
...
+            %td= human_datetime(od.order.ordered_at)
+            %td= od.order.user.full_name
+            %td.centered= OrderDetailPresenter.new(od).wrapped_quantity
             = render :partial => 'shared/order_detail_cell', :locals => { :od => od }
             %td= od.assigned_user
             %td= od.order_status
             %td.currency
-              - od.send(:extend, PriceDisplayment)
-              = od.wrapped_total
+              =  OrderDetailPresenter.new(od).wrapped_total
```
waggles: should we be handling object creation like this in the controller code? it feels heavy for view logic

app/views/timed_service_price_policies/_table.html.haml
```diff
@@ -0,0 +1 @@
+= render "time_based_price_policies/table", price_policies: price_policies, url_date: url_date, cancellation: false, minimum_cost: false
```
waggles: it'd be cool if we could find a way to not have to have this extra layer and just render this partial directly hmm

app/views/facility_orders/index.html.haml
jhanggi: I don't think so. Since this is using a presenter, I would actually consider it view code. We're in a loop over many order details, so it would be tricky to get this into a controller. There's also a problem Meara and I ran into yesterday where cancan doesn't like presenters/decorators/delegators. `can?(:something, decorated_object)` will return false while the `can?(:something, object)` will return true. So we do want to keep the objects unwrapped where we don't need them.

config/locales/views/admin/en.products.yml
meara: The text in this spot for instruments doesn't specify units--the pricing units are specified in the form (for both Instruments and TimedServices). I think it would be kind of confusing to include the units here, honestly--because we charge by the minute, but they set the rates by hour. It's much clearer in the form.

app/models/product.rb
meara: Let's punt for now.

app/presenters/order_detail_presenter.rb
meara: I think that's a great idea, but I'll make a separate task for it if that's cool.

app/views/timed_service_price_policies/_adjustment_row.html.haml
meara: Waggles and I discussed, and we concluded that having the `cancellation` and `minimum_cost` locals explicitly assigned here helps to make it more clear what the purpose of these partials are.

app/views/timed_service_price_policies/_table.html.haml
meara: Resolved by the discussion in https://github.com/tablexi/nucore-open/pull/1178#discussion_r129665919

meara:
These two bullet points I think belong in a separate story:
- The email receipt should probably be displayed as H:MM. I'd be ok with this being a followup since it might be a little tricky since I'm pretty sure we're relying on JS to switch to time-based displays:.
- The statement PDF for open does not have quantity, but all 3 schools have quantity on their PDF, so we'll want to display as a time there. I would vote for adding quantity to open's default PDF and then we'll adjust the individual schools when we merge this down.

We'll also add stories for:
- handling product accessories so that services/timed services can only be added with the correct charge mode.
- cleaning up some of the shared product views (linting etc)

Other than what will be handled in other stories, I think this should be good to go, unless anyone has any other comments.

app/views/timed_service_price_policies/_adjustment_row.html.haml
jhanggi: 👍

meara:
@jhanggi the default 30 minutes is something Waggles and I actually added on purpose--the previous default was 1 minute, which seemed awkward. We're setting it here: https://github.com/tablexi/nucore-open/pull/1178/files#diff-1946e0a50c928addd9705c36a3b7dd5cR4

Is there another default that we think would make more sense?

jhanggi:
I agree the one minute was awkward. 30 minutes is a definitely a reasonable default for the cart because you have the opportunity to edit it before you purchase it. My concern with the add to existing order workflow is that if there is no order form, the order detail gets automatically purchased (and possibly even completed). This might be fine assuming we don't send any emails (I'm pretty sure we don't) because you can still edit it right away.

It might be worth doing a short demo of this overall feature with NU this afternoon to get their feedback, especially around some of the decisions we made on the order form and whatnot.

jhanggi:
I think if we can drop the 30 minute default here, we can deal with everything else in separate PRs.

meara:
@jhanggi updated

app/support/orders/item_adder.rb
```diff
@@ -1,5 +1,8 @@
 class Orders::ItemAdder
 
+  # time based services default to 1 minute (arbitrary)
+  DEFAULT_DURATION = 1
```
jhanggi: Should we give a more specific name for this constant? `DEFAULT_TIMED_SERVICE_DURATION` or something similar? Mainly because it doesn't apply to instruments, but you could expect it to.

PR #1179, opened 2017-07-26
[#137095] Add Global UI for Log Events

noelrappin:
* Added a separate event_time to the log_events table rather than depending on `created_at`

* Added a new global settings for viewing log events, which can filter on dates or kind of event (text search of object names is the next step)

* Added search functionality to the `LogEvent` class

* Added a script to partially seed the log event table from Account, User, and AccountUser. Mostly for testing, but it could be run in production if needed.


app/models/log_event.rb
```diff
@@ -3,8 +3,44 @@ class LogEvent < ActiveRecord::Base
...
+
+  def self.search(start_date: nil, end_date: nil, events: [])
+    result = all
+    result = result.filter_date(start_date, end_date) if start_date || end_date
+    result = result.filter_event(events) if events.present?
+    result
+  end
+
+  def self.filter_date(start_date, end_date)
+    dates = [(start_date || 10.years.ago), (end_date || Date.current)]
```
jhanggi: Will this filter to the end of day for `end_date`? If I search for 7/11-7/24, we should include 7/24. That keeps it consistent with other searches in nucore.

app/controllers/account_users_controller.rb
```diff
@@ -31,7 +31,8 @@ def create
     @account_user.created_by = session_user.id
 
     if @account_user.save
-      LogEvent.log(@account_user, :create, current_user)
+      LogEvent.log(
+        @account_user, :create, @account_user.created_at, current_user)
```
jhanggi: I get why we may want to be able to log an event to the past to deal with importing, but I'm not sure I see the value in requiring passing a timestamp in every time. What if we made the timestamp an optional keyword argument that defaults to `Time.current`?

app/controllers/users_controller.rb
```diff
@@ -65,7 +65,7 @@ def create
   def create_external
     @user = User.new(params[:user])
     @user.password = generate_new_password
-    LogEvent.log(@user, :create, current_user)
+    LogEvent.log(@user, :create, @user.created_at, current_user)
```
jhanggi: On that same note about timestamps, I don't think `created_at` will have been set yet because it hasn't been saved. It's possible I'm wrong.

app/models/log_event.rb
```diff
@@ -3,8 +3,44 @@ class LogEvent < ActiveRecord::Base
...
+    where(event_time: (dates.min..dates.max))
+  end
+
+  # When this gets updated to Rails 5, you can use ActiveRecord#or directly
+  def self.filter_event(*events)
```
jhanggi: Can we add some comments around what this is expecting? This part kind of confused me at first, especially the use of the double underscore.

app/models/log_event.rb
```diff
@@ -3,8 +3,44 @@ class LogEvent < ActiveRecord::Base
...
+    where_strings = events.flatten.map do |event|
+      loggable_type, event_type = event.split("__")
+      "(`loggable_type` = '#{loggable_type.camelize}' AND `event_type` = '#{event_type}')"
```
jhanggi: Is there a chance of SQL-injection here? I'm guessing these values are coming from user input.

config/locales/views/admin/en.log_events.yml
```diff
@@ -0,0 +1,18 @@
+en:
+  views:
+    log_events:
+      index:
+        account__create: Payment Source created
```
jhanggi: I wonder if it might be easier to think about if we structured this as
```
log_events:
  index: 
    account:
      create: Payment Source created
      update: Payment Source updated
```

And I could even see this moving into the `LogEvent#to_s` or similar. It doesn't quite feel like it's necessarily `index`-specific. But I could see punting on that.

lib/tasks/seed_log_events.rake
```diff
@@ -0,0 +1,44 @@
...
+
+      AccountUser.all.each do |account_user|
+        next if LogEvent.where(loggable: account_user).present?
+        LogEvent.log(
+          account_user, :create, account_user.created_at,
+          User.find_by(id: account_user.created_by))
+
+        next if account_user.deleted_at.blank?
+        LogEvent.log(
+          account, :delete, account.deleted_at,
```
jhanggi: Should this be `account_user`? I'm getting an error here when I try to run this.

app/views/log_events/index.html.haml
```diff
@@ -0,0 +1,52 @@
...
+    class: "js--chosen",
+    data: { placeholder: text(:event_placeholder) })
+
+  = label_tag(text(:start_date))
+  = text_field_tag(:start_date, params[:start_date], class: :datepicker)
+
+  = label_tag(text(:end_date))
+  = text_field_tag(:end_date, params[:end_date], class: :datepicker)
+
+  = submit_tag(text(:filter))
```
jhanggi: Can we add `class: "btn"` on this so it matches our button styling elsewhere.

app/views/log_events/index.html.haml
```diff
@@ -0,0 +1,52 @@
...
+                        [text("account__update"), "account__update"],
+                        [text("account_user__create"), "account_user__create"],
+                        [text("account_user__delete"), "account_user__delete"],
+                        [text("user__create"), "user__create"]], params[:events]),
+    multiple: true,
+    class: "js--chosen",
+    data: { placeholder: text(:event_placeholder) })
+
+  = label_tag(text(:start_date))
+  = text_field_tag(:start_date, params[:start_date], class: :datepicker)
```
jhanggi: We have a `datepicker__data` class which will get loaded no matter what page it's on. It has a nice feature of being able to set min/max dates for the picker via data attributes (we don't necessarily need that here).

app/models/log_event.rb
```diff
@@ -3,8 +3,92 @@ class LogEvent < ActiveRecord::Base
...
+  def loggable_to_s
+    case loggable
+    when AccountUser
+      "#{loggable.account} / #{loggable.user}"
+    else
+      loggable.to_s
+    end
+  end
+
+  class LogEventSearcher
```
meara: Should this be in its own file, or is there a reason to keep it here?

noelrappin: No strong opinion, it's pretty much a subordinate class, so it seemed to make sense to have it nest, but I can move it out if we prefer.

app/models/log_event.rb
```diff
@@ -3,8 +3,92 @@ class LogEvent < ActiveRecord::Base
   belongs_to :user
   belongs_to :loggable, polymorphic: true
 
-  def self.log(loggable, event_type, user)
-    create(loggable: loggable, event_type: event_type, user_id: user.id)
+  def self.log(loggable, event_type, event_time, user)
```
jhanggi: I feel moderately strongly that we should make the event time an optional keyword argument and get rid of it in most of the calls. I think it really clutters up the API.

app/models/log_event.rb
jhanggi: This class is large enough, I would vote for moving it into its own file.

app/models/log_event.rb
```diff
@@ -3,8 +3,92 @@ class LogEvent < ActiveRecord::Base
...
+    def initialize(start_date: nil, end_date: nil, events: [], query: nil)
+      @start_date = start_date
+      @end_date = end_date
+      @events = filter_events(events)
+      @query = query
+    end
+
+    def filter_events(events)
+      return events if events.blank?
+      events.select { |event| event.in?(ALLOWED_EVENTS) }
```
jhanggi: `Array(events).select {...}` would remove the need for guard clause.

app/models/log_event.rb
```diff
@@ -3,8 +3,92 @@ class LogEvent < ActiveRecord::Base
...
+      where_strings = events.flatten.map do |event|
+        loggable_type, event_type = event.split(".")
+        "(`loggable_type` = '#{loggable_type.camelize}' AND `event_type` = '#{event_type}')"
+      end
+      LogEvent.where(where_strings.join(" OR "))
+    end
+
+    def filter_query
+      accounts = Account.where("account_number LIKE ?", "%#{query}%")
+      users = UserFinder.search(query, 1000)[0]
```
jhanggi: This feels like an arbitrary magic number. Would it be worth updating UserFinder to have the limit be optional and default to no limit?

app/models/log_event.rb
```diff
@@ -3,8 +3,92 @@ class LogEvent < ActiveRecord::Base
...
+    def search
+      result = LogEvent.all
+      result = result.merge(filter_date) if start_date || end_date
+      result = result.merge(filter_event) if events.present?
+      result = result.merge(filter_query) if query.present?
+      result
+    end
+
+    def dates
+      [(start_date || 10.years.ago), (end_date || Date.current)]
```
jhanggi: 10.years.ago feels like an arbitrary magic number. Is there a way to do this range without the magic number?

app/models/log_event.rb
```diff
@@ -3,8 +3,92 @@ class LogEvent < ActiveRecord::Base
...
+        filter_loggable_string("Account", accounts),
+        filter_loggable_string("User", users),
+      ].compact.join(" OR ")
+      LogEvent.where(where_string)
+    end
+
+    def filter_loggable_string(loggable_type, loggable_objects)
+      return nil if loggable_objects.empty?
+      id_string = loggable_objects.map(&:id).join(",")
+      "(`loggable_type` ='#{loggable_type}' AND `loggable_id` IN (#{id_string}))"
```
jhanggi: Will the quoting with backticks cause problems in oracle? Is there a way to get the `to_sql` of a `where(loggable_type: loggable_type, loggable_id: loggable_objects.map(&:id))`.  Is rails smart enough to just do `LogEvent.where(loggable: loggable_objects)`, without needing to do id/type yourself?

lib/tasks/seed_log_events.rake
```diff
@@ -0,0 +1,44 @@
...
+
+      AccountUser.all.each do |account_user|
+        next if LogEvent.where(loggable: account_user).present?
+        LogEvent.log(
+          account_user, :create, account_user.created_at,
+          User.find_by(id: account_user.created_by))
+
+        next if account_user.deleted_at.blank?
+        LogEvent.log(
+          account, :delete, account_user.deleted_at,
```
jhanggi: Shouldn't we be logging to the `account_user`, not the `account`?

config/locales/views/admin/en.log_events.yml
```diff
@@ -0,0 +1,22 @@
+en:
+  views:
+    log_events:
+      index:
+        account:
+          create: Payment Source created
```
jhanggi: I like this structure of the locales much better.

app/models/log_event.rb
noelrappin: Okay. New API is (loggable, event_type, user, event_time: Time.current)

app/models/log_event.rb
noelrappin: done

app/models/log_event.rb
noelrappin: It is arbitrary, the obvious other choice would be `LogEvent.minimum(:event_time)`.

app/models/log_event.rb
noelrappin: I didn't want to touch `UserFinder` if I could avoid it. But yeah, from the standpoint of this consumer, the limit is unneeded. I don't need user_finder to default to no limit as much as it would be useful to have an option to bypass the limit (for example, by passing in `nil`).

app/models/log_event.rb
noelrappin: I have no idea about the backticks in Oracle, we could probably take them out, it's there because it's how ActiveRecord generates code.

I think what I have here is pretty close to the to_sql of `where(loggable_type: loggable_type, loggable_id: loggable_objects.map(&:id))`, at least that was my goal. The tricky part is you have to OR the two types together. 

Rails is not smart enough to do `LogEvent.where(loggable: loggable_objects)` if the objects are of multiple different types (it was the first thing I tried). 

Another option would be something like this:

~~~
accounts = Account.where("account_number LIKE ?", "%#{query}%").map(&:id)
users = UserFinder.search(query, 1000)[0].map(&:id)
account_logs = LogEvent.where(loggable_type: "Account", loggable_id: accounts).map(&:id)
user_logs =  LogEvent.where(loggable_type: "User", loggable_id: users).map(&:id)
where(id: account_logs + user_logs)
~~~

That preserves the scope-ness of the result, but it adds an extra database call. Though give that other calls are simpler, and the final call has an index, it might not be noticeably slower in practice.

app/services/log_event_searcher.rb
```diff
@@ -0,0 +1,62 @@
...
+    where_strings = events.flatten.map do |event|
+      loggable_type, event_type = event.split(".")
+      "(`loggable_type` = '#{loggable_type.camelize}' AND `event_type` = '#{event_type}')"
+    end
+    LogEvent.where(where_strings.join(" OR "))
+  end
+
+  def filter_query
+    accounts = Account.where("account_number LIKE ?", "%#{query}%")
+    users = UserFinder.search(query, 1000)[0]
```
jhanggi: I think you might be able to just pass `nil` in as the second argument to `UserFinder` for now. `ActiveRecord::Relation.limit(nil)` doesn't do anything.

app/services/log_event_searcher.rb
```diff
@@ -0,0 +1,62 @@
...
+  def search
+    result = LogEvent.all
+    result = result.merge(filter_date) if start_date || end_date
+    result = result.merge(filter_event) if events.present?
+    result = result.merge(filter_query) if query.present?
+    result
+  end
+
+  def dates
+    [(start_date || LogEvent.minimum(:event_time)), (end_date || Date.current)]
```
jhanggi: I don't really like the `LogEvent.minimum(:event_time)` because of the extra database call, which once this table gets big will get slow, especially if event_time is not indexed. What if you introduced a `beginning_of_time = 10.years.ago`. Then, while it may be arbitrary, it at least is saying what it is meaning to be.

app/services/log_event_searcher.rb
```diff
@@ -0,0 +1,62 @@
...
+  def filter_date
+    LogEvent.where(event_time: (dates.min.beginning_of_day..dates.max.end_of_day))
+  end
+
+  # When this gets updated to Rails 5, you can use ActiveRecord#or directly
+  # the event name comes in as <loggable_type>.<event_type>
+  def filter_event
+    where_strings = events.flatten.map do |event|
+      loggable_type, event_type = event.split(".")
+      "(`loggable_type` = '#{loggable_type.camelize}' AND `event_type` = '#{event_type}')"
```
jhanggi: I just confirmed the backticks won't work in oracle. Is this something we could get with arel?  It or maybe we just remove the backticks and move on.

app/models/log_event.rb
jhanggi: That's not a bad option. I would do it with a `pluck` over a `map(&:id)`, but it'll be a couple queries, but is probably fast enough. I do feel like this might be clearer, and if we're only doing one type at a time, then could we do `LogEvent.where(loggable: loggable_objects)`.

noelrappin: Yes, pluck for accounts, I don't think I can pluck from UserFinder without modifying UserFinder.

noelrappin: I don't think `LogEvent.where(loggable: loggable_objects)` will work still (at least not and also pluck)

app/services/log_event_searcher.rb
```diff
@@ -0,0 +1,60 @@
...
+  def filter_event
+    where_strings = events.flatten.map do |event|
+      loggable_type, event_type = event.split(".")
+      "(loggable_type = '#{loggable_type.camelize}' AND event_type = '#{event_type}')"
+    end
+    LogEvent.where(where_strings.join(" OR "))
+  end
+
+  def filter_query
+    account_ids = Account.where("account_number LIKE ?", "%#{query}%").pluck(:id)
```
jhanggi: I'm thinking we should probably also include any `AccountUser`s that match the account query. That way if I search for chart string X, I get the creation/updates as well as the add/remove users for that account. Maybe we want to do the same for the users, but I care a little bit less about that. What do you think?

app/views/log_events/index.html.haml
```diff
@@ -0,0 +1,51 @@
...
+    class: "js--chosen",
+    data: { placeholder: text(:event_placeholder) })
+
+  = label_tag(text(:start_date))
+  = text_field_tag(:start_date, params[:start_date], class: :datepicker__data)
+
+  = label_tag(text(:end_date))
+  = text_field_tag(:end_date, params[:end_date], class: :datepicker__data)
+
+  = label_tag(text(:query))
```
jhanggi: Can we improve the styling of the form at all? I don't want to spend a lot of time on it, but the fact that the button and the text input aren't aligned bugs me a little. 

I'm also thinking the query should be first item, and then the type would be the least important filter. I'm guessing (I could of course be wrong) that the most common use case will be to see the history of a single account.

screen shot 2017-07-27 at 5 17 26 pm

noelrappin: How do you feel about 

global_settings_-_nucore

I wonder if the horizontal spacing needs to be closer?

jhanggi: Maybe do the start/end on the right side (we use the date range in a single column on the billing tab) and move the event to the left. I think that will alleviate the spacing problem.

PR #1181, opened 2017-07-27
[#137257] charge mode matches accessory

meara:
When adding accessories to products, the scaling type chosen must make sense with the accessory that is chosen (i.e. items must have quantity, not a time-based scaling type).

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -10,7 +10,7 @@ class Product < ActiveRecord::Base
   has_many   :stored_files
   has_many   :price_groups, through: :price_group_products
   has_many   :price_group_products
-  has_many :product_accessories, -> { where(deleted_at: nil) }, dependent: :destroy
+  has_many   :product_accessories, -> { where(deleted_at: nil) }, dependent: :destroy
   has_many   :accessories, through: :product_accessories, class_name: "Product"
   has_many   :price_policies
   has_many   :training_requests, dependent: :destroy
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -5,15 +5,15 @@ class Product < ActiveRecord::Base
   belongs_to :facility
   belongs_to :initial_order_status, class_name: "OrderStatus"
   belongs_to :facility_account
-  has_many   :product_users
-  has_many   :order_details
-  has_many   :stored_files
-  has_many   :price_groups, through: :price_group_products
-  has_many   :price_group_products
-  has_many   :product_accessories, -> { where(deleted_at: nil) }, dependent: :destroy
-  has_many   :accessories, through: :product_accessories, class_name: "Product"
-  has_many   :price_policies
-  has_many   :training_requests, dependent: :destroy
+  has_many :product_users
+  has_many :order_details
+  has_many :stored_files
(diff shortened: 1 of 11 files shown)


app/models/product.rb
```diff
@@ -10,7 +10,7 @@ class Product < ActiveRecord::Base
   has_many   :stored_files
   has_many   :price_groups, through: :price_group_products
   has_many   :price_group_products
-  has_many :product_accessories, -> { where(deleted_at: nil) }, dependent: :destroy
+  has_many   :product_accessories, -> { where(deleted_at: nil) }, dependent: :destroy
```
jhanggi: If this is just a spacing change, I would not make it, as ideally, the not aligning becomes the default.

app/models/product_accessory.rb
```diff
@@ -9,10 +13,17 @@ class ProductAccessory < ActiveRecord::Base
   ## validations
   validates :product, presence: true
   validates :accessory, presence: true
-  validates :scaling_type, presence: true, inclusion: SCALING_TYPES
+  validates :scaling_type, presence: true
+  validate :scaling_type_matches_product?
```
jhanggi: I would probably lean away from using the `?` in the method name since it's not really a predicate.

app/models/product_accessory.rb
```diff
@@ -1,6 +1,10 @@
 class ProductAccessory < ActiveRecord::Base
 
-  SCALING_TYPES = %w(quantity manual auto).freeze
+  SCALING_TYPES = {
+    item: ["quantity"],
+    service: ["quantity"],
+    timed_service: ["manual", "auto"]
+  }.with_indifferent_access
```
jhanggi: Part of me feels like the scaling types should be the keys and the values the product types. It might also be nice to have that hash be able to default to "quantity". That way if we add a new product type, we don't have to worry about accessories unless it's timed.

app/models/product_accessory.rb
```diff
@@ -9,10 +13,17 @@ class ProductAccessory < ActiveRecord::Base
...
 
   def self.scaling_types
-    SCALING_TYPES
+    SCALING_TYPES.values.flatten.uniq
+  end
+
+  def scaling_type_matches_product
+    return if SCALING_TYPES[accessory.type.underscore].include?(scaling_type)
+
+    errors.add(:scaling_type, "does not match product type")
```
jhanggi: I18n

app/models/product_accessory.rb
```diff
@@ -9,10 +14,17 @@ class ProductAccessory < ActiveRecord::Base
...
 
   def self.scaling_types
-    SCALING_TYPES
+    SCALING_TYPES.values.flatten.uniq
+  end
+
+  def scaling_type_matches_product
+    return if SCALING_TYPES[accessory.type.underscore].include?(scaling_type)
+
+    errors.add(:scaling_type)
```
jhanggi: Hmm. I was kind of surprised by this since you didn't add any locales for it. But it turns out to be ok in terms of messaging:

```
> pa.errors.full_messages
=> ["Scaling Type is invalid"]
```

PR #1184, opened 2017-07-27
[#137106] update font-awesome-rails gem

meara:
- https://pm.tablexi.com/issues/137106


meara:
@jhanggi so, as far as I have found, the calendar icons still work but have changed a bit in appearance. I'm not sure what other icons we use, but I'm guessing the story is similar for those as well.

We don't need to change this right now, but we will before Rails 5--there is no 3.x version of font awesome compatible with Rails 5.

meara:
Found an upgrade guide, https://github.com/FortAwesome/Font-Awesome/wiki/Upgrading-from-3.2.1-to-4

Will upgrade our icon calls in this PR as well.

app/assets/stylesheets/app/sortable.scss
```diff
@@ -1,19 +1,19 @@
 .sortable {
-  background-image: image-url("icon-sort.gif");
+  background-image: image-url("icon-sort.gif"); #
```
meara: @jhanggi not sure what to do about these--I'm pretty sure they depend on font-awesome, but I don't know how this `.gif`thing works or whether I can do it with the new version. Any suggestions?

jhanggi: Honestly, I have no idea. Is font-awesome now all SVG or fonts? If we can't figure it out, maybe we just pull these files out of the old version of font-awesome and include them in nucore.

jhanggi:
Overall this looks pretty good. I haven't noticed anywhere we've missed. There are a couple places where the icons are now different sizes. 

On the main facility list, the instrument not available triangle is smaller than it was.
Stage:
screen shot 2017-07-31 at 3 01 47 pm

Local:
screen shot 2017-07-31 at 3 01 51 pm

The calendar icon feels overly big now. It also seems to mess up the line height/vertical alignment:
Stage:
screen shot 2017-07-31 at 3 02 23 pm
Local:
screen shot 2017-07-31 at 3 02 27 pm

It also looks like we're missing the add accessories icon on facility orders show.
Stage:
screen shot 2017-07-31 at 3 05 03 pm
(it's possible Im' looking at the wrong thing).

There was also a note on #132894 about the underscore.

meara:
@jhanggi -- I updated the size/alignment for the first two issues you flagged.

For the third issue, I'm having trouble getting the add accessories icon to show up even on master. I added an instrument with possible accessories to an order, but it's not showing the icon--is there something particular I need to do with the setup?

jhanggi:
The add accessories icon looks fine. I hadn't looked into what made that visible when I first commented. The order has to be complete in order for that icon to appear. I'm fine if we want to address the underscore on the calendar there in a separate PR.

PR #1185, opened 2017-07-27
[#137103] bundle update awesome-nested-set gem

meara:
https://pm.tablexi.com/issues/137103


jhanggi:
Maybe the removal of the monkey patch would be good to include here?

meara:
@jhanggi monkey patch?

jhanggi:
What waggles did in https://github.com/tablexi/nucore-dartmouth/pull/126

meara:
Yep, will remove that in #1186 I think.

PR #1190, opened 2017-07-31
[#136040] Allow reruns of completed submissions

waggles:
* Adds `complete` status to scope of batchable order_details; the remaining options `reconciled` and `canceled` are not valid for re-run
* Renames and fixes the existing delete spec

### DIFF
--- a/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/submission.rb
+++ b/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/submission.rb
@@ -22,8 +22,14 @@ module SangerSequencing
     alias purchased_at ordered_at
 
     scope :purchased, -> { joins(:order).merge(Order.purchased.order(ordered_at: :desc)) }
-    scope :ready_for_batch, -> { purchased.merge(OrderDetail.new_or_inprocess).where(batch_id: nil) }
     scope :for_facility, ->(facility) { where(orders: { facility_id: facility.id }) }
+
+    BATCHABLE_STATES = %w(new inprocess complete).freeze
+    scope :ready_for_batch, lambda {
+      purchased
+        .merge(OrderDetail.where(state: BATCHABLE_STATES))
+        .where(batch_id: nil)
+    }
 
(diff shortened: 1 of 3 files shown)


vendor/engines/sanger_sequencing/app/models/sanger_sequencing/submission.rb
```diff
@@ -22,9 +22,15 @@ class Submission < ActiveRecord::Base
     alias purchased_at ordered_at
 
     scope :purchased, -> { joins(:order).merge(Order.purchased.order(ordered_at: :desc)) }
-    scope :ready_for_batch, -> { purchased.merge(OrderDetail.new_or_inprocess).where(batch_id: nil) }
     scope :for_facility, ->(facility) { where(orders: { facility_id: facility.id }) }
 
+    BATCHABLE_STATES = %w(new inprocess complete).freeze
```
meara: What happens to the UI if there are a ton of completed submissions? Will that be messy/make it difficult to find the submissions that haven't been run yet?

waggles: If they are attached to a batch, they shouldn't show up in this scope, which I believe is the way this works..? does that make sense

meara: Ah okay, so this makes sense if a submission gets completed accidentally without getting attached to a batch. What if the submission is run as part of a batch, but for some reason or other needs to be re-run? If I'm understanding the story correctly, that's in scope...

waggles: so I think that's covered in ticket 136038, which is highly tied to this one

meara: Okay. I think it may be worth checking with the client whether that ticket covers (or could be made to cover) a case where a batch was actually run, but the samples may need to be run again (i.e. if the data was bad). I'm not sure if that's actually a situation they have but that was my impression.

PR #1191, opened 2017-07-31
update deprecated calls

meara:
Noticed these while doing some Rails 5 prep/gem upgrades. Seemed worth cleaning up.

Doing https://pm.tablexi.com/issues/137255 as well, while I'm at it.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -18,21 +18,19 @@ module ApplicationHelper
     (full_title + app_name).html_safe
   end
 
-  def order_detail_description(order_detail) # TODO: deprecate in favor of OrderDetailPresenter#description_as_html
+  def order_detail_description(order_detail)
+    warn "Depecrated order_detail_description called. Use OrderDetailPresenter#description_as_html instead. At #{caller(1..1)}"
     OrderDetailPresenter.new(order_detail).description_as_html
   end
 
-  def order_detail_description_as_html(order_detail) # TODO: deprecate in favor of OrderDetailPresenter#description_as_html
+  def order_detail_description_as_html(order_detail)
+    warn "Depecrated order_detail_description_as_html called. Use OrderDetailPresenter#description_as_html instead. At #{caller(1..1)}"
     OrderDetailPresenter.new(order_detail).description_as_html
   end
 
-  def order_detail_description_as_text(order_detail) # TODO: move this into OrderDetailPresenter
-    name = order_detail.product.to_s
-    if order_detail.bundle
-      name.prepend("#{order_detail.bundle} -- ")
-    else
-      name
-    end.html_safe
+  def order_detail_description_as_text(order_detail)
+    warn "Depecrated order_detail_description_as_text called. Use OrderDetailPresenter#description_as_text instead. At #{caller(1..1)}"
+    OrderDetailPresenter.new(order_detail).description_as_text
   end
 
   def sortable(column, title = nil)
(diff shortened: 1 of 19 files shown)


app/views/facility_journals/_problem_transactions.html.haml
```diff
@@ -31,7 +31,7 @@
             %td= format_usa_date(order_detail.public_send(:"#{@date_range_field}"))
             %td.user-order-detail.order-note
               .order-detail-description
-                = order_detail_description(order_detail)
+                = OrderDetailPresenter.new(order_detail).description_as_html
                 - order_detail.extend(PriceDisplayment)
```
jhanggi: Should we kill this `PriceDisplayment` while we're here?

app/helpers/application_helper.rb
```diff
@@ -18,21 +18,19 @@ def html_title(title = nil)
     (full_title + app_name).html_safe
   end
 
-  def order_detail_description(order_detail) # TODO: deprecate in favor of OrderDetailPresenter#description_as_html
+  def order_detail_description(order_detail)
```
jhanggi: If there aren't anymore of these left, should we get rid of the methods?

meara: Not sure if there might be any in school specific engines, which is why I've got the deprecation warnings here.

jhanggi: 👍

app/presenters/order_detail_presenter.rb
```diff
@@ -18,6 +18,15 @@ def description_as_html
...
   end
 
+  def description_as_text
+    name = product.to_s
+    if bundle
+      name.prepend("#{bundle} -- ")
+    else
+      name
+    end.html_safe
+  end
```
waggles: this saves duplication but results in some slightly-too-complex code paths, what do you think about flattening some of the logic?

meara: You're right this is not the most beautiful method. I basically copy and pasted this from the application_helper, don't feel super motivated to change it now. 😜

app/views/notifier/order_receipt.html.haml
```diff
@@ -35,8 +35,7 @@
       %th= OrderDetail.human_attribute_name("#{label_key_prefix}_total")
       %th= OrderDetail.human_attribute_name(:note).pluralize
   %tbody
-    - @order.order_details.each do |order_detail|
-      - order_detail.send(:extend, PriceDisplayment)
+    - OrderDetailPresenter.wrap(@order.order_details).each do |order_detail|
```
jhanggi: Looks like we're double wrapping in line 42 now.

app/views/orders/receipt.html.haml
```diff
@@ -31,8 +31,7 @@
         %th.currency= t('.th.adjust')
       %th.currency= t('.th.total')
   %tbody
-    - @order_details.each do |order_detail|
-      - order_detail.send(:extend, PriceDisplayment)
+    - OrderDetailPresenter.wrap(@order_details).each do |order_detail|
```
jhanggi: Doing some double wrapping in line 40.

app/models/reports/account_transactions_report.rb
```diff
@@ -72,14 +72,14 @@ def headers
...
-    order_detail.extend(PriceDisplayment)
+    order_detail = OrderDetailPresenter.new(order_detail)
 
     [
       order_detail.order.id,
       order_detail.id,
       format_usa_date(order_detail.send(:"#{@date_range_field}")),
       order_detail.order.facility,
-      order_detail_description(order_detail),
+      OrderDetailPresenter.new(order_detail).description_as_html,
```
jhanggi: Is this double wrapping?

meara:
Updated. Just checked again for `can?` calls anywhere within the blocks and we should be good.

PR #1192, opened 2017-07-31
[#136038] Update default group scope for Batch

waggles:
If a user visits the Batch index page without specifying a `group` param, Batches with `group == ""` are not displayed.

### DIFF
--- a/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/batch.rb
+++ b/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/batch.rb
@@ -2,7 +2,7 @@ module SangerSequencing
 
   class Batch < ActiveRecord::Base
 
-    DEFAULT_GROUP = [nil, ""].freeze
+    DEFAULT_PRODUCT_GROUP_NAME = [nil, ""].freeze
 
     self.table_name = "sanger_sequencing_batches"
 
@@ -17,9 +17,9 @@ module SangerSequencing
       where(facility: facility)
     end
 
-    def self.for_product_group(product_group)
-      product_group = DEFAULT_GROUP if product_group.blank?
-      where(group: product_group)
+    def self.for_product_group(product_group_name)
+      product_group_name = DEFAULT_PRODUCT_GROUP_NAME if product_group_name.blank?
+      where(group: product_group_name)
     end
 
     def well_plates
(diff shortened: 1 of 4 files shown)


meara:
@waggles what problem is this solving?

waggles:
oops I had a better explanation in the dupe, here it is:
This will treat both the empty string and `nil` as the "default" (non-fragment?) case. This solves the problem of some `Batch`es missing from the index page when the `group` params is not sent.

So on the sanger "tab" you wouldn't see all of the Batches  since they were a mix of group `nil` and group `""`

vendor/engines/sanger_sequencing/spec/models/sanger_sequencing/batch_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    subject { described_class.for_product_group(group) }
+
+    before { FactoryGirl.create(:product_group, :fragment) }
+    let!(:blank_batch) { FactoryGirl.create(:sanger_sequencing_batch, group: "") }
+    let!(:nil_batch) { FactoryGirl.create(:sanger_sequencing_batch, group: nil) }
+    let!(:grouped_batch) { FactoryGirl.create(:sanger_sequencing_batch, group: "fragment") }
+
+    context "with group given as nil" do
+      let(:group) { nil }
+      it { is_expected.to eq [blank_batch, nil_batch] }
```
jhanggi: Just a minor suggestion: use `match_array` or `contain_exactly`. We can't always assume ordering by ID in Oracle.

vendor/engines/sanger_sequencing/app/models/sanger_sequencing/batch.rb
```diff
@@ -15,8 +17,9 @@ def self.for_facility(facility)
       where(facility: facility)
     end
 
-    def self.for_product_group(group)
-      where(group: group)
+    def self.for_product_group(product_group)
```
jhanggi: Is this a ProductGroup or a string? Can we be a little clearer?

jhanggi:
Can you update the description so that's a little more user-focused? I would also agree it might be worth tracking down why we sometimes have nils and sometimes blanks. I think it's old because I don't think DC has any nulls, only blanks.

waggles:
So, as far as I can tell, it's not possible to create a record with group `nil` through the form. However, we do switch between links that use either `nil` or `""`.

waggles:
@jhanggi needs another look when you have the chance~

PR #1193, opened 2017-08-01
[#137315] feature flag validation for journals spanning fiscal years

meara:
- NU wants to prevent journals being created with order details spanning fiscal years.
- Dartmouth wants to be able to create journals with order details spanning fiscal years.

Added a feature flag to enable/disable this validation per fork.

### DIFF
--- a/spec/models/journal_spec.rb
+++ b/spec/models/journal_spec.rb
@@ -18,6 +18,23 @@ RSpec.describe Journal do
 
   describe "validations" do
     it { is_expected.to validate_length_of(:reference).is_at_most(50) }
+
+    describe "#must_not_span_fiscal_years" do
+      before do
+        order.add(product)
+        order.reload.order_details.first.update_attributes(fulfilled_at: Time.current)
+        order.order_details.last.update_attributes(fulfilled_at: 1.year.ago)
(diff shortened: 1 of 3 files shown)


waggles:
👍  pending spec

spec/models/journal_spec.rb
```diff
@@ -18,6 +18,23 @@
...
+        order.order_details.last.update_attributes(fulfilled_at: 1.year.ago)
+        journal.order_details_for_creation = order.order_details
+      end
+
+      context "when journals may span fiscal years", feature_setting: { journals_may_span_fiscal_years: true } do
+        it { is_expected.to be_valid }
+      end
+
+      context "when journals may not span fiscal years", feature_setting: { journals_may_span_fiscal_years: false } do
+        it { is_expected.not_to be_valid }
```
jhanggi: Usually when checking for invalid, I like to check that it is invalid for the right reason. I think `expect(model.errors).to be_added(:base, :fiscal_year_span)` or `expect(model.errors(:base)).to be_added(:fiscal_year_span)` would do it.

But since we're right next to the alternate case, it's probably fine as is.

meara: 👍 added

PR #1194, opened 2017-08-01
[#136038] Update default group scope for Batch

waggles:
This will treat both the empty string and `nil` as the "default" (non-fragment?) case. This solves the problem of some `Batch`es missing from the index page when the `group` params is not sent.


waggles:
oops dupe

PR #1195, opened 2017-08-01
[#137332] bundle update thin

meara:
Fix deprecation warnings for older version of thin from Ruby 2.4 upgrade.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -105,7 +105,7 @@ group :development, :test do
   gem "spring-commands-rspec"
   gem "teaspoon-jasmine"
   gem "test-unit", "~> 3.0"
-  gem "thin", "~> 1.7.2"
+  gem "thin", ">= 1.7.2"
 end
 
 group :test do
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -105,7 +105,7 @@ group :development, :test do
   gem "spring-commands-rspec"
   gem "teaspoon-jasmine"
   gem "test-unit", "~> 3.0"
-  gem "thin"
+  gem "thin", "~> 1.7.2"
```
jhanggi: Do we need to lock it down this specifically, or can we get away with just a bundle update?

meara: Ah maybe we don't. I got in the habit of doing that because I thought it helped keep things in sync across forks--when do and don't we want to lock gemfile versions?

jhanggi: I usually only like using it where we're worried about breaking changes (so things like devise, simple_form) on minor versions, where we wouldn't want to go from say 1.4.2 to 1.5.0 withouts some testing and validating the newer version works.

In general, I don't like to lock down things in the development group because those should theoretically be able to upgrade as much as they want and break very little, and at the very least will never break across production. 

But you're right in that it sometimes does help us make sure we always increase the versions in the downstreams. If we're worried about making sure we're at a minimum version, I would consider using `>= 1.7.2`, which will also allow going up to 1.8 if that gets released.

meara: Thanks for the explanation!

PR #1199, opened 2017-08-03
[#136048] Notify by product

waggles:
Adds a the `order_notification_recipient` field used on `Facility` field to `Product` and iterates through an order's order_details to send to each product recipient if necessary.

In addition, attaches the ical file to the Facility-level order_notification email.


app/mailers/notifier.rb
```diff
@@ -77,15 +86,21 @@ def order_detail_status_change(order_detail, old_status, new_status, to)
...
-      calendar = ReservationCalendar.new(reservation)
-      attachments[calendar.filename] = {
-        mime_type: "text/calendar", content: [calendar.to_ical]
-      }
+      attach_reservation_ical(reservation)
     end
   end
 
+  def attach_reservation_ical(reservation)
+    return if reservation.blank?
```
jhanggi: Do we need this? Looks like we're already compacting.

jhanggi: And I'll be honest, I'm not sure extracting this method adds much value.

app/services/order_purchaser.rb
```diff
@@ -35,6 +35,12 @@ def purchase!
 
     Notifier.delay.order_receipt(user: order.user, order: order) if send_receipt?
     Notifier.delay.order_notification(order, order_notification_recipient) if send_facility_notification?
+
+    order.order_details.each do |order_detail|
+      next unless order_detail.product.order_notification_recipient?
+      Notifier.delay.product_order_notification(order_detail, order_detail.product.order_notification_recipient)
```
jhanggi: Can you check and see how this serializes the order detail into the delayed_jobs table? If it uses globalid, then it's fine, but if not, we should probably switch to using `Notifier.product_order_notification(...).deliver_later`. And maybe switch the other notifiers as well.

app/services/order_purchaser.rb
```diff
@@ -35,6 +35,12 @@ def purchase!
 
     Notifier.delay.order_receipt(user: order.user, order: order) if send_receipt?
     Notifier.delay.order_notification(order, order_notification_recipient) if send_facility_notification?
+
+    order.order_details.each do |order_detail|
```
jhanggi: Would it make sense to extract all these notifications into a separate method? This `purchase!` method is getting long.

app/mailers/notifier.rb
```diff
@@ -33,8 +33,17 @@ def account_update(args)
     send_nucore_mail args[:user].email, text("views.notifier.account_update.subject")
   end
 
+  # Notifies the specified recipient if an order is placed including a product
+  def product_order_notification(order_detail, recipient)
```
jhanggi: I wouldn't mind this being a separate class (or on another mailer class). This notifier is pretty giant.

This is one of those places where it seems logical to follow existing apparent patterns, but we've decided we don't really like the pattern and have started to move away from it for new stuff, but the old stuff is still the old way.

app/mailers/notifier.rb
waggles: the way the logic is set up is confusing, I'll update it--the extraction only happened as a consequence of needing to attach only the ical stuff for a specific product though

app/mailers/purchase_notifier.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  # Notifies the specified recipient if an order is placed including a product
+  def product_order_notification(order_detail, recipient)
+    @order = order_detail.order
+    @order_detail = OrderDetailPresenter.new(order_detail)
+    attach_reservation_ical(order_detail.reservation) if order_detail.reservation.present?
+    send_nucore_mail recipient, text("views.purchase_notifier.product_order_notification.subject", product: order_detail.product)
+  end
+
+  def order_notification(order, recipient)
```
meara: Can we add a comment on this method on what it's doing that's different from product_order_notification?

app/mailers/purchase_notifier.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  include DateHelper
+  add_template_helper ApplicationHelper
+  add_template_helper TranslationHelper
+  add_template_helper OrdersHelper
+  add_template_helper ViewHookHelper
+
+  default from: Settings.email.from, content_type: "multipart/alternative"
+
+  # Notifies the specified recipient if an order is placed including a product
```
meara: Can we replace recipient with "facility staff member" or something similar, just to be super clear?

app/services/order_purchaser.rb
```diff
@@ -79,12 +79,19 @@ def quantities
     order.order_details.order("order_details.id").pluck(:quantity)
   end
 
-  def send_facility_notification?
-    order_notification_recipient.present? && !acting_as?
-  end
+  def send_purchase_notifications
+    if !acting_as? || params[:send_notification] == "1"
```
meara: I think I would prefer all these conditionals as separate private methods, for ease of reading and also more descriptiveness about what each section is doing.

i.e. `send_facility_notification_if_appropriate`, `send_product_specific_facility_notifaction_if_appropriate` etc.

Or maybe better, `send_facility_notification if should_send_facility_notification?`

Thoughts?

app/mailers/purchase_notifier.rb
```diff
@@ -0,0 +1,54 @@
+class PurchaseNotifier < ActionMailer::Base
```
meara: I like this separation.

app/views/purchase_notifier/product_order_notification.html.haml
```diff
@@ -0,0 +1,50 @@
+%table{ border: 0 }
+  %tr
+    %td
+      %strong= Order.human_attribute_name(:ordered_at)
+    %td= l(@order.ordered_at, format: :receipt)
```
meara: Is this a shortcut for a link helper or something?

waggles: the `l` is an alias for `localize` I believe

app/services/order_purchaser.rb
waggles: how about just locally assigned variables? I prefer not to add methods unless they actually need to be called more than just one place

app/mailers/purchase_notifier.rb
waggles: good point, will do

app/services/order_purchaser.rb
waggles: hmm do you think we should still send the product-level notification if `acting_as?`

meara: Hmm, I'm definitely in the school of skinny methods and reusable pieces, even though at the moment we might not actually be reusing the pieces elsewhere. I won't insist on it though.

My biggest issue with this method in its current form is that it's hard for me to tell at a glance what each piece is supposed to be doing. Whatever you want to do to clarify the function of each chunk would be great, whether that's using local variables or just adding some code comments.

Re sending product-level notifications--I'm not sure. I would default towards sending the notification even if acting-as, because I would rather they tell us they don't need the notification than have them be upset because they were relying on the notification and didn't get it.  Also, f there are multiple staff members for a facility--I imagine the person who places the order on behalf of might not be the same person who would be expecting the product specific notification.

PR #1200, opened 2017-08-11
[#137003] update email

meara:
Update new user creation email to allow customization for Dartmouth. Companion to https://github.com/tablexi/nucore-dartmouth/pull/128


app/views/notifier/new_user.html.haml
```diff
@@ -2,8 +2,19 @@
 
 - if @password
   %p
+    - if text(".username_and_password_text").present?
+      = text(".username_and_password_text")
+      %br
```
waggles: would line 5 still be necessary if we just move this `br` tag into the locale string?

app/views/notifier/new_user.text.haml
```diff
@@ -1,7 +1,17 @@
-= text(".intro", first_name: @user.first_name, last_name: @user.last_name, login_url: new_user_session_url)
+= text(".intro_text", first_name: @user.first_name, last_name: @user.last_name, login_url: new_user_session_url)
+= "" # new line
```
waggles: is there perhaps another way to do this in a `text.haml`? I know nothing about text hamls haha

meara: There might be, but I couldn't figure out anything less stupid that worked haha.

app/views/notifier/new_user.html.haml
meara: Can't move the `br` into the locale string unless we want to have separate strings for the html and text versions.

jhanggi: What would you think of something like this and rely on text_helper's markdown capabilities for the linebreaks/formatting?

```
- if @password
  = html("username_and_password_html" username: @user.username, password: @password)
- else
  = html("only_username_html", username: @user.username)
```

app/views/notifier/new_user.text.haml
jhanggi: Could it be cleaner/easier to do the text version as `.erb`?

app/views/notifier/new_user.html.haml
meara: Will experiment with this and also the erb thing.

meara:
@jhanggi updated

app/views/notifier/new_user.html.haml
```diff
@@ -1,9 +1,8 @@
...
 
 - if @password
-  %p
-    = "#{User.human_attribute_name(:username)}: #{@user.username}"
-    %br
-    = "#{User.human_attribute_name(:password)}: #{@password}"
+  = html(".username_and_password_html", username: @user.username, password: @password)
 - else
-  %p= text(".no_password")
+  = text(".only_username_html", username: @user.username)
```
jhanggi: We might not need it, but we should probably use `html()` here for consistency.

config/locales/en.notifier.yml
```diff
@@ -5,9 +5,18 @@ en:
         subject: "!app_name! Payment Method Updated"
       new_user:
         intro_html: "Congratulations, %{first_name}. A !app_name! account has been created for you. You may log in at <a href=\"%{login_url}\">%{login_url}</a>."
```
jhanggi: Protip: with the markdown parser, you can use it's syntax for links: `You may log in at [%{login_url}][%{login_url}].` (there's a chance I have the parens/square brackets reversed:).

app/views/notifier/new_user.text.erb
```diff
@@ -0,0 +1,10 @@
+<%= text(".intro_text", first_name: @user.first_name, last_name: @user.last_name, login_url: new_user_session_url) %>
+
+<!-- non-standard indentation below is to prevent indents appearing in the text formatting of this email -->
```
jhanggi: Not a big deal, but should we use a ruby comment instead of an html comment so it doesn't get rendered in the output.

PR #1201, opened 2017-08-11
[#137409] handle missing reserve end at

meara:
Don't try to estimate reservation cost unless both reservation start and end times exist.


spec/models/instrument_price_policy_calculations_spec.rb
```diff
@@ -43,6 +43,10 @@
       expect(policy.estimate_cost_and_subsidy(end_at, start_at)).to be_nil
     end
 
+    it "returns nil if start_at or end_at is nil" do
```
jhanggi: Should we do both sides of this, i.e. nil for start, and then another test for nil for end? Also, using start_at as the end_at argument is a little confusing (the previous test does it for the sake of intentionally inverting the times).

meara:
@jhanggi thoughts on merging this for now and making a separate task for date validation? Both seem worth doing IMO.

PR #1203, opened 2017-08-14
update deploy documentation

meara:


doc/HOWTO_forks.md
```diff
@@ -41,3 +41,19 @@ the commits in your branch so you only need to bring one commit.
 git fetch nu
 git cherry-pick XXXXX
 ```
+
+# Deploy
+From the root of your fork:
+
+- `bundle exec cap [staging_name] deploy`
+
+Find `staging_name` by the name of the `.rb` files in `config/deploy`
```
waggles: does this mean like... if the file is named `staging.rb` then `[staging_name]` is staging?

doc/HOWTO_forks.md
```diff
@@ -41,3 +41,19 @@ the commits in your branch so you only need to bring one commit.
...
+
+To confirm deploy was successful:
+
+- `ssh nucore@nucore.rrc.uic.edu`
```
waggles: since this looks like it's uic-specific, should we have it be more generic / say that this is an example

waggles:
oops I just saw the wip tag! 😳

doc/HOWTO_forks.md
meara: Yeah these are kinda bad notes---why this is still marked WIP haha.

I actually kinda forget what this means, as I recall some of the forks use "stage" and some use "staging". I think this is referring to what that file sets the rails_env as.

doc/HOWTO_forks.md
meara: Yep

doc/HOWTO_forks.md
jhanggi: It's a little funky. For NU we using `staging` and `prod`, UIC, we have `uic_stage`/`uic_prod` (I forget exactly why), then DC we have `staging` and `txi_staging` (because we have the stage that's auto-deployed in our environment, and then the one in theirs). It is unrelated to `rails_env`.

doc/HOWTO_forks.md
```diff
@@ -41,3 +41,19 @@ the commits in your branch so you only need to bring one commit.
...
+- `ls -al`
+- `cd nucore.rrc.uic.edu/current`
```
jhanggi: you can also check out what `current` symlinks to. It should symlink to `sitedomain.edu/releases/20170814XXXXXX`, and the timestamp of the release should look like it's very recent. That doesn't really tell you if the release was successful, though.

PR #1204, opened 2017-08-16
[#137451] Extend journal row description field length

jhanggi:
`OrderDetail#long_description` is what gets put into the
`JournalRow#description` field. If you have a really long product name
(that field is limited to 200 characters), then the description is too
long for the journal row field.

The 512 is somewhat arbitrary, but should be more than enough for
fitting long product names, the order detail and a user’s name.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -18,7 +18,8 @@ class Product < ActiveRecord::Base
   # Allow us to use `product.hidden?`
   alias_attribute :hidden, :is_hidden
 
-  validates_presence_of :name, :type
+  validates :type, presence: true
+  validates :name, presence: true, length: { maximum: 200 }
   validate_url_name :url_name, :facility_id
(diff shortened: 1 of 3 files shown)


app/models/product.rb
```diff
@@ -18,7 +18,8 @@ class Product < ActiveRecord::Base
   # Allow us to use `product.hidden?`
   alias_attribute :hidden, :is_hidden
 
-  validates_presence_of :name, :type
+  validates :type, presence: true
+  validates :name, presence: true, length: { maximum: 200 }
```
waggles: where does the 200 come from?

meara:
@jhanggi looks like you need to run your migration and push the schema up

app/models/product.rb
jhanggi: It's what's in the database. Simple_form seems to read that info from the database, because it does add `maxlength=200` to the input. This just adds the validation in case someone messes with the form.

PR #1207, opened 2017-08-18
[#137409] fix reservation validations

meara:
- https://pm.tablexi.com/issues/137409
- Certain time-related validations should only be run if reserve_start_at and reserve_end_at have valid values to avoid hard-errors.

### DIFF
--- a/app/support/reservations/validations.rb
+++ b/app/support/reservations/validations.rb
@@ -6,13 +6,14 @@ module Reservations::Validations
     delegate :editing_time_data, to: :order_detail, allow_nil: true
 
     validates_uniqueness_of :order_detail_id, allow_nil: true
-    validates :product_id, :reserve_start_at, presence: true
+    validates :product_id, presence: true
     validates :reserve_end_at, presence: true, if: :end_at_required?
+    validates :reserve_start_at, presence: true
     validate :does_not_conflict_with_other_reservation,
              :instrument_is_available_to_reserve,
              :satisfies_minimum_length,
              :satisfies_maximum_length,
-             if: :reserve_start_at && :reserve_end_at && :reservation_changed?,
+             if: -> (r) { r.reserve_start_at && r.reserve_end_at && r.reservation_changed? },
              unless: :admin?
 
     validates_each [:actual_start_at, :actual_end_at] do |record, attr, value|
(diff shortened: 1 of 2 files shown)


app/support/reservations/validations.rb
```diff
@@ -6,13 +6,14 @@ module Reservations::Validations
     delegate :editing_time_data, to: :order_detail, allow_nil: true
 
     validates_uniqueness_of :order_detail_id, allow_nil: true
-    validates :product_id, :reserve_start_at, presence: true
```
waggles: weird, did this not work for both of them? or am I missing something

spec/models/reservations_validations_spec.rb
```diff
@@ -6,6 +6,8 @@
 
   let(:now) { Time.zone.now }
 
+  it { is_expected.to validate_presence_of(:reserve_start_at) }
```
waggles: should we add in validation expectations for `product_id`, `order_detail_id` while we're in here?

spec/models/reservations_validations_spec.rb
```diff
@@ -60,4 +62,20 @@
       end
     end
   end
+
+  describe "conditionally run validations" do
+    context "with reserve_start_at, reserve_end_at, and reservation_changed? is true" do
```
waggles: the context this block uses is created outside of this file (in the `setup_reservation` factory)--what do you think about making this context explicit by inserting a `let(:reservation)` that specifically defines the fields used by this condition?

then we can depend on the spec at least having the context we state it has, even if someone updates the factory

(I'm sure this happens a lot but I just happened to think about it here, so up to you)

app/support/reservations/validations.rb
meara: Oh, it worked, I just ended up separating them because I thought it was easier to read.

spec/models/reservations_validations_spec.rb
meara: I thought about doing that, but the factory setup needed to make that work is pretty fussy.

spec/models/reservations_validations_spec.rb
meara: This is an interesting question. I sort of feel like we can depend on `setup_reservation` not changing, because so many tests depend on it. Whether or not that is desirable in our test structure seems debatable, but given that we're pretty firmly placed in that paradigm, I feel like we might as well take advantage of the abstraction.

PR #1208, opened 2017-08-22
[#137464] Allow user to edit order detail note

waggles:
https://pm.tablexi.com/issues/137105

The story mentions "We should review statuses and editing capability to make sure we keep this editable appropriately. For example, as long as the reservation is not yet complete."

As far as I can tell, the edit page is still accessible for a completed reservation (go directly to the `/edit`), but it is not linked to anymore. Is that a problem? If so, it seems like a different story, given that other fields are also editable from there.

### DIFF
--- a/app/views/reservations/edit.html.haml
+++ b/app/views/reservations/edit.html.haml
@@ -25,6 +25,13 @@
 
   = render "reservation_fields", f: f
 
+  .well
+    .container
+      .row
+        .span
+          %h4= OrderDetail.model_name.human
+          = f.input :note
+
   %ul.inline
     %li= f.submit 'Save', class: 'btn'
     - if @order.purchased?


jhanggi:
I haven't played too much with it, but can you actually save any changes to the times once the reservation is complete? Just wondering if we should put an extra guard condition on edit/update?

waggles:
it seems like I can, though lemme double check to make sure everything I'm testing it with is in the right state

waggles:
update: digging more specifically, it appears that only admin users can modify both fields, others just get redirected to show page, so I think we are good

app/views/reservations/edit.html.haml
```diff
@@ -25,6 +25,13 @@
 
   = render "reservation_fields", f: f
 
+  .well
```
jhanggi: I probably should have caught this sooner. We should match the style of the `new` form where this is above the times. 

Also, it should not _always_ appear. It is contingent on the user_notes_field_mode being either optional or required.

The `new` form uses this:

```haml
      - if show_note_input_to_user?(@order_detail)
        .span6
          = f.input :note,
            label: @order_detail.product.user_notes_label.presence,
            required: @order_detail.product.user_notes_field_mode.required?
```

app/views/reservations/edit.html.haml
```diff
@@ -25,6 +25,13 @@
...
+    .container
+      .row
+        .span
+          %h4= OrderDetail.model_name.human
```
jhanggi: I would remove this. Is there a reason for it?

waggles: There's a field `admin_note` on `Reservation` and a field `note` on `OrderDetail`--since this is `reserverations/edit` and not `order_details/edit`, I wasn't sure if this would be confusing with no additional clarification. I can see it not being a problem though, users prob have more context/understanding?

app/views/reservations/edit.html.haml
waggles: Ah okay, I had moved it because it looked really weird especially before adding the extra label below, but the consistency makes sense

app/views/reservations/edit.html.haml
jhanggi: This shouldn't cause any confusion. `admin_note` is only used for admin and offline reservations. The order detail note is for user orders.

PR #1211, opened 2017-08-24
[#137398] temp feature flag timed services

meara:
Feature flag for timed services while feature is being worked on.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -64,12 +64,20 @@ class Product < ActiveRecord::Base
   scope :not_archived, -> { where(is_archived: false) }
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
 
+  def self.default_types
+    if SettingsHelper.feature_on?(:timed_services)
+      [Instrument, Item, Service, TimedService, Bundle]
+    else
+      [Instrument, Item, Service, Bundle]
+    end
+  end
+
   def self.types
-    @types ||= [Instrument, Item, Service, TimedService, Bundle]
+    @types ||= default_types
   end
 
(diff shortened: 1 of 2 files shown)


meara:
I think we'll probably want to remove this feature flag once the feature is complete--it's a little awkward code-wise, and there's a few spots it doesn't cover if a timed service already exists. For example, if a timed service has been added as an accessory on an instrument, it will still be shown as addable to a reservation.

For a temporary flag, while no timed services exist in stage/prod, I think this is fine, but we wouldn't want to leave this lying around once the feature is complete.

jhanggi:
I agree this should go away once we're ready with the feature. In the meantime, this feels like it's touching a few more places than it really needs to assuming that there aren't existing TimedServices. For example, the facility show view should not show a section if there are no products of that type. And many of the scopes shouldn't do anything different if there aren't any TimedServices.

I would expect the only place we really need to flag is the sidebar for the Products tab in order to prevent creating a TimedService (I don't care too much if the route is protected or not). What's the reason we need to touch these other sections?

meara:
I think I effectively decided in the middle that it wasn't practical to comprehensively flag everything--if we're comfortable with this being more of a temporary, "let's prevent people from accidentally interacting with this feature before it's ready" than a true comprehensive feature flag, I think that makes sense. I'll scale back what this PR is doing accordingly.

jhanggi:
Yeah, I think the prevent people from using it before it's ready is the way to think about this feature flag. I'm fine making it available for everyone once it's ready.

meara:
@jhanggi good to merge on this then?

PR #1213, opened 2017-08-30
[#137455] user preferences

meara:
Companion PR to https://github.com/tablexi/nucore-uic/pull/151

### DIFF
--- a/app/controllers/user_preferences_controller.rb
+++ b/app/controllers/user_preferences_controller.rb
@@ -4,6 +4,7 @@ class UserPreferencesController < ApplicationController
   before_action :check_acting_as
 
   def index
+    UserPreference.create_appropriate_user_preferences(current_user)
     @user_preferences = current_user.user_preferences
   end
 
--- a/app/controllers/user_preferences_controller.rb
+++ b/app/controllers/user_preferences_controller.rb
@@ -4,7 +4,7 @@ class UserPreferencesController < ApplicationController
   before_action :check_acting_as
 
   def index
-    UserPreference.create_appropriate_user_preferences(current_user)
+    UserPreference.create_missing_user_preferences(current_user)
     @user_preferences = current_user.user_preferences
   end
 
@@ -15,7 +15,7 @@ class UserPreferencesController < ApplicationController
 
   def update
     @user_preference = current_user.user_preferences.find(params[:id])
-    if @user_preference.update_attributes(value: user_preference_params[:value])
+    if @user_preference.update_attributes(user_preference_params)
       redirect_to user_user_preferences_path(current_user)
     else
       render :edit
--- a/app/controllers/user_preferences_controller.rb
+++ b/app/controllers/user_preferences_controller.rb
@@ -4,19 +4,24 @@ class UserPreferencesController < ApplicationController
   before_action :check_acting_as
 
   def index
+    @user_preferences = current_user.user_preferences
   end
 
   def edit
-    @user_preference = UserPreference.find params[:id]
+    @user_preference = current_user.user_preferences.find params[:id]
   end
 
   def update
-    @user_preference = UserPreference.find params[:id]
-    if @user_preference.update_attributes(value: params[:user_preference][:value])
+    @user_preference = current_user.user_preferences.find params[:id]
+    if @user_preference.update_attributes(value: user_preference_params[:value])
       redirect_to user_user_preferences_path(current_user)
     else
       render :edit
     end
+  end
+
+  def user_preference_params
+    params.require(:user_preference).permit(:value)
   end
 
 end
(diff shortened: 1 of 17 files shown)


app/controllers/user_preferences_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def index
+  end
+
+  def edit
+    @user_preference = UserPreference.find params[:id]
+  end
+
+  def update
+    @user_preference = UserPreference.find params[:id]
```
jhanggi: We should probably limit this to `current_user.user_preferences.find`. That way we don't have to worry about abilities too much.

app/controllers/user_preferences_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+    if @user_preference.update_attributes(value: params[:user_preference][:value])
```
jhanggi: Can we do strong params here?

app/models/user_preference.rb
```diff
@@ -0,0 +1,10 @@
+class UserPreference < ActiveRecord::Base
+
+  belongs_to :user
+  validates :name, :value, presence: true
+
+  OPTIONS = {
+    "Facility Home Page" => ["Dashboard", "Timeline View"],
```
jhanggi: Hmm. This option feels like it belongs in UIC/dashboard engine. Not quite sure the best way to get it there.

app/views/user_preferences/index.html.haml
```diff
@@ -0,0 +1,12 @@
+%h1 User Preferences
```
jhanggi: I18n this view.

config/routes.rb
```diff
@@ -23,6 +23,10 @@
 
   post "global_search" => 'global_search#index', as: "global_search"
 
+  resources :users, only: [] do
```
jhanggi: I would expect that we already have a `users` route. Am I wrong?

db/migrate/20170829193038_create_user_preferences.rb
```diff
@@ -0,0 +1,11 @@
+class CreateUserPreferences < ActiveRecord::Migration
+  def change
+    create_table :user_preferences do |t|
+      t.references :user, index: true, foreign_key: true
```
jhanggi: Would a uniqueness constraint/validation on user_id/name be a good idea?

config/routes.rb
meara: We do, but it's nested under facilities, whereas we want user preferences to exist regardless of facilities.

db/migrate/20170829193038_create_user_preferences.rb
meara: Added.

app/views/shared/_header.html.haml
```diff
@@ -22,7 +22,7 @@
               %li.navbar-text= "#{acting_user.full_name} (#{acting_user.username})"
               %li= link_to "#{t('pages.cart')} (#{current_cart.order_details.count})", :cart
             - else
-              %li.navbar-text= "#{session_user.full_name} (#{session_user.username})"
+              %li.visible-with-link= link_to "#{session_user.full_name} (#{session_user.username})", user_user_preferences_path(session_user)
```
jhanggi: This feels a little awkward to link to if you have no preferences (as there will be for NU/UIC for now). I wonder if it might be better to have it not link until you have some settings? I'm open to ideas.

meara: Which school has the password reset option available? I remember that being floated as something that might make sense to put on this page.

I like the idea of not linking unless there is something to do on the page, though.

jhanggi: NU does if you're an external (non-netid) user, and dartmouth does for everyone since everyone is "external".

meara: On reflection, I'm going to call putting the reset password stuff onto the user preferences page out of scope for now.

meara:
@jhanggi are we good to merge this? I've still got a couple changes to make in the UIC fork but would have to rebase that once this is merged in anyway.

app/controllers/user_preferences_controller.rb
```diff
@@ -0,0 +1,29 @@
+class UserPreferencesController < ApplicationController
+
+  before_action :authenticate_user!
+  before_action :check_acting_as
+
+  def index
+    UserPreference.create_appropriate_user_preferences(current_user)
```
jhanggi: Maybe `create_missing_user_preferences` might be a better name? "Appropriate" doesn't have much semantic meaning to me here. Not a big deal, though.

app/controllers/user_preferences_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def edit
+    @user_preference = current_user.user_preferences.find(params[:id])
+    @user_preference_option = @user_preference.option
+  end
+
+  def update
+    @user_preference = current_user.user_preferences.find(params[:id])
+    if @user_preference.update_attributes(value: user_preference_params[:value])
```
jhanggi: I thought we had changed this. Maybe we're missing a commit?

PR #1215, opened 2017-09-08
[#137282] Add timed service to existing order

jhanggi:
When adding a TimedService to an existing order, we will gray out the quantity field (and always adjust it back to 1) and display a new time input field.

screen shot 2017-09-08 at 11 58 31 am
screen shot 2017-09-08 at 11 58 23 am

### DIFF
--- a/app/assets/javascripts/app/merge_orders.coffee
+++ b/app/assets/javascripts/app/merge_orders.coffee
@@ -6,20 +6,21 @@ class window.MergeOrder
     return unless @$form.length
 
     @$quantity_field = @$form.find(".js--edit-order__quantity")
-    @$timed_quantity_display_field = @$form.find(".js--edit-order__duration")
+    @$duration_display_field = @$form.find(".js--edit-order__duration")
 
     # clockpunch converts the original field into a field with name _display.
     # We will need to disable both the visible display field and the hidden field
-    # so they don't get sent as part of the POST.
-    @$timed_quantity_display_field.timeinput()
-    @$timed_quantity_hidden_field = @$timed_quantity_display_field.data("timeparser").$hidden_field
+    # so they don't get sent as part of the POST unless a timed service is
+    # selected.
+    @$duration_display_field.timeinput()
+    @$duration_hidden_field = @$duration_display_field.data("timeparser").$hidden_field
 
     @$form.find(".js--edit-order__product").on "change", (event) =>
       is_timed = $(event.target).find(":selected").data("timed-product")
 
-      @$timed_quantity_display_field.toggle(is_timed)
-      @$timed_quantity_display_field.prop("disabled", !is_timed)
-      @$timed_quantity_hidden_field.prop("disabled", !is_timed)
+      @$duration_display_field.toggle(is_timed)
+      @$duration_display_field.prop("disabled", !is_timed)
+      @$duration_hidden_field.prop("disabled", !is_timed)
 
       @$quantity_field.val(1) if is_timed
       @$quantity_field.prop("disabled", is_timed)
--- a/app/assets/javascripts/app/merge_orders.coffee
+++ b/app/assets/javascripts/app/merge_orders.coffee
@@ -3,10 +3,11 @@ class window.MergeOrder
     # empty body
 
   initTimeBasedServices: ->
+    return unless @$form.length
+
     @$quantity_field = @$form.find(".js--edit-order__quantity")
-    console.debug "quantity field", @$quantity_field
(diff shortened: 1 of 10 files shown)


app/models/product.rb
```diff
@@ -253,10 +253,16 @@ def product_type
     self.class.name.underscore.pluralize
   end
 
+  # Used when displaying quantities throughout the site and when editing an order.
   def quantity_as_time?
     false
   end
 
+  # Primarily used when adding to an existing order (merge orders)
+  def order_quantity_as_time?
```
meara: So this separate method is used to distinguish TimedService behavior from Instrument behavior then?

app/services/order_appender.rb
```diff
@@ -12,14 +12,18 @@ def initialize(original_order, user)
...
 
-      order.add(product, quantity, created_by: user.id).each do |order_detail|
-        update_order_detail!(order_detail, note: note, order_status: order_status, fulfilled_at: params[:fulfilled_at])
+      attributes = {
+        note: params[:note].presence,
+        duration: params[:duration],
+        created_by: user.id,
+      }
+
+      order.add(product, quantity, attributes).each do |order_detail|
```
meara: This is a nice refactor.

app/support/orders/item_adder.rb
```diff
@@ -9,20 +9,23 @@ def initialize(order)
 
   def add(product, quantity = 1, attributes = {})
     check_for_mixed_facility! product
```
meara: While we're in here, mind adding parens to this argument?

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -1,10 +1,12 @@
 - if @merge_orders.blank?
   .well
     = simple_form_for [current_facility, @order], method: :put, html: { class: "js--edit-order" } do |f|
-      = text_field_tag :product_add_quantity, 1
+      = text_field_tag :product_add_quantity, 1, class: "js--edit-order__quantity"
```
meara: Is the double underscore a convention in the codebase?

app/assets/javascripts/app/merge_orders.coffee
```diff
@@ -0,0 +1,30 @@
+class window.MergeOrder
+  constructor: (@$form) ->
+    # empty body
+
+  initTimeBasedServices: ->
+    return unless @$form.length
+
+    @$quantity_field = @$form.find(".js--edit-order__quantity")
```
meara: I'm having some trouble following your variables here. Can we change timed_quantity_display_field to duration_display_field for clarity?

app/assets/javascripts/app/merge_orders.coffee
```diff
@@ -0,0 +1,30 @@
...
+    @$timed_quantity_display_field = @$form.find(".js--edit-order__duration")
+
+    # clockpunch converts the original field into a field with name _display.
+    # We will need to disable both the visible display field and the hidden field
+    # so they don't get sent as part of the POST.
```
meara: Can we add something like "unless a timed service is selected" to this comment?

meara:
Is this rebased on master? Looks like you're still getting that sporadic spec failure here.

app/models/product.rb
jhanggi: Yes. Just using the existing `quantity_as_time?` was causing instruments to show the duration field.

app/views/facility_orders/_merge_order_form.html.haml
jhanggi: We use it in a couple places, mostly for CSS. It's definitely a convention in many of our newer projects.

coffeelint.json
```diff
@@ -55,7 +55,7 @@
   },
   "newlines_after_classes": {
     "value": 1,
-    "level": "error"
+    "level": "ignore"
```
jhanggi: According to http://www.coffeelint.org/ this rule is ignore by default. Not sure why we had it marked as an error. We have errors on this cop in other files, and it seems to me to be a weird one (especially the 1 value). It would mean that the `$ ->` line would come directly after the last line of the class.

app/models/product.rb
```diff
@@ -77,7 +77,7 @@ def self.types
   end
 
   def self.mergeable_types
-    @mergeable_types ||= default_types
+    default_types
```
meara: Depending on what order we put things in, I think we'll have some merge conflicts here.

coffeelint.json
meara: 👍

PR #1217, opened 2017-09-08
rm timed services feature flag

meara:
Remove temporary feature flag added in #1211.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -64,20 +64,14 @@ class Product < ActiveRecord::Base
   scope :not_archived, -> { where(is_archived: false) }
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
 
-  def self.default_types
-    if SettingsHelper.feature_on?(:timed_services)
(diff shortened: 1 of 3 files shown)


app/models/product.rb
```diff
@@ -64,20 +64,14 @@ class Product < ActiveRecord::Base
...
-      [Instrument, Item, Service, TimedService, Bundle]
-    else
-      [Instrument, Item, Service, Bundle]
-    end
-  end
+  DEFAULT_TYPES = [Instrument, Item, Service, TimedService, Bundle].freeze
 
   def self.types
-    @types ||= default_types
+    @types ||= DEFAULT_TYPES
```
jhanggi: I think these need to be `DEFAULT_TYPES.dup`. The original constant is frozen, so we get an error when we try to add `SecureRoom` through the engine.

jhanggi:
Can I request one additional change? Can you remove the `feature_setting` from this spec?
 https://github.com/tablexi/nucore-open/pull/1215/files#diff-cfd295cfb22f2912c8580eaf6b1f0786R101

PR #1220, opened 2017-09-12
add additional user roles to demo seed

meara:
Add facility administrator, account manager, senior staff to demo seed.

### DIFF
--- a/lib/tasks/demo_seed.rake
+++ b/lib/tasks/demo_seed.rake
@@ -246,6 +246,17 @@ namespace :demo do
     end
     UserRole.grant(user_staff, UserRole::FACILITY_STAFF, facility)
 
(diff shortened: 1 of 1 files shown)


lib/tasks/demo_seed.rake
```diff
@@ -246,6 +246,17 @@ namespace :demo do
     end
     UserRole.grant(user_staff, UserRole::FACILITY_STAFF, facility)
 
+    user_facility_administrator = User.find_by(username: "mfa123@example.com")
+    unless user_facility_administrator
+      user_facility_administrator = User.new(username: "mfa123@example.com",
+                                             email: "mfa123@example.com",
+                                             first_name: "Macy",
+                                             last_name: "Administator")
```
jhanggi: Can we say "Facility Administrator" to make it clear from the name that it's a facility admin, and not a global admin?

PR #1225, opened 2017-09-27
update nokogiri for security warning

rdunlop:


Gemfile
```diff
@@ -26,7 +26,7 @@ gem "aasm", "~> 4.11.1"
 gem "paperclip", "~> 4.2.0"
 gem "paper_trail"
 gem "awesome_nested_set", "~> 3.1.3"
-gem "nokogiri", "~> 1.7.1"
+gem "nokogiri"
```
meara: Hi Robin--in regards to your email about specifying the gem version in the Gemfile, one of the main reasons we do it this way in NUCore is so that when we update a gem in open, the update also gets to the dependent forks (nucore-nu, nucore-dartmouth, nucore-uic). Since the various forks don't share a Gemfile.lock, pinning it in the Gemfile has been a good way to manage it.

If you don't like `~>`, maybe `>=` would be better?

meara:
Also, FYI @rdunlop, on nucore we try to use the "request a review" function on github--that's definitely the best way to make sure that I at least see that a PR needs reviewing. 😀

rdunlop:
Updated @meara thanks.

spec/controllers/global_search_controller_spec.rb
```diff
@@ -168,7 +168,12 @@ def it_should_have_customer_paths
       end
 
       context "when there is an order and order detail with same ids" do
-        let!(:order2) { FactoryGirl.create(:purchased_order, id: order_detail.id, product: product) }
+        let!(:order2) do
+          # random test failure, if by chance there is an order with the same ID
+          # in the spec suite.
+          Order.find_by(id: order_detail.id) || FactoryGirl.create(:purchased_order, id: order_detail.id, product: product)
```
jhanggi: I need to think through this, but I'm not quite convinced this is correct. Either way, can we pull this out into a separate PR? It doesn't really belong with the nokogiri upgrade.

PR #1227, opened 2017-09-30
If ROLLBAR_ACCESS_TOKEN is defined on CI, we still don't want to inform rollbar

rdunlop:


meara:
I don't have any specific objections to this, but this seems like the kind of thing Jason probably has opinions on, so probably good to wait until he's back and can take a look.

PR #1228, opened 2017-10-01
attempt to fix spurious test failure.

rdunlop:
Would fail if an order with the same ID randomly already existed

first seen: https://circleci.com/gh/tablexi/nucore-open/4869

### DIFF
--- a/spec/controllers/global_search_controller_spec.rb
+++ b/spec/controllers/global_search_controller_spec.rb
@@ -168,7 +168,12 @@ RSpec.describe GlobalSearchController do
       end
 
       context "when there is an order and order detail with same ids" do
-        let!(:order2) { FactoryGirl.create(:purchased_order, id: order_detail.id, product: product) }
+        let!(:order2) do
+          # random test failure, if by chance there is an order with the same ID
+          # in the spec suite.
+          Order.find_by(id: order_detail.id) || FactoryGirl.create(:purchased_order, id: order_detail.id, product: product)
+        end
+
         before :each do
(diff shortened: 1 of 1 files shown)


meara:
I think we'd better wait for Jason on this one as well, I don't have a ton of context on the desired behavior of the search functionality. The risk I see here is that the pre-existing order that is found might still be in a state that would change how it shows up in search, which could then cause another, different sporadic test failure.

spec/controllers/global_search_controller_spec.rb
```diff
@@ -168,7 +168,12 @@ def it_should_have_customer_paths
       end
 
       context "when there is an order and order detail with same ids" do
-        let!(:order2) { FactoryGirl.create(:purchased_order, id: order_detail.id, product: product) }
+        let!(:order2) do
+          # random test failure, if by chance there is an order with the same ID
+          # in the spec suite.
+          Order.find_by(id: order_detail.id) || FactoryGirl.create(:purchased_order, id: order_detail.id, product: product)
```
jhanggi: Our order details are referred to by OrderId-OrderDetailId. 

In the normal case, we would have something like Orders: 1-2 and 2-3. If I search for "2", I want to get both orders back.

With this change, I would expect this to still fail when the order already exists: the search results would only return one order detail since there is only one that matches. That would not `match_array([order2.order_details.first, order_detail])`.

Does my logic seem right? This is a tricky one to reason about.

rdunlop:
closing due to inactivity

PR #1229, opened 2017-10-03
[#139062] track created_by for admin and offline reservations

meara:
- Capture user creating the Administrative Hold and display in hover over

### DIFF
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -111,7 +111,7 @@ class FacilityReservationsController < ApplicationController
     @instrument =
       current_facility.instruments.find_by!(url_name: params[:instrument_id])
     @reservation = @instrument.admin_reservations.new
-    @reservation.assign_attributes(params[:admin_reservation])
+    @reservation.assign_attributes(params[:admin_reservation].merge(created_by: current_user.id))
     @reservation.assign_times_from_params(params[:admin_reservation])
 
     if @reservation.save
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -13,6 +13,7 @@ class Reservation < ActiveRecord::Base
   #####
   belongs_to :product
   belongs_to :order_detail, inverse_of: :reservation
+  belongs_to :created_by_user, class_name: "User", foreign_key: :created_by
   has_one :order, through: :order_detail
 
   ## Virtual attributes
--- a/app/controllers/offline_reservations_controller.rb
+++ b/app/controllers/offline_reservations_controller.rb
@@ -17,6 +17,7 @@ class OfflineReservationsController < ApplicationController
 
   def create
     @reservation = @instrument.offline_reservations.new(new_offline_reservation_params)
+    @reservation.assign_attributes(created_by: current_user.id)
 
     if @reservation.save
       flag_ongoing_reservations_as_problem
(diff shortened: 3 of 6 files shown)


spec/controllers/facility_reservations_controller_spec.rb
```diff
@@ -141,6 +141,10 @@
           expect(assigns[:reservation].admin_note).to eq "Testing"
         end
 
+        it "sets created_by" do
```
jhanggi: We should probably have a spec for the offline creation as well.

app/models/reservation.rb
```diff
@@ -13,6 +13,7 @@ class Reservation < ActiveRecord::Base
   #####
   belongs_to :product
   belongs_to :order_detail, inverse_of: :reservation
+  belongs_to :created_by_user, class_name: "User", foreign_key: :created_by
```
jhanggi: I'm not a fan of this pattern, even though we use it elsewhere because `created_by` on some tables is really old and didn't originally have an association on it. So when we added the association, we went with `created_by_user` rather than changing the column name.

I'm partial to calling the column `created_by_id` or (actually probably better, but nothing matches this pattern) `created_by_user_id`, and then the association becomes `created_by`. We use a similar patter in `OrderDetails#dispute_by`, `Payment#paid_by`, `SangerSequencing::Batch#created_by`, `NuResearchSafety::Certificate#deleted_by`, and a few others.

It's one of those things where I much prefer the newer style, but we still have a bunch of the older style that it's somewhat inconsistent.

Thoughts?

app/controllers/facility_reservations_controller.rb
```diff
@@ -111,7 +111,7 @@ def create
     @instrument =
       current_facility.instruments.find_by!(url_name: params[:instrument_id])
     @reservation = @instrument.admin_reservations.new
-    @reservation.assign_attributes(params[:admin_reservation])
+    @reservation.assign_attributes(params[:admin_reservation].merge(created_by: current_user.id))
```
jhanggi: While we're at it, should we capture this for normal reservations as well? It would kind of be duplicating the `OrderDetail#created_by`, but it would make our data more consistent.

app/models/reservation.rb
meara: Ah, I originally had it as `created_by_id` but then moved away from that because I wasn't seeing that elsewhere in the codebase. I'll switch it.

app/controllers/facility_reservations_controller.rb
meara: I decided not to, as normal reservations go through ReservationCreator, which is rather more complicated. Seemed like more work than it was worth. Possibly there's a simpler way to handle capturing this though, idk if you have a suggestion.

spec/controllers/facility_reservations_controller_spec.rb
meara: Added.

meara:
Will add to export raw.

db/migrate/20171010205831_add_created_by_id_to_reservations.rb
```diff
@@ -0,0 +1,6 @@
+class AddCreatedByIdToReservations < ActiveRecord::Migration
+  def change
+    add_column :reservations, :created_by_id, :integer
+    add_foreign_key :reservations, :users, column: :created_by_id
```
jhanggi: I think we need to manually add the index because, well, oracle. Oracle doesn't automatically add an index for foreign keys (or at least it didn't at one point). Might be worth confirming before we merge this.

meara: From a quick google search, it sounds like Oracle still doesn't automatically create the index. Added one.

PR #1230, opened 2017-10-03
[#137631, #139373] admin hold expiration and UI addition

meara:
https://pm.tablexi.com/issues/139373 - soft delete expired admin reservations
https://pm.tablexi.com/issues/137631 - add end times to admin reservation scheduler

- [x] add expiration info to hover-overs in calendar views
- [x] try out in NU fork (verify things work with oracle)
- [x] fix expiration styling


app/controllers/facility_reservations_controller.rb
```diff
@@ -141,6 +141,7 @@ def update_admin
 
     @reservation.assign_times_from_params(params[:admin_reservation])
     @reservation.admin_note = params[:admin_reservation][:admin_note]
+    @reservation.expires_mins_before = params[:admin_reservation][:expires_mins_before]
     @reservation.category = params[:admin_reservation][:category]
```
jhanggi: If we can use strong params to make this a little cleaner, I would vote for that.

jhanggi: It's a also a little weird the stylistic difference from `create_admin`. Maybe there's a reason they're different, but if we can make them more similar, that'd be great.

config/locales/en.models.yml
```diff
@@ -330,6 +330,7 @@ en:
         canceled_by: Canceled By
         canceled_at: Canceled At
         type: Type
+        expires_mins_before: Expires Time Prior
```
jhanggi: I'm not sure what, but I feel like we should try to come up with a better label for this field. Maybe a hint would help?

db/migrate/20171003201727_add_expires_minutes_before_to_reservations.rb
```diff
@@ -0,0 +1,5 @@
+class AddExpiresMinutesBeforeToReservations < ActiveRecord::Migration
+  def change
+    add_column :reservations, :expires_mins_before, :integer
```
jhanggi: We should validate the numericality of this field and that it's only integers. There's a part of me that wants to allow negative numbers (if the timeinput will allow that), although that might be moot with being allowed to schedule over an admin reservation (I'm thinking that time opens up in the past so you can retroactively reserve for people).

db/schema.rb
```diff
@@ -753,7 +755,7 @@
     t.datetime "updated_at",             null: false
   end
 
-  add_index "user_preferences", ["user_id", "name"], name: "index_user_preferences_on_user_id_and_name", unique: true, using: :btree
+  add_index "user_preferences", ["user_id"], name: "index_user_preferences_on_user_id", using: :btree
```
jhanggi: This seems like it doesn't belong in this PR.

meara: https://github.com/tablexi/nucore-open/pull/1247

meara: Also, apparently `expires_mins_before` has already snuck into the schema (in https://github.com/tablexi/nucore-open/pull/1229/files#diff-1acd2e7e27a227829d5d14a91c863bb6R551). Is that going to cause an issue?

jhanggi: See my comment on #1247, the original is what the schema _should_ be.

The `expires_mins_before` sneaking in shouldn't cause any problems as long as later PRs didn't revert it.

db/migrate/20171003201727_add_expires_minutes_before_to_reservations.rb
meara: Added.

config/locales/en.models.yml
meara: We'll see what NU thinks.

app/helpers/date_helper.rb
```diff
@@ -56,6 +56,14 @@ def human_time(dt)
     end
   end
 
+  def display_as_time(minutes)
```
jhanggi: I don't have a problem with adding a view helper method for it, but we have `MinutesToTimeFormatter` which includes this logic.

app/presenters/reservations/calendar_presenter.rb
```diff
@@ -23,6 +23,12 @@ def reservation_options(options)
           title: "Instrument Offline",
           email: created_by.try(:full_name),
         }
+      elsif expires_mins_before.present?
```
jhanggi: It's not bad, but I might do something like

```
else
  hash = {
    title: model_name.human,
    email: created_by.try(:full_name),
  }
  hash[:expiration] = "Expires #{display_as_time(expires_mins_before)} prior" if expires_mins_before.present?
  hash
```

We don't necessarily need to do it, but what would happen if we changed `OfflineReservations`'s model name to "Instrument Offline" and collapse even another conditional?

app/views/facility_reservations/_admin_reservation_fields.html.haml
```diff
@@ -0,0 +1,46 @@
...
+      .controls
+        .string.optional.control-label &nbsp;
+        = time_select f, :reserve_end, minute_step: 1
+
+.clearfix
+
+= f.input :admin_note, input_html: { class: "span6" } unless @order_detail
+
+.checkboxControl
+  = f.input :expires?, as: :boolean, input_html: { data: { "disables" => ".admin_reservation_expires_mins_before" } }, wrapper_html: {class: "checkboxControl__checkbox"}
```
jhanggi: Some spacing issues inside of `{}`, and can we use the symbol `disables:`?

app/views/facility_reservations/edit.html.haml
```diff
@@ -24,7 +24,38 @@
...
+          .string.optional.control-label &nbsp;
+          = time_select f, :actual_start
+      = f.input :actual_duration_mins, input_html: { class: "timeinput" }
+    - unless @reservation.has_actuals?
+      .copy_actual_from_reservation
+        = link_to "Copy from reservation", "#"
+        %span.start_date.hidden= @reservation.reserve_start_at
+        %span.end_date.hidden= @reservation.reserve_end_at
+    - if @instrument.reservation_only?
+      %li.alert This is a reservation only instrument. Pricing is determined by the reservation times.
```
jhanggi: We probably should i18n this at some point, but doesn't have to be now.

app/views/facility_reservations/edit_admin.html.haml
```diff
@@ -1,5 +1,5 @@
 = content_for :head_content do
-  = render :partial => 'shared/headers/calendar'
+  = render 'shared/headers/calendar'
```
jhanggi: We're cleaning up the hash syntax here. Can we clean up the single v double quotes?

app/assets/javascripts/app/disabling_by_checkbox.coffee
```diff
@@ -0,0 +1,8 @@
+$ ->
+  $("[data-disables]").on "change", ->
```
jhanggi: Code climate is complaining about `this` (aka `@`) without a fat arrow. If we make this a fat arrow, I think we would need to add something like `.on "change", (el) ->` and then `$(el)` (I may be wrong on the exact syntax).

app/helpers/date_helper.rb
meara: Ah thank you, I thought we had something like that but couldn't find it.

app/assets/javascripts/app/disabling_by_checkbox.coffee
meara: Yeah I kinda don't want to mess with the javascript right now if that's okay with you.

app/presenters/reservations/calendar_presenter.rb
meara: I don't entirely know all the spots where we use OfflineReservation's model name so am a little hesitant to do that now, but you make good sense.

app/views/facility_reservations/edit.html.haml
meara: I vote for later. ;)

PR #1236, opened 2017-10-20
[#139351] add note for timed service quantity

meara:
Add note to mitigate confusion over the meaning of the quantity field for timed services.


app/views/facilities/_product_list.html.haml
```diff
@@ -1,6 +1,8 @@
 - if products.any?
   .product_list{class: products.first.class.model_name.to_s.pluralize.downcase}
     %h3= product_list_title products, local_assigns[:title_extra]
+    - if products.first.class.model_name == "TimedService"
+      %p Please note: Quantity below are individual orders, not minutes per order.
```
jhanggi: I feel like we should probably I18n this.

PR #1239, opened 2017-10-25
[#137629] admin hold schedule over

meara:
What the flash messages look like:

screen shot 2017-11-01 at 2 33 50 pm
screen shot 2017-11-01 at 2 34 04 pm
screen shot 2017-11-01 at 2 39 20 pm
screen shot 2017-11-02 at 4 14 27 pm

to do:

- [x] add feature specs for placing reservations over admin reservations
- [x] write a test to verify earliest possible (whichever method that is) gives expected results (reservation_spec.rb:589)
- [x] After the reservation is submitted, include a flash message notifying the facility staff member that they've scheduled over an admin reservation. Text: Warning, you've scheduled over an administrative hold.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -259,6 +259,9 @@ class OrdersController < ApplicationController
     end
 
     if order_purchaser.success?
+      should_show_admin_hold_warning = @order.order_details.map(&:reservation).compact.map(&:conflicting_admin_reservation).any?
+      flash[:error] = I18n.t("controllers.reservations.create.admin_hold_warning") if should_show_admin_hold_warning
+
       if single_reservation? && !acting_as?
         flash[:notice] = I18n.t("controllers.orders.purchase.reservation.success")
         if can_switch_instrument_on?
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -259,7 +259,7 @@ class OrdersController < ApplicationController
     end
 
     if order_purchaser.success?
-      should_show_admin_hold_warning = @order.order_details.map(&:reservation).compact.map(&:conflicting_admin_reservation).any?
+      should_show_admin_hold_warning = @order.order_details.map(&:reservation).compact.any?(&:conflicting_admin_reservation)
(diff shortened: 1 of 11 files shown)


meara:
Yep, planning on it. Still marked WIP. :)

meara:
Feature's not done (see to-dos in description), but if you have any comments feel free to add them.

app/controllers/orders_controller.rb
```diff
@@ -259,6 +259,9 @@ def purchase
     end
 
     if order_purchaser.success?
+      should_show_admin_hold_warning = @order.order_details.map(&:reservation).compact.map(&:conflicting_admin_reservation).any?
```
jhanggi: Does `@order.order_details.map(&:reservation).compact.any?(&:conflicting_admin_reservation)` work? It's a minor thing, but that would be faster in the average case because it would stop looking up conflicts as soon as it finds one.

app/controllers/reservations_controller.rb
```diff
@@ -99,6 +99,9 @@ def create
       @reservation = creator.reservation
       flash[:notice] = I18n.t "controllers.reservations.create.success"
 
+      should_show_admin_hold_warning = creator.reservation.conflicting_admin_reservation.present?
+      flash[:error] = I18n.t('controllers.reservations.create.admin_hold_warning') if should_show_admin_hold_warning
```
jhanggi: Double quotes. I don't have a strong feeling on this, but I probably wouldn't use the intermediate variable.

I think it might be worth adding a `Reservation#conflicting_admin_reservation?` method to make this and the other places a little shorter.

app/support/reservations/validations.rb
```diff
@@ -81,6 +82,27 @@ def conflicting_reservation(options = {})
...
+    conflicting_admin_reservation.nil?
+  end
+
+  def does_not_conflict_with_admin_reservation
+    errors.add(:base, :conflict) if conflicting_admin_reservation
+  end
+
+  def conflicting_admin_reservation
+    conflicting_reservations =
+      Reservation
```
jhanggi: Part of me wants to DRY up this scope, but I'm not sure if it's worth it.

meara: I'm gonna vote for not worth it right now.

PR #1245, opened 2017-11-03
[#139318] Ignore ActionController::UnknownFormat in ExceptionNotifier

jhanggi:
There's no real need to be notified of this error. This happens if a js/json path meant for ajax is called without a format header. It is already ignored by Rollbar.

This also ignores `ActionController::UnknownHttpMethod` in rollbar. We sometimes see those errors when we're getting vulnerability scanned.

This will get a little funky going into the schools. NU already has the `ActionController::UnknownHttpMethod` ignore and Dartmouth doesn't use ExceptionNotifier. But this should help get things closer in line.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -75,6 +75,7 @@ Nucore::Application.configure do
   # config.force_ssl = true
 
   config.middleware.use ExceptionNotification::Rack,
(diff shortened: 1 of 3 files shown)


meara:
I'm fine with this in general--what's the plan for managing merging this into the various forks, given their different setups?

jhanggi:
We'll just have to deal with the merge conflicts as they arise.

PR #1246, opened 2017-11-03
[#139135] Disable cross-facility journal creation

jhanggi:
DC accidentally created one and it caused confusion. NU times out on this
view and they don't really use the billing admin role. UIC does their journaling
using an outside system (Banner Feeder).

### DIFF
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -2,7 +2,7 @@ require "set"
 
 class Journal < ActiveRecord::Base
 
-  class CreationError < StandardError; end
+  class CreationError < NUCore::Error; end
 
   module Overridable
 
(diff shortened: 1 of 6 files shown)


jhanggi:
@meara Updated. There are still a few places where the code still seems to support cross-facility journals, but cleaning up all those places seemed more effort that this request warrants.

app/models/journal.rb
```diff
@@ -2,7 +2,7 @@
 
 class Journal < ActiveRecord::Base
 
-  class CreationError < StandardError; end
+  class CreationError < NUCore::Error; end
```
meara: Nice.

PR #1251, opened 2017-11-16
Upgrade eye-patch to support memory checks

jhanggi:
We will need to update the `eye.yml.erb` on each of the schools' servers. We will probably tune the checks for each school based on how much memory they have available.


meara:
While we're messing with eye-patch, I happened to notice this ticket: https://pm.tablexi.com/issues/134965

@jhanggi is that worth doing together with this?

jhanggi:
That's on the imserc/mmpf app. Those have very low memory footprints, so I'm not that worried about them right now.

PR #1252, opened 2017-11-17
[#126403] create multiple admin holds

meara:
https://pm.tablexi.com/issues/126403

### DIFF
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -183,6 +183,7 @@ class FacilityReservationsController < ApplicationController
     admin_params = params.require(:admin_reservation).permit(:admin_note,
                                                              :expires_mins_before,
                                                              :category,
+                                                             :repeats,
                                                              :repeat_frequency,
                                                              :repeat_end_date)
     admin_params[:expires_mins_before] = nil if params[:admin_reservation][:expires] == "0"
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -102,6 +102,7 @@ class FacilityReservationsController < ApplicationController
                    @instrument.admin_reservations.build(duration_mins: @instrument.min_reserve_mins)
     @reservation = @reservation.becomes(AdminReservation)
     @reservation.round_reservation_times
+    @reservation_form = AdminReservationForm.new(@reservation)
     set_windows
 
     render layout: "two_column"
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -114,10 +114,13 @@ class FacilityReservationsController < ApplicationController
     @instrument =
       current_facility.instruments.find_by!(url_name: params[:instrument_id])
     @reservation = @instrument.admin_reservations.new
-    @reservation.assign_attributes(admin_reservation_params.merge(created_by: current_user))
-    @reservation.assign_times_from_params(params[:admin_reservation])
+    @reservation_form = AdminReservationForm.new(@reservation)
+    @reservation_form.assign_attributes(admin_reservation_params.merge(created_by: current_user))
+    # @reservation.assign_attributes(admin_reservation_params.merge(created_by: current_user))
+    # @reservation.assign_times_from_params(params[:admin_reservation])
+    @reservation_form.assign_times_from_params(params[:admin_reservation])
 
-    if @reservation.save
+    if @reservation_form.save
       flash[:notice] = "The reservation has been created successfully."
       redirect_to facility_instrument_schedule_url
     else
@@ -179,8 +182,12 @@ class FacilityReservationsController < ApplicationController
   def admin_reservation_params
     admin_params = params.require(:admin_reservation).permit(:admin_note,
                                                              :expires_mins_before,
-                                                             :category)
+                                                             :category,
+                                                             :repeat_frequency,
+                                                             :repeat_end_date)
     admin_params[:expires_mins_before] = nil if params[:admin_reservation][:expires] == "0"
+    admin_params[:repeat_frequency] = nil if params[:admin_reservation][:repeats] == "0"
+    admin_params[:repeat_end_date] = nil if params[:admin_reservation][:repeats] == "0"
     admin_params
   end
 
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -102,8 +102,9 @@ class FacilityReservationsController < ApplicationController
                    @instrument.admin_reservations.build(duration_mins: @instrument.min_reserve_mins)
     @reservation = @reservation.becomes(AdminReservation)
     @reservation.round_reservation_times
+
     @reservation_form = AdminReservationForm.new(@reservation)
-    @reservation_form.repeat_end_date = format_usa_date(@reservation.reserve_end_at)
+    @reservation_form.repeat_end_date = @reservation.reserve_end_at
     set_windows
 
     render layout: "two_column"
--- a/app/controllers/facility_reservations_controller.rb
+++ b/app/controllers/facility_reservations_controller.rb
@@ -103,6 +103,7 @@ class FacilityReservationsController < ApplicationController
     @reservation = @reservation.becomes(AdminReservation)
     @reservation.round_reservation_times
     @reservation_form = AdminReservationForm.new(@reservation)
+    @reservation_form.repeat_end_date = format_usa_date(@reservation.reserve_end_at)
     set_windows
 
     render layout: "two_column"
@@ -116,8 +117,6 @@ class FacilityReservationsController < ApplicationController
     @reservation = @instrument.admin_reservations.new
     @reservation_form = AdminReservationForm.new(@reservation)
     @reservation_form.assign_attributes(admin_reservation_params.merge(created_by: current_user))
-    # @reservation.assign_attributes(admin_reservation_params.merge(created_by: current_user))
-    # @reservation.assign_times_from_params(params[:admin_reservation])
     @reservation_form.assign_times_from_params(params[:admin_reservation])
 
     if @reservation_form.save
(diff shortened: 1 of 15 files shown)


app/views/facility_reservations/_admin_reservation_repeat_fields.html.haml
```diff
@@ -0,0 +1,4 @@
+.checkboxControl
+  = f.input :repeats?, as: :boolean, input_html: { data: { disables: ".admin_reservation_repeat_frequency, .admin_reservation_repeat_end_date" } }, wrapper_html: { class: "checkboxControl__checkbox" }
+  = f.input :repeat_frequency, collection: AdminReservation::REPEAT_OPTIONS.map(&:titleize), selected: "Weekly"
+  = f.input :repeat_end_date, input_html: { class: "datepicker" }
```
meara: Can't figure out how to get this input to start with the default value of today like the other datepickers on the page do.

app/controllers/facility_reservations_controller.rb
```diff
@@ -113,10 +114,13 @@ def create
     @instrument =
       current_facility.instruments.find_by!(url_name: params[:instrument_id])
     @reservation = @instrument.admin_reservations.new
-    @reservation.assign_attributes(admin_reservation_params.merge(created_by: current_user))
-    @reservation.assign_times_from_params(params[:admin_reservation])
+    @reservation_form = AdminReservationForm.new(@reservation)
+    @reservation_form.assign_attributes(admin_reservation_params.merge(created_by: current_user))
+    # @reservation.assign_attributes(admin_reservation_params.merge(created_by: current_user))
+    # @reservation.assign_times_from_params(params[:admin_reservation])
+    @reservation_form.assign_times_from_params(params[:admin_reservation])
```
jhanggi: I feel like this could probably be pulled into the form object.

app/controllers/facility_reservations_controller.rb
```diff
@@ -178,8 +182,12 @@ def destroy
   def admin_reservation_params
     admin_params = params.require(:admin_reservation).permit(:admin_note,
                                                              :expires_mins_before,
-                                                             :category)
+                                                             :category,
+                                                             :repeat_frequency,
+                                                             :repeat_end_date)
     admin_params[:expires_mins_before] = nil if params[:admin_reservation][:expires] == "0"
+    admin_params[:repeat_frequency] = nil if params[:admin_reservation][:repeats] == "0"
```
jhanggi: I feel like now that we're using a form object, these things could be pulled into the form.

app/forms/admin_reservation_form.rb
```diff
@@ -0,0 +1,44 @@
...
+    if [@reservation, self].map(&:valid?).all?
+      @reservation.save
+      create_recurring_reservations
+    else
+      self.errors.merge(@registration.errors)
+      false
+    end
+  end
+
+  def repeats?
```
jhanggi: Is this going to be used anywhere?

app/models/admin_reservation.rb
```diff
@@ -10,6 +10,13 @@ class AdminReservation < Reservation
     other
   ).freeze
 
+  REPEAT_OPTIONS = %w(
```
jhanggi: I feel like these probably belong on the form object if that's the only place they're used.

app/views/facility_reservations/_admin_reservation_repeat_fields.html.haml
jhanggi: Not sure quite where the right place to do it without thinking a little more, but you need to set the `repeat_end_date` attribute before you pass it to the partial. The form object initializer might be an ok place: `@reservation_form.repeat_end_date = reservation.reserve_end_at`, or something like that.

app/forms/admin_reservation_form.rb
```diff
@@ -0,0 +1,72 @@
...
+    if [@reservation, self].map(&:valid?).all?
+      @reservation.save
+      create_recurring_reservations
+    else
+      self.errors.merge(@registration.errors)
+      false
+    end
+  end
+
+  def create_recurring_reservations
```
meara: So this is super WIP-y, feel free to totally restructure this entire create_recurring business if you do end up picking this up, there are probably much better ways of doing this than my half-formed thought here.

jhanggi:
Did a little bit of cleanup, but didn't really get into the recurring event creation. I would recommend using an external library. I think it'll end up being quicker to implement and will allow more flexibility in the future. started looking into various options and put a few options in our dev slack channel to see what other people recommend.

app/controllers/facility_reservations_controller.rb
```diff
@@ -178,8 +182,13 @@ def destroy
   def admin_reservation_params
     admin_params = params.require(:admin_reservation).permit(:admin_note,
                                                              :expires_mins_before,
-                                                             :category)
+                                                             :category,
+                                                             :repeats,
+                                                             :repeat_frequency,
+                                                             :repeat_end_date)
     admin_params[:expires_mins_before] = nil if params[:admin_reservation][:expires] == "0"
+    admin_params[:repeat_frequency] = nil if params[:admin_reservation][:repeats] == "0"
```
jhanggi: Just an option: `ActiveRecord::Type::Boolean.new.type_cast_from_database("0") == false`

However, this moves in Rails 5.

https://gist.github.com/equivalent/3825916

app/forms/admin_reservation_form.rb
```diff
@@ -0,0 +1,105 @@
...
+    # errors does not support `merge`
+    @reservation.errors.each do |k, error_messages|
+      errors.add(k, error_messages)
+    end
+    errors.none?
+  end
+
+  def save
+    if valid?
+      # @reservation.save
```
jhanggi: Leftover comment?

app/forms/admin_reservation_form.rb
```diff
@@ -0,0 +1,105 @@
...
+      reservations = build_recurring_reservations
+      reservations.map(&:save).all?
```
jhanggi: Is there a scenario where one might fail, but not others? Should we wrap this in a transaction and use `save!`?

app/models/recurrence.rb
```diff
@@ -0,0 +1,38 @@
...
+    self
+  end
+
+  def monthly
+    @schedule.add_recurrence_rule IceCube::Rule.monthly.until(@until_time)
+    self
+  end
+
+  def to_enum
+    @schedule.all_occurrences_enumerator
```
jhanggi: For another time, to fully abstract this adapter class, this should probably return a class of our own making rather than `IceCube::Occurrence`. Then we could abide by our naming standards rather than theirs, e.g. `start_time` to `start_at`.

I think calling `map { |t| Occurrence.new(start_at: t.start_time, end_at: t.end_time) }` on this would work (or it might need `.lazy.map`.

But like I said, for another time.

spec/features/admin/creating_admin_reservation_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+    visit new_facility_instrument_reservation_path(facility, instrument)
+
+    fill_in "Reserve start date", with: "10/17/#{fiscal_year}"
+    fill_in "Duration mins", with: "30"
+    check "Repeats"
+    select "Daily", from: "Repeat frequency"
+    fill_in "Repeat end date", with: "10/20/#{fiscal_year}"
+    click_button "Create"
+
+    expect(page).to have_content "10/19/#{fiscal_year} 9:30 AM - 10:00 AM"
```
jhanggi: For completeness, we should probably test that it has `10/17` and `10/18`.

spec/forms/admin_reservation_form_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+  let(:instrument) { create(:setup_instrument) }
+  let(:start_date) { Time.zone.parse("2017-10-17 12:00:00") }
+  let(:admin_reservation) { build(:admin_reservation, product: instrument, reserve_start_at: start_date, duration: 1.hour) }
+  let!(:form) { AdminReservationForm.new(admin_reservation) }
+
+  describe "validations" do
+    describe "cannot_exceed_max_end_date" do
+      it "invalid past max end date" do
+        form.assign_attributes(repeats: "1", repeat_frequency: "weekly", repeat_end_date: format_usa_date(start_date + 14.weeks))
+        expect(form.valid?).to eq false
```
jhanggi: In addition to this validity check, we should also check `expect(form.errors).to be_added(:repeat_end_date, :too_far_in_future)`. That way we see it's invalid for the right reason.

And, it's a matter of taste, but I lean towards `expect(form).not_to be_valid`.

spec/forms/admin_reservation_form_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+    context "weekdays_only" do
+      before do
+        form.assign_attributes(repeats: "1", repeat_frequency: "weekdays_only", repeat_end_date: format_usa_date(start_date + 7.days))
+      end
+
+      it "builds the right number of reservations" do
+        expect(form.build_recurring_reservations.length).to eq 6
+      end
+
+      it "sets the group id", :aggregate_failures do
```
jhanggi: Maybe change this description to say they're all the same group_id?

spec/forms/admin_reservation_form_spec.rb
```diff
@@ -0,0 +1,108 @@
...
+
+      it "builds the right number of reservations" do
+        expect(form.build_recurring_reservations.length).to eq 4
+      end
+
+      context "on the 31st" do
+        let(:start_date) { Time.zone.parse("2017-10-31 12:00:00") }
+
+        it "builds reservations with expected dates" do
+          expect(form.build_recurring_reservations.map { |t| t.reserve_start_at.strftime("%F") }).to eq ["2017-10-31", "2017-11-30", "2017-12-31", "2018-01-31"]
```
jhanggi: For completeness, can we include February in here? And what do you think of checking a Feb w/ a leap day?

app/forms/admin_reservation_form.rb
meara: I think it's really unlikely--I can't think of a situation where they wouldn't either succeed or fail all together.

app/controllers/facility_reservations_controller.rb
meara: I feel like if this changes in Rails 5, not worth changing now.

PR #1253, opened 2017-11-21
[#139725] Public calendar icon should respect admin reservations

jhanggi:
* Rename Instrument#available? to #walkup_available? to better clarify its use-case.
* Refactoring to how reservation validations are performed
* Remove Reservation#save_extended_reservation by using a :user_purchase context
* I18n reservation validation messages


meara:
I poked around some more to check the validations and yeah, things look good to me.

PR #1254, opened 2017-12-01
Update eye template to use SMTP

jhanggi:
Most of nucore installations use SMTP, so use that instead of aws for
sending notifications. I've already made these updates to the server on NU & UIC. We're getting crash/flapping alerts into opsgenie now.

### DIFF
--- a/config/eye.yml.erb.template
+++ b/config/eye.yml.erb.template
@@ -17,14 +17,16 @@ triggers:
       within: 5 minutes
 
 notifications:
-  - name: monitoring
-    type: ses
+  - name: crash
+    type: mail
     level: info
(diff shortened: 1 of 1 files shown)


config/eye.yml.erb.template
```diff
@@ -17,14 +17,16 @@ triggers:
...
-    contact: nucore+eye@example.com
+    contact: eye@nucore.example.com
     config:
-      from: noreply@example.com
-      access_key_id: ~
-      secret_access_key: ~
+      from_mail: noreply@example.com
+      host: smtp.example.com
+      port: 25
+      # domain: ~
```
pturley: that doesnt need to be there

PR #1255, opened 2017-12-01
[#139070] retroactive admin reservations

meara:
https://pm.tablexi.com/issues/139070
https://pm.tablexi.com/issues/139097

### DIFF
--- a/app/assets/javascripts/reservations.js
+++ b/app/assets/javascripts/reservations.js
@@ -9,10 +9,10 @@ $(document).ready(function() {
     if (typeof minDaysFromNow == "undefined") {
       window['minDaysFromNow'] = 0;
     }
-    $("#datepicker").datepicker({'minDate': minDaysFromNow, 'maxDate': maxDaysFromNow});
+    $("#datepicker").datepicker();
 
     $('.datepicker').each(function(){
-      $(this).datepicker({'minDate': minDaysFromNow, 'maxDate': maxDaysFromNow})
+      $(this).datepicker()
       		.change(function() {
       			var d = new Date(Date.parse($(this).val()));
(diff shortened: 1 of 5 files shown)


meara:
The javascript duration/end time updater seems to rely on `set_windows` because removing `set_windows` breaks it. I feel like refactoring that is probably out of scope for this change.

app/assets/javascripts/reservations.js
```diff
@@ -9,10 +9,10 @@ $(document).ready(function() {
     if (typeof minDaysFromNow == "undefined") {
       window['minDaysFromNow'] = 0;
     }
-    $("#datepicker").datepicker({'minDate': minDaysFromNow, 'maxDate': maxDaysFromNow});
+    $("#datepicker").datepicker();
 
     $('.datepicker').each(function(){
-      $(this).datepicker({'minDate': minDaysFromNow, 'maxDate': maxDaysFromNow})
```
jhanggi: This is also used by the user reservation form, so we can't eliminate these options w/o changing to the `__data` version (probably not worth it right now).

I think you can probably do something for `maxDaysFromNow` like what we did for `minDaysFromNow` so that it's at least defined on the pages that don't use it.

app/assets/javascripts/reservations.js
```diff
@@ -8,6 +8,7 @@ $(document).ready(function() {
   function init_datepickers() {
     if (typeof minDaysFromNow == "undefined") {
       window['minDaysFromNow'] = 0;
+      window['maxDaysFromNow'] = 365;
```
jhanggi: It's probably fine since we'll probably always have neither or both, but to be on the safe side, can we do the two variables in separate `typeof` `undefined` blocks.

PR #1257, opened 2017-12-04
[#137694] email footer

meara:
https://pm.tablexi.com/issues/137694

### DIFF
--- a/app/views/purchase_notifier/order_receipt.html.haml
+++ b/app/views/purchase_notifier/order_receipt.html.haml
@@ -45,9 +45,9 @@
             = order_detail.reservation
         %td= order_detail.wrapped_quantity
         %td= order_detail.order_status.name
-        %td= order_detail.display_cost
+        %td= order_detail.display_cost unless order_detail.canceled?
         %td= order_detail.display_subsidy
-        %td= order_detail.display_total
+        %td= order_detail.display_total unless order_detail.canceled?
         %td
           - if order_detail.note.present?
             = order_detail.note
@@ -56,7 +56,7 @@
       %td{colspan: 6}
         %strong= text(".total")
       %td.currency
-        = number_to_currency(@order.order_details.to_a.sum { |d| d.total || 0 })
+        = number_to_currency(@order.total)
       %td
 
 - if @order.any_details_estimated?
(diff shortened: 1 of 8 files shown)


app/views/purchase_notifier/order_receipt.html.haml
```diff
@@ -45,9 +45,9 @@
             = order_detail.reservation
         %td= order_detail.wrapped_quantity
         %td= order_detail.order_status.name
-        %td= order_detail.display_cost
+        %td= order_detail.display_cost unless order_detail.canceled?
         %td= order_detail.display_subsidy
```
jhanggi: We should probably hide `subsidy` as well if it's canceled.

spec/models/order_spec.rb
```diff
@@ -54,70 +54,55 @@ def define_purchasable_instrument
...
-                                                    actual_subsidy: subsidy,
-                                                    price_policy_id: price_policy.id,
-                                                   ))
-        end
-
-        @total = @cost - @subsidy
-      end
+      let!(:order_detail_1) { create(:order_detail, order: order, product: item, account: account, actual_cost: 10, actual_subsidy: 5, price_policy: price_policy) }
+      let!(:order_detail_2) { create(:order_detail, order: order, product: item, account: account, actual_cost: 12, actual_subsidy: 0, price_policy: price_policy) }
+      let(:cost) { 22 }
```
jhanggi: If they're only used once, I would just put these inline in the examples.

spec/models/order_spec.rb
```diff
@@ -54,70 +54,55 @@ def define_purchasable_instrument
...
-                                                   ))
-        end
+      it "does not include canceled state order details in its total" do
+        create(:order_detail, order: order, product: item, account: account, actual_cost: 5, actual_subsidy: 0, price_policy: price_policy, state: 'canceled')
+
+        expect(order.total).to eq total
+      end
+
+      it "does include canceled with fee order details in its total" do
+        create(:order_detail, order: order, product: item, account: account, actual_cost: 5, actual_subsidy: 0, price_policy: price_policy, state: 'complete', canceled_at: Time.current)
```
jhanggi: I'm surprised we don't have them, but it might be nice to add factory traits for `complete`, `canceled`, and `canceled_with_cost`.

spec/models/order_spec.rb
jhanggi: Looks like you put the other expected values inline. Do we want to put these inline as well?

PR #1258, opened 2017-12-05
[#139509] basic instruments index

meara:
https://pm.tablexi.com/issues/139509

Currently looks something like this:
screen shot 2017-12-22 at 1 46 12 pm
screen shot 2017-12-22 at 1 46 23 pm


jhanggi:
But there are test failures.

app/controllers/instruments_controller.rb
```diff
@@ -1,6 +1,6 @@
 class InstrumentsController < ProductsCommonController
 
-  customer_tab :show, :public_schedule
+  customer_tab :show, :public_schedule, :list
```
jhanggi: This is definitely a place where I'm not crazy about the mixing of admin and regular user paths in the same controller. It would be really nice to split that out at some point. What if we called this action `public_list` just to clarify its role? It can still route to `/list`.

app/controllers/instruments_controller.rb
```diff
@@ -9,7 +9,14 @@ class InstrumentsController < ProductsCommonController
   skip_before_action :authenticate_user!, only: [:public_schedule]
```
jhanggi: Do we want to put this new action outside of login?

app/controllers/instruments_controller.rb
```diff
@@ -9,7 +9,14 @@ class InstrumentsController < ProductsCommonController
...
   skip_authorize_resource only: [:public_schedule]
 
-  skip_before_action :init_product, only: [:instrument_statuses]
+  skip_before_action :init_product, only: [:instrument_statuses, :list]
+
+  # GET /facilities/:facility_id/instruments/list
+  def list
+    @instruments = Instrument.active.in_active_facility.order(:name).includes(:facility)
```
jhanggi: I think you could do `joins(:facility).merge(Facility.active)` instead. I'm not sure how valuable the scope is.

app/assets/stylesheets/nucore.scss
```diff
@@ -377,3 +377,16 @@ textarea.wide {
 .checkboxControl__checkbox {
   margin-right: 2.5em;
 }
+
+.productListLine {
+  display: flex;
+}
+
+.productListLine__productName {
```
jhanggi: I'm not quite sure what these do, but would `productListLine__first` and `last` be easier to work with in the future?

PR #1264, opened 2017-12-08
[#137157] training requests are sortable

meara:
https://pm.tablexi.com/issues/137157

### DIFF
--- a/app/controllers/training_requests_controller.rb
+++ b/app/controllers/training_requests_controller.rb
@@ -1,5 +1,7 @@
 class TrainingRequestsController < ApplicationController
 
+  include NewInprocessController
+
   admin_tab :index
 
   before_action :authenticate_user!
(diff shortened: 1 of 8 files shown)


app/controllers/training_requests_controller.rb
```diff
@@ -1,5 +1,7 @@
 class TrainingRequestsController < ApplicationController
 
+  include NewInprocessController
```
jhanggi: I don't love this module, especially with the name it currently has: definitely feels out of place here. It currently has two responsibilities: sorting and `index`. I think they should probably be split apart.

I think it would be nice to wrap up some of the sort functionality into an object.

Just throwing this out there:
```
class SortingObject # better name?
  def initialize(sort_lookup_hash)
    @sort_lookup_hash = sort_lookup_hash
  end

  def sort(objects, sort_key, direction)
     #... all the magic
  end
end
```

Then in the controller, you could reference that object, and contain the lookup hash. Or you could have that as a base class, and then have a `OrderDetailSorter` and a `TrainingRequestSorter` that contain the hash.

I'm not opposed to keeping a module that you can include to wrap up the `helper_method`s.

PR #1287, opened 2017-12-15
[#137159] add training requested filter to bulk emailer

meara:
https://pm.tablexi.com/issues/137159

### DIFF
--- a/vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js.coffee
+++ b/vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js.coffee
@@ -4,8 +4,8 @@ class window.BulkEmailSearchForm
     @_initDateRangeSelectionHandlers()
 
   hideNonRestrictedProducts: ->
-    user_types = @selectedUserTypes().toArray()
-    required_selected = $(user_types).filter(['authorized_users', 'training_requested']).toArray()
+    user_types = @selectedUserTypes()
+    required_selected = user_types.filter(['authorized_users', 'training_requested'])
     # if either authorized_users or training_requested is selected (and nothing else), we should display only restricted products
     required_selected.length > 0 && required_selected.length == user_types.length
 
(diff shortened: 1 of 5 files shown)


vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js.coffee
```diff
@@ -9,6 +9,23 @@ class window.BulkEmailSearchForm
 
   selectedUserTypes: -> @$userTypeCheckboxes().filter(':checked').map -> @.value
 
+  dateIrrelevantSelectedOnly: ->
+    user_types = @selectedUserTypes().toArray()
+    (user_types.includes('authorized_users') || user_types.includes('training_requested')) &&
```
jhanggi: This is kind of a confusing and very verbose check.

What if we add a data attributes to these fields  marking them as date relevant/irrelevant and then look to make sure our logic passes? I'd prefer if we can avoid that mapping in the JS if we can.

vendor/engines/bulk_email/app/services/bulk_email/recipient_searcher.rb
```diff
@@ -62,6 +62,14 @@ def search_authorized_users
         .reorder(*self.class::DEFAULT_SORT)
     end
 
+    def search_training_requested
+      users
+        .joins(:training_requests)
+        .uniq
+        .where(training_requests: { product_id: product_ids_from_params })
```
jhanggi: I think we handle it, but can we make sure this does not include users who issued a training request, but then were later approved? We actually destroy training requests on the "destroy" action, so this shouldn't include users who's request was denied.

vendor/engines/bulk_email/app/assets/javascripts/bulk_email/bulk_email.js.coffee
```diff
@@ -3,33 +3,45 @@ class window.BulkEmailSearchForm
     @_initUserTypeChangeHandler()
     @_initDateRangeSelectionHandlers()
 
-  authorizedUsersSelectedOnly: ->
+  hideNonRestrictedProducts: ->
     user_types = @selectedUserTypes()
-    user_types.length == 1 && user_types[0] == 'authorized_users'
+    required_selected = user_types.filter(['authorized_users', 'training_requested'])
+    # if either authorized_users or training_requested is selected (and nothing else), we should display only restricted products
```
jhanggi: Maybe split the comment? I'm seeing a lint violation.

meara:
@jhanggi what kind of tests were you thinking? I thought we had decided unit testing the javascript wasn't worth the time required to set up?

jhanggi:
Let's not test the JS piece around disabling/enable other fields, but let's at least test that if you choose the new checkbox, it returns the people we expect.

PR #1294, opened 2017-12-20
[#136545] Add Filtering to Reconciliation Pages

jhanggi:
As part of this feature, this creates a parallel path for doing transaction searching. The end goal is this will replace the current `TransactionSearch` module that is included into controllers and used with the `_with_search` action. That will give us more control over each individual action while also hopefully much easier to reason about.

That is why there is a `_search2` view and a `TransactionChosen2Input`. Ideally, they will replace the original, but it seemed more prudent to create this parallel method than to try to refactor the original.

screen shot 2017-12-20 at 12 40 28 pm

The new chosen input has some additional features that will come to play later like being able to render `data` attributes on individual `options`. This is used in the product filter. I did not leave the product filter in this PR, because we're not using it here, but I did test with it.


jhanggi:
Code Climate warning is about complexity of a method I moved. I don't think it's worth refactoring that right now.

app/services/transaction_search/searcher.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def search(order_details, params)
+      @searchers.reduce(Results.new(order_details)) do |results, searcher_class|
+        searcher = searcher_class.new(results.order_details)
+
+        search_params = params[searcher_class.key.to_sym]
+        search_params = Array(search_params).reject(&:blank?) unless searcher.multipart?
+
+        # TODO: Collapse optimized into search within the searchers
```
meara: Would it make sense to clarify what this todo is waiting on?

PR #1295, opened 2017-12-20
Replace "All Transactions" search with newer search method

jhanggi:
This continues from #1294 and replaces the "All Transactions" search logic, e.g. http://localhost:3000/facilities/example/transactions with the new search. It gets rid of the `_with_search` metaprogramming. Hopefully, this is easier to reason about.

The feature spec I wrote works for both master and this branch. I think more specific unit tests are going to be better for testing the search criteria than the existing controller specs, which is why I did one happy-path feature spec and killed the controller specs for this action.

### DIFF
--- a/app/services/transaction_search/searcher.rb
+++ b/app/services/transaction_search/searcher.rb
@@ -5,12 +5,12 @@ module TransactionSearch
     # Do not modify this directly. Use TransactionSearch.register instead.
     cattr_accessor(:default_searchers) do
       [
-        :facilities,
-        :accounts,
-        :products,
-        :account_owners,
-        :order_statuses,
-        :date_ranges,
+        TransactionSearch::FacilitySearcher,
+        TransactionSearch::AccountSearcher,
+        TransactionSearch::ProductSearcher,
+        TransactionSearch::AccountOwnerSearcher,
+        TransactionSearch::OrderStatusSearcher,
+        TransactionSearch::DateRangeSearcher,
       ]
     end
 
--- a/app/services/transaction_search/searcher.rb
+++ b/app/services/transaction_search/searcher.rb
@@ -2,8 +2,22 @@ module TransactionSearch
 
   class Searcher
 
+    # Do not modify this directly. Use TransactionSearch.register instead.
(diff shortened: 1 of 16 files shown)


meara:
@jhanggi is there a reason this is marked WIP? Were you planning to add more to this, or is this ready for review?

jhanggi:
I think I wanted to run with the change to the reconciliation for a little to make sure that was solid before moving forward with this one. There's nothing holding it back other than the risk that something might go wrong in this change.

app/services/transaction_search/searcher.rb
```diff
@@ -2,8 +2,22 @@ module TransactionSearch
 
   class Searcher
 
+    # Do not modify this directly. Use TransactionSearch.register instead.
```
meara: Why?

app/services/transaction_search/searcher.rb
```diff
@@ -44,6 +61,12 @@ def options
         @search_options.dup
       end
 
+      def to_h
```
meara: This seems like a weird method name to me. Not what I really think `to_h` does...

app/services/transaction_search/searcher.rb
jhanggi: Adding a new searcher also needs to add an `attr_accessor` to `SearchForm`. The `register` method does that for you.

app/services/transaction_search/searcher.rb
jhanggi: It is converting itself to a hash... Might be doing too much to be named that. Maybe `to_options` or `to_options_by_searcher`? This is in a `Results` object so it's a hash keyed by the searcher's key and the values are the array of options (e.g. facilities, products) to be used on the resulting form.

app/services/transaction_search/searcher.rb
meara: Mind adding that to the comment?

app/services/transaction_search/searcher.rb
meara: I think I like `to_options_by_searcher`. That seems much clearer IMO.

PR #1316, opened 2018-01-04
[#136545] restyle filter button

meara:
https://pm.tablexi.com/issues/136545

Update button name from "Search" to "Filter", move it beneath the statements field, and add padding to disconnect the search form from the table below.

Before: 
screen shot 2018-01-04 at 2 57 14 pm

After:
screen shot 2018-01-04 at 2 56 07 pm

### DIFF
--- a/app/assets/stylesheets/app/transaction_search.scss
+++ b/app/assets/stylesheets/app/transaction_search.scss
@@ -10,9 +10,15 @@
 }
 
 .search_form input[type=submit] {
-  clear:both;
-  float:right;
+  clear: both;
+  float: right;
   margin-bottom: 0.5em;
+}
+
+.search_form input[type=submit].submit--float-left {
+  float: left;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/app/transaction_search.scss
```diff
@@ -10,11 +10,17 @@
...
 
 .search_form input[type=submit] {
-  clear:both;
-  float:right;
+  clear: both;
+  float: right;
   margin-bottom: 0.5em;
 }
 
+.search_form input[type=submit].submit--float-left {
```
meara: @wtfluckey my question here is probably, is there a less awkward way to select the specific element I want (in the html file below), without changing the styling on line 12? Unfortunately the styling on line 12 appears to be re-used enough that it's not worth hunting down all the places that styling is applied to replace it with a more sensible selection scheme.

PR #1323, opened 2018-01-12
[#137168] price change reason

meara:
https://pm.tablexi.com/issues/137168

### DIFF
--- a/app/assets/javascripts/app/manage_orders.coffee
+++ b/app/assets/javascripts/app/manage_orders.coffee
@@ -86,12 +86,6 @@ class OrderDetailManagement
       total = row.find('.cost input').val() - row.find('.subsidy input').val()
       row.find('.total input').val(total.toFixed(2))
       self.notify_of_update $(row).find('input[name*=total]')
-      enablePriceChangeReasonField()
-
-  enablePriceChangeReasonField = ->
-    input_field = $("#order_detail_price_change_reason")
-    input_field.prop("disabled", false)
-    # input_field.parent().parent().toggle()
 
   disableSubmit: ->
     @waiting_requests ||= 0
--- a/app/assets/javascripts/app/manage_orders.coffee
+++ b/app/assets/javascripts/app/manage_orders.coffee
@@ -86,6 +86,12 @@ class OrderDetailManagement
       total = row.find('.cost input').val() - row.find('.subsidy input').val()
       row.find('.total input').val(total.toFixed(2))
       self.notify_of_update $(row).find('input[name*=total]')
+      enablePriceChangeReasonField()
+
+  enablePriceChangeReasonField= ->
+    input_field = $("#order_detail_price_change_reason")
+    input_field.prop("disabled", false)
+    # input_field.parent().parent().toggle()
 
   disableSubmit: ->
     @waiting_requests ||= 0
--- a/app/assets/javascripts/app/manage_orders.coffee
+++ b/app/assets/javascripts/app/manage_orders.coffee
@@ -88,7 +88,7 @@ class OrderDetailManagement
       self.notify_of_update $(row).find('input[name*=total]')
       enablePriceChangeReasonField()
 
-  enablePriceChangeReasonField= ->
+  enablePriceChangeReasonField = ->
     input_field = $("#order_detail_price_change_reason")
     input_field.prop("disabled", false)
     # input_field.parent().parent().toggle()
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -46,6 +46,7 @@ class OrderDetail < ActiveRecord::Base
   belongs_to :assigned_user, class_name: "User", foreign_key: "assigned_user_id"
   belongs_to :created_by_user, class_name: "User", foreign_key: :created_by
   belongs_to :dispute_by, class_name: "User"
+  belongs_to :price_changed_by_user, class_name: "User"
   belongs_to :order_status
   belongs_to :account
   belongs_to :bundle, foreign_key: "bundle_product_id"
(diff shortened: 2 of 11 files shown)


db/migrate/20180111200920_add_price_change_reason_to_order_details.rb
```diff
@@ -0,0 +1,8 @@
+class AddPriceChangeReasonToOrderDetails < ActiveRecord::Migration
+  def change
+    add_column :order_details, :price_change_reason, :string
+    add_column :order_details, :price_changed_by_user_id, :integer
+
+    add_index :order_details, :price_changed_by_user_id
```
jhanggi: Should we do this with an actual foreign key?

app/services/order_details/param_updater.rb
```diff
@@ -58,7 +59,8 @@ def update_attributes(params)
 
     user_newly_assigned = @order_detail.assigned_user_id_changed? && @order_detail.assigned_user.present?
 
-    @order_detail.manually_priced!
+    @order_detail.manually_priced! # this seems like a misleading method name
```
jhanggi: I'm not opposed to renaming this. This is probably a good place to update set the price_changed_by_user, though this method is now getting pretty big.

One thing we'll want to take into account is that if a user manually changes the price, then another user comes and changes something else (like the note), it should not overwrite the price changed by user.

app/assets/javascripts/app/manage_orders.coffee
```diff
@@ -86,6 +86,12 @@ class OrderDetailManagement
       total = row.find('.cost input').val() - row.find('.subsidy input').val()
       row.find('.total input').val(total.toFixed(2))
       self.notify_of_update $(row).find('input[name*=total]')
+      enablePriceChangeReasonField()
+
+  enablePriceChangeReasonField = ->
```
jhanggi: I'm not quite set on this, but I feel like we shouldn't even show the field until you adjust the price. And then it should go away if the price is recalculated. My gut feeling is the enabled status should be based on field value `!=` calculated (i.e. expected) value.

app/models/order_detail.rb
```diff
@@ -46,6 +46,7 @@ def set_problem_order
   belongs_to :assigned_user, class_name: "User", foreign_key: "assigned_user_id"
   belongs_to :created_by_user, class_name: "User", foreign_key: :created_by
   belongs_to :dispute_by, class_name: "User"
+  belongs_to :price_changed_by_user, class_name: "User"
```
jhanggi: per codeclimate: "Specify an `:inverse_of` option."

meara: I'm not seeing this in codeclimate for some reason? Confused.

jhanggi: hmm. I have the codeclimate chrome extension, which marks certain things. Maybe there's a rubocop version discrepancy? Don't worry about it, then.

app/models/order_detail.rb
```diff
@@ -107,6 +108,19 @@ def journal_or_statement_date
   validate :account_usable_by_order_owner?, if: ->(o) { o.account_id_changed? || o.order.nil? || o.order.ordered_at.nil? }
   validates_length_of :note, maximum: 1000, allow_blank: true, allow_nil: true
   validate :valid_manual_fulfilled_at
+  validate :require_pricing_note
+
+  def require_pricing_note
+    return unless @manually_priced # && feature flag is off
+    return if cost_estimated?
+
+    dup_od = self.dup
```
meara: Split into separate method

app/services/order_details/param_updater.rb
```diff
@@ -58,7 +59,8 @@ def update_attributes(params)
 
     user_newly_assigned = @order_detail.assigned_user_id_changed? && @order_detail.assigned_user.present?
 
-    @order_detail.manually_priced!
+    @order_detail.manually_priced! # don't auto-reassign price
+    # is this a good place to save price_changed_by_user?
```
meara: Rm comment

app/models/order_detail.rb
```diff
@@ -107,6 +108,28 @@ def journal_or_statement_date
...
+  def actual_costs_differ_from_calculated?
+    dup_od = dup
+    dup_od.reservation = reservation.dup
+    dup_od.assign_price_policy
+
+    dup_od.actual_cost != actual_cost || dup_od.actual_subsidy != actual_subsidy
+  end
+
+  def actual_costs_match_calculated?
+    !actual_costs_differ_from_calculated?
```
jhanggi: Would it be easier to read if we moved the logic into this method with `dup_od.actual_cost == actual_cost && dup_od.actual_subsidy == actual_subsidy` as the boolean logic and then made `actual_costs_differ_from_calculated?` the `!` of this? It's very minor, so it's not a big deal.

app/models/reports/export_raw.rb
```diff
@@ -86,6 +86,8 @@ def default_report_hash
         journal_date: ->(od) { od.journal.journal_date if od.journal },
         reconciled_note: :reconciled_note,
         reconciled_at: :reconciled_at,
+        price_change_reason: :price_change_reason,
+        price_changed_by_user: ->(od) { od.price_changed_by_user.try(:full_name) },
```
jhanggi: I don't have a strong feeling on this, but what if we used `username`? We could also just rely on the `to_s` of `User`, though that includes `(Deactivated)` if the user is deactivated (only UIC has deactivated users currently, but NU does have that feature in the backlog).

meara: Is there a reason we would prefer username over full name? We use full name elsewhere for whodunnits, i.e. in Admin Reservations etc.

PR #1332, opened 2018-01-17
[#139390] Allow extending reservations inside lock window

jhanggi:
You are now allowed to edit a reservation inside the lock window, but
you cannot change the start time and you cannot shorten the duration,
though you may extend it.

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -82,7 +82,7 @@ class Reservation < ActiveRecord::Base
   scope :ongoing, -> { not_ended.where("actual_start_at <= ?", Time.current) }
 
   def self.today
-    for_date(Time.zone.now)
+    for_date(Time.current)
   end
 
   def self.for_date(date)
@@ -94,7 +94,7 @@ class Reservation < ActiveRecord::Base
       .where("reserve_start_at < ?", end_time)
   end
 
-  def self.upcoming(t = Time.zone.now)
+  def self.upcoming(t = Time.current)
     # If this is a named scope differences emerge between Oracle & MySQL on #reserve_end_at querying.
     # Eliminate by letting Rails filter by #reserve_end_at
     joins("LEFT JOIN order_details ON order_details.id = reservations.order_detail_id")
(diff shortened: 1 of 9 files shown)


spec/app_support/reservations/duration_change_validations_spec.rb
```diff
@@ -63,16 +61,32 @@
...
+              reservation.reserve_end_at += 10.minutes
+              expect(validator).to be_valid
+              expect(validator.errors).to be_empty
+            end
+
+            it "is not allowed to shorten the reservation" do
+              reservation.reserve_end_at -= 5.minutes
+              expect(validator).not_to be_valid
+              expect(validator.errors.full_messages)
+                .to include("Reserve end at cannot be shortened inside the lock window")
```
giladshanan: I'm wondering about the wording of the error messages here.  Would 'reservation end time cannot be shortened...' be clearer than 'reserve end at cannot be shortened...' ?

app/models/reservation.rb
```diff
@@ -230,7 +230,7 @@ def reserve_start_at_editable?
   end
 
   def reserve_end_at_editable?
-    outside_lock_window? && Time.zone.now <= reserve_end_at && extendable? && actual_end_at.blank?
+    Time.zone.now <= reserve_end_at && extendable? && actual_end_at.blank?
```
meara: Should this also be `Time.current` since we're using that below?

spec/app_support/reservations/duration_change_validations_spec.rb
meara: @jhanggi I'm not sure what the history is behind the human-readable version of `reserve_end_at` is. Idk if it's worth changing though (i.e. would it cause more user confusion to change now).

jhanggi:
Good call on the messages. I moved the message to `duration_mins`, and now the message shows "Duration cannot be shortened...". It also highlights the duration field in red. I added the `(X hours)` to make that error message wrap below; otherwise it just overflowed the gray box.

meara:
👍  thanks Jason! I think moving the message to duration makes a lot of sense.

PR #1334, opened 2018-01-18
fix specs

meara:
Specs were failing in UIC because the training_requests feature was turned off.

https://circleci.com/gh/tablexi/nucore-uic/725?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link


vendor/engines/bulk_email/spec/controllers/bulk_email/bulk_email_controller_spec.rb
```diff
@@ -11,7 +11,7 @@
   let!(:restricted_item) { FactoryBot.create(:item, facility_account: facility_account, requires_approval: true) }
   let!(:service) { FactoryBot.create(:service, facility_account: facility_account) }
 
-  describe "POST #search" do
+  describe "POST #search", feature_setting: { training_requests: true } do
```
jhanggi: Can we push this change down so it's more granular? I feel like a lot of these tests don't need it. I think think this tag slows down each individual spec. Not significantly, but enough that it could start to add up.

PR #1335, opened 2018-01-18
[#137632] add 'add accessories' + to problem order table

meara:
https://pm.tablexi.com/issues/137632


jhanggi:
I think the second commit doesn't belong here. I think we accidentally committed something from the cross-facility report branch here.

PR #1336, opened 2018-01-18
[#139681] link to cross facility reports, fix export raw in cross facility

meara:
https://pm.tablexi.com/issues/139681

### DIFF
--- a/app/views/facilities/list.html.haml
+++ b/app/views/facilities/list.html.haml
@@ -11,6 +11,10 @@
         %th= Facility.human_attribute_name(:name).titlecase
         %th= Facility.human_attribute_name(:status).titlecase
     %tbody
+      -if current_user.administrator?
+        %tr
+          %td= link_to Facility.cross_facility.to_s, facility_general_reports_path(Facility.cross_facility, report_by: :product)
+          %td= Facility.cross_facility.status_string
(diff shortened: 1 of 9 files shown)


jhanggi:
Let's add a couple tests around this. There are also circle failures and code climate changes to be made.

meara:
@jhanggi updated

app/views/facilities/list.html.haml
```diff
@@ -11,6 +11,10 @@
         %th= Facility.human_attribute_name(:name).titlecase
         %th= Facility.human_attribute_name(:status).titlecase
     %tbody
+      -if current_user.administrator?
```
jhanggi: Should we feature flag this for now? Hopefully, eventually we can open it up at some point, but I'm not sure I want to give this to NU/UIC just yet without making sure it doesn't kill the servers.

meara: What are we worrying about that would kill the servers? Happy to feature flag it, but not sure I understand exactly what the concern is.

jhanggi: The products/accounts/etc tabs could eat up a huge chunk of memory and or time. I'd like to see what sort of timings we get from this once we're actually on a production machine. As long as we can still access the page via url hacking.

PR #1338, opened 2018-01-19
[#139955] fix price group params bug

meara:
https://pm.tablexi.com/issues/139955

Steps to reproduce:

Log in as an account administrator
Click "Users" tab. URL should be `/facilities/all/users`
Click on a user's name
Click "Edit" button
Click "Update"
This happens because we are expecting a value for `params[:internal]`, which global admins would see as a dropdown for allowing them to manually override if the user should get internal or external pricing. Account managers don't have that dropdown.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -169,7 +169,7 @@ class UsersController < ApplicationController
   end
 
   def price_group_params
-    params.require(:user).permit(:internal) if current_user.administrator?
+    current_user.administrator? ? params.require(:user).permit(:internal) : {}
   end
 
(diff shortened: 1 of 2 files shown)


jhanggi:
Should we have a test for this? Might be ok without it. Can you update the description to include the steps to reproduce and other info. For anyone looking who doesn't have access to the redmine ticket.

meara:
Yeah Gilad and I were debating whether it was worth adding a test. Seems kind of like it would belong in a feature spec, but a feature spec just for that seems a bit overkill, setup/runtime wise. We can add one if you feel strongly though.

PR #1340, opened 2018-01-19
[#102967] feature gate global price group management

meara:
https://pm.tablexi.com/issues/102967

### DIFF
--- a/app/controllers/concerns/price_group_members_controller.rb
+++ b/app/controllers/concerns/price_group_members_controller.rb
@@ -34,7 +34,7 @@ module PriceGroupMembersController
   end
 
   def destroy
-    raise "Permission Denied" unless @price_group.can_manage_price_group_members?
+    raise NUCore::PermissionDenied unless @price_group.can_manage_price_group_members?
 
     if destroy_price_group_member!
       set_flash(:notice, :destroy)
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/price_group_members_controller.rb
```diff
@@ -23,6 +23,8 @@ def initialize
   end
 
   def create
+    raise "Permission Denied" unless @price_group.can_manage_price_group_members?
```
meara: What is a more sensible thing to raise than this? Or should we redirect back with an error message?

jhanggi: I would say raise `NUCore::PermissionDenied`. That'll render a 403. They shouldn't be able to get here otherwise.

app/models/price_group.rb
```diff
@@ -40,6 +40,10 @@ def global?
     facility.nil?
   end
 
+  def can_manage_price_group_members?
+    is_not_global || SettingsHelper.feature_on?(:can_manage_global_price_groups)
```
jhanggi: can we use `is_not_global?` with the question mark?

meara: Not unless we change the method name; it'll hit an undefined method error.

I was assuming there was some reason relating to the `before_destroy` hook above that `is_not_global` doesn't have a question mark already--does `?` play oddly with symbols?

PR #1337, opened 2018-01-19
[#139681] optimize export raw

meara:
https://pm.tablexi.com/issues/139681

PR #1341, opened 2018-01-20
Optimize Export Raw performance

jhanggi:
Extends from #1336 and replaces #1337.

Clean diff: https://github.com/tablexi/nucore-open/compare/139681-cross-facility-export-raw...export_raw_optimize_jh?expand=1

I mostly tested this on NU/Oracle since I have a large and realistic dataset locally.

With a 5500 row export (about 2 months of data for IMSERC), it originally took 450-500 seconds. Now it takes ~200 seconds. That same report, the memory in delayed_job would spike to around 250-300mb (this is locally). In this new version, it stays around 150mb.

I'm still experimenting with the batch size. In oracle, at least, we get significantly faster queries, but there are of course more of them.

One of the things I did was to enable the query cache, which is not on by default in delayed_job. This will have minimal effect on open, but for NU which has PMU, and likely DC with user affiliations it should save some trips to the database.

### DIFF
--- a/app/models/concerns/reports/csv_exporter.rb
+++ b/app/models/concerns/reports/csv_exporter.rb
@@ -65,7 +65,10 @@ module Reports
 
     def csv_body
       CSV.generate do |csv|
-        report_data.each { |row| csv << format_row(row) }
+        # When used in delayed_job, this will make sure the query cache is on
+        ActiveRecord::Base.cache do
+          report_data.each { |row| csv << format_row(row) }
+        end
       end
     end
 
--- a/app/services/reports/querier.rb
+++ b/app/services/reports/querier.rb
@@ -2,17 +2,22 @@ module Reports
 
   class Querier
 
+    # This seems to be ideal for NU/Oracle, but we can consider tweaking for MySQL
+    BATCH_SIZE = 500
+
     attr_reader :order_status_id, :current_facility, :date_range_field,
-                :date_range_start, :date_range_end, :extra_includes
+                :date_range_start, :date_range_end, :extra_includes,
+                :extra_joins, :extra_preloads, :batch_size, :options
 
     def initialize(options = {})
+      @options = options
       @order_status_id = options[:order_status_id]
       @current_facility = options[:current_facility]
       @date_range_field = options[:date_range_field]
       @date_range_start = options[:date_range_start]
       @date_range_end = options[:date_range_end]
-      @extra_includes = options[:includes]
       @transformer_options = options[:transformer_options]
+      @batch_size = options[:batch_size] || BATCH_SIZE
     end
 
     def perform
@@ -20,20 +25,30 @@ module Reports
     end
 
     def order_details
-      return [] if order_status_id.blank?
-
-      includes = default_includes
-      includes.concat(extra_includes) if extra_includes.present?
+      return OrderDetail.none if order_status_id.blank?
 
       OrderDetail.where(order_status_id: order_status_id)
                  .for_facility(current_facility)
                  .action_in_date_range(date_range_field, date_range_start, date_range_end)
+                 .joins(*joins)
                  .includes(*includes)
+                 .preload(*preloads)
                  .merge(Order.purchased)
+                 .find_each(batch_size: batch_size)
     end
 
-    def default_includes
-      [:account, :order, :order_status, :price_policy, :product]
+    private
+
+    def joins
+      [:order, :account] + Array(options[:joins])
+    end
+
+    def includes
+      [:order, :price_policy] + Array(options[:includes])
+    end
+
+    def preloads
+      [:product, :order_status, :account] + Array(options[:preloads])
     end
 
   end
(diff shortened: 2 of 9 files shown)


app/models/concerns/reports/csv_exporter.rb
```diff
@@ -65,7 +65,10 @@ def csv_header
 
     def csv_body
       CSV.generate do |csv|
-        report_data.each { |row| csv << format_row(row) }
+        # When used in delayed_job, this will make sure the query cache is on
```
meara: 👍

app/services/reports/querier.rb
```diff
@@ -2,38 +2,53 @@ module Reports
 
   class Querier
 
+    # This seems to be ideal for NU/Oracle, but we can consider tweaking for MySQL
```
meara: Have we tested the batching against MySQL datasets? Just want to make sure this doesn't slow things down for them.

app/services/reports/querier.rb
```diff
@@ -2,38 +2,53 @@ module Reports
...
                  .includes(*includes)
+                 .preload(*preloads)
                  .merge(Order.purchased)
+                 .find_each(batch_size: batch_size)
+    end
+
+    private
+
+    def joins
+      [:order, :account] + Array(options[:joins])
```
meara: I don't object to us having `Array(options[:joins])` here, and similar below, but I'm wondering what the expected use case is.

vendor/engines/split_accounts/app/services/split_accounts/order_detail_list_transformer.rb
```diff
@@ -10,13 +10,36 @@ def initialize(order_details = [])
...
         else
-          results << order_detail
+          order_detail
+        end
+      end
+    end
+
+    private
+
+    # Wraps a lazy enumerator with another enumerator so that if the block returns
```
meara: Can you clarify this comment a bit? I'm a little confused, and also it looks like the lazy enumerator is on this outside, not the inside, of the wrapping.

I'm assuming this is good for speed reasons--does this save a query when you're trying to do the splits above?

app/services/reports/querier.rb
jhanggi: I have not yet. That is something I do want to test out before we release this.

app/services/reports/querier.rb
jhanggi: Wrapping with `Array()` allows `options[:joins]` to be `nil`, a single item, or an array. The big advantage is handling the `nil` case, which will be converted to an empty array so the `+` will still work.

meara: Oh, I meant when are we expecting to be passing in options[:joins] that is not nil, essentially?

vendor/engines/split_accounts/app/services/split_accounts/order_detail_list_transformer.rb
jhanggi: It wraps one lazy enumerator with another lazy enumerator. `find_each` works as a lazy enumerator, meaning if you loop `each` (or another enumerable method--most `Enumerable` methods use `each` under the hood) on it, it won't hit the database until you need it to. So if you have `OrderDetail.find_each(batch_size: 1000).first(500)`, it'll only do the one query.

How it affects the splits above is this: before, even with `find_each`, the `each_with_object([])` is what is considered "terminal" meaning that it will render out the entire array doing as many queries as necessary and loading them all into memory before actually doing the loop. It's similar to `to_a`. You can't call `(1..Float::Infinity).to_a` because it'll go on forever, but you can call `(1..Float::Infinity).first(100)`.

What this basically does is each time you loop over the `each`, if it sees the next item is as an array (as in the case of split charges), it'll get the first thing out of the array on this loop, and the next thing on the next loop. If it's just a single item, it'll return that thing.

Here's what it does to the whole process:

Before:
* Create the query
* Load every order detail into memory (not sure if this was queried in batches or not, but it definitely loaded everything into memory)
* Loop over and create split order details into a new array
* Loop over the new array and write to CSV

Now: 
* Create the query
* For each 1000 order details
  * For each order detail
    * If non-split, write the row to CSV
    * If split, generate the multiple fake order-details, loop over them and write them to CSV

So really, the major advantage with this is memory: we only ever have 1000 (plus split fake-order-details) in un-freeable memory at once. Also, it doesn't affect this very much because we're sending it via email, but it is significantly faster to writing the first byte to CSV.

meara: Okay, that's very helpful. Can you clarify this comment that they're both lazy enums, and also point out briefly that the purpose of this is to save memory? It's helpful for me to have why's in comments, not just what's.

app/services/reports/querier.rb
jhanggi: I'm not 100%. I had `reservation` being passed in there, but turns out that means the query would exclude non-reservation orders. I do think I like having the option there so the three methods are consistent.

meara:
@jhanggi looks like a spec broke

PR #1349, opened 2018-01-22
[#139851] Audit gemfile

meara:
https://pm.tablexi.com/issues/139851


meara:
Cool.  Will plan to open additional PRs to upgrade everything locked to an old version over the next week or so.

PR #1370, opened 2018-01-29
[#139183] display calendar time range header

meara:
Now looks like this:

screen shot 2018-01-29 at 5 37 50 pm
screen shot 2018-01-29 at 5 38 11 pm
screen shot 2018-01-29 at 5 38 26 pm

### DIFF
--- a/app/assets/stylesheets/calendar/fullcalendar.extensions.scss
+++ b/app/assets/stylesheets/calendar/fullcalendar.extensions.scss
@@ -46,12 +46,12 @@
 }
 
 #calendar .fc-header-left, #calendar .fc-header-right,  {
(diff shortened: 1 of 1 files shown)


jhanggi:
Does this get all the places where we use this calendar?
* Admin reservation index
* New reservation
* Public Schedule
* Anywhere else I'm missing?

meara:
New reservation apparently uses a different version of the calendar--you can't switch to monthly view, you only see weekly. (I checked that view and there's no difference between this branch and master.)

The change does show up on the admin reservation page as well as the public schedule, and based on the CSS if there's anywhere else this same calendar element is re-used it should show up there as well.

PR #1372, opened 2018-01-30
[#140077] product name search

meara:
Search on product name.

Searching on "example":

screen shot 2018-01-30 at 1 04 09 pm

Searching "scope" against Dartmouth data:
screen shot 2018-01-30 at 1 18 32 pm

### DIFF
--- a/app/views/global_search/_products.html.haml
+++ b/app/views/global_search/_products.html.haml
@@ -19,5 +19,5 @@
         %td.span6
           = truncate(strip_tags(product.description), length: 300)
 
-- if results.count >= 20
-  %p Additional results not shown. Please try a more specific search.
+- if results.count >= 1
+  %p= t('.too_many_results')
--- a/app/views/global_search/_products.html.haml
+++ b/app/views/global_search/_products.html.haml
@@ -3,8 +3,8 @@
 %table.table.table-striped.table-hover.js--responsive_table
   %thead
     %tr
-      %th= "#{Product.model_name.human} Name"
-      %th= "Product Type"
+      %th= Product.model_name.human
+      %th= Product.human_attribute_name("type")
       %th= Facility.model_name.human
       %th= Product.human_attribute_name("description")
   %tbody
--- a/app/views/global_search/_products.html.haml
+++ b/app/views/global_search/_products.html.haml
@@ -1,0 +1,20 @@
+%h3= Product.model_name.human(count: 2)
+
+%table.table.table-striped.table-hover.js--responsive_table
+  %thead
+    %tr
+      %th= "#{Product.model_name.human} Name"
+      %th= "Product Type"
+      %th= Facility.model_name.human
+      %th= Product.human_attribute_name("description")
+  %tbody
+    - results.each do |product|
+      %tr
+        %td.nowrap.span2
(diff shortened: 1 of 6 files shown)


app/views/global_search/_products.html.haml
```diff
@@ -0,0 +1,20 @@
...
+    %tr
+      %th= "#{Product.model_name.human} Name"
+      %th= "Product Type"
+      %th= Facility.model_name.human
+      %th= Product.human_attribute_name("description")
+  %tbody
+    - results.each do |product|
+      %tr
+        %td.nowrap.span2
+          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(product.facility, product)
```
jhanggi: We need to do some restriction to only show hidden products for members of the facility. And keep in mind, a user might be part of one facility but not another so we need to show the products from one, but not the other.

app/views/global_search/_products.html.haml
```diff
@@ -0,0 +1,20 @@
...
+        %td.span2
+          = product.model_name.human
+        %td.span2
+          = link_to product.facility, facility_path(product.facility)
+        %td.span6
+          = product.description
```
jhanggi: My general feeling is we should not display the entire thing. For facilities we have a short description. We could also use the rails `truncate` helper.

app/views/global_search/_products.html.haml
```diff
@@ -0,0 +1,20 @@
+%h3= Product.model_name.human(count: 2)
+
+%table.table.table-striped.table-hover.js--responsive_table
+  %thead
+    %tr
+      %th= "#{Product.model_name.human} Name"
+      %th= "Product Type"
```
jhanggi: I18n

app/views/global_search/_products.html.haml
```diff
@@ -0,0 +1,20 @@
+%h3= Product.model_name.human(count: 2)
+
+%table.table.table-striped.table-hover.js--responsive_table
+  %thead
+    %tr
+      %th= "#{Product.model_name.human} Name"
```
jhanggi: We could probably just say "Name" or "Product" and not worry about needing to I18n this further.

app/views/global_search/_products.html.haml
jhanggi: I also wonder if we should show two links for the people who can both purchase _and_ manage the product.

app/views/global_search/_products.html.haml
meara: I kinda like displaying the whole thing, honestly, but am willing to be convinced otherwise. It looks okay to me with Dartmouth data.

If we were to truncate, what length should we truncate at? 300 characters?

app/views/global_search/_products.html.haml
meara: I was thinking what was shown would be handled by the ability call in the base searcher:

https://github.com/tablexi/nucore-open/blob/master/app/services/global_search/base.rb#L19

Are there cases that wouldn't be handled by that?

jhanggi: You're probably right. I would definitely add a test for it, though.

PR #1373, opened 2018-01-30
[#140005] Prevent error on removing product/user access if already removed

jhanggi:
Steps to reproduce:

* Add a user to a restricted product
* Open a second tab of the access list
* Click "Remove" to remove the user on one tab
* Click "Remove" for the same user on the second tab

As part of this, I cleaned up the controller spec, so it's probably worth viewing the two commits separately.

### DIFF
--- a/app/models/product_user.rb
+++ b/app/models/product_user.rb
@@ -3,6 +3,7 @@ class ProductUser < ActiveRecord::Base
   belongs_to :user
   belongs_to :product
   belongs_to :product_access_group
+  belongs_to :approved_by_user, class_name: "User", foreign_key: "approved_by"
 
   validates_numericality_of :product_id, :user_id, :approved_by, only_integer: true
   validates_uniqueness_of :user_id, scope: :product_id, message: "is already approved"
--- a/app/models/product_user.rb
+++ b/app/models/product_user.rb
@@ -3,7 +3,7 @@ class ProductUser < ActiveRecord::Base
   belongs_to :user
   belongs_to :product
   belongs_to :product_access_group
-  belongs_to :approved_by_user, class_name: "User", foreign_key: "approved_by"
+  belongs_to :approved_by_user, class_name: "User", foreign_key: "approved_by", inverse_of: false
 
   validates_numericality_of :product_id, :user_id, :approved_by, only_integer: true
   validates_uniqueness_of :user_id, scope: :product_id, message: "is already approved"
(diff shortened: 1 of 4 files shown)


spec/controllers/product_users_controller_spec.rb
```diff
@@ -4,72 +4,85 @@
...
-    expect(@price_policy).to be_valid
-    @params = { facility_id: @authable.url_name, instrument_id: @instrument.url_name }
-
-    @rule = @instrument.schedule_rules.create(FactoryBot.attributes_for(:schedule_rule))
-    @level = FactoryBot.create(:product_access_group, product_id: @instrument.id)
-    @level2 = FactoryBot.create(:product_access_group, product_id: @instrument.id)
-  end
+  let(:facility) { create(:setup_facility) }
+  let(:instrument) { create(:setup_instrument, facility: facility, requires_approval: true) }
+  let(:guest) { create(:user) }
```
meara: I'm not entirely following why this user is named as "guest"--are they not a regular logged in user? Why not just "user"?

jhanggi: Yeah, I should clean that up. Before the changes it was `@guest`, which isn't a very good name, but that's one of the default users that gets created by `create_users`. 

I'll do that when I clean up the rubocop stuff.

spec/controllers/product_users_controller_spec.rb
```diff
@@ -4,72 +4,60 @@
...
 
   end
 
-  context "update_restrictions" do
-    before :each do
-      @guest_product = ProductUser.create(product: @instrument, user: @guest, approved_by: @admin.id, approved_at: Time.zone.now)
-      @staff_product = ProductUser.create(product: @instrument, user: @staff, approved_by: @admin.id, approved_at: Time.zone.now)
+  describe "update_restrictions" do
+    let(:level) { create(:product_access_group, product: instrument) }
+    let(:level2) { create(:product_access_group, product: instrument) }
```
giladshanan: What about :level1 and :level2 ?  Or :guest_level and :staff_level?

app/models/product_user.rb
```diff
@@ -3,7 +3,7 @@ class ProductUser < ActiveRecord::Base
   belongs_to :user
   belongs_to :product
   belongs_to :product_access_group
-  belongs_to :approved_by_user, class_name: "User", foreign_key: "approved_by"
+  belongs_to :approved_by_user, class_name: "User", foreign_key: "approved_by", inverse_of: false
```
meara: What's this `inverse_of: false` for?

jhanggi: It means there is no inverse_of. Mainly, it's a way to appease rubocop.

PR #1374, opened 2018-01-30
[#139481] Add email address to training request notification

jhanggi:
* Change to use GlobalID
* Share markdown between text and HTML emails
* Add a way to get a random item from the database

### DIFF
--- a/spec/mailers/previews/training_request_mailer_preview.rb
+++ b/spec/mailers/previews/training_request_mailer_preview.rb
@@ -1,0 +1,9 @@
+class TrainingRequestMailerPreview < ActionMailer::Preview
+
+  def notify_facility_staff
+    product = NUCore::Database.random(Product.requiring_approval.where.not(training_request_contacts: [nil, ""]))
+    user = NUCore::Database.random(User.all)
+    TrainingRequestMailer.notify_facility_staff(user, product)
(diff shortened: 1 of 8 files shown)


spec/mailers/previews/training_request_mailer_preview.rb
```diff
@@ -0,0 +1,11 @@
+class TrainingRequestMailerPreview < ActionMailer::Preview
+
+  def notify_facility_staff
+    # Is not null and not blank. Oracle does not support `where.not(traning_request_contacts: "")` for CLOBS.
+    products_with_contacts = Product.requiring_approval.where("LENGTH(training_request_contacts) > 0")
+    product = NUCore::Database.random(products_with_contacts)
+    user = NUCore::Database.random(User.all)
```
meara: Fancy.

jhanggi: I've done this on other projects. In some projects we've even gone as far as to put a `sample`/`random` scope on ActiveRecord::Base so you could call `User.random`, but since we're just using this in mailer previews it didn't seem worth putting it that globally.

giladshanan: 👏

PR #1378, opened 2018-02-01
[#140234] cancel does not require pricing note

meara:
Steps to reproduce:
- Start with a complete order
- Go to the order detail modal
- Choose cancel
- We get the "Pricing Note is required" message

Make sure to handle both a normal cancelation where the actual costs are set to $0 and instrument cancelation with a fee (which technically keeps the order in "complete" state)

### DIFF
--- a/spec/features/admin/manage_order_detail_spec.rb
+++ b/spec/features/admin/manage_order_detail_spec.rb
@@ -51,7 +51,8 @@ RSpec.describe "Managing an order detail" do
     context "canceling a reservation" do
       before do
         instrument.price_policies.update_all(cancellation_cost: 5)
-        instrument.update_attribute(:min_cancel_hours, 1)
+        # reservation is set for tomorrow, we need min_cancel_hours to be longer than time to reserve_start_at
+        instrument.update_attribute(:min_cancel_hours, 48)
       end
 
       it "cancels without a fee" do
@@ -67,9 +68,7 @@ RSpec.describe "Managing an order detail" do
         check "Add reservation cost"
         click_button "Save"
 
-        # save_and_open_page
-
-        expect(page).to have_content("Canceled")
+        expect(page).to have_content("Complete")
         expect(page).to have_css('tfoot .currency', text: "$5.00", count: 2)
       end
     end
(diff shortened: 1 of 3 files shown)


spec/features/admin/manage_order_detail_spec.rb
```diff
@@ -32,4 +32,46 @@
...
+
+      it "cancels with a fee" do
+        select "Canceled", from: "Status"
+        check "Add reservation cost"
+        click_button "Save"
+
+        # save_and_open_page
+
+        expect(page).to have_content("Canceled")
+        expect(page).to have_css('tfoot .currency', text: "$5.00", count: 2)
```
meara: @jhanggi I cannot for the life of me figure out what is wrong with the setup for this spec. Everything seems right as far as I can tell--I don't know where the fee is getting lost.

jhanggi: I looked pretty quickly. It's that the `assign_price_policy` within `cancellation_fee` is returning `nil`. I don't know why yet, though. One thing to try (even on master with the require feature turned off) is canceling a New/future reservation with the checkbox checked, both in and outside the cancelation window. There may be a problem there as well.

jhanggi: Ah, I figured it out. The UI is confusing in how in works in this case. There are a few situations where even if you check this box, the fee will never get applied:

* There is no `min_cancel_hours ` set on the product, even if there is a reservation cost on the price policy
* There is no reservation cost on the price policy
* The reservation is still outside the cancelation period (e.g. cancelation period of 12 hours, the reservation starts 24 hours from now). 

That last one is what's happening here. The default reservation factory is for "tomorrow at 10am". Since we're setting the `min_cancel_hours` to 1, we're still outside that window. If you change the value to 48 hours, or change the reservation_start_at, this test will pass. Or, if you make a complete reservation in the past (since the past is inherently out of the cancelation window). I'd be fine with using just about any of these scenarios. It probably would be worth explicitly setting a `reserve_start_at`/`_end_at` in the test to make the test clearer.

I don't think these are actual problems or bad logic, though the interface is confusing. If the user themself were to cancel the reservation, they would not be charged the fee, so I think it's reasonable that the admin wouldn't charge that fee either. We could probably hide the checkbox in these scenarios or show a warning next to the box, but I'm not sure it's worth the effort. We could check in with NU and see if they've had any reported confusion and if so, then we could schedule something in.

meara: Ahhhhhh okay. I added a comment to clarify how `min_cancel_hours` operates.

PR #1379, opened 2018-02-02
[#140078] description search

meara:
As a user, I want to search on the description of an instrument/product, so that it's easy to find what I'm looking for.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,6 +15,7 @@ group :oracle do
   gem "ruby-oci8"
   gem "activerecord-oracle_enhanced-adapter"
 end
+gem "where-or"
 
 ## auth
 gem "cancancan"
--- a/db/migrate/20180202191333_add_full_text_index_to_products.rb
+++ b/db/migrate/20180202191333_add_full_text_index_to_products.rb
@@ -1,0 +1,5 @@
+class AddFullTextIndexToProducts < ActiveRecord::Migration
+  def change
+    add_index :products, [:name, :description], type: :fulltext
+  end
+end
--- a/db/migrate/20180202191333_add_full_text_index_to_products.rb
+++ b/db/migrate/20180202191333_add_full_text_index_to_products.rb
@@ -1,5 +1,0 @@
-class AddFullTextIndexToProducts < ActiveRecord::Migration
-  def change
-    add_index :products, [:name, :description], type: :fulltext
-  end
-end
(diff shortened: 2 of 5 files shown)


meara:
@jhanggi Looks like there's some implementation tradeoffs here.

The first version I did (just using a `LIKE` search) should be compatible with Oracle, but there's potentially performance concerns. Some searching examples of how it works:
- Search for: "scope"
  - Matches: "microscope"
- Search for: "electron microscope"
  - Matches: "electron microscope",
  - Does not match: "electron" OR "microscope"

The second version (using MySQL's full-text search) is potentially more performant, but is obviously not compatible with oracle, and has some different search behavior. Examples:
- Search for: "scope"
  - Does not match: "microscope"
- Search for: "electron microscope"
  - Matches: "electron microscope" OR "electron" OR "microscope"

I'm kind of leaning towards the `LIKE` search version (first commit), honestly, because I think that is more intuitive for users. I'm not sure how big of a performance hit it's going to end up being, however.  If you have time, can you try clocking some searches against your NU snapshot data?

meara:
Clocking on Dartmouth snapshot data:

| Search term | `MATCH` (ms) | `LIKE` (ms) |
| ------------- |---------------| -------------|
| dna | 366 | 365 |
| rna | 378 |  432 |
| scope | 380 | 555 |
| cell | 453 | 495 |
| scanning electron microscope | 490 | 632 |
| fl | 369 | 402 |
| op | 383 | 505 |

Keep in mind that these searches are returning slightly different things (as illustrated in the previous comment), so we're not exactly comparing apples to apples.

db/migrate/20180202191333_add_full_text_index_to_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddFullTextIndexToProducts < ActiveRecord::Migration
+  def change
+    add_index :products, [:name, :description], type: :fulltext
```
jhanggi: This index type is not valid for oracle. If we use the plain like search, does this make a difference on mysql?

Gemfile
```diff
@@ -15,6 +15,7 @@ group :oracle do
   gem "ruby-oci8"
   gem "activerecord-oracle_enhanced-adapter"
 end
+gem "where-or"
```
jhanggi: Should we look into where we can use this other places? (can definitely be separate PRs, and we should figure out how much effort we want to spend on it)

meara: Sure, I'll add a ticket to Rails 5 upgrade land.

PR #1384, opened 2018-02-05
[#137101] rm protected attributes

meara:


app/controllers/orders_controller.rb
```diff
@@ -336,7 +336,7 @@ def load_statuses
   end
 
   def order_params
-    return if params[:order].blank?
+    return {} if params[:order].blank?
```
jhanggi: Should this be `ActionController::Parameters.new` or something rather than a hash for consistency.

PR #1388, opened 2018-02-07
Fix transaction search N+1s

jhanggi:
Extends from #1387. 

Gets rid of some N+1s in transaction search by adding a new array of optimizers. This also allows us to prevent the n+1 on secure room occupancies.


meara:
I think this needs a rebase?

jhanggi:
This has been rebased.

PR #1392, opened 2018-02-07
Add a before_login_form view hook

jhanggi:
This will be used to add a button for linking out to single sign on


meara:
Is this PR still relevant given https://github.com/tablexi/nucore-open/pull/1395?

jhanggi:
This is included in #1395. I think we can just live with it there.

PR #1396, opened 2018-02-08
[#140237] Strong params part 1

meara:


app/controllers/affiliates_controller.rb
```diff
@@ -5,14 +5,12 @@ def new
   end
 
   def create
-    attrs = params[:affiliate]
-
-    if attrs.blank?
+    if affiliate_params.blank?
```
jhanggi: Do we think this is a legitimate code path? Could we get away with relying on the fact the `Affiliate` will be invalid if we're missing the params?

app/controllers/affiliates_controller.rb
```diff
@@ -38,14 +36,12 @@ def update
     rescue ActiveRecord::RecordNotFound
       flash[:error] = "Affiliate with id #{id} not found!"
     else
-      attrs = params[:affiliate]
-
-      if attrs.blank?
+      if affiliate_params.blank?
```
jhanggi: Same here

app/controllers/file_uploads_controller.rb
```diff
@@ -37,7 +37,7 @@ def download
 
   # POST /facilities/1/services/3/files
   def create
-    @file = @product.stored_files.new(params[:stored_file].merge(created_by: session_user.id))
+    @file = @product.stored_files.new(create_params.merge(created_by: session_user.id))
```
jhanggi: Do we have a spec around this? I'm only about 50% confident that `merge` will keep `created_by` allowed.

app/controllers/product_access_groups_controller.rb
```diff
@@ -50,4 +50,14 @@ def destroy
     end
   end
 
+  private
+
+  def create_params
```
jhanggi: Since create/update are identical, should they be just one method?

meara: cancancan looks for the methods to match the controller action names, so I think we need both.

app/controllers/file_uploads_controller.rb
meara: Added a spec. Looks like the merge is fine.

PR #1398, opened 2018-02-09
Strong params part 2

meara:
https://pm.tablexi.com/issues/140237

### DIFF
--- a/app/controllers/account_users_controller.rb
+++ b/app/controllers/account_users_controller.rb
@@ -25,9 +25,9 @@ class AccountUsersController < ApplicationController
   # POST /accounts/:account_id/account_users
   def create
     ## TODO add security
-    @user                    = User.find(params[:user_id])
-    @account_user            = @account.account_users.new(create_params)
-    @account_user.user       = @user
+    @user = User.find(params[:user_id])
+    @account_user = @account.account_users.new(create_params)
+    @account_user.user = @user
     @account_user.created_by = session_user.id
 
     if @account_user.save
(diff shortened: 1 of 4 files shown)


app/controllers/account_users_controller.rb
```diff
@@ -26,7 +26,7 @@ def new
   def create
     ## TODO add security
     @user                    = User.find(params[:user_id])
-    @account_user            = @account.account_users.new(params[:account_user])
+    @account_user            = @account.account_users.new(create_params)
```
jhanggi: Can we rubocop this and get rid of the extra spaces?

app/controllers/bundle_products_controller.rb
```diff
@@ -24,7 +24,7 @@ def index
 
   # POST /facilities/:facility_id/bundles/:bundle_id/bundle_products
   def create
-    @bundle_product = @bundle.bundle_products.new(params[:bundle_product])
+    @bundle_product = @bundle.bundle_products.new(create_params)
```
jhanggi: Does this use `load_and_authorize_resource` if so, can we get rid of this line?

app/controllers/facility_facility_accounts_controller.rb
```diff
@@ -26,7 +26,7 @@ def new
 
   # POST /facilities/:facility_id/facility_accounts(.:format)
   def create
-    @facility_account = current_facility.facility_accounts.new(params[:facility_account])
+    @facility_account = current_facility.facility_accounts.new(facility_facility_account_params)
```
jhanggi: Can we use `load_and_authorize_resource through: :current_facility` and then get rid of this line and the `find` in update?

meara: I can't get that to work--I think it's confused about the class name, despite it being passed explicitly in the `load_and_authorize_resource`. It looks for `facility_facility_account` on current_facility.

PR #1407, opened 2018-02-15
[#137156] Redirect to In Review after disputing

jhanggi:
After a BA or account owner goes in to the Transactions in Review page
and they dispute an order, they should be returned to Transactions in
Review rather than the My Orders/My Reservations page.

Also I18n the controller flashes

### DIFF
--- a/app/controllers/order_details_controller.rb
+++ b/app/controllers/order_details_controller.rb
@@ -21,10 +21,10 @@ class OrderDetailsController < ApplicationController
   # Put /orders/:order_id/order_details/:id
   def update
     if @order_detail.update_attributes(order_detail_params)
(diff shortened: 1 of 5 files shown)


app/controllers/order_details_controller.rb
```diff
@@ -35,9 +35,9 @@ def cancel
     if @order_detail.reservation && @order_detail.reservation.can_cancel?
       @order_detail.transaction do
         if @order_detail.cancel_reservation(session_user)
-          flash[:notice] = "The reservation has been canceled successfully." # TODO: I18n
+          flash[:notice] = text("cancel.success") # TODO: I18n
```
meara: Should this `TODO` still be there?

PR #1409, opened 2018-02-15
[#87522] add filtering orders and reservations tabs

meara:
https://pm.tablexi.com/issues/87522


app/services/transaction_search/date_range_searcher.rb
```diff
@@ -8,7 +8,11 @@ class DateRangeSearcher < BaseSearcher
     attr_reader :date_range_field
 
     def self.options
-      FIELDS.map { |field| [I18n.t(field, scope: "admin.transaction_search.date_range_fields"), field] }
+      if @billing_context
```
meara: I'm guessing there's a better way to do this than passing the `billing_context` flag around, but I'm not sure what it would be. Also not crazy about the name of that flag.

app/services/transaction_search/date_range_searcher.rb
```diff
@@ -25,9 +29,14 @@ def search(params)
       end_date = to_date(params[:end])
       @date_range_field = extract_date_range_field(params)
 
-      order_details
-        .action_in_date_range(@date_range_field, start_date, end_date)
-        .ordered_by_action_date(@date_range_field)
+      if start_date.present? || end_date.present?
+        order_details
+          .action_in_date_range(@date_range_field, start_date, end_date)
+          .ordered_by_action_date(@date_range_field)
```
jhanggi: We shouldn't have the conditional around this. We _always_ want to search by the date range field. and `action_in_date_range` is smart enough to deal with `nil`s. What are you trying to accomplish with the conditional?

app/services/transaction_search/ordered_for_searcher.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    def label_method
+      :full_name
+    end
+
+    def label
+      Order.human_attribute_name(:user)
+    end
+
+
```
jhanggi: Probably some rubocop to be done to this class.

app/services/transaction_search/date_range_searcher.rb
jhanggi: I'm not a huge fan of this option. I get that "fulfilled_at" doesn't really make sense in the New/In Process context, but that doesn't really bother me that much if the options are there. And I think problem orders have a fulfilled_at, so "billing context" is not necessarily the right name for the option.

If we really wanted to make this nicer, I would see if we can find a way to pass those options more explicitly. Something more like `allowed_date_fields: [:ordered_at]`. I'm not quite sure how that would work, though. Could do a `select` from that option in the view. `TransactionSearch::DateRangeSearcher.options.select { |label, value| value.in?(@search_form.allowed_date_fields) }`

These options are just me throwing out ideas off the cuff; not sure if they're great ones or not.

app/views/facility_orders/index.html.haml
```diff
@@ -18,7 +23,7 @@
           %th.order_num_width.centered= sortable "order_number", "Order"
           %th.order_num_width.centered= "Order Detail"
           %th.ordered_on_width= sortable "date", "Ordered On"
-          %th User Name
+          %th= Order.human_attribute_name(:user)
           %th Quantity
```
jhanggi: Can we I18n some of these other columns while we're here?

jhanggi:
We also should have a few specs around this.

app/services/transaction_search/date_range_searcher.rb
meara: Ah okay we can take this out actually. We were seeing no results again when the date filter was added, but I think it was because the field was set to "Fulfilled". Now that it defaults to "Ordered" it's returning results again.

app/services/transaction_search/date_range_searcher.rb
meara: Okay. I mean, if you're not bothered by the extra options being there, I guess it's not a huge deal for me either as long as NU doesn't mind.

jhanggi: Looks like we had some of the same thoughts. If we just don't have this option, what's the worst case? That the new/inprocess has a couple options that will result in 0 results. I'm kind of ok due to what the effort would likely be to get rid of them.

jhanggi:
Did you test that the page renders with zero orders in the scope? Just want to make sure that doesn't throw off the search form giving us one of those weird sql errors.

meara:
It's fine with zero orders as long as the join with orders is there. The view does look a little janky though, which is something on my to-do list to fix.

jhanggi:
Don't forget to hit secure_rooms/occupancies tab with this as well

app/models/order_detail.rb
```diff
@@ -306,6 +306,7 @@ def can_be_viewed_by?(user)
   scope :for_accounts, ->(accounts) { where("order_details.account_id in (?)", accounts) unless accounts.nil? || accounts.empty? }
   scope :for_facilities, ->(facilities) { joins(:order).where("orders.facility_id in (?)", facilities) unless facilities.nil? || facilities.empty? }
   scope :for_products, ->(products) { where("order_details.product_id in (?)", products) unless products.blank? }
+  scope :for_users, ->(user_ids) { joins(:order).where(orders: { user_id: user_ids }) unless user_ids.blank? }
```
jhanggi: Could be `if user_ids.present?`, but we probably should keep this for consistency.

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
...
-          %th= sortable "status"
-          %th.currency Price
+.span12
+  = render "shared/transactions/search2"
 
-      -## THE DETAIL ROWS
-      %tbody
-        - @order_details.each do |od|
+.span12
+  - if @order_details.empty?
```
jhanggi: I would lean towards flipping the conditional so it's `if @order_details.any?` and the positive case is first.

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
...
+    %p.notice There are no "New" or "In Process" orders.
```
jhanggi: Can we I18n this?

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
...
+  - else
+    = form_tag(batch_update_facility_orders_path, method: :post) do
+      - facility_users = User.find_users_by_facility(current_facility)
```
jhanggi: Where is this used? If it is used (possibly for assignment), can we pull it up into the controller?

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
...
+      %table.table.table-striped.table-hover.old-table
+        = render partial: "table_controls"
+        %thead
+          -## THE HEADER ROW
```
jhanggi: This comment doesn't feel necessary since it's inside a `<thead>`

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
...
-            = render partial: "shared/order_detail_cell", locals: { od: od }
-            %td= od.assigned_user
-            %td= od.order_status
-            %td.currency
-              =  OrderDetailPresenter.new(od).wrapped_total
-        -## THE FOOTER CONTROL ROW
-      = render partial: "table_controls"
+            %th
+            %th.order_num_width.centered= sortable "order_number", "Order"
+            %th.order_num_width.centered= "Order Detail"
```
jhanggi: Can we I18n these other column headers?

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
...
+            %th.ordered_on_width= sortable "date", "Ordered On"
+            %th= Order.human_attribute_name(:user)
+            %th Quantity
+            %th.order-note= sortable "product"
+            %th= sortable "assigned_to"
+            %th= sortable "status"
+            %th.currency Price
+
+        -## THE DETAIL ROWS
```
jhanggi: Once again, this comment doesn't really add much.

app/views/facility_orders/show_problems.html.haml
```diff
@@ -1,13 +1,19 @@
...
-  %p.notice= t(".no_problem_orders")
-- else
-  %p.notice= t(".instructions")
-  = render partial: "shared/problem_order_details",
-    locals: { assign_price_policies_path: assign_price_policies_to_problem_orders_facility_orders_path, show_reservation_start_at: false }
+.span12
+  = render "shared/transactions/search2"
+
+.span12
+  - if @order_details.empty?
```
jhanggi: Can we flip the conditional?

app/views/facility_reservations/index.html.haml
```diff
@@ -1,59 +1,65 @@
...
-- if @order_details.empty?
-  %p.notice There are no "New" or "In Process" Reservations.
-- else
-  = form_tag(batch_update_facility_reservations_path, method: :post) do
+.span12
+  = render "shared/transactions/search2"
+
+.span12
+  - if @order_details.empty?
+    %p.notice There are no "New" or "In Process" Reservations.
```
jhanggi: I18n.

app/views/facility_reservations/index.html.haml
```diff
@@ -1,59 +1,65 @@
...
 
-- if @order_details.empty?
-  %p.notice There are no "New" or "In Process" Reservations.
-- else
-  = form_tag(batch_update_facility_reservations_path, method: :post) do
+.span12
+  = render "shared/transactions/search2"
+
+.span12
+  - if @order_details.empty?
```
jhanggi: Flip the conditional

app/views/facility_reservations/index.html.haml
```diff
@@ -1,59 +1,65 @@
...
+            %th
+            %th Order
+            %th Order Detail
+            %th= sortable "reserved_by"
+            %th{colspan: 2}= sortable "date", "Reserved On"
+            %th{colspan: 2}= sortable "reserve_range", "Reserved For"
+            %th.order-note= sortable "product_name", "Product"
+            %th= sortable "assigned_to", "Assigned To"
+            %th= sortable "status", "Status"
+            %th.currency Price
```
jhanggi: I18n

app/views/facility_reservations/index.html.haml
```diff
@@ -1,59 +1,65 @@
...
+                  = link_to res, edit_facility_order_order_detail_reservation_path(current_facility, order, od, res)
+                - else
+                  = link_to res, facility_order_order_detail_reservation_path(current_facility, order, od, res)
+                - if is_current
+                  %br
+                  %span.very_dense
+                    = link_to t("reservations.switch.start"), order_order_detail_reservation_switch_instrument_path(od.order, od, od.reservation, switch: "on") if od.reservation.can_switch_instrument_on?
+                    = link_to t("reservations.switch.end"), order_order_detail_reservation_switch_instrument_path(od.order, od, od.reservation, switch: "off"), class: end_reservation_class(od.reservation) if od.reservation.can_switch_instrument_off?
+              = render partial: "shared/order_detail_cell", locals: { od: od, show_reservation: false }
+              %td= od.assigned_user.nil? ? "" : od.assigned_user.full_name
```
jhanggi: Maybe do `od.assigned_user&.full_name`?

app/views/facility_reservations/show_problems.html.haml
```diff
@@ -1,14 +1,20 @@
...
-- else
-  %p.notice= t(".instructions")
-  = render partial: "shared/problem_order_details",
-    locals: { assign_price_policies_path: assign_price_policies_to_problem_orders_facility_reservations_path,
-    show_reservation_start_at: true }
+.span12
+  = render "shared/transactions/search2"
+
+.span12
+  - if @order_details.empty?
```
jhanggi: Flip conditional

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/index.html.haml
```diff
@@ -1,37 +1,43 @@
+= render 'shared/transactions/headers'
```
jhanggi: Use double quotes

vendor/engines/secure_rooms/app/views/secure_rooms/facility_occupancies/show_problems.html.haml
```diff
@@ -1,12 +1,18 @@
+= render 'shared/transactions/headers'
```
jhanggi: Double quotes

app/views/facility_orders/index.html.haml
```diff
@@ -1,47 +1,53 @@
+= render 'shared/transactions/headers'
```
jhanggi: Double quotes

app/views/facility_orders/show_problems.html.haml
```diff
@@ -1,13 +1,19 @@
+= render 'shared/transactions/headers'
```
jhanggi: Double quotes

app/services/transaction_search/ordered_for_searcher.rb
```diff
@@ -0,0 +1,25 @@
+module TransactionSearch
+
+  class OrderedForSearcher < BaseSearcher
+
+    def options
+      User.select("users.id, users.first_name, users.last_name")
```
jhanggi: Just curious, does this still work if you used symbols? `User.select(:id, :first_name, :last_name)`. I feel like that'll be a little safer (assuming it works).

PR #1412, opened 2018-02-16
Support transaction search with strong params enabled

jhanggi:
With strong params enabled, `to_h` will return a hash with only
permitted parameters included. We were using `to_h` to support `nil`,
i.e. `nil.to_h => {}`.

```
Hash({}) == {}
Hash(nil) == {}
Hash(normal_hash) == normal_hash
Hash(strong_params) == strong_params
```


meara:
This did break specs when you had permit_all_params turned off, right?

jhanggi:
Definitely `features/admin/all_transactions_search_spec.rb` was broken if `permit_all_params = false` . This fixes it. 

I think that's the only one where we go full controller -> search so far (other than the other PRs I've been waiting to merge #1399, #1401, #1402)

PR #1414, opened 2018-02-19
[#121458] Fix admin ability to edit past reservations in cart

jhanggi:
To reproduce old issue:
* On a facility with an order form
* Order on behalf of another user
* Add "2" of an instrument to the cart
* "Create Reservation" on the other to the future. Click on the time to get to the edit page and change the start time. This one should have been working fine.
* "Create Reservation" on one of them to the past. 
* Click on the time. This is either not editable (if you're a director/staff), or if you're a global admin changing the start time would not persist.

This also adds authorizations for the reservation controller. Previously, you could copy a link, sign in as another user and still access the pages.

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -263,7 +263,9 @@ class ReservationsController < ApplicationController
       :actual_start_meridian,
     )
 
-    reservation_params.merge!(reservation_start_as_params) if fixed_start_time?
+    # If the start time is locked (only after purchase, don't allow the params
+    # to override the existing values)
+    reservation_params.merge!(reservation_start_as_params) if fixed_start_time? && !@reservation.in_cart?
 
     reservation_params
   end
@@ -333,7 +335,7 @@ class ReservationsController < ApplicationController
   end
 
   def editable_by_current_user?
-    if current_user.administrator? && @reservation.admin_editable?
+    if current_ability.can?(:manage, @reservation) && @reservation.admin_editable?
       true
     elsif @reservation.can_customer_edit?
       true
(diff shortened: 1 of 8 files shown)


spec/features/purchasing_a_reservation_on_behalf_of_spec.rb
```diff
@@ -62,6 +63,96 @@
...
+      click_link "Make a Reservation", match: :first
+      fill_in "Reserve Start", with: I18n.l(1.day.ago.to_date, format: :usa)
+      click_button "Create"
+
+      all(:link, "Make a Reservation").last.click
+      fill_in "Reserve Start", with: I18n.l(2.days.ago.to_date, format: :usa)
+      click_button "Create"
 
+      # this would normally be hidden and the default would be to complete backdated
+      # reservations. This might change in #121458
```
meara: I'm not following this comment. Doesn't this PR cover #121458?

app/controllers/reservations_controller.rb
```diff
@@ -256,7 +263,9 @@ def reservation_params
       :actual_start_meridian,
     )
 
-    reservation_params.merge!(reservation_start_as_params) if fixed_start_time?
+    # If the start time is locked (only after purchase, don't allow the params
```
meara: Can you clarify this comment? Should this be something like "Prevent override of start time params after purchase if start time is locked?"

Also, under what circumstances is the start time locked? If it's inside the lock window? Should we just say that?

jhanggi: It could be within the lock window, or after the reservation has started. Basically it's any of the times where you can extend the reservation, but not move the start time. But I can work on cleaning up the wording.

spec/features/purchasing_a_reservation_on_behalf_of_spec.rb
jhanggi: Oops. I meant 140462.

The "More Options" is hidden by css/js, which capybara is ignoring. But since it's visible, it'll get submitted. In that #140462 ticket, we're going to figure out what to do with past reservations but you want to select "New" as the order status.

The other option, which I'm not sure if it's worth the effort is to make these fields `disabled` by default so they don't get submitted, then enable them in the JS.

PR #1419, opened 2018-02-23
Strong params part 3

meara:


app/controllers/facility_reservations_controller.rb
```diff
@@ -43,42 +41,25 @@ def update
...
-        additional_notice = "  Order detail actual cost has been updated as well."
-      end
-    else
-      # We're updating a reservation before it's been completed
-      if reserve_changed?
-        @order_detail.assign_estimated_price
-        additional_notice = " Estimated cost has been updated as well."
-      end
+    # We're updating a reservation before it's been completed
+    if reserve_changed?
```
jhanggi: Where did we change this? I feel like I remember it and it's probably a good change, but it doesn't feel like it should be part of the strong params changes.

app/controllers/product_users_controller.rb
```diff
@@ -69,14 +69,15 @@ def destroy
...
-    product_param_name = @product.class.name.underscore
-    unless params[product_param_name]
+    permitted_params = params.require(@product.class.name.underscore).require(:product_users)
+
+    unless permitted_params
       redirect_to action: :index
       return
     end
-    params[product_param_name][:product_users].each do |key, value|
+    permitted_params.each do |key, value|
```
jhanggi: I remember there being some confusion around this. Would it be worth renaming `key` and `value` to whatever they actual are? `product_user_id`, I think?

app/controllers/products_common_controller.rb
```diff
@@ -157,12 +152,28 @@ def translation_scope
...
+                                                      :requires_approval, :training_request_contacts,
+                                                      :is_archived, :is_hidden, :order_notification_recipient,
+                                                      :user_notes_field_mode, :user_notes_label, :show_details,
+                                                      :schedule_id, :control_mechanism, :reserve_interval,
+                                                      :min_reserve_mins, :max_reserve_mins, :min_cancel_hours,
+                                                      :auto_cancel_mins, :lock_window, :cutoff_hours,
+                                                      relay_attributes: [:ip, :port, :username, :password, :type, :instrument_id])
+  end
+
+  def update_params
```
jhanggi: Did we figure out you could use something like `resource_params` and have cancan use it for both create and update? Not sure if it's worth changing that, though.

app/controllers/users_controller.rb
```diff
@@ -136,11 +136,11 @@ def create_params
   end
 
   def edit_user_params
-    @user.admin_editable? ? params.require(:user).permit(:email, :first_name, :last_name, :username) : empty_params
+    @user.admin_editable? ? params.require(:user).except(:internal).permit(:email, :first_name, :last_name, :username) : with_dropped_params { empty_params }
```
jhanggi: This seems like an odd construct. Why do we need to wrap `with_dropped_params { empty_params }` when the params should have nothing inside them?

app/services/price_policy_updater.rb
```diff
@@ -50,7 +50,7 @@ def price_group_attributes(price_group)
   end
 
   def initial_price_group_attributes(price_group)
-    @params["price_policy_#{price_group.id}"] || { can_purchase: false }
+    @params["price_policy_#{price_group.id}"]&.permit(:can_purchase, :usage_rate, :usage_subsidy, :minimum_cost, :cancellation_cost, :unit_cost, :unit_subsidy) || { can_purchase: false }
```
jhanggi: Can we make this line a little shorter?

app/controllers/facility_reservations_controller.rb
meara: It's this commit.

https://github.com/tablexi/nucore-open/pull/1419/commits/4d3f1c19cb1b9a5ad516b0c61e70ab9d0e692ad0

Do you feel strongly enough that it's worth separating out the tech debt cleanup part from the strong params change? Strong params is itself a bit of a tech debt cleanup task--I feel like I'm okay with this being here but can pull it out if it really bothers you.

jhanggi: I don't feel strongly enough to be worth the effort.

spec/services/accessorizer_spec.rb
```diff
@@ -94,7 +94,7 @@
 
     context "a quantity accessory" do
       let(:params) do
-        HashWithIndifferentAccess.new(quantity_accessory.id.to_s => { enabled: "true", quantity: "3" })
+        ActionController::Parameters.new(quantity_accessory.id.to_s => { enabled: "true", quantity: "3" })
```
jhanggi: Do we need the `to_s` on the ID?

jhanggi:
FYI, I marked the codeclimate warning on a complex method as wontfix since it's not really part of this unit of work.

spec/services/accessorizer_spec.rb
meara: Specs fail without it, so I'm thinking yes.

PR #1421, opened 2018-03-01
Upgrade ruby-saml gem for security warning.

rdunlop:
ref: https://github.com/onelogin/ruby-saml/commit/048a544730930f86e46804387a6b6fad50d8176f


jhanggi:
@rdunlop Thanks for this. While we're at it, can we upgrade https://github.com/apokalipto/devise_saml_authenticatable/releases

Ticket: https://pm.tablexi.com/issues/140609

PR #1422, opened 2018-03-01
[#140279] clean up queries

meara:
https://pm.tablexi.com/issues/140279

Cleaned up a few queries to use the `.or` operator or otherwise modernize their syntax. There weren't as many spots where it made sense to use the `or` operator as expected, in part because of the very specific parallel structure requirements--gets very messy with joins etc.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -231,7 +231,7 @@ class OrderDetail < ActiveRecord::Base
   end
 
   def self.unreconciled
-    where.not(state: ["reconciled"])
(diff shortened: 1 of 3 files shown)


app/models/order_detail.rb
```diff
@@ -231,15 +231,15 @@ def self.joins_relay
   end
 
   def self.unreconciled
-    where("order_details.state <> ?", "reconciled")
+    where.not(state: ["reconciled"])
```
jhanggi: I think this doesn't need the brackets

PR #1425, opened 2018-03-02
[#139338] cross payment source filtering

meara:
https://pm.tablexi.com/issues/139338

Adds a new view in a subnav under My Payment Sources for My Transactions, with filtering.


app/controllers/transactions_controller.rb
```diff
@@ -11,6 +11,33 @@ def initialize
     super
   end
 
+  def index
+    account_ids = current_user.account_users.map(&:account_id)
+    order_details = OrderDetail.where(account_id: account_ids).joins(:order)
```
jhanggi: Just curious, can we use `User#administered_order_details`? That's what the in dispute uses. Not sure if that means the same thing or not?

app/controllers/transactions_controller.rb
```diff
@@ -11,6 +11,33 @@ def initialize
...
+    )
+
+    @search = TransactionSearch::Searcher.new(TransactionSearch::FacilitySearcher,
+                                              TransactionSearch::AccountSearcher,
+                                              TransactionSearch::ProductSearcher,
+                                              TransactionSearch::DateRangeSearcher,
+                                              TransactionSearch::OrderStatusSearcher,
+                                              TransactionSearch::OrderedForSearcher).search(order_details, @search_form)
+    @date_range_field = @search_form.date_params[:field]
+    @order_details = @search.order_details.preload(:order_status)
```
jhanggi: I believe the preload is covered by `OrderStatusSearcher`.

app/presenters/nav_tab/link_collection.rb
```diff
@@ -38,10 +38,21 @@ def default
 
   private
 
+  def payment_sources
+    NavTab::Link.new(
+        tab: :payment_sources,
+        subnav: [:accounts, :transactions].map { |method_name| send(method_name) },
```
jhanggi: I think this would be clearer: `subnav: [accounts, transactions]`. I don't think we need the metaprogramming.

I might even be ok with going fully inline:

```ruby
tab: :payment_sources,
subnav:
  [
    NavTab::Link.new(tab: :accounts, text: t_my(Account), url: accounts_path),
    NavTab::Link.new(tab: :transactions, text: I18n.t("pages.transactions"), url: transactions_path),
  ]
```

app/views/transactions/index.html.haml
```diff
@@ -0,0 +1,7 @@
+= render :partial => "shared/transactions/headers"
```
jhanggi: Can we go with `render "shared/transactions/headers"`?

app/views/transactions/index.html.haml
```diff
@@ -0,0 +1,7 @@
...
+- content_for :h1 do
+  Transaction History
```
jhanggi: I18n

app/views/transactions/index.html.haml
```diff
@@ -0,0 +1,7 @@
...
+
+= render "shared/transactions/search2"
+
+= render :partial => 'shared/transactions/table', locals: { show_statements: true }
```
jhanggi: Can we go with `render "shared/transactions/table", show_statements: true`?

jhanggi:
And can you update the description to explain a little more what this feature is adding and what it's changing?

app/controllers/transactions_controller.rb
meara: That means a more specific thing--based on the way we're displaying account transactions currently, users should be able to see transactions on any account they have access to. `User#administered_order_details`, by contrast, would show transactions for only the accounts where the user has an admin user role (i.e. not just "purchaser").

I made the call towards preserving current behavior, but it is inconsistent with how things work with `in_dispute`--if this is a concern we could run it past the client and see what they think.

jhanggi: If it maintains current behavior, then let's stick with that. Thanks for double checking; I wasn't sure the difference.

config/locales/en.yml
```diff
@@ -69,6 +69,8 @@ en:
     logout: Logout
     manage: Manage %{model}
     my_tab: My %{model}
+    payment_sources: My Payment Sources
```
jhanggi: Can we clean up the use of this vs `t_my(Account)` in `LinkCollection#accounts`? Haven't tried it out, but they seem like they should be the same translation.

app/controllers/accounts_controller.rb
```diff
@@ -3,11 +3,11 @@ class AccountsController < ApplicationController
   customer_tab  :all
   before_action :authenticate_user!
   before_action :check_acting_as
-  before_action :init_account, only: [:show, :user_search, :transactions, :transactions_in_review, :suspend, :unsuspend]
+  before_action :init_account, only: [:show, :user_search, :transactions, :suspend, :unsuspend]
```
jhanggi: We don't need to do anything here, but the rubocop/codeclimate warning is interesting about explicitly defining actions in the controller when using the `before_action`. Just something to keep in mind going forward.

app/presenters/nav_tab/link_collection.rb
```diff
@@ -38,10 +38,21 @@ def default
 
   private
 
+  def payment_sources
```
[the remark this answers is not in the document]
meara: I don't like that naming convention only because none of the other navlinks methods use it. If we change this name to `my_payment_sources`, I'd want to change all the method names referenced in line 32 above.

app/views/shared/transactions/_table_inside.html.haml
```diff
@@ -62,5 +62,5 @@
           = order_detail.order_status
           = order_detail_status_badges(order_detail)
         - if local_assigns[:show_statements]
-          %td= link_to "Download", account_statement_path(@account, order_detail.statement_id, format: :pdf) if order_detail.statement
+          %td= link_to "Download", account_statement_path(order_detail.account, order_detail.statement_id, format: :pdf) if order_detail.statement
```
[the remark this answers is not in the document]
meara: Yep, it's included in the `account_searcher`.

PR #1428, opened 2018-03-05
[#140462] order on behalf of reservation statuses determined by dates

meara:
https://pm.tablexi.com/issues/140462

Instrument order details should always have order status determined by their dates.

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -162,6 +162,8 @@ class Order < ActiveRecord::Base
     order_status = order_status.is_a?(OrderStatus) ? order_status : OrderStatus.find(order_status)
 
     order_details.each do |od|
+      next if od.reservation # reservations should always have order_status dictated by their dates
+
       if order_status.root == OrderStatus.complete
         od.backdate_to_complete!(ordered_at)
       else
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -153,31 +153,6 @@ class Order < ActiveRecord::Base
     errors.empty?
   end
 
-  def can_backdate_order_details?
-    ordered_at <= Time.zone.now
-  end
-
-  def backdate_order_details!(update_by, order_status)
-    # can accept either an order status or an id
-    order_status = order_status.is_a?(OrderStatus) ? order_status : OrderStatus.find(order_status)
-
-    order_details.each do |od|
-      next if od.reservation # reservations should always have order_status dictated by their dates
-
-      if order_status.root == OrderStatus.complete
-        od.backdate_to_complete!(ordered_at)
-      else
-        od.update_order_status!(update_by, order_status, admin: true)
-      end
-    end
-  end
-
-  def complete_past_reservations!
-    order_details.select { |od| od.reservation && od.reservation.reserve_end_at < Time.zone.now }.each do |od|
-      od.backdate_to_complete! od.reservation.reserve_end_at
-    end
-  end
-
   def max_group_id
     order_details.maximum(:group_id).to_i + 1
   end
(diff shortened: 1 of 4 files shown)


app/models/order.rb
```diff
@@ -162,6 +162,8 @@ def backdate_order_details!(update_by, order_status)
     order_status = order_status.is_a?(OrderStatus) ? order_status : OrderStatus.find(order_status)
 
     order_details.each do |od|
+      next if od.reservation # reservations should always have order_status dictated by their dates
```
jhanggi: Since this is the main path everything goes through, is there a way to make all of the logic live within this block to make the path a little more readable? Somehow pull the `complete_past_reservations!` into this loop?

I'm probably missing something here, but something like this. Otherwise, we're kind of doing two loops one to (possibly) backdate reservations and then another to backdate orders based on the status chosen in the dropdown. I'm open to a discussion but I feel like one loop with three (well, technically four) outcomes is easier to grok than two loops where one has two outcomes.

```ruby
if od.reservation 
  od.backdate_to_complete! od.reservation.reserve_end_at if od.reservation.reserve_end_at < Time.current
elsif order_status.root == OrderStatus.complete
  od.backdate_to_complete!(ordered_at)
else
  od.update_order_status!(update_by, order_status, admin: true)
end
```

meara: We can't do this refactoring because we want `complete_past_reservations!` to run regardless of whether `acting_as? && order_status_id.present?` is true.

app/services/order_purchaser.rb
```diff
@@ -62,9 +62,8 @@ def order_in_the_past!
     # update order detail statuses if you've changed it while acting as
     if acting_as? && order_status_id.present?
       order.backdate_order_details!(user, order_status_id)
```
jhanggi: It seems like the only place `complete_past_reservations!` and `backdate_order_details!` are used is here. Just curious, what would that look like if we pulled those methods out into `OrderPurchaser`? Would that add any clarity?

app/services/order_purchaser.rb
```diff
@@ -55,15 +55,39 @@ def order_detail_updater
...
+      backdate_order_details!(user, order_status_id)
+    end
+    complete_past_reservations!
+  end
+
+  def can_backdate_order_details?
+    order.ordered_at <= Time.zone.now
+  end
+
+  def backdate_order_details!(update_by, order_status)
```
jhanggi: I think I do like these methods better here rather than on `Order`. Do you agree?

app/services/order_purchaser.rb
```diff
@@ -55,15 +55,39 @@ def order_detail_updater
...
+    # can accept either an order status or an id
+    order_status = order_status.is_a?(OrderStatus) ? order_status : OrderStatus.find(order_status)
```
jhanggi: Since this is only used in one scenario, can we get rid of this flexibility and just accept either an id or an `OrderStatus` object?

app/services/order_purchaser.rb
```diff
@@ -55,15 +55,39 @@ def order_detail_updater
   end
 
   def order_in_the_past!
-    unless order.can_backdate_order_details?
+    unless can_backdate_order_details?
```
jhanggi: Weird, my codeclimate plugin is telling me `Favor modifier `unless` usage when having a single-line body.`, but I don't see i in codeclimate itself.

app/services/order_purchaser.rb
meara: Yeah, I think it makes sense to keep this functionality all in one file, and it is nice to keep the `Order` model a bit skinnier.

app/services/order_purchaser.rb
meara: I feel like if we made that change the line length would be a bit unwieldy.

meara:
Testing-wise I think the change on orders_controller_spec covers it adequately.

PR #1429, opened 2018-03-06
[#137627] add links to 10 most recently purchased products to home page

giladshanan:
# Release Notes

Add a list of the user's 10 most recently purchased products to the user's home page. 

# Screenshots
screen shot 2018-03-16 at 1 58 11 pm

# Context

Allow users to access frequently used products faster.

# Refactorings

- Rename `sorted` scope on Facilitiy model to `alphabetized`
- Add `MostRecentlyUsedSearcher` service object for Products and Facilities
- Refactor `public_calendar` view helper to handle non-specific facility context

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -31,8 +31,8 @@ class FacilitiesController < ApplicationController
 
   # GET /facilities
   def index
-    @facilities = Facility.active.sorted
-    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.sorted
+    @facilities = Facility.active.alphabetized
+    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
     @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.alphabetized
     render layout: "application"
@@ -53,7 +53,7 @@ class FacilitiesController < ApplicationController
     # show list of operable facilities for current user, and admins manage all facilities
     @active_tab = "manage_facilites"
     if session_user.administrator?
-      @facilities = Facility.sorted
+      @facilities = Facility.alphabetized
       flash.now[:notice] = "No facilities have been added" if @facilities.empty?
     else
       @facilities = operable_facilities
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -32,9 +32,9 @@ class FacilitiesController < ApplicationController
   # GET /facilities
   def index
     @facilities = Facility.active.sorted
-    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities
+    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.sorted
     @active_tab = "home"
-    @products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products
+    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.alphabetized
     render layout: "application"
   end
 
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -34,8 +34,7 @@ class FacilitiesController < ApplicationController
     @facilities = Facility.active.alphabetized
     @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
-    # recent products get sorted alphabetically in the product_list partial
-    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.includes(:facility)
+    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.includes(:facility).alphabetized
     render layout: "application"
   end
 
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -39,7 +39,7 @@ class FacilitiesController < ApplicationController
   def index
     @facilities = Facility.active.sorted
     @active_tab = "home"
-    @products = Product.active.in_active_facility.recently_purchased
+    @products = Product.active.in_active_facility.recently_purchased(acting_user)
     render layout: "application"
   end
 
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -34,7 +34,7 @@ class FacilitiesController < ApplicationController
     @facilities = Facility.active.alphabetized
     @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
-    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.alphabetized
+    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products
     render layout: "application"
   end
 
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -35,7 +35,7 @@ class FacilitiesController < ApplicationController
     @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
     # recent products get sorted alphabetically in the product_list partial
-    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products
+    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.includes(:facility)
     render layout: "application"
   end
 
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -8,7 +8,6 @@ class FacilitiesController < ApplicationController
   before_action :check_acting_as, except: [:index, :show]
   before_action :load_order_details, only: [:confirm_transactions, :move_transactions, :reassign_chart_strings]
   before_action :set_admin_billing_tab, only: [:confirm_transactions, :disputed_orders, :movable_transactions, :transactions]
-  before_action :set_recently_used_facilities, only: [:index]
   before_action :store_fullpath_in_session, only: [:index, :show]
 
   load_and_authorize_resource find_by: :url_name
@@ -23,11 +22,6 @@ class FacilitiesController < ApplicationController
     action_name.in?(%w(disputed_orders movable_transactions transactions)) ? "two_column_head" : "two_column"
   }
 
-  def set_recently_used_facilities
-    @recently_used_facilities =
-      acting_user.present? && acting_user.recently_used_facilities.sorted
-  end
-
   cattr_accessor(:facility_homepage_redirector) { DefaultFacilityHomepageRedirector }
 
   # GET /facilities/:facility_url/dashboard
@@ -38,8 +32,9 @@ class FacilitiesController < ApplicationController
   # GET /facilities
   def index
     @facilities = Facility.active.sorted
+    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities
     @active_tab = "home"
-    @products = Product.active.in_active_facility.recently_purchased(acting_user)
+    @products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products
     render layout: "application"
   end
 
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -34,6 +34,7 @@ class FacilitiesController < ApplicationController
     @facilities = Facility.active.alphabetized
     @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
+    # recent products get sorted alphabetically in the product_list partial
     @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products
     render layout: "application"
   end
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -39,6 +39,7 @@ class FacilitiesController < ApplicationController
   def index
     @facilities = Facility.active.sorted
     @active_tab = "home"
+    @products = Product.active.in_active_facility.recently_purchased
     render layout: "application"
   end
 
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -28,7 +28,7 @@ module ProductsHelper
   def public_calendar_link(product)
     if product.respond_to? :reservations
       opts = public_calendar_options(product)
-      link_to "", facility_instrument_public_schedule_path(current_facility, product), opts
+      link_to "", facility_instrument_public_schedule_path( product.facility, product), opts
     end
   end
 
@@ -43,7 +43,7 @@ module ProductsHelper
   private
 
   def public_calendar_options(product)
-    if current_facility.show_instrument_availability?
+    if current_facility&.show_instrument_availability?
       public_calendar_availability_options(product)
     else
       { class: ["fa fa-calendar fa-2x"],
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -43,7 +43,7 @@ module ProductsHelper
   private
 
   def public_calendar_options(product)
-    if current_facility&.show_instrument_availability?
+    if current_facility&.show_instrument_availability? || product.facility.show_instrument_availability?
       public_calendar_availability_options(product)
     else
       { class: ["fa fa-calendar fa-2x"],
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -28,7 +28,7 @@ module ProductsHelper
   def public_calendar_link(product)
     if product.respond_to? :reservations
       opts = public_calendar_options(product)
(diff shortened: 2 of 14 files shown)


app/models/product.rb
```diff
@@ -64,6 +64,7 @@ class Product < ActiveRecord::Base
   scope :not_archived, -> { where(is_archived: false) }
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
   scope :in_active_facility, -> { joins(:facility).where(facilities: { is_active: true }) }
+  scope :recently_purchased, -> { joins(order_details: :order).order('ordered_at DESC').distinct.limit(10) }
```
meara: Does `.order(ordered_at: :desc)` work here? If so, it would be better to use that format to take advantage of ActiveRecord.

app/views/facilities/index.html.haml
```diff
@@ -21,3 +21,19 @@
 
 - else
   .alert.alert-info= t(".empty", items: Facility.model_name.human(count: 2))
+
+- if @products.any?
+  %h4= text(".recently_purchased")
+  %table.table.table-striped.table-hover.product_list
```
meara: I definitely get why you're using a table here; however, we don't typically use tables when listing products in this type of view, so it kind of breaks the style idiom for users, if that makes sense.

I would suggest modeling this after https://github.com/tablexi/nucore-open/blob/master/app/views/facilities/_product_list.html.haml#L7 -- that's what we use in `facilities/show`, so the link locations and style will match what the user expects.

app/views/facilities/index.html.haml
```diff
@@ -21,3 +21,19 @@
...
+    %thead
+      %tr
+        %th
+    %tbody
+      - @products.each do |product|
+        %tr
+          %td
+            - if product.is_a?(Instrument)
+              = link_to product.name, [product.facility, product, :schedule]
```
meara: So this link location, and the link on line 38, are actually linking to the admin views for the product. We want to link to the ordering views (see the product list partial I linked above), since this feature is for regular users trying to buy things, not admins wanting to manage products.

meara:
Oh also, one more small note: I think we want the "Recently Used Products" to appear at the top of the page, above the list of all facilities.

app/models/product.rb
giladshanan: That syntax results in a  `Unknown column 'products.orders'` error, but I think I figured out how to do it with `.merge`

giladshanan:
@jhanggi can you give this a look-through when you have a chance?  @meara and I both worked on it so it would be good to have another pair of eyes on it.

app/models/product.rb
```diff
@@ -64,6 +64,7 @@ class Product < ActiveRecord::Base
   scope :not_archived, -> { where(is_archived: false) }
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
   scope :in_active_facility, -> { joins(:facility).where(facilities: { is_active: true }) }
+  scope :recently_purchased, ->(user) { joins(order_details: :order).merge(Order.where(user: user, state: :purchased).group(:product_id).order('MAX(orders.ordered_at) DESC')).limit(10) }
```
meara: We're not still using this, right?

meara:
Also @giladshanan looks like you probably want to run `rubocop -a` on that new spec file.

app/services/most_recently_used_searcher.rb
```diff
@@ -0,0 +1,21 @@
...
+  attr_reader :user
+
+  def initialize(user)
+    @user = user
+  end
+
+  def recently_used_facilities(limit = 5)
+    return [] unless user
+
+    Facility.active.joins(products: { order_details: :order }).merge(Order.for_user(user).group("orders.facility_id").order('MAX(orders.ordered_at) DESC')).limit(limit)
```
jhanggi: Is there a good way to add some line breaks in here to make the line not so long and easier to read?

Also, use single quotes. And can you say `group(orders: :facility_id)`? (It’s very likely that doesn’t work, but would probably be cleaner)

app/views/facilities/index.html.haml
```diff
@@ -6,6 +6,21 @@
 
 .alert.alert-info= text(".welcome")
 
+- if @products.any?
```
jhanggi: I think it might be nicer to call this `@recent_products`.

spec/services/most_recently_used_searcher_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+        
+        it "returns only 5 products" do
+          expect(recently_used_products.length).to eq 5
+        end        
+      end
+
+    end
+  end
+
+end
```
jhanggi: You use sublime, right? I would recommend adding the `"ensure_newline_at_eof_on_save": true` and `”trim_trailing_white_space_on_save": true` settings.

app/views/facilities/index.html.haml
```diff
@@ -6,6 +6,21 @@
...
+  %h4= text(".recently_purchased")
+  .product_list{ class: @products.first.class.model_name.to_s.pluralize.downcase }
```
jhanggi: Is there a reason we can’t use the `product_list` partial itself?

Part of it may be related to sorting, and if that’s the case, we could pull the sorting out of that and do it in `facilities/show`, though to be honest, I think we should sort the products alphabetically. I think it’ll be easier to use that way. 

If we do decide to decide to leave this here, there’s probably some cleanup we can do: I don’t think the classes here actually do anything.

It might also be nice to add the facility name. But I could go either way.

Also, as noted in the ticket, we want to show recent facilities first, then products.

app/services/most_recently_used_searcher.rb
giladshanan: So `group(orders: :facility_id)` doesn't work (the group method doesn't accept hash arguments) but `group(:id)` does the trick.

spec/services/most_recently_used_searcher_spec.rb
giladshanan: Good call, thanks!

app/views/facilities/index.html.haml
giladshanan: The `product_list` partial expects a current_facility, which we don't have in this context. Is there an easy workaround that would let us use the partial anyway?

Sorting alphabetically seems fine to me.  What do you mean about adding the facility name?

giladshanan: Add the facility name to the product link?  Something like: `Product name (Facility name)`?

giladshanan: So facilities have a `sorted` scope and products have an `alphabetized` scope, they do very similar things.  I lean towards `alphabetized` and `sorted` is only used in a handful of places.  Do we want to change this or leave it alone?

giladshanan:
screen shot 2018-03-12 at 10 43 40 am

app/services/most_recently_used_searcher.rb
jhanggi: I struggle with understanding why `group(:id)` works. Might just be how the two scopes get merged. I think I would lean towards keeping the string in that case.

app/controllers/facilities_controller.rb
```diff
@@ -38,7 +32,9 @@ def dashboard
   # GET /facilities
   def index
     @facilities = Facility.active.sorted
+    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.sorted
```
jhanggi: Should we create an aliased scope `alphabetized` for facilities? I do like that better than `sorted`. Probably want to leave `sorted` around for now because other stuff probably uses it. I'd be fine if we did that separately, though.

app/views/facilities/index.html.haml
```diff
@@ -13,11 +13,28 @@
...
-    %h4= text(".all")
 
+  - if @recent_products.any?
+
+    %h4= text(".recently_purchased")
+    .product_list
+      %ul
+        - @recent_products.each do |product|
+          %li
+            = link_to "#{product.name}" + " - #{product.facility}", facility_product_path(product.facility, product)
```
jhanggi: Nitpicky, but I would do this all in one string: `"#{product.name} - #{product.facility}"

giladshanan:
@meara @jhanggi It looks like `.sorted` was only used in a handful of places so I went ahead and added the change here.  I also included a commit that makes a few adjustments to the product_list partial so that we can use it here.  If we go that route we won't get the facility name in the list.  I wasn't sure what your thoughts were on that so I figured I'd include it here so you can give me some feedback.

app/controllers/facilities_controller.rb
```diff
@@ -37,8 +31,10 @@ def dashboard
 
   # GET /facilities
   def index
-    @facilities = Facility.active.sorted
+    @facilities = Facility.active.alphabetized
+    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
+    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.alphabetized
```
meara: Are we sure that including the `alphabetized` scope here doesn't affect which products are selected as recently used? I'm a little leery of adding additional sorting in here because of the way active record chains scopes. Can we test this?

app/views/facilities/_product_list.html.haml
```diff
@@ -1,18 +1,22 @@
...
+    - if current_facility.present?
+      %h3= product_list_title products, local_assigns[:title_extra]
+
     - if products.first.class.model_name == "TimedService"
       %p= t("facilities.show.timed_services_note")
     %ul
       - products.sort.each do |product|
         %li{ class: product.class.model_name.to_s.downcase }
-          = public_calendar_link(product)
+          - if current_facility.present?
```
meara: Does `public_calendar_link` require the facility? I think we can leave that in, I don't think having the calendar icon appear hurts anything, and it might be useful.

app/views/facilities/_product_list.html.haml
```diff
@@ -1,18 +1,22 @@
 - if products.any?
   .product_list{ class: products.first.class.model_name.to_s.pluralize.downcase }
-    %h3= product_list_title products, local_assigns[:title_extra]
+
+    - if current_facility.present?
+      %h3= product_list_title products, local_assigns[:title_extra]
+
     - if products.first.class.model_name == "TimedService"
```
meara: I think the TimedService note ought to be nested under the `if current_facility.present?` as well. Try adding a Timed Service product type to see how it shows up.

app/views/facilities/_product_list.html.haml
```diff
@@ -1,18 +1,22 @@
...
-          = public_calendar_link(product)
+          - if current_facility.present?
+            = public_calendar_link(product)
           - if local_assigns[:f]
             = f.fields_for :order_details do |builder|
               - if session_user.can_override_restrictions?(product) || product.can_be_used_by?(acting_user)
                 = builder.text_field :quantity, value: 0, class: "product_quantity", index: nil
                 = builder.hidden_field :product_id, value: product.id, index: nil
-          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility, product)
+          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product)
```
meara: I think this can just be `product.facility`, rather than using the or.

meara: Also we should check for N+1s here.

app/services/most_recently_used_searcher.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def recently_used_facilities(limit = 5)
+    return [] unless user
+
+    Facility.active
+            .joins(products: { order_details: :order })
+            .merge(Order.for_user(user)
+                        .group(:id)
+                        .order('MAX(orders.ordered_at) DESC'))
```
jhanggi: Use double quotes for strings.

app/services/most_recently_used_searcher.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def recently_used_products(limit = 10)
+    return [] unless user
+
+    Product.active.in_active_facility
+           .joins(order_details: :order)
+           .merge(Order.for_user(user)
+                       .group(:id)
+                       .order('MAX(orders.ordered_at) DESC'))
```
jhanggi: Double quotes

app/views/facilities/_product_list.html.haml
jhanggi: I agree, if we can include it here, I say we do it.

giladshanan: The `public_calendar_link` does require a facility, it links to `facility_instrument_public_schedule_path`.  There's a way to make it work though. I'll put it in my next commit with the other changes so you can let me know what you think.

giladshanan:
screen shot 2018-03-13 at 2 10 28 pm
Partial view

giladshanan:
screen shot 2018-03-13 at 2 10 49 pm
Partial view with calendar (still working on the other suggested changes, but this is how it looks)

app/controllers/facilities_controller.rb
giladshanan: The `alphabetized` scope adds a secondary column to sort by, in this case the end of the query looks like this `ORDER BY MAX(orders.ordered_at) DESC, lower(products.name)`.  Testing it out it doesn't seem to change the order at all.  In fact, if you take a closer look at the first screenshot above, you'll see the products are not listed in alpha order.  The ordering in the two more recent screenshots comes from the partial, which uses Array#sort on the products collection returned by `@recent_products`.

giladshanan: Should we ditch the `alphabetized` scope for products since it isn't doing much?

meara: I was just about to suggest that 😀 . Leaving it in seems misleading, since the actual ordering is provided by the Array#sort in the partial.

It might be nice to pull the sort out of the partial and into the controller (and add a comment on why we're doing the sort in ruby and not in SQL)--you'd need to make sure you made the change for everywhere the partial is used, though.

app/services/most_recently_used_searcher.rb
```diff
@@ -0,0 +1,31 @@
...
+            .merge(Order.for_user(user)
+                        .group(:id)
+                        .order("MAX(orders.ordered_at) DESC"))
+            .limit(limit)
+  end
+
+  def recently_used_products(limit = 10)
+    return [] unless user
+
+    Product.active.in_active_facility
```
meara: I would put `includes(:facility)` explicitly here, rather than in the `in_active_facility` scope.

app/views/facilities/_product_list.html.haml
```diff
@@ -12,7 +15,7 @@
               - if session_user.can_override_restrictions?(product) || product.can_be_used_by?(acting_user)
                 = builder.text_field :quantity, value: 0, class: "product_quantity", index: nil
                 = builder.hidden_field :product_id, value: product.id, index: nil
-          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility, product)
+          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(product.facility, product)
```
meara: On second thought, I think what you had with `current_facility || product.facility` is better, because it lets us avoid all the `includes` we would in the facility-specific context to prevent N+1s.

Specifically on this page: https://github.com/tablexi/nucore-open/blob/master/app/views/facilities/show.html.haml

app/helpers/products_helper.rb
```diff
@@ -43,7 +43,7 @@ def product_url_hint(product)
   private
 
   def public_calendar_options(product)
-    if current_facility.show_instrument_availability?
+    if current_facility&.show_instrument_availability?
```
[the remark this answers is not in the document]
giladshanan: does `current_facility&.show_instrument_availability? || product.facility.show_instrument_availability?` do the trick?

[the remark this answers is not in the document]
meara: The issue is we only want to call `product.facility` if `current_facility` is not present, rather than any time `current_facility&.show_instrument_availability?` evaluates to false. Does that make sense?

app/controllers/facilities_controller.rb
giladshanan: Hm, it looks like this would mean putting 10 `.sort`s in the `facilities/show` view.  Is there another way to clarify intent here?  Maybe a code comment in the facilities controller where we call the MostRecentlyUsedSearcher's?

app/services/most_recently_used_searcher.rb
giladshanan: @meara it looks like Oracle doesn't like the new AR queries so we're gonna have to break it up into two steps like it was before.

app/controllers/facilities_controller.rb
```diff
@@ -37,8 +31,11 @@ def dashboard
 
   # GET /facilities
   def index
-    @facilities = Facility.active.sorted
+    @facilities = Facility.active.alphabetized
+    @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = "home"
+    # recent products get sorted alphabetically in the product_list partial
+    @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.includes(:facility)
```
jhanggi: Is there any downside to adding the alphabetization here, even if it's still re-sorted in the partial. We don't need to do it now, but I think it would be nice to not do the sort in the `product_list` partial (`facilities/show` is kind of nasty already, but could defiitely be cleaned up), but that can be for another time.

app/helpers/products_helper.rb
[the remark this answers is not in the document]
giladshanan: Yup, that does make sense, thanks.

giladshanan:
🙌

PR #1430, opened 2018-03-06
[#87522] fix subquery ordering bug in oracle, fix order links in new/inprocess

meara:


app/services/transaction_search/product_searcher.rb
```diff
@@ -3,7 +3,7 @@ module TransactionSearch
   class ProductSearcher < BaseSearcher
 
     def options
-      Product.where(id: order_details.select("distinct order_details.product_id")).order(:name)
+      Product.where(id: order_details.select("distinct order_details.product_id")).reorder(:name)
```
jhanggi: I think this can just be a plain `order` now.

meara:
@jhanggi if you have another minute to look at this, I'd super appreciate it--I can't figure out my way around this test failure with the sorting on `assigned_user`. Something weird seems to be happening with the join to `assigned_user`--if I use preload, it seems not to be joining to `assigned_user`, which makes some sense, but if I try including `assigned_user` instead of preloading, I get unknown column errors on everything in `assigned_users_order_details`.

Help?

jhanggi:
I don't know the full answer, but here's what I can tell you: as is, the query does not set an alias when it `joins(:assigned_user)`, it's just using `users`. If you add a `joins(order: :user)`, then AR is aware there are multiple joins to `users` and it aliases the assigned user `joins`, so the alias we use in the sort will work again. 

You also need to use `includes(:assigned_user)` vs `preload`. `includes` has some smarts in it that decides whether to `eager_load` or `preload` cf https://blog.bigbinary.com/2013/07/01/preload-vs-eager-load-vs-joins-vs-includes.html

What I'm most confused by is how the `user` association is getting preloaded in the normal case to avoid the N+1.

If you can figure that out, then it might make it so you don't need to add the `joins(order: :user)` explicitly. I would recommend you still use `includes` vs `preload` in this case.

Hope that helps.

jhanggi:
That also means that `joins(:assigned_user)` is not the correct option in that it uses an `INNER JOIN` meaning if there's no assigned user, then it won't be returned.

app/controllers/concerns/new_inprocess_controller.rb
```diff
@@ -9,7 +9,10 @@ def index
     @search = TransactionSearch::Searcher.new(TransactionSearch::ProductSearcher,
                                               TransactionSearch::DateRangeSearcher,
                                               TransactionSearch::OrderedForSearcher).search(order_details, @search_form)
-    @order_details = @search.order_details.preload(:order_status, :assigned_user)
+    @order_details = @search.order_details
+                            .includes(:order_status)
+                            .joins("LEFT OUTER JOIN users AS assigned_users ON assigned_users.id = order_details.assigned_user_id")
```
meara: @jhanggi This is a little inelegant, but it works.

jhanggi: Yeah, maybe pull it into a scope on `OrderDetail` so that it's kind of hidden away from the controller?

app/controllers/facility_reservations_controller.rb
```diff
@@ -229,7 +228,7 @@ def sort_lookup_hash
       "reserve_range" => "CONCAT(reservations.reserve_start_at, reservations.reserve_end_at)",
       "product_name"  => "products.name",
       "status"        => "order_statuses.name",
-      "assigned_to"   => "CONCAT(assigned_users_order_details.last_name, assigned_users_order_details.first_name)",
+      "assigned_to"   => "CONCAT(assigned_users.last_name, assigned_users.first_name)",
```
jhanggi: It might be nice to make the reservations and orders controllers a little more consistent by using the same format, either `CONCAT` or the array. I think I lean towards the array.

PR #1433, opened 2018-03-09
[#140405] payment source transactions link to main transaction filter page

meara:
Add links to account details pages from My Payment Sources, remove transactions tab from accounts, link "Transactions" link to the My Transactions filter page instead.


meara:
We have a spec for the transactions search page--https://github.com/tablexi/nucore-open/blob/master/spec/features/transactions_search_spec.rb--it seemed a bit overkill to me to add a feature spec for the transactions links, but I can if you feel strongly.

app/views/accounts/index.html.haml
```diff
@@ -23,14 +23,17 @@
       - @account_users.each do |au|
         - account = au.account
         %tr
-          %td= account.account_number_to_s
+          - if au.can_administer? || session_user.administrator?
```
jhanggi: Is it possible to use a cancan ability check here rather than an explicit check? I see that's what we use below, so if it's too much effort to figure out, it's fine as is.

meara: I'd rather not change this for now if that works, just to make sure we aren't changing the existing behavior.

PR #1434, opened 2018-03-11
Add settings override for default ruby-saml params

gluxon:
The University of Connecticut Shibboleth server prefers the HTTP Redirect SAML binding for sign-in. Changing this required overriding the idp_sso_target_url SAML setting passed to ruby-saml.

This is my first pull request for the project, so let me just note the following: I am perfectly fine modifying/improving anything that may not be up to maintainer expectations. I wouldn't call myself a Ruby expert, so if I'm doing something that's not a best practice, please feel free to be blunt and call me out on it. This follows for any future pull requests I make.

### DIFF
--- a/vendor/engines/saml_authentication/lib/saml_authentication/devise_configurator.rb
+++ b/vendor/engines/saml_authentication/lib/saml_authentication/devise_configurator.rb
@@ -19,6 +19,9 @@ module SamlAuthentication
         config.saml_configure do |settings|
           settings.assertion_consumer_service_url = Rails.application.routes.url_helpers.auth_saml_user_session_url
           settings.issuer = Rails.application.routes.url_helpers.metadata_saml_user_session_url
+          Hash(Settings.saml.driver).each do |key, value|
+            settings.public_send("#{key}=", value)
+          end
 
           configure_security(settings)
         end
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -145,6 +145,8 @@ paperclip:
 # saml:
 #   idp_metadata: "https://websso.example.com/idp/metadata"
 #   # certificate_file: path/to/file.p12 # Optional. Do not check in to version control.
+#   # driver: # Optional. Useful to override inferred SAML settings if need be.
+#   #   "idp_sso_target_url": "https://websso.example.com/idp/profile/SAML2/Redirect/SSO"
 #   attribute_map:
 #     "PersonImmutableID": "username"
 #     "User.email": "email"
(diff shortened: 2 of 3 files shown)


mcab:
To provide context for what `driver` may be used for, an example use has been added to [config/settings.yml](https://github.com/tablexi/nucore-open/blob/55d0256978fdfbe3d90c93b9a5c826d280052ca5/config/settings.yml).

In regards to spec testing, would it make sense to test this change in [sessions_controller_spec.rb](https://github.com/tablexi/nucore-open/blob/55d0256978fdfbe3d90c93b9a5c826d280052ca5/vendor/engines/saml_authentication/spec/controllers/saml_authentication/sessions_controller_spec.rb)? It's main purpose is to test account creation and updating, but not for overriding possible issues with IdP servers.

vendor/engines/saml_authentication/lib/saml_authentication/devise_configurator.rb
```diff
@@ -19,6 +19,9 @@ def configure!
         config.saml_configure do |settings|
           settings.assertion_consumer_service_url = Rails.application.routes.url_helpers.auth_saml_user_session_url
           settings.issuer = Rails.application.routes.url_helpers.metadata_saml_user_session_url
+          Settings.saml.driver&.each do |key, value|
```
jhanggi: I can't quite decide if this should go before or after the call to`configure_security`.  On one hand, it would add the flexibility of overriding the security settings in the YAML, but on the other hand, I'm not sure I want those settings to be overridable. I think we can leave it as is for now and cross that bridge if we need to.

I might also consider `Hash(Settings.saml.driver).each`. It'll still handle a `nil` case, and it's a little more explicit about what it expects.

```ruby
Hash(nil) => {}
Hash({}) => {}
Hash({ a: :b }) => { a: :b }
```

jhanggi:
One other thing, can you push something else, even if just an empty commit? I'd like to get CircleCI to run builds from the fork. I just made a change to our settings that I believe will do that, but it probably needs to be re-triggered.

vendor/engines/saml_authentication/lib/saml_authentication/devise_configurator.rb
gluxon: > I might also consider Hash(Settings.saml.driver).each. It'll still handle a nil case, and it's a little more explicit about what it expects.

That's a good idea; I like that too. Just pushed the amended commit.

gluxon:
> I'm ok if we don't have a spec for this (don't get used to that response, though :)

Don't worry, we won't!

> One other thing, can you push something else, even if just an empty commit? I'd like to get CircleCI to run builds from the fork. I just made a change to our settings that I believe will do that, but it probably needs to be re-triggered.

Sure thing. I added the amended commit a few days ago, but I just realized that it looks like no response happened on my end when scrolling through comments. (It got buried under "outdated diffs.) So let this comment serve as an official response from me.

PR #1440, opened 2018-03-12
[#87522] update new/inprocess/problem tabs for orders/reservations/occupancies…

meara:
… per NU reqs

From NU:
```
We need a few changes on the Orders/Reservations/Occupancies tabs:

Orders, Reservations: New/In Process: Lock billing status at Ordered (gray out selection), add Order Statuses filter
Problem Orders: Lock billing status at Ordered (gray out selection), Add Payment Source filter

Occupancies: New/In Process: Remove everything except Products filter
Problem Orders: Lock billing status at Ordered (gray out selection), Add Payment Source filter
```

### DIFF
--- a/app/services/transaction_search/order_status_searcher.rb
+++ b/app/services/transaction_search/order_status_searcher.rb
@@ -3,10 +3,10 @@ module TransactionSearch
   class OrderStatusSearcher < BaseSearcher
 
     def options
-      # Unlike the other lookups, this query is much faster this way than using a subquery
-      OrderStatus.find_by_sql(order_details.joins(:order_status)
-                                           .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                                           .reorder("order_statuses.lft").to_sql).uniq
+      OrderStatus.select("order_statuses.id, order_statuses.facility_id, order_statuses.name, order_statuses.lft")
+                 .where(id: order_details.select("distinct order_details.order_status_id"))
+                 .order("order_statuses.lft")
+
     end
 
     def search(params)
--- a/app/services/transaction_search/order_status_searcher.rb
+++ b/app/services/transaction_search/order_status_searcher.rb
@@ -6,8 +6,7 @@ module TransactionSearch
       # Unlike the other lookups, this query is much faster this way than using a subquery
       OrderStatus.find_by_sql(order_details.joins(:order_status)
                                            .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                                           .reorder("order_statuses.lft").to_sql)
-                                           .uniq
+                                           .reorder("order_statuses.lft").to_sql).uniq
     end
 
     def search(params)
--- a/app/services/transaction_search/order_status_searcher.rb
+++ b/app/services/transaction_search/order_status_searcher.rb
@@ -3,8 +3,7 @@ module TransactionSearch
   class OrderStatusSearcher < BaseSearcher
 
     def options
-      OrderStatus.select("order_statuses.id, order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                 .where(id: order_details.select("distinct order_details.order_status_id"))
+      OrderStatus.where(id: order_details.select("distinct order_details.order_status_id"))
                  .order("order_statuses.lft")
     end
 
--- a/app/services/transaction_search/order_status_searcher.rb
+++ b/app/services/transaction_search/order_status_searcher.rb
@@ -5,8 +5,9 @@ module TransactionSearch
     def options
       # Unlike the other lookups, this query is much faster this way than using a subquery
       OrderStatus.find_by_sql(order_details.joins(:order_status)
-                                                         .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                                                         .reorder("order_statuses.lft").to_sql)
+                                           .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
+                                           .reorder("order_statuses.lft").to_sql)
+                                           .uniq
     end
 
     def search(params)
--- a/app/services/transaction_search/order_status_searcher.rb
+++ b/app/services/transaction_search/order_status_searcher.rb
@@ -6,7 +6,6 @@ module TransactionSearch
       OrderStatus.select("order_statuses.id, order_statuses.facility_id, order_statuses.name, order_statuses.lft")
                  .where(id: order_details.select("distinct order_details.order_status_id"))
                  .order("order_statuses.lft")
-
     end
 
     def search(params)
--- a/app/services/transaction_search/order_status_searcher.rb
+++ b/app/services/transaction_search/order_status_searcher.rb
@@ -3,8 +3,10 @@ module TransactionSearch
   class OrderStatusSearcher < BaseSearcher
 
     def options
-      OrderStatus.where(id: order_details.select("distinct order_details.order_status_id"))
-                 .order("order_statuses.lft")
+      # Unlike the other lookups, this query is much faster this way than using a subquery
+      OrderStatus.find_by_sql(order_details.joins(:order_status)
+                                           .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
+                                           .reorder("order_statuses.lft").to_sql)
     end
 
     def search(params)
--- a/app/views/shared/transactions/_search2.html.haml
+++ b/app/views/shared/transactions/_search2.html.haml
@@ -9,7 +9,7 @@
     %fieldset.span2#calendar_filter
       - if @search.options.map(&:key).include?("date_ranges")
         %br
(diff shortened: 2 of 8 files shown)


app/services/transaction_search/order_status_searcher.rb
```diff
@@ -5,8 +5,8 @@ class OrderStatusSearcher < BaseSearcher
     def options
       # Unlike the other lookups, this query is much faster this way than using a subquery
       OrderStatus.find_by_sql(order_details.joins(:order_status)
-                                                         .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                                                         .reorder("order_statuses.lft").to_sql)
+                                           .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
+                                           .reorder("order_statuses.lft").to_sql).uniq
```
jhanggi: I feel like the `uniq` shouldn't be necessary. What caused it to become necessary? The `.select("distinct(order_statuses.id)` should have it covered.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
```diff
@@ -20,9 +20,18 @@ def initialize
...
 
     # GET /facilities/:facility_id/occupancies
-    # Provided by NewInprocessController
-    # def index
-    # end
+    def index
+      order_details = new_or_in_process_orders.joins(:order)
+
+      @search_form = TransactionSearch::SearchForm.new(params[:search], defaults: { date_range_field: "ordered_at" })
+      @search = TransactionSearch::Searcher.new(TransactionSearch::ProductSearcher).search(order_details, @search_form)
```
jhanggi: I'll be honest, I'm a little confused by them wanting to remove the ordered for filter. I'm also a little surprised they wouldn't want the payment source filter on both in/inprocess and problems.

app/views/shared/transactions/_search2.html.haml
```diff
@@ -9,7 +9,7 @@
     %fieldset.span2#calendar_filter
       - if @search.options.map(&:key).include?("date_ranges")
         %br
-        = f.input :date_range_field, collection: TransactionSearch::DateRangeSearcher.options, label: false, include_blank: false
+        = f.input :date_range_field, collection: TransactionSearch::DateRangeSearcher.options.select { |label, value| value.in?(@search_form.allowed_date_fields) }, label: false, include_blank: false
```
jhanggi: Just wondering if it would be worth it to add an `only:` argument to `TransactionSearch::DateRangeSearcher.options` and pull the `select` logic in there. Might clean up the view a bit.

app/services/transaction_search/order_status_searcher.rb
meara: I'm confused by that too, but for some reason in the new/in process tabs we were getting multiple (literally as many as there were order details) instances of "new" and "in process" appearing as options. Weirdly, when I tried to filter by one of the "new"s, it would do the filter correctly (find 22 of 22 new order details), but then the selector would have all 22 "new"s shown as selected.

I ended up throwing the uniq on it because I figured that wouldn't break anything and it didn't seem worth doing a deep dive into why I was getting those results when it was behaving as expected in the billing tab, but if it bothers you excessively I can look into it more.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
meara: I thought that was weird too, but I wasn't sure it was worth pushing back on.

app/services/transaction_search/order_status_searcher.rb
jhanggi: It only concerns me if the query itself returns an OrderStatus object for _every single_ order detail, and then it gets `uniq`ed. If that's thousands of order details (especially if it's not paginated), that could balloon memory. If the `uniq` changes the SQL, then it's probably fine.

vendor/engines/secure_rooms/app/controllers/secure_rooms/facility_occupancies_controller.rb
jhanggi: Yeah, probably not worth fighting. And it's easy enough to add back in when they realize they want it.

app/services/transaction_search/order_status_searcher.rb
meara: Ah, it definitely is returning an OrderStatus object for every single order detail. They're distinct objects too, not just the same one multiple times. I'll play around with restructuring the query and see if there's a better way to do this. I don't get why the distinct isn't working...

app/services/transaction_search/order_status_searcher.rb
```diff
@@ -3,10 +3,9 @@ module TransactionSearch
   class OrderStatusSearcher < BaseSearcher
 
     def options
-      # Unlike the other lookups, this query is much faster this way than using a subquery
-      OrderStatus.find_by_sql(order_details.joins(:order_status)
-                                                         .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                                                         .reorder("order_statuses.lft").to_sql)
+      OrderStatus.select("order_statuses.id, order_statuses.facility_id, order_statuses.name, order_statuses.lft")
```
jhanggi: This is probably fine enough. I worry about the comment around the performance of the query, though. If we do go this way, probably don't need the `.select("order_statuses.id, order_statuses.facility_id, order_statuses.name, order_statuses.lft")`, and we can just select everything. OrderStatus doesn't have a lot of cruft on it.

meara: Admittedly I don't have massive amounts of seed data I'm testing against, but I'm not seeing any noticeable slowness, and this is demonstrably better memory-wise. We can always optimize for speed later if it turns out to be an issue.

jhanggi: I think that's fine. My only concern is that we did optimize this at one point. Let me do a quick test again NU data and make sure this doesn't cause a big impact.

app/services/transaction_search/order_status_searcher.rb
```diff
@@ -3,10 +3,8 @@ module TransactionSearch
   class OrderStatusSearcher < BaseSearcher
 
     def options
-      # Unlike the other lookups, this query is much faster this way than using a subquery
-      OrderStatus.find_by_sql(order_details.joins(:order_status)
-                                                         .select("distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft")
-                                                         .reorder("order_statuses.lft").to_sql)
+      OrderStatus.where(id: order_details.select("distinct order_details.order_status_id"))
```
jhanggi: It's a significant difference.

Searching on All Transactions with no date filters, it's the difference of 1/4 second and 5 seconds. The queries are definitely different. I think we do need to spend a little more time on it.

Run 1:
Old:
```
OrderStatus Load (217.5ms)  SELECT distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft FROM "ORDER_DETAILS" INNER JOIN "ORDERS" ON "ORDERS"."ID" = "ORDER_DETAILS"."ORDER_ID" INNER JOIN "ORDER_STATUSES" ON "ORDER_STATUSES"."ID" = "ORDER_DETAILS"."ORDER_STATUS_ID" WHERE "ORDERS"."STATE" = 'purchased' AND "ORDERS"."FACILITY_ID" = 21  ORDER BY order_statuses.lft
```
New:
```
  OrderStatus Load (4989.1ms)  SELECT "ORDER_STATUSES".* FROM "ORDER_STATUSES" WHERE "ORDER_STATUSES"."ID" IN (SELECT distinct order_details.order_status_id FROM "ORDER_DETAILS" INNER JOIN "ORDERS" ON "ORDERS"."ID" = "ORDER_DETAILS"."ORDER_ID" WHERE "ORDERS"."STATE" = 'purchased' AND "ORDERS"."FACILITY_ID" = :a1)  ORDER BY order_statuses.lft  [["facility_id", 21]]
```

Run 2:
Old:
```
OrderStatus Load (177.7ms)  SELECT distinct(order_statuses.id), order_statuses.facility_id, order_statuses.name, order_statuses.lft FROM "ORDER_DETAILS" INNER JOIN "ORDERS" ON "ORDERS"."ID" = "ORDER_DETAILS"."ORDER_ID" INNER JOIN "ORDER_STATUSES" ON "ORDER_STATUSES"."ID" = "ORDER_DETAILS"."ORDER_STATUS_ID" WHERE "ORDERS"."STATE" = 'purchased' AND "ORDERS"."FACILITY_ID" = 21  ORDER BY order_statuses.lft
```
New:
```
  OrderStatus Load (4204.8ms)  SELECT "ORDER_STATUSES".* FROM "ORDER_STATUSES" WHERE "ORDER_STATUSES"."ID" IN (SELECT distinct order_details.order_status_id FROM "ORDER_DETAILS" INNER JOIN "ORDERS" ON "ORDERS"."ID" = "ORDER_DETAILS"."ORDER_ID" WHERE "ORDERS"."STATE" = 'purchased' AND "ORDERS"."FACILITY_ID" = :a1)  ORDER BY order_statuses.lft  [["facility_id", 21]]
```

PR #1443, opened 2018-03-13
[#140080] enable product search without logging in or when acting as

meara:
https://pm.tablexi.com/issues/140080

### DIFF
--- a/app/controllers/global_search_controller.rb
+++ b/app/controllers/global_search_controller.rb
@@ -12,13 +12,12 @@ class GlobalSearchController < ApplicationController
   end
 
   def index
-    @searchers = if session_user.nil? || session_user != acting_user
-                   # if user is not logged in, or if ordering on behalf, only product searching is available
-                   [GlobalSearch::ProductSearcher.new(acting_user, current_facility, params[:search])]
-                 else
+    @searchers = if user_signed_in? && !acting_as?
                    self.class.searcher_classes.map do |searcher_class|
                      searcher_class.new(acting_user, current_facility, params[:search])
                    end
+                 else
+                   [GlobalSearch::ProductSearcher.new(acting_user, current_facility, params[:search])]
                  end
   end
 
--- a/app/controllers/global_search_controller.rb
+++ b/app/controllers/global_search_controller.rb
@@ -1,8 +1,6 @@
 class GlobalSearchController < ApplicationController
 
   customer_tab :all
-  before_action :authenticate_user!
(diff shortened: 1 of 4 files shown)


app/controllers/global_search_controller.rb
```diff
@@ -14,9 +12,14 @@ def self.searcher_classes
   end
 
   def index
-    @searchers = self.class.searcher_classes.map do |searcher_class|
-      searcher_class.new(current_user, current_facility, params[:search])
-    end
+    @searchers = if session_user.nil? || session_user != acting_user
```
jhanggi: Looks like we have a `acting_as?` on ApplicationController that does something similar. Could we use that? I think Devise gives us a `logged_in?` helper. Would `if logged_in? && !acting_as?` be more intention-revealing? (I think what I said requires would flip the conditional)

app/controllers/global_search_controller.rb
```diff
@@ -14,9 +12,13 @@ def self.searcher_classes
   end
 
   def index
-    @searchers = self.class.searcher_classes.map do |searcher_class|
-      searcher_class.new(current_user, current_facility, params[:search])
-    end
+    @searchers = if user_signed_in? && !acting_as?
```
jhanggi: I am somewhat of the opinion that you should be able to do a little more than you currently can while acting as another user. If we included other searchers, like the order detail or statement searchers, would that lead to some 403s or "you can't do that while acting as another user" from the resulting links? If it does, then this is fine.

If it doesn't I would vote towards including the searchers if you're signed in and using the user you're acting as as the permission check (I forget who is acting_user and who is current_user in this context).

spec/controllers/global_search_controller_spec.rb
```diff
@@ -40,15 +40,23 @@ def it_should_have_customer_paths
...
   let(:statement_results) { assigns[:searchers].find { |s| s.template == "statements" }.results }
+  let(:product_results) { assigns[:searchers].find { |s| s.template == "products" }.results }
 
   describe "index" do
-    describe "permissions" do
-      before :each do
-        @action = :index
-        @method = :get
+
+    context "when not signed in" do
```
jhanggi: We should probably add a spec for when you're acting as another user.

app/controllers/global_search_controller.rb
meara: You'll get a `This function is unavailable while you are acting as another user.` message if you try to click on an order detail link after searching for it.

PR #1444, opened 2018-03-14
[#140353] replace problematic test

meara:
Fix sporadically failing test. Example: https://circleci.com/gh/tablexi/nucore-open/5517


meara:
Oh good catch, I hadn't seen that spec.

PR #1453, opened 2018-03-16
[#140196, #140197] extract secrets from settings.yml

meara:
# Release Notes

Secret information such as API keys should exist in `secrets.yml` rather than in `settings.yml` or by being overridden in `settings.local.yml`.

# Deploy Note

Will need to update `settings.yml` and `secrets.yml` to match the updated templates on staging and production servers in conjunction with the release.


config/settings.yml
```diff
@@ -135,7 +125,7 @@ split_accounts:
     - administrator
     # - account_manager
 
-# This may be overridden in settings.local.yml if your fork is using S3, so
+# This may be overridden in secrets.yml if your fork is using S3, so
```
jhanggi: This part isn't quite true as it stands. I think it might be nice to think about how this can work. 

What might work is to do a `Settings.paperclip.merge(Rails.application.secrets.paperclip)` and then we can get the best of both worlds. Thoughts on that?

meara: Hmm. Where would we actually do the merge?

jhanggi: Looks like we use the paperclip settings in `app/models/concerns/downloadable_file.rb`. So probably in there.

giladshanan:
Looks good.  Does it make sense to add a "Deploy Note" that changes will need to be made to `secrets.yml` on the stage and production servers?

meara:
Good call on adding a "Deploy Note" @giladshanan.

app/models/concerns/downloadable_file.rb
```diff
@@ -3,14 +3,14 @@ module DownloadableFile
   extend ActiveSupport::Concern
 
   included do
-    has_attached_file :file, Settings.paperclip.to_hash.merge(validate_media_type: false)
+    has_attached_file :file, PaperclipSettings.config.to_hash.merge(validate_media_type: false)
```
jhanggi: The `to_hash` here might now be unnecessary.

PR #1455, opened 2018-03-19
[#137627] recently used products/facilities still works when no logged in user

meara:
# Release Notes
Fix error on recently used products/facilities when there is no logged in user, follow up to https://github.com/tablexi/nucore-open/pull/1429.

# Additional Context

Error was occurring because we were trying to chain additional scopes onto `recently_used_facilities` and `recently_used_products` in `facilities#index`, which was an empty array rather than an empty AR relation when no user is present.

### DIFF
--- a/app/services/most_recently_used_searcher.rb
+++ b/app/services/most_recently_used_searcher.rb
@@ -7,14 +7,14 @@ class MostRecentlyUsedSearcher
   end
 
   def recently_used_facilities(limit = 5)
-    return [] unless user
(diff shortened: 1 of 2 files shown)


jhanggi:
Do we want to add a spec that should have caught this? I feel like the link to the PR that caused it could be linked in the release notes. That way if we can essentially squash this note in our email notes if the guilty PR had not been released.

giladshanan:
Adding a spec seems like a good idea to me.

PR #1457, opened 2018-03-20
[#140997] feature flag saml create/update user

meara:
# Release Notes

Feature flag new user creation/updating via SAML.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -128,7 +128,6 @@ feature:
   price_change_reason_required_on: true
   can_manage_global_price_groups_on: true
   cross_facility_reports_on: false
-  saml_create_user_on: false
 
 split_accounts:
   # Roles are allowed to create Split Accounts
@@ -144,6 +143,7 @@ paperclip:
   path: ":rails_root/public/:attachment/:id_partition/:style/:safe_filename"
 
 # saml:
+#   create_user: false
 #   idp_metadata: "https://websso.example.com/idp/metadata"
 #   # certificate_file: path/to/file.p12 # Optional. Do not check in to version control.
 #   # driver: # Optional. Useful to override inferred SAML settings if need be.
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -153,4 +153,4 @@ paperclip:
 #     "User.email": "email"
 #     "User.FirstName": "first_name"
 #     "User.LastName": "last_name"
-
+#  # user_updater_class_name: DartmouthRadar::UserUpdater
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -128,6 +128,7 @@ feature:
   price_change_reason_required_on: true
   can_manage_global_price_groups_on: true
   cross_facility_reports_on: false
+  saml_create_user_on: false
 
 split_accounts:
   # Roles are allowed to create Split Accounts
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -143,7 +143,7 @@ paperclip:
   path: ":rails_root/public/:attachment/:id_partition/:style/:safe_filename"
 
 # saml:
-#   create_user: false
+#   create_user: true
 #   idp_metadata: "https://websso.example.com/idp/metadata"
 #   # certificate_file: path/to/file.p12 # Optional. Do not check in to version control.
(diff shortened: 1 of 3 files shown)


vendor/engines/saml_authentication/lib/saml_authentication/devise_configurator.rb
```diff
@@ -9,8 +9,8 @@ def configure!
       Devise.setup do |config|
         config.saml_session_index_key = :session_index
         config.saml_default_user_key = :username
-        config.saml_create_user = true
-        config.saml_update_user = true
+        config.saml_create_user = SettingsHelper.feature_on?(:saml_create_user)
+        config.saml_update_user = SettingsHelper.feature_on?(:saml_create_user)
```
meara: @jhanggi I know we explicitly don't want to update the email, but we seem to be depending on this to do updating of the attributes as well.

The specs need updating to take into account the feature being on and off, obviously, but I'd like to talk through with you what we want to be the expected behavior for each state when you get a chance.

meara: Do this for only create, move from feature flag into the saml settings.

vendor/engines/saml_authentication/spec/controllers/saml_authentication/sessions_controller_spec.rb
```diff
@@ -24,6 +24,17 @@
     end
 
     describe "the user does not exist already" do
+      around :each do |example|
```
meara: @jhanggi maybe you can help me debug--something weird is going on here. When I have `create_user: false` set locally, the specs will fail, even though if I throw a pry binding into the spec the around wrapper seems to be working properly, because inside the example it shows `Settings.saml.create_user` as `true`. I can't figure out what the issue is--any ideas?

jhanggi: The `config.saml_create_user = saml_create_user?` only gets called at initialization, so changing the `Settings` value here won't affect the actual devise configuration. I think if you do `original_value = Devise.saml_create_user`. And since that already should be a boolean (since you did the nil check in the initializer), you shouldn't need the nil check here.

config/settings.yml
```diff
@@ -152,4 +153,4 @@ paperclip:
 #     "User.email": "email"
 #     "User.FirstName": "first_name"
 #     "User.LastName": "last_name"
-
+#  # user_updater_class_name: DartmouthRadar::UserUpdater
```
jhanggi: Let's leave the dartmouth specific naming out of this. I think we can have `SamlAuthentication::UserUpdater`.

config/settings.yml
```diff
@@ -143,6 +143,7 @@ paperclip:
   path: ":rails_root/public/:attachment/:id_partition/:style/:safe_filename"
 
 # saml:
+#   create_user: false
```
jhanggi: I feel like the "default" we show should be the actual default. Or we should add some documentation around it.

If we start wanting to add more documentation, then maybe we do a thing in the engine where we add another Settings source, like a `saml.yml`.

vendor/engines/saml_authentication/lib/saml_authentication/devise_configurator.rb
```diff
@@ -30,6 +30,18 @@ def configure!
 
     private
 
+    def saml_updater
+      Settings.saml.user_updater_class_name.presence.try(:constantize).try(:new) || SamlAuthentication::UserUpdater.new
+    end
+
+    def saml_create_user?
+      if !Settings.saml.create_user.nil?
```
jhanggi: I would probably flip the conditional. It gets rid of a double negative, and I think it clarifies that it's a default.

```ruby
if Settings.saml.create_user.nil?
  true
else
  Settings.saml.create_user
end
```

or an alternative construction:

```ruby
Settings.saml.create_user = true if Settings.saml.create_user.nil?
Settings.saml.create_user
```

config/settings.yml
meara: A `saml.yml` file is an interesting idea. I don't know if we have enough saml-specific stuff to justify that yet, but good thing to keep in mind going forward.

PR #1464, opened 2018-03-28
Fix cap deploy on stage

meara:
# Release Notes

TECH TASK: Continue work on reorganizing secrets/settings.

### DIFF
--- a/app/controllers/api/order_details_controller.rb
+++ b/app/controllers/api/order_details_controller.rb
@@ -3,7 +3,7 @@ class Api::OrderDetailsController < ApplicationController
   respond_to :json
   rescue_from ActiveRecord::RecordNotFound, with: :order_detail_not_found
 
-  http_basic_authenticate_with name: Rails.application.secrets.api.basic_auth_name, password: Rails.application.secrets.api.basic_auth_password
+  http_basic_authenticate_with name: Rails.application.secrets.api["basic_auth_name"], password: Rails.application.secrets.api["basic_auth_password"]
 
   # Qualtrics requires that we use a query parameter rather than being part of
   # the URL.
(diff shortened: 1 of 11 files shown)


config/initializers/devise.rb
```diff
@@ -12,7 +12,7 @@
   # The secret key used by Devise. Devise uses this key to generate
   # random tokens. Changing this key will render invalid all existing
   # confirmation, reset password and unlock tokens in the database.
-  config.secret_key = Settings.devise.secret_key
+  config.secret_key = Rails.application.secrets.devise.secret_key
```
meara: This is clearly broken when the devise key does not exist, but I'm not sure what makes sense to do here. Should this key always exist?

app/controllers/api/order_details_controller.rb
```diff
@@ -3,7 +3,7 @@ class Api::OrderDetailsController < ApplicationController
   respond_to :json
   rescue_from ActiveRecord::RecordNotFound, with: :order_detail_not_found
 
-  http_basic_authenticate_with name: Settings.api.basic_auth_name, password: Settings.api.basic_auth_password
+  http_basic_authenticate_with name: Rails.application.secrets.api["basic_auth_name"], password: Rails.application.secrets.api["basic_auth_password"]
```
jhanggi: Just a warning, in Rails 5.1, the keys become symbols. http://guides.rubyonrails.org/upgrading_ruby_on_rails.html#application-secrets-now-loaded-with-all-keys-as-symbols

config/settings/stage.yml
```diff
@@ -0,0 +1,6 @@
+email:
+  from: "nucore-noreply@nucore.stage.tablexi.com"
+  fake:
+    enabled: true
+    to: ['jason@tablexi.com', "meara@tablexi.com"]
```
jhanggi: Let's use double quotes here for consistency.

app/controllers/api/order_details_controller.rb
meara: Ah okay, will make sure to update that when we move to 5.1.

PR #1469, opened 2018-03-30
Update to Rails 5.0

meara:
# Release Notes

TECH TASK: Update to Rails 5.0


config/initializers/filter_parameter_logging.rb
```diff
@@ -0,0 +1,4 @@
+# Be sure to restart your server when you modify this file.
+
+# Configure sensitive parameters which will be filtered from the log file.
+Rails.application.config.filter_parameters += [:password]
```
jhanggi: I think this can be removed from `application.rb` now.

app/models/facility.rb
```diff
@@ -60,7 +60,7 @@ def self.cross_facility
 
   def products(type = nil)
     if type
-      super.where(type: type)
+      self.public_send(type.to_s.pluralize.underscore.to_sym)
```
meara: There's probably a better way to handle this, would be happy to hear suggestions if you have one.

config/application.rb
```diff
@@ -37,6 +37,8 @@ class Application < Rails::Application
     config.active_record.observers = :order_detail_observer
     #:cacher, :garbage_collector, :forum_observer
 
+    config.active_record.belongs_to_required_by_default = false
```
jhanggi: I feel like this belongs in `config/initializers/new_framework_defaults.rb`, but we don't have that file, so I can't say for sure.

https://blog.bigbinary.com/2016/08/18/new-framework-defaults-in-rails-5-to-make-upgrade-easier.html

meara: Let's consider this when we go through the `bin/rails app:update` together. The first time I did it I may have missed some things.

config/initializers/filter_parameter_logging.rb
meara: Yup, done.

meara:
Code looks good to me!

Something that might be prudent to do before merging this is to try this branch out on each of the forks and see what needs fixing, since all the forks have at least one fork-specific engine that won't have been updated as part of the changes in open. Would be nice to have the kinks worked out there so we're ready when we pull the Rails 5 changes in from from open.

jhanggi:
I think we can remove the WIP label?

PR #1471, opened 2018-04-02
[#140361] refactor user creation/updating to use user form object (open)

meara:
# Release Notes
Refactor user creation and updating to use a form object, enabling custom validations etc on a per-school basis.

# Additional Context
Changes in open to enable requested Dartmouth features, see https://github.com/tablexi/nucore-dartmouth/pull/179 for additional context. Companion to https://github.com/tablexi/nucore-dartmouth/pull/181.

### DIFF
--- a/app/forms/user_form.rb
+++ b/app/forms/user_form.rb
@@ -1,0 +1,49 @@
+class UserForm < SimpleDelegator
+
+  include ActiveModel::Validations
+
+  def user
+    __getobj__
+  end
+
+  def valid?
+    success = [super, user.valid?].all?
+
+    user.errors.each do |k, error_messages|
+      errors.add(k, error_messages)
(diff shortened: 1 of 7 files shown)


app/forms/user_form.rb
```diff
@@ -0,0 +1,49 @@
...
+    false
+  end
+
+  private
+
+  def set_password
+    user.password = generate_new_password
+  end
+
+  def generate_new_password
```
jhanggi: I feel like this becomes a little awkward overriding this to return `nil` in DC, but I'm not sure I have a better solution. I was thinking [ActiveModel::Callbacks](http://api.rubyonrails.org/classes/ActiveModel/Callbacks.html), but that's probably overkill.

PR #1474, opened 2018-04-04
[#141023] Avoid error when submitting order form with no file

jhanggi:
# Release Notes

Prevent hard error when submitting an order form from the cart when the user does not specify a file.

# Original error

```
An ActionController::ParameterMissing occurred in order_detail_stored_files#upload_order_file:
param is missing or the value is empty: stored_file
  app/controllers/order_detail_stored_files_controller.rb:35:in `upload_order_file'
```

### DIFF
--- a/app/controllers/order_detail_stored_files_controller.rb
+++ b/app/controllers/order_detail_stored_files_controller.rb
@@ -32,7 +32,7 @@ class OrderDetailStoredFilesController < ApplicationController
 
   # POST /orders/:order_id/order_details/:order_detail_id/upload_order_file
   def upload_order_file
-    @file = @order_detail.stored_files.new(params.require(:stored_file).permit(:file))
+    @file = @order_detail.stored_files.new(params[:stored_file]&.permit(:file))
     @file.file_type  = "template_result"
     @file.name       = "Order File"
     @file.product = @order_detail.product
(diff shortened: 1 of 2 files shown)


app/controllers/order_detail_stored_files_controller.rb
```diff
@@ -32,7 +32,7 @@ def order_file
 
   # POST /orders/:order_id/order_details/:order_detail_id/upload_order_file
   def upload_order_file
-    @file = @order_detail.stored_files.new(params.require(:stored_file).permit(:file))
+    @file = @order_detail.stored_files.new(params[:stored_file]&.permit(:file))
```
meara: Do we need both the hash syntax and the `&.` in order to handle the possible nil? I would have thought require would fail more elegantly than that...

meara:
Also looks like CodeClimate is still grumpy...

app/controllers/order_detail_stored_files_controller.rb
jhanggi: `require` raises an error when `:stored_file` doesn't exist in the params. If you don't select a file, the browser doesn't send the param `stored_file[file]`, so we don't get a `stored_file` key.

Another option would be `.new(file: params.dig(:stored_file, :file))`, but that's not really a great pattern for when you need more params.

Or a `rescue`, but that's about it if we want to keep the `require`.

I'm not really sure of a better option.

meara: I think it's fine as is then. Thanks for the additional context!

PR #1477, opened 2018-04-05
rails 4 compatible prep for rails 5

meara:
# Release Notes

TECH TASK: Changes needed for Rails 5 that are compatible with Rails 4.

- Models inherit from ApplicationRecord
- Replace deprecated render nothing calls
- Replace method overriding an AR relation with scope
- Remove redundant calls to include ForbiddenAttributesProtection (now standard)
- Replace deprecated `uniq` calls with `distinct`
- Use ActionController::Parameters objects instead of hashes
- Replace deprecated `before_filter` with `before_action`

### DIFF
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -1,7 +1,5 @@
 class Facility < ApplicationRecord
 
-  include ActiveModel::ForbiddenAttributesProtection
-
   before_validation :set_journal_mask, on: :create
 
   has_many :items
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -56,14 +56,6 @@ class Facility < ApplicationRecord
       new(url_name: "all", name: "Cross-Facility", abbreviation: "ALL", is_active: true)
   end
 
-  def products(type = nil)
-    if type
-      self.public_send(type.to_s.pluralize.underscore.to_sym)
-    else
-      super
-    end
-  end
-
   def destroy
     # TODO: can you ever delete a facility? Currently no.
     # super
--- a/app/models/facility.rb
+++ b/app/models/facility.rb
@@ -1,4 +1,4 @@
-class Facility < ActiveRecord::Base
+class Facility < ApplicationRecord
 
   include ActiveModel::ForbiddenAttributesProtection
 
@@ -60,7 +60,7 @@ class Facility < ActiveRecord::Base
 
   def products(type = nil)
     if type
-      super.where(type: type)
+      self.public_send(type.to_s.pluralize.underscore.to_sym)
     else
       super
     end
--- a/app/services/order_details/reconciler.rb
+++ b/app/services/order_details/reconciler.rb
@@ -11,7 +11,7 @@ module OrderDetails
     validate :all_journals_and_statements_must_be_before_reconciliation_date
 
     def initialize(order_detail_scope, params, reconciled_at)
-      @params = params
+      @params = params || ActionController::Parameters.new
       @order_details = order_detail_scope.readonly(false).find_ids(to_be_reconciled.keys)
       @reconciled_at = reconciled_at
     end
--- a/app/services/order_details/reconciler.rb
+++ b/app/services/order_details/reconciler.rb
@@ -37,7 +37,7 @@ module OrderDetails
     # The params hash comes in with the unchecked IDs as well. Filter out to only
     # those we're going to reconcile. Returns an array of IDs.
     def to_be_reconciled
-      Hash(@params).select { |_order_detail_id, params| params[:reconciled] == "1" }
+      @params.select { |_order_detail_id, params| params[:reconciled] == "1" }
     end
 
     def reconcile(order_detail, params)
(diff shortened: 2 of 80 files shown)


app/models/facility.rb
```diff
@@ -60,7 +60,7 @@ def self.cross_facility
 
   def products(type = nil)
     if type
-      super.where(type: type)
+      self.public_send(type.to_s.pluralize.underscore.to_sym)
```
meara: Rails 5 really didn't like the previous way this method was structured. This refactoring is a little clunky--let me know if you have a better idea.

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/events_controller.rb
```diff
@@ -10,7 +10,7 @@ def create
         ),
       )
 
-      render nothing: true
+      render head :ok
```
meara: This line is causing test failures that we're double-rendering--I'm not quite following why. Nothing else seems to be calling render in this action. Any insight?

app/views/shared/transactions/_search2.html.haml
```diff
@@ -1,6 +1,6 @@
 -# This view, replaces _search, which relies on the TransactionSearch controller/concern.
 -# This view, along with TransactionSearch::Searcher is the more modern method. See FacilityAccountsReconciliationController
-= simple_form_for @search_form, url: url_for(params), method: :get, html: { class: "search_form" }, defaults: { required: false } do |f|
+= simple_form_for @search_form, url: url_for(params.permit(:search)), method: :get, html: { class: "search_form" }, defaults: { required: false } do |f|
```
jhanggi: I'm surprised this works without additional `permit(:product_id, ...)`, but hey, if it works, awesome.

vendor/engines/secure_rooms/app/models/secure_rooms/facility_extension.rb
```diff
@@ -0,0 +1,13 @@
+module SecureRooms
+
+  module FacilityExtension
+
+    extend ActiveSupport::Concern
+
+    included do
+      has_many :secure_rooms
```
jhanggi: Why did this work before without this relation?

app/services/order_details/reconciler.rb
```diff
@@ -37,7 +37,7 @@ def full_errors
     # The params hash comes in with the unchecked IDs as well. Filter out to only
     # those we're going to reconcile. Returns an array of IDs.
     def to_be_reconciled
-      Hash(@params).select { |_order_detail_id, params| params[:reconciled] == "1" }
+      @params.select { |_order_detail_id, params| params[:reconciled] == "1" }
```
jhanggi: Is there any case where `@params` might be nil? I think that's what the `Hash()` was mitigating.

app/models/facility.rb
jhanggi: I'll need to think about this. I think I ran into something similar with an association pointing at an STI table on another project. I need to remember how I handled that. Might also be worth looking into how this method is used.

app/models/facility.rb
```diff
@@ -1,4 +1,4 @@
-class Facility < ActiveRecord::Base
+class Facility < ApplicationRecord
 
   include ActiveModel::ForbiddenAttributesProtection
```
jhanggi: Are we ready to kill these `include`s out? I think it's added by default now.

app/models/facility.rb
jhanggi: Now that I see what this is doing, it's definitely something hacky. I would vote for killing this method off and instead replacing usages of it with a new scope on Product. 

```ruby
# current
current_facility.products("Instrument").find_by!(url_name: params[id_param])

# proposed
current_facility.products.of_type("Instrument").find_by!(url_name: params[id_param])
```

vendor/engines/secure_rooms/app/controllers/secure_rooms_api/events_controller.rb
jhanggi: I'm not sure why you're not seeing this elsewhere, but it's just `head :ok`. No `render`.

app/services/order_details/reconciler.rb
meara: Handling that above in the initializer.

app/views/shared/transactions/_search2.html.haml
```diff
@@ -1,6 +1,6 @@
 -# This view, replaces _search, which relies on the TransactionSearch controller/concern.
 -# This view, along with TransactionSearch::Searcher is the more modern method. See FacilityAccountsReconciliationController
-= simple_form_for @search_form, url: url_for(params), method: :get, html: { class: "search_form" }, defaults: { required: false } do |f|
+= simple_form_for @search_form, url: url_for(params.permit!), method: :get, html: { class: "search_form" }, defaults: { required: false } do |f|
```
meara: So, turns out `permit(:search)` did not in fact work. Are you opposed to `permit!` here? It seems reasonably safe, since we're effectively filtering out only the search params for which we already have a searcher defined on whichever particular controller action.

jhanggi: I was just trying it out. It looks like we might be able to get away with just `url: url_for,`. The empty `url_for` will link to the current path with controller/action. I'm pretty sure whatever query params are included in this URL are lost anyways.

I know we now maintain the filter params in the sort headers (#1466), but if we want to maintain the sort after filtering, then I think some hidden fields are probably the way to go.

Speaking of, how does the change in #1466 not get flagged by Rails 5?

meara: I think #1466 might actually get flagged by Rails 5, but I figure we'll cross that bridge when we get there.

PR #1478, opened 2018-04-06
[#140407] Reservation calendar drag and drop

jhanggi:
# Release Notes

Adds a block onto the calendar during reservation creation based off the default time. Drag and dropping the block will update the normal date/time selectors, and updating the selectors will redraw the block.

# Screenshot

screen shot 2018-04-13 at 6 21 40 pm

# Features and Caveats

Includes an upgrade to FullCalendar.

* Feature flagged: _off_ by default
* Available on reservation new and edit
* Drag and drop
* Click on a blank space, and it will move the reservation to that point, maintaining previous duration
* You cannot drag and drop an overlapping reservation
* You cannot change the duration to overlap with another reservation
* On edit view, will allow you to change the duration of an ongoing reservation, but not move it
* On edit view, you can move a reservation if it has not started

Caveats:

* You can cause an overlapping reservation by modifying the form. This may cause some weirdness, but you'll get an error if you submit it
* Drag and drop in 30 minute intervals, if you want more precise than that, you'll need to use the form
* If the default reservation starts at 11:35, the drag & drop will move it in 30 minute blocks, e.g. 12:05, 12:35.


jhanggi:
Notes: we want to add click to move, i.e. when a user clicks on an empty space, we should move the reservation to that point.

Gemfile
```diff
@@ -52,6 +52,7 @@ gem "text_helpers"
 gem "chosen-rails"
 gem "fine_uploader", path: "vendor/engines/fine_uploader"
 gem "rubyzip"
+gem "momentjs-rails"
```
meara: What are we using this gem for?

app/assets/stylesheets/application.scss
```diff
@@ -30,7 +30,7 @@ $baseFontSize: 12px;
 
 @import "fine-uploader/fine-uploader-new";
 @import "fine-uploader-override";
-@import "calendar/fullcalendar";
+@import "fullcalendar";
 @import "calendar/fullcalendar.extensions";
```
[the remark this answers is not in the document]
jhanggi: Yes, it's our custom styling. I am hoping we can clean up that file, though.

Gemfile
jhanggi: It's a dependency of the new version of FullCalendar. Probably worth a comment around it. I was hoping to use a gem version of full calendar, but the existing one out there is out of date (https://github.com/bokmann/fullcalendar-rails). I might switch to using the rails-assets version, though, though I doubt that has the moment dependency.

jhanggi: There is a PR in that gem for the upgrade to 3.9, but it hasn't been merged, but we could track to it.

jhanggi:
@meara this could probably use another look, both at the code and how it works. It should support both new and edit on the user facing side.

app/assets/javascripts/app/reservation_calendar.js
```diff
@@ -0,0 +1,68 @@
...
+  renderCurrentEvent: function(start, end) {
+    if (this.currentEvent) {
+      this.$calendar.fullCalendar("removeEvents", [this.currentEvent.id]);
+    }
+    this.currentEvent = {
+      id: "_currentEvent",
+      title: "My Reservation",
+      start: start,
+      end: end,
+      color: '#378006',
```
meara: What about a color a little more like #207065?

Gemfile
```diff
@@ -51,6 +51,7 @@ gem "nested_form_fields"
 gem "text_helpers"
 gem "chosen-rails"
 gem "fine_uploader", path: "vendor/engines/fine_uploader"
+gem "fullcalendar", path: "vendor/engines/fullcalendar"
```
meara: I like the idea of using an engine to wrap the assets.

app/assets/javascripts/app/reservation_calendar.js
```diff
@@ -0,0 +1,68 @@
...
+   this.$calendar.fullCalendar('renderEvent', this.currentEvent, true);
+  },
+
+  // When in edit mode,this will remove the current reservation from the JSON
+  // source so we can replace it with our currentEvent.
+  _removeSelfFromSource: function() {
+    self.$calendar.fullCalendar("removeEvents", [self.id]);
+  },
+
+  _addReservationFormListerer: function() {
```
meara: Think you mean `Listener`

meara:
Also looks like the specs need an update.

meara:
This seems like not the most user-friendly date/time formatting--is this what we usually use?

screen shot 2018-04-12 at 4 00 33 pm

jhanggi:
@meara I think I fixed the specs. 

I ended up removing the "now" indicator because it caused problems with timezone. If I changed the timezone to eastern (a la Dartmouth), it rendered everything as it should (9-5 is still 9-5). And it's 6:30 Chicago, so the default time was 7:30 (correct), but the red line was showing up at 6:30. We didn't have it before, so I don't think it'll be missed.

meara:
Oh also, can you update the screenshot in the description?

PR #1479, opened 2018-04-06
Fix migration from scratch caused by paranoia

jhanggi:
# Release Notes

Tech task: Migration from scratch became broken after we introduced Paranoia.

# Additional Context

Errors because `deleted_at` column does not exist on earlier migrations of `Reservation`

### DIFF
--- a/db/migrate/20160614185348_add_type_to_reservations.rb
+++ b/db/migrate/20160614185348_add_type_to_reservations.rb
@@ -1,9 +1,12 @@
 class AddTypeToReservations < ActiveRecord::Migration
 
+  class Reservation < ActiveRecord::Base
+  end
+
   def up
     add_column :reservations, :type, :string
     change_column :reservations, :reserve_end_at, :datetime, null: true
-    admin_reservations = Reservation.where(order_detail: nil)
+    admin_reservations = Reservation.where(order_detail_id: nil)
     admin_reservations.where(reserve_end_at: nil).update_all(type: "OfflineReservation")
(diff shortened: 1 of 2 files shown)


meara:
@jhanggi I don't follow exactly what the problem was or why these changes fix it. Can you explain further?

jhanggi:
Sure. Before this change, the existing `Reservation` class gets loaded and used. That class now has `acts_as_paranoid` on it which adds a default scope of `where(deleted_at: nil)`. So if you run migrations from scratch, when it gets to this migration and it tries to run `Reservation.where(...).each`, default scope is applied, but the `deleted_at` column does not yet exist. That doesn't happen until a later migration.

What this change does is creates a new `AddTypeToReservations::Reservation` (namespaced to the migration class) class that doesn't have any baggage (but still ties to the same table as `::Reservation`), and that's what gets used inside the migration.

It's generally a bad idea to rely on application classes in our migrations since they might get deleted or renamed. However, sometimes the likelihood of that happens is low (like in this case: would we really ever delete something so foundational to the app?) and you're not relying on any scopes or associations (which could also disappear), so we just use the AR model. But, turns out, default scopes can mess with that.

db/migrate/20160614185348_add_type_to_reservations.rb
```diff
@@ -1,9 +1,12 @@
 class AddTypeToReservations < ActiveRecord::Migration
 
+  class Reservation < ActiveRecord::Base
```
meara: Can we add a comment here and below that this class is being added to avoid issues with the default scope from paranoia?

PR #1481, opened 2018-04-06
[#73287] Add ability to suspend a user

jhanggi:
# Release Notes

Adds the ability for a global administrator to suspend a user.

Suspended users are:
* no longer able to log into NUcore
* kicked out of their session if they have an active one
* tagged in most places with "(SUSPENDED)"
* excluded from bulk email searches
* unable to be ordered on behalf of

# Screenshot

screen shot 2018-04-02 at 1 50 35 pm

screen shot 2018-04-02 at 2 09 01 pm

# Additional Context

The support for suspended users was originally added in #534 and #538, but never had an interface before.

I tried to exclude the "(SUSPENDED)" only when it seemed like it wouldn't make sense (e.g. journal rows and statements), but opted to show it most places because it seemed less confusing to see SUSPENDED and understand what it means than to not see it and wonder why they're still active.

Originally, I planned on updating the searches to exclude suspended user in situations like adding a new payment source or adding to a product access list, but ultimately decided it would be less confusing to see the result with a "SUSPENDED" tag than to not see the tag at all. Instead I opted to disable the link in those cases.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -19,7 +19,7 @@ class UsersController < ApplicationController
   before_action :check_acting_as
 
   load_and_authorize_resource except: [:create, :password, :password_reset, :edit, :update, :show], id_param: :user_id
-  load_and_authorize_resource only: [:edit, :update, :show], id_param: :id
+  load_and_authorize_resource only: [:edit, :update, :show, :suspend, :activate], id_param: :id
 
   layout "two_column"
 
@@ -129,6 +129,17 @@ class UsersController < ApplicationController
       flash[:error] = text("update.error", message: @user_form.errors.full_messages.to_sentence)
       render action: "edit"
     end
+  end
+
+  def suspend
+    @user.suspended_at ||= Time.current
+    @user.save!
+    redirect_to facility_user_path(current_facility, @user), notice: "User suspended"
+  end
+
+  def activate
+    @user.update!(suspended_at: nil)
+    redirect_to facility_user_path(current_facility, @user), notice: "User activated"
   end
 
   private
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -19,7 +19,7 @@ class UsersController < ApplicationController
   before_action :check_acting_as
 
   load_and_authorize_resource except: [:create, :password, :password_reset, :edit, :update, :show], id_param: :user_id
-  load_and_authorize_resource only: [:edit, :update, :show, :suspend, :activate], id_param: :id
+  load_and_authorize_resource only: [:edit, :update, :show, :suspend, :unsuspend], id_param: :id
 
   layout "two_column"
 
@@ -137,9 +137,9 @@ class UsersController < ApplicationController
     redirect_to facility_user_path(current_facility, @user), notice: "User suspended"
   end
 
-  def activate
+  def unsuspend
     @user.update!(suspended_at: nil)
-    redirect_to facility_user_path(current_facility, @user), notice: "User activated"
+    redirect_to facility_user_path(current_facility, @user), notice: "User re-activated"
   end
 
   private
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -718,8 +718,7 @@ class OrderDetail < ActiveRecord::Base
   end
 
   def long_description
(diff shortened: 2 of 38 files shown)


app/models/order_detail.rb
```diff
@@ -718,8 +718,7 @@ def description
   end
 
   def long_description
-    "##{self}: #{order.user}: #{I18n.l(fulfilled_at.to_date, format: :usa)}: "\
-    "#{product} x#{quantity}"
+    OrderDetailJournalDescriptionPresenter.new(self).long_description
```
meara: It seems a little overkill to me to create a presenter just for one method--can you explain why you made that choice? Do you think other methods will eventually be added on that presenter?

app/models/user.rb
```diff
@@ -31,6 +31,9 @@ class User < ActiveRecord::Base
   # Gem ldap_authenticatable expects User to respond_to? :login. For us that's #username.
   alias_attribute :login, :username
 
+  # TODO: Update downstreams to use `suspended_at`
```
meara: What is the reason we are leaving this as a todo rather than including that change in this PR?

app/presenters/users/name_presenter.rb
```diff
@@ -0,0 +1,49 @@
...
+
+    def username_label?
+      @username_label
+    end
+
+    def suspended_label?
+      @suspended_label
+    end
+
+    def append_suspended_string(string)
```
meara: I don't see where this method gets used...

spec/features/admin/suspending_a_user_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "rails_helper"
+
+RSpec.describe "User deletion" do
```
meara: This should be "User suspension", not "User deletion", no?

app/controllers/users_controller.rb
```diff
@@ -131,6 +131,17 @@ def update
     end
   end
 
+  def suspend
+    @user.suspended_at ||= Time.current
+    @user.save!
+    redirect_to facility_user_path(current_facility, @user), notice: "User suspended"
+  end
+
+  def activate
```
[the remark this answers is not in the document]
jhanggi: You're right. I went this way thinking that it was the terminology we used for accounts. Turns out, it's just the button that's labeled "Activate". The controller action is `unsuspend`. I'll go that way.

app/models/user.rb
jhanggi: UIC uses `deactivated_at` in some of its synchronization jobs. We'll want to change that once this merges down, but this provides a way to not break in the meantime.

app/models/order_detail.rb
jhanggi: `long_description` is only used in journals. It's not really intrinsic to the OrderDetail. I would have gotten rid of `OrderDetail#long_description` and used the presenter in the journal, but the downstream forks's engines also rely on this method. So this was kind of a middle ground.

app/presenters/users/name_presenter.rb
jhanggi: Oops. Looks like it's left over from a refactoring.

app/models/user.rb
```diff
@@ -142,56 +146,33 @@ def administered_order_details
...
     super && active?
   end
 
-  def deactivate
-    update_attribute(:deactivated_at, deactivated_at || Time.current)
-  end
-
-  def activate
-    update_attribute(:deactivated_at, nil)
+  def active?
```
meara: Should this and `self.active` below be renamed as `unsuspended` as well? Also, how come we're using a class method below and not a scope?

jhanggi: I actually think `active` is accurate here and is definitely less clunky than `unsuspended` (and semantically, I wouldn't consider something unsuspended unless it had been first suspended). It definitely is more of an meaningful term when being used to determine if you are allowed to log in or not.

It could become broader at some point, but right now, it essentially means not suspended.

As far as the class method, that's how we used to do a lot of our scopes. We can definitely make it a scope, though.

jhanggi: We mix and match. In User, it seems like we always use the class method. I could go either way.

meara: Okay, if we're viewing this as a broader term that may eventually be expanded, I think this is fine then.

PR #1486, opened 2018-04-10
remove ordering clause that causes oracle compatibility issue

meara:
# Release Notes

Remove order clause from query that causes an oracle compatibility issue.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -171,7 +171,6 @@ class OrderDetail < ActiveRecord::Base
     where("dispute_at IS NOT NULL")
       .where(dispute_resolved_at: nil)
       .where("order_details.state != ?", "canceled")
-      .order("dispute_at")
   end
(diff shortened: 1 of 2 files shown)


app/models/order_detail.rb
```diff
@@ -171,7 +171,6 @@ def self.in_dispute
     where("dispute_at IS NOT NULL")
       .where(dispute_resolved_at: nil)
       .where("order_details.state != ?", "canceled")
-      .order("dispute_at")
```
jhanggi: Does this ordering still get applied where it's needed (i.e. the controller)?

meara: Added that back in.

jhanggi:
Unclear if the CI failure is legit or not. I would suspect, not.

meara:
That test isn't failing locally so I think should be fine on rebuild.

PR #1487, opened 2018-04-10
Add restart_grace for unicorn

jhanggi:
# Release Notes

TECH TASK: Prevent `eye restart unicorn` from thinking unicorn has crashed

This change has been made on txi stage and NU servers.

## Pulled up from

https://github.com/tablexi/nucore-nu/pull/361/files


meara:
@jhanggi do we need to update this on Dartmouth and UIC as well?

jhanggi:
I've updated all the servers.

PR #1494, opened 2018-04-12
[#141446] Fix display issue with "X" in chosen selectors

jhanggi:
# Release Notes

Fix "X" display issue on multi-select filters.

# Screenshot
screen shot 2018-04-05 at 11 01 18 am

# Additional Context

Hopefully, this is a temporary fix until the next release of chosen-rails. cf https://github.com/tsechingho/chosen-rails/issues/97

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -4,6 +4,10 @@ $baseFontSize: 12px;
 @import "bootstrap";
 @import "bootstrap-responsive";
 @import "font-awesome";
+// Temporary fix until the chosen gem is fixed
+// https://github.com/tsechingho/chosen-rails/issues/97
+$chosen-sprite: image-url("chosen-sprite.png");
+$chosen-sprite-retina: image-url("chosen-sprite@2x.png");
 @import "chosen";
 
 @import "nucore-variables";


app/assets/stylesheets/application.scss
```diff
@@ -4,6 +4,10 @@ $baseFontSize: 12px;
 @import "bootstrap";
 @import "bootstrap-responsive";
 @import "font-awesome";
+// Temporary fix until the chosen gem is fixed
+// https://github.com/tsechingho/chosen-rails/issues/97
+$chosen-sprite: image-url("chosen-sprite.png");
```
meara: Huh, is this really the best place for this, or are we okay with putting this here because it's intended as strictly temporary?

jhanggi: Yeah, I put it here because it's meant to be temporary, and a fairly loud place to say: hey, check if this is fixed yet; if so get rid of me. If it were meant to be permanent I would find a better home for it.

meara:
Oh btw, did you mean to attach a screenshot? Looks like it's missing from the PR description.

jhanggi:
Fixed the screenshot.

PR #1495, opened 2018-04-13
[#140404] Replace filtering in journals/statements tabs

meara:
# Release Notes

Replace filtering code in journals/statements tabs with newer, more maintainable standard.

### DIFF
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -7,12 +7,13 @@ class FacilityJournalsController < ApplicationController
   before_action :authenticate_user!
   before_action :check_acting_as
   before_action :check_billing_access
-  before_action :init_journals, except: :create_with_search
-  helper_method :has_pending_journals?
+  before_action :init_journals, except: :create
 
   include TransactionSearch
 
-  layout "two_column"
+  layout lambda {
+    action_name.in?(%w(new)) ? "two_column_head" : "two_column"
+  }
 
   def initialize
     @subnav     = "billing_nav"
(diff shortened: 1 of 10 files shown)


app/controllers/facility_journals_controller.rb
```diff
@@ -7,12 +7,13 @@ class FacilityJournalsController < ApplicationController
...
   before_action :check_acting_as
   before_action :check_billing_access
-  before_action :init_journals, except: :create_with_search
-  helper_method :has_pending_journals?
+  before_action :init_journals, except: :create
 
   include TransactionSearch
 
-  layout "two_column"
+  layout lambda {
```
jhanggi: I really wish there was a nicer way to write this 😞

app/controllers/facility_statements_controller.rb
```diff
@@ -23,11 +25,15 @@ def index
   end
 
   # GET /facilities/:facility_id/statements/new
-  def new_with_search
-    @order_details = @order_details.need_statement(@facility)
+  def new
+    order_details = OrderDetail.need_statement(@facility)
     @order_detail_action = :create
-    set_default_start_date if SettingsHelper.feature_on?(:set_statement_search_start_date)
```
jhanggi: Can the `set_default_start_date` method go away now?

app/controllers/facility_journals_controller.rb
```diff
@@ -26,23 +27,31 @@ def index
...
+      @action_date_field = { journal_date: @earliest_journal_date }
+    end
 
-      respond_to do |format|
-        format.csv do
-          # used for "Export as CSV" link for order details with expired accounts
-          @order_details = @invalid_order_details
-        end
-        format.any {}
+    @valid_order_details, @invalid_order_details = ValidatorFactory.partition_valid_order_details(@order_details.unexpired_account)
```
jhanggi: Wow, we're doing a lot in this action. Would there be in value in extracting some smaller methods?

spec/features/admin/facility_journals_search_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  let(:order_details) { orders.map(&:order_details).flatten }
+
+  before do
+    order_details.each do |detail|
+      detail.update(reviewed_at: 1.day.ago)
+    end
+  end
+
+  it "can do a basic filter" do
```
jhanggi: Would it be worth adding a spec around invalid order details or pending journals? Or do we think the existing controller specs will cover that?

meara: I think the existing controller specs are fairly thorough there.

app/controllers/facility_journals_controller.rb
meara: I'm all ears if you have a suggestion--I borrowed that from a few other controllers where we have a similar pattern.

app/controllers/facility_journals_controller.rb
meara: I agree there's a lot going on here. I'm a little leery of pulling out smaller methods though, because a lot of what's happening here is setting instance variables that are used in the various views, and I think if we hide those in other methods the code becomes a lot more confusing. I'm willing to be convinced otherwise though.

PR #1497, opened 2018-04-13
SAML Single Logout

jhanggi:
# Release Notes

Support SAML single logout. When you are logged in via SAML, when you click the "Logout" button, you will also be logged out of your identity provider. This will be critical for security on shared computers.

# Additional Context

This involved a little more hackery into the devise_saml_authenticatable gem than I really wanted to do. I might open some PRs or issues around that once we know everything is working.

One difference between OneLogin (where I was developing against) and Dartmouth's IDP is that DC requires as part of the logout request that `NameID` matches the logged in user's netid. This was not supported out of the box by the gem, so I had to hack that in. There's one setting change that also needs to happen for DC, but it'll be easy because of UConn's changes: `driver.name_identifier_format: "urn:oasis:names:tc:SAML:1.1:nameid-format:unspecified"`

I also wasn't quite sure the best way to handle the different paths needed for the logout link, so I'm open to suggestions there.

### DIFF
--- a/app/views/shared/_header.html.haml
+++ b/app/views/shared/_header.html.haml
@@ -32,8 +32,7 @@
               %li.visible-with-nav= link_to t('pages.cart') + " (#{current_cart.order_details.count})", :cart
               %li.divider-vertical
               = render "shared/message_summary"
-              %li= link_to t('pages.logout'), sign_out_user_path
-              %li= link_to "SSO Logout", destroy_saml_user_session_path
+              %li= render "shared/logout/link_#{warden.session(:user)[:strategy] || 'default'}"
               - if responsive? && !acting_as? && current_user
                 = form_tag global_search_path, class: "navbar-search pull-right hidden-with-nav" do
                   = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2"
--- a/app/views/shared/_header.html.haml
+++ b/app/views/shared/_header.html.haml
@@ -33,6 +33,7 @@
               %li.divider-vertical
               = render "shared/message_summary"
               %li= link_to t('pages.logout'), sign_out_user_path
+              %li= link_to "SSO Logout", destroy_saml_user_session_path
               - if responsive? && !acting_as? && current_user
                 = form_tag global_search_path, class: "navbar-search pull-right hidden-with-nav" do
                   = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2"
--- a/vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
+++ b/vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
@@ -4,7 +4,8 @@ module SamlAuthentication
 
     # Some SSO providers are strict and require that the name_identifier matches
     # what's logged in on their system for SLO.
-    before_action(only: :destroy) { @signed_out_username = current_user.public_send(Devise.saml_default_user_key) }
+    before_action :store_logging_out_username, only: :destroy
+    after_action :store_winning_strategy, only: :create
 
     protected
 
@@ -19,6 +20,18 @@ module SamlAuthentication
       config = saml_config(idp_entity_id).dup
       config.name_identifier_value = @signed_out_username
       request.create(config)
+    end
+
+    private
+
+    def store_logging_out_username
+      @signed_out_username = current_user.public_send(Devise.saml_default_user_key)
+    end
+
+    # This will be used to choose which Logout link is shown to the user.
+    # https://github.com/apokalipto/devise_saml_authenticatable/wiki/Supporting-multiple-authentication-strategies
+    def store_winning_strategy
+      warden.session(resource_name)[:strategy] = warden.winning_strategies[resource_name].class.name.demodulize.underscore.to_sym
     end
 
   end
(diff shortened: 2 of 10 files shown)


app/views/shared/_header.html.haml
```diff
@@ -32,7 +32,7 @@
               %li.visible-with-nav= link_to t('pages.cart') + " (#{current_cart.order_details.count})", :cart
               %li.divider-vertical
               = render "shared/message_summary"
-              %li= link_to t('pages.logout'), sign_out_user_path
+              %li= render "shared/logout/link_#{warden.session(:user)[:strategy] || 'default'}"
```
meara: This seems like a place where I might actually prefer a view helper rather than having this or statement directly in the view.

vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
```diff
@@ -2,13 +2,38 @@ module SamlAuthentication
...
+    end
+
+    private
+
+    def store_logging_out_username
+      @signed_out_username = current_user.public_send(Devise.saml_default_user_key)
+    end
+
+    # This will be used to choose which Logout link is shown to the user.
+    # https://github.com/apokalipto/devise_saml_authenticatable/wiki/Supporting-multiple-authentication-strategies
```
meara: Ew. Thank you for commenting this and providing the link to the wiki.

vendor/engines/saml_authentication/lib/saml_authentication/engine.rb
```diff
@@ -14,6 +14,8 @@ class Engine < Rails::Engine
       ViewHook.add_hook "devise.sessions.new",
                         "before_login_form",
                         "saml_authentication/sessions/new"
+
+      OneLogin::RubySaml::Logging.logger.level = Logger::DEBUG
```
meara: I assume we won't want this here once this PR is no longer WIP.

vendor/engines/saml_authentication/lib/saml_authentication/idp_entity_id_reader.rb
```diff
@@ -0,0 +1,40 @@
+module SamlAuthentication
+
+  # The default IdpEntityIdReader triggers an error on SLO because that request to
```
meara: I'm not entirely following what the error is that we're trying to handle here--might be good to walk through it together.

vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
```diff
@@ -2,13 +2,38 @@ module SamlAuthentication
...
     protected
 
     # Remove recall, so failures redirect back to sign_in page
     def auth_options
       { scope: resource_name }
     end
 
+    def after_sign_out_path_for(_)
+      idp_entity_id = get_idp_entity_id(params)
+      request = OneLogin::RubySaml::Logoutrequest.new
```
meara: It's weird to me that we're referencing OneLogin directly here--shouldn't we be getting the SAML provider from settings or something?

vendor/engines/saml_authentication/db/migrate/20180413194031_add_saml_session_index_to_user.rb
```diff
@@ -0,0 +1,6 @@
+class AddSamlSessionIndexToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :saml_session_index, :string
```
meara: I'm confused by what we need this for...

app/views/shared/_header.html.haml
jhanggi: View helper is a good call.

vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
jhanggi: We're using `devise_saml_authenticatable` which provides the devise adapter. Under the hood, it uses https://github.com/onelogin/ruby-saml which is built by OneLogin. And as opposed to most gems, they've namespaced all their classes with `OneLogin`. So it's not really a reference to OneLogin as a provider, it's just part of the class name.

This particular method is an override of the devise_saml_authenticatable's session controller.

vendor/engines/saml_authentication/db/migrate/20180413194031_add_saml_session_index_to_user.rb
jhanggi: I don't exactly understand the details, but according to https://github.com/apokalipto/devise_saml_authenticatable "This stores the session index defined by the IDP during login.  If provided it will be used as a salt for the user's session to facilitate an IDP initiated logout request." It kind of seems like it's not fully fleshed out (they don't even tell you it needs to be a database column), but I think the idea is if you initiate the logout from the IdP, this is used to help log you out.

I might play around a little more with it to see if we can get away without it. My suspicion is that we would need to do some more overriding of the devise saml gem.

vendor/engines/saml_authentication/lib/saml_authentication/engine.rb
jhanggi: Yeah, we won't, but it's been super helpful in development.

PR #1502, opened 2018-04-16
[#141487] replace filtering for my transactions in review

meara:
# Release Notes

Replace filtering code with newer paradigm for the My Transactions/In Review page.

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -38,9 +38,21 @@ class TransactionsController < ApplicationController
     end
   end
 
-  def in_review_with_search
+  def in_review
     @recently_reviewed = current_user.administered_order_details.recently_reviewed.paginate(page: params[:page])
-    @order_details = current_user.administered_order_details.in_review
+    order_details = current_user.administered_order_details.in_review
+
+    @search_form = TransactionSearch::SearchForm.new(params[:search])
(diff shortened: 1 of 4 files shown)


app/controllers/transactions_controller.rb
```diff
@@ -38,9 +38,21 @@ def index
...
   end
 
-  def in_review_with_search
+  def in_review
     @recently_reviewed = current_user.administered_order_details.recently_reviewed.paginate(page: params[:page])
-    @order_details = current_user.administered_order_details.in_review
+    order_details = current_user.administered_order_details.in_review
+
+    @search_form = TransactionSearch::SearchForm.new(params[:search])
+    @search = TransactionSearch::Searcher.new(TransactionSearch::FacilitySearcher,
```
jhanggi: Do we use this exact list elsewhere? I'm just wondering if it makes sense to wrap up some of the common filter lists like we have the default, so we could do something like `TransactionSearch::Searcher.new(:my_transactions)` or a `TransactionSearch::Searcher.my_transactions` class method and have some sort of mapping of keys to the the list? Just a thought; you probably have a better sense of where we're using custom lists. I wouldn't have a problem punting on it for now, either.

meara: I vote we punt for now, the degree of duplication isn't large enough to make this a big concern IMO.

PR #1504, opened 2018-04-17
Fix missing translations and enable raising in errors in test/CI

### LINKED ISSUE #1496: Missing Translation Strings for Partials
**Problem**: When trying to edit a reconcilable order, the edit order dialog never loads due to a missing translation string.

screen shot 2018-04-13 at 6 41 16 pm

I believe it's because the partial view isn't aware of the current `order_management.order_details.edit.remove_from_journal` translation string. The quick fix would be to write one for the `shared_details` partial view, but could possibly be overlooked when refactoring. I'm not sure how to have the partial view aware of this existing translation string, so I don't have a better suggestion.

jhanggi:
# Release Notes

Fix missing translation errors in development. Tech task: raise missing translation errors in test.

# Additional Context

By turning on `raise_on_missing_translation` in test, we'll hopefully catch these issues sooner.

Recharge chart string issue was ntroduced in #1493 because I thought this was duplicated, but turns
out this is used by `t("#{account_class.name.underscore}.account_fields.label.account_number
.#{section}”),`

User activate was a typo in #1481

"Remove from journal" has been around for a while. It's never been a problem in production because `raise_on_missing_translation` is set to false on stage and prod, so it just falls back to the titleized version, "Remove From Journal", which is what the translation was anyways.

Resolves #1496

### DIFF
--- a/config/locales/en.account_components.yml
+++ b/config/locales/en.account_components.yml
@@ -1,0 +1,16 @@
+# Labels for account components. These are often overridden/added to in the schools
+# to match their chart string fields (e.g. Fund, Org, Department, etc.)
+
+en:
+  nufs_account:
+    account_fields:
+      label:
+        account_number:
+          account_number: Account Number
+
+  facility_account:
+    # Components of the account
+    account_fields:
+      label:
+        account_number:
+          account_number: Account Number
--- a/app/views/users/show.html.haml
+++ b/app/views/users/show.html.haml
@@ -26,5 +26,5 @@
       - if @user.active?
         %li= link_to t("shared.suspend"), suspend_facility_user_path(current_facility, @user), class: "btn", method: :patch
       - else
-        %li= link_to t("share.activate"), unsuspend_facility_user_path(current_facility, @user), class: "btn", method: :patch
+        %li= link_to t("shared.activate"), unsuspend_facility_user_path(current_facility, @user), class: "btn", method: :patch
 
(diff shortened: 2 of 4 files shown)


config/locales/en.account_components.yml
```diff
@@ -0,0 +1,16 @@
+# Labels for account components. These are often overridden/added to in the schools
+# to match their chart string fields (e.g. Fund, Org, Department, etc.)
```
meara: Seems weird this is in its own super specific yaml file--what's the rationale there?

config/locales/en.account_components.yml
```diff
@@ -0,0 +1,16 @@
...
+      label:
+        account_number:
+          account_number: Account Number
+
+  facility_account:
+    # Components of the account
+    account_fields:
+      label:
+        account_number:
+          account_number: Account Number
```
meara: Should this be Payment Source Number to match where you moved it from, or is that an intentional change?

config/locales/en.account_components.yml
jhanggi: Over time, I've been trying to find a better organization for the locales, and a lot of that is shortening up the base `en.yml`. One of the things I wanted to do was consolidate these two pieces since they're very related. There might be a better place, but this was the best I came up with.

config/locales/en.account_components.yml
jhanggi: The "Payment Source" string wasn't actually being used prior to the change that broke this, it was using "Account Number". It's not really that important because these components are only used in open since NU and DC both use their own chart string components.

PR #1505, opened 2018-04-17
[#140406] Remove old filtering code

meara:
# Release Notes

Remove old-style filtering code that has now been generally replaced throughout the codebase.


app/inputs/transaction_chosen_old_input.rb
```diff
@@ -0,0 +1,63 @@
+# This class is deprecated in favor of TransactionChosenInput. This class
+# is still being used by BulkEmailController for searching.
+class TransactionChosenOldInput < SimpleForm::Inputs::Base # CollectionSelectInput
```
jhanggi: I think we can switch this out when we do some other cleanup to the new views (like getting rid of extraneous instance variables).

PR #1506, opened 2018-04-17
SAML Single Logout take 2

jhanggi:
# Release Notes

Support Single Logout for SAML. When you are signed into NUcore via SAML, when you log out, you are also logged out of the SSO/IdP provider as well as any other applications you were signed into as part of the same session.

# Additional Context

This is a second take at #1497 that is hopefully a little clearer. It probably could still use some specs, though.


vendor/engines/saml_authentication/README.md
```diff
@@ -73,12 +75,56 @@ OneLogin application configuration:
...
+   param.
+6. We redirect the user to NUcore's homepage (or facility's homepage)
+
+### IdP-initiated Signout
+
+1. User clicks "Logout" on the IdP's site
+2. IdP generates a `SAMLRequest` parameter and redirects the user's browser to
+   `SamlAuthentication::SessionsController#idp_sign_out` with the parameter as part
+   of the query string.
+3. NUcore invalidate's the user's session
```
meara: Typo.

vendor/engines/saml_authentication/README.md
```diff
@@ -73,12 +75,56 @@ OneLogin application configuration:
...
+4. NUcore generates an SLO logout response (saying we successfully logged the user out),
+   and redirects the user's browser back to the IdP.
+5. The IdP logs the user out of any other applications that were authenticated as part of
+   the same IdP session.
+6. User is redirected to somewhere on the IdP's (e.g. the login page)
```
meara: Think you mean "somewhere on the IdP's site".

vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
```diff
@@ -2,13 +2,69 @@ module SamlAuthentication
 
   class SessionsController < Devise::SamlSessionsController
 
+    # Some SSO providers are strict and require that the name_identifier matches
+    # what's logged in on their system for SLO.
+    before_action :store_logging_out_username, only: :destroy
+    after_action :store_winning_strategy, only: :create
+
+    # Overrides SamlSessionsController because we don't want to use saml_session_index_key
+    # because it logs you out of other browsers/sessions when you log in on another
```
meara: Can you clarify in the docs exactly what the behavior is with the respect to this override? Can you have more than one session with the IdP going at the same time, i.e. I'm signed in to email and/or NUcore on computer A, but signed into to email and NUcore on computer B, so that when I sign out of NUcore with SLO I am signed out of both email and NUcore on B, but my sessions on A are not affected?

vendor/engines/saml_authentication/app/controllers/saml_authentication/sessions_controller.rb
```diff
@@ -2,13 +2,69 @@ module SamlAuthentication
 
   class SessionsController < Devise::SamlSessionsController
 
+    # Some SSO providers are strict and require that the name_identifier matches
```
meara: I think this comment can go down by the method definition, to match with how you're doing it for `:store_winning_strategy`.

vendor/engines/saml_authentication/lib/saml_authentication/engine.rb
```diff
@@ -14,6 +14,8 @@ class Engine < Rails::Engine
       ViewHook.add_hook "devise.sessions.new",
                         "before_login_form",
                         "saml_authentication/sessions/new"
+
+      OneLogin::RubySaml::Logging.logger.level = Logger::DEBUG
```
meara: Are you planning to actually merge this and then take it out once the epic is complete, or...?

vendor/engines/saml_authentication/lib/saml_authentication/idp_entity_id_reader.rb
```diff
@@ -0,0 +1,40 @@
+module SamlAuthentication
+
+  # The default IdpEntityIdReader triggers an error on SLO because that request to
+  # us has a SAMLResponse parameter, which does not get parsed by OneLogin::RubySaml::Response
+  # correctly: "Issuer of the Response not found or multiple."
+  # Overrides DeviseSamlAuthenticatable::DefaultIdpEntityIdReader.
```
meara: I would put this line at the top of the comments (have the explanation of why we're doing this override below the fact that we are doing it).

Also, do you want to put a note in the comments that we're probably eventually going to open a PR/issue with the gem, and this will hopefully then be able to be taken out?

jhanggi: Good calls on all the comment suggestions. I'm probably not going to merge this for a little bit (definitely not until I hear back from DC on their configuration), but I'll address your thoughts.

vendor/engines/saml_authentication/lib/saml_authentication/engine.rb
jhanggi: I'm not sure yet. I'm thinking I might leave it in at least through testing on stage. Or we could do something like `OneLogin::RubySaml::Logging.logger.level = Rails.env.production? Logger::INFO : Logger::DEBUG`.

meara: Yeah I'm pretty okay with this ternary as a concept, that seems reasonably practical.

meara:
Btw @jhanggi what's the plan for specs here?

jhanggi:
I think my plan is do some controller specs where I capture some fixture files containing the SAML responses/requests like I did for the login. I'm open to other ideas if you have them. I'm still a little disinclined from adding vcr.

vendor/engines/saml_authentication/spec/controllers/saml_authentication/sessions_controller_spec.rb
```diff
@@ -135,4 +140,66 @@
...
+      doc = Nokogiri::XML(xml)
+      expect(doc.css("saml|NameID").first.text).to eq(user.username)
+    end
+  end
+
+  describe "#idp_sign_out" do
+    let(:user) { create(:user) }
+    before do
+      request.env["devise.mapping"] = Devise.mappings[:user]
+      request.host = "localhost:3000" # without this, we're getting redirected to http://test.host/
```
jhanggi: I could not figure out what was going on here, and eventually gave up. I was expecting it to work with what we have defined in `Rails.application.routes.default_url_options` in `environments/test.rb`

vendor/engines/saml_authentication/lib/saml_authentication/idp_entity_id_reader.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    def self.entity_id(params)
+      if params[:action] == "idp_sign_out" && params[:SAMLResponse]
+        OneLogin::RubySaml::Logoutresponse.new(
+          params[:SAMLResponse],
+          settings: Devise.saml_config,
+          allowed_clock_drift: Devise.allowed_clock_drift_in_seconds,
+        ).issuer
+      else
+        original_implementation(params)
```
vandrijevik: I’m not sure I quite follow the comment [here](https://github.com/tablexi/nucore-open/pull/1506/files#diff-74fdc61f331e681b5a8aa680f72296c6R25) and why this way of wrapping the original behavior improves clarity.

Perhaps I am missing something, but can’t we replace this line with a call to:

```
DeviseSamlAuthenticatable::DefaultIdpEntityIdReader.entity_id(params)
```

To me, that very clearly shows what we preempt and then fall back to, without a need to copy the method in this class, or rename it to `original_implementation`. We’re also not taking on an additional burden of maintaining the copy of `original_implementation` up to date with what the gem may change to upstream, and there is no need for comments to explain that we’re copying a method instead of just calling it.

vendor/engines/saml_authentication/spec/controllers/saml_authentication/sessions_controller_spec.rb
vandrijevik: FWIW, this is because `ActionDispatch::TestRequest` [defaults `HTTP_HOST` to `test.host`](https://github.com/rails/rails/blob/e53f45c57263e4e996bff3d4ba26342cadd3cca5/actionpack/lib/action_dispatch/testing/test_request.rb#L7). As a result, when a request like this is processed by Rails in a functional test, it uses that request’s host for any URL generation, because `default_url_options` kicks in only when the host is unknown (like in mailer contexts).

vendor/engines/saml_authentication/lib/saml_authentication/idp_entity_id_reader.rb
jhanggi: That's a fair question. I think I originally had this subclass `DefaultIdpEntityIdReader` and  call to `super`, but I felt that lost why we were needing to override it without needing to go into a `bundle open` to see what was missing. I also was feeling that with the override that I'd rather be in control of the whole thing and avoid them doing something under the hood that would break this.

I could go either way, though.

vendor/engines/saml_authentication/spec/controllers/saml_authentication/sessions_controller_spec.rb
jhanggi: What I was getting confused by is we have other controller specs that don't have this issue, and we don't specifically test `request.host`.

PR #1511, opened 2018-04-18
[#140001] update logo

meara:
# Release Notes

Update logo.

# Screenshots

screen shot 2018-04-19 at 1 51 56 pm
screen shot 2018-04-19 at 1 53 09 pm


jhanggi:
I'm especially worried about the height CSS not giving us a merge conflict. Also, seeing a merge conflict on a binary file is not the easiest thing to deal with. We have a kind of weird locale called `logo_file` that maybe we could use to deal with the different logos? Maybe it's not worth it and we just deal with the merge conflict on the image file.

Also, there's a `line-height` on the copyright of 75px to match the height of the image so the text is vertically centered to the logo. Looks like we didn't do any adjustments for UIC/Dartmouth, but it would be nice to do here.

meara:
Ended up resizing the logo so we don't need to adjust the height with CSS. I took out the line-height because it looks good on NU/open without it, and I think it looks pretty okay on Dartmouth and UIC as well, even though it's not centered with respect to the logo (see screenshots).

As far as merge conflicts go, Dartmouth has a different filename for the logo image, so the only messy one will be UIC, and I think it's probably more efficient to just manage that change during the latest-from-open merge than engineer some solution here.

PR #1512, opened 2018-04-18
Update rubocop

meara:
# Release Notes

TECH TASK: Update rubocop to latest version available on CodeClimate.


jhanggi:
The rubocop issues codeclimate sees I believe are legit, even though the message seems wrong. I think as we have some of the things, it could be ambiguous.

I'm pretty sure bad:
`cattr_accessor :common_permitted_account_params { [:description] }`

Good:
`cattr_accessor(:common_permitted_account_params) { [:description] }`

app/models/facility.rb
```diff
@@ -126,8 +126,8 @@ def complete_problem_order_details
 
   def set_journal_mask
     f = Facility.all.order(journal_mask: :desc).first
-    self.journal_mask = if f && f.journal_mask.match(/^C(\d{2})$/)
-                          sprintf("C%02d", Regexp.last_match(1).to_i + 1)
+    self.journal_mask = if f&.journal_mask&.match(/^C(\d{2})$/)
```
jhanggi: Fun fact about this I learned recently: it means we can never have more than 100 facilities.

PR #1515, opened 2018-04-19
[#140938] add specs to ensure case-insensitive matching

meara:
# Release Notes

Ensure matching on email/username to migrate existing email users to SSO via SAML is case-insensitive.


jhanggi:
I think you probably have `saml.create_user` set to false, maybe in settings.local.yml. I get the failure when it's not set, and the test passes when I set it to `false`.

I think the nesting of this spec might be a little messed up. This is not getting the `around` block that explicitly sets this.

PR #1516, opened 2018-04-19
[#137627] Homepage and facility page styling updates

meara:
# Release Notes
- Remove "Facilities" heading from home page
- Make calendar icon smaller to match text size
- Use consistent indentation and move the calendar icon to the right of the order form
- Center order form inputs vertically with respect to text

# Screenshots

screen shot 2018-04-19 at 5 30 32 pm
screen shot 2018-04-19 at 5 30 49 pm

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -46,20 +46,20 @@ module ProductsHelper
     if current_facility&.show_instrument_availability? || !current_facility && product.facility.show_instrument_availability?
       public_calendar_availability_options(product)
     else
-      { class: ["fa fa-calendar fa-2x"],
+      { class: ["fa fa-calendar fa-lg fa-fw"],
         title: t("instruments.public_schedule.icon") }
     end
   end
 
   def public_calendar_availability_options(product)
     if product.offline?
-      { class: ["fa fa-calendar fa-2x", "in-use"],
+      { class: ["fa fa-calendar fa-lg fa-fw", "in-use"],
         title: text("instruments.offline.note") }
     elsif product.walkup_available?
-      { class: ["fa fa-calendar fa-2x", "available"],
+      { class: ["fa fa-calendar fa-lg fa-fw", "available"],
         title: text("instruments.public_schedule.available") }
     else
-      { class: ["fa fa-calendar fa-2x", "in-use"],
+      { class: ["fa fa-calendar fa-lg fa-fw", "in-use"],
         title: text("instruments.public_schedule.unavailable") }
     end
   end
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -25,10 +25,12 @@ module ProductsHelper
     }
   end
 
-  def public_calendar_link(product)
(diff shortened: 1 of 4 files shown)


app/helpers/products_helper.rb
```diff
@@ -25,10 +25,12 @@ def options_for_relay
     }
   end
 
-  def public_calendar_link(product)
+  def public_calendar_link(product, should_show_icon)
```
jhanggi: Maybe just `show_icon` and make it a keyword arg with a default?

app/helpers/products_helper.rb
```diff
@@ -25,10 +25,12 @@ def options_for_relay
...
     if product.respond_to? :reservations
       opts = public_calendar_options(product)
       link_to "", facility_instrument_public_schedule_path(product.facility, product), opts
+    elsif should_show_icon
+      content_tag :span, "", class: "fa-lg fa-fw", style: "display: inline-block"
```
jhanggi: We don't have to, but it might be nice to get the `style` piece out of this and into CSS. I originally suggested it here just as a quick and dirty way to test it.

app/helpers/products_helper.rb
```diff
@@ -46,20 +48,20 @@ def public_calendar_options(product)
...
-      { class: ["fa fa-calendar fa-2x"],
+      { class: ["fa fa-calendar fa-lg fa-fw"],
         title: t("instruments.public_schedule.icon") }
     end
   end
 
   def public_calendar_availability_options(product)
     if product.offline?
-      { class: ["fa fa-calendar fa-2x", "in-use"],
+      { class: ["fa fa-calendar fa-lg fa-fw", "in-use"],
```
jhanggi: I feel like there might be a way to dry up all these shared class lists.

app/views/facilities/_product_list.html.haml
```diff
@@ -7,14 +7,15 @@
       - if products.first.class.model_name == "TimedService"
         %p= t("facilities.show.timed_services_note")
     %ul
+      - should_show_cal_icon = products.any? {|p| p.respond_to? :reservations }
```
jhanggi: Add a space after the curly brace in`{|p|`

app/helpers/products_helper.rb
meara: Probably, but literally everywhere we make calls to font-awesome it looks like this, so I think it might make more sense just to live with this. Especially if we're going to end up changing everything again when font-awesome-rails moves to font-awesome 5.

app/helpers/products_helper.rb
meara: Honestly, this whole thing is kind of a hack, but I think it's more clear to keep it all together in one place rather than extracting this super odd special-case CSS into a separate file somewhere.

app/helpers/products_helper.rb
jhanggi: That kind of why I was thinking about DRYing it up, so it would ideally be simpler to upgrade to fa5. I'm fine if we do the cleanup at that point.

PR #1518, opened 2018-04-19
[#137711] reservations sort by date orders by ordered_at

meara:
# Release Notes

The "Ordered Date" sort now sorts by ordered_at, not by reservation start time.


jhanggi:
Could we make the final commit message a little more non-dev friendly?

PR #1535, opened 2018-04-30
[#137552] Better error message when beginning a reservation in grace period

jhanggi:
# Release Notes

Improve error message when attempting to start a reservation in the grace period while the previous reservation is still running.

# Additional Context

Example: 
User 1 creates a reservation from 9-10 and starts their reservation
User 2 creates a reservation from 10-11

At 9:55, when the grace period begins, User 2 sees the "Begin Reservation" link. When clicking the link, they previously saw a confusing error generated by AASM. Now, display a more useful error message.

Unchanged behavior: once the grace period is over at 10:00 (i.e. the previous reservation is
technically over), “Begin Reservation” will start the reservation for User 2 and
move User 1's reservation to the problem queue.

Old message:
unknown

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -244,11 +244,11 @@ class ReservationsController < ApplicationController
         switch_instrument_off!
       end
     rescue AASM::InvalidTransition => e
-      if e.failures.include?(:time_data_completeable?)
-        flash[:error] = text("switch_instrument.prior_is_still_running")
-      else
-        raise
-      end
+      flash[:error] = if e.failures.include?(:time_data_completeable?)
+                        text("switch_instrument.prior_is_still_running")
+                      else
+                        e.message
+                      end
     rescue => e
       flash[:error] = e.message
     end
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -243,6 +243,12 @@ class ReservationsController < ApplicationController
       when params[:switch] == "off"
         switch_instrument_off!
       end
+    rescue AASM::InvalidTransition => e
+      if e.failures.include?(:time_data_completeable?)
+        flash[:error] = text("switch_instrument.prior_is_still_running")
+      else
+        raise
+      end
     rescue => e
       flash[:error] = e.message
     end
--- a/config/locales/en.controllers.yml
+++ b/config/locales/en.controllers.yml
@@ -132,6 +132,8 @@ en:
         Please add it again to make a reservation.
       update:
         failure: The reservation cannot be updated.
+      switch_instrument:
+        prior_is_still_running: Cannot "Begin Reservation" when a previously scheduled reservation is ongoing.
 
     general_reports:
       headers:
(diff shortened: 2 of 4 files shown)


app/controllers/reservations_controller.rb
```diff
@@ -243,6 +243,12 @@ def switch_instrument
       when params[:switch] == "off"
         switch_instrument_off!
       end
+    rescue AASM::InvalidTransition => e
```
jhanggi: I went back and forth on where to put the logic for this change, but ultimately decided this was the least disruptive spot.

app/controllers/reservations_controller.rb
```diff
@@ -243,6 +243,12 @@ def switch_instrument
...
+      if e.failures.include?(:time_data_completeable?)
+        flash[:error] = text("switch_instrument.prior_is_still_running")
+      else
+        raise
```
jhanggi: This will cause it to fall back to the default `rescue => e` a couple lines down.

config/locales/en.controllers.yml
```diff
@@ -132,6 +132,8 @@ en:
         Please add it again to make a reservation.
       update:
         failure: The reservation cannot be updated.
+      switch_instrument:
+        prior_is_still_running: Cannot "Begin Reservation" when a previously scheduled reservation is ongoing.
```
vandrijevik: The quoted `“Begin Reservation”` here seems strange — maybe consider something like:

> Could not begin a reservation because a previously-scheduled one is still ongoing.

spec/controllers/reservations_controller_spec.rb
```diff
@@ -1130,6 +1130,8 @@
         @params[:reservation_id] = @reservation.id
         create(:relay, instrument: @instrument)
         @random_user = create(:user)
+        @instrument.update_attributes(min_reserve_mins: 30)
```
vandrijevik: Can we make this a `update!` so it alerts us loudly if the setup doesn’t succeed?

spec/controllers/reservations_controller_spec.rb
```diff
@@ -1197,26 +1197,68 @@
...
-              expect(assigns(:reservation)).to_not be_changed
-              expect(assigns(:reservation).reserve_start_at).to eq(start_at)
               expect(assigns(:reservation).actual_start_at).to eq(Time.zone.now)
             end
           end
         end
 
+        context "when there is a prior reservation overrunning its time" do
+          let(:start_at) { 3.minutes.ago.change(usec: 0) }
+          let(:end_at) { 57.minutes.from_now.change(usec: 0) }
```
vandrijevik: Just curious: why are the `change(usec: 0)` calls here needed?

app/controllers/reservations_controller.rb
jhanggi: Yeah, that's a good call. I like that better; definitely clearer.

config/locales/en.controllers.yml
jhanggi: I don't disagree it's a little funky but it's the text they asked for.

spec/controllers/reservations_controller_spec.rb
jhanggi: Good call.

spec/controllers/reservations_controller_spec.rb
jhanggi: It's a copy and paste from the a spec or two down. I'm not 100% sure, but I'm pretty sure it has to do with database vs ruby timestamp precision--I'm pretty sure mysql truncates the milliseconds, or at least it did at one point. I can probably get away without it here since I'm not directly comparing timestamps in any of my new tests.

PR #1538, opened 2018-05-03
Fix time-travel bug in spec

jhanggi:
# Release Notes

Tech task: Fix time-travel bug in spec and add add a deprecation warning when using rails’s built-in time travel helpers.

## Additional Context

Hopefully I have explained the situation well enough in the comments. Let me know if it doesn't make sense or how I can clarify it. Ultimately, `travel_to` and its ilk don't work well when nested inside itself (I can't find a reference right now, but I remember it being a problem. Maybe it gets resolved in a future version of rails?)

Here's a spec that verifies the problem:
```ruby
  it "can nest inside itself" do
    expect(Time.current.iso8601).to eq("2018-09-11T09:30:00-05:00")
    travel_to(Time.parse("2019-09-11T09:30:00-05:00")) do
      expect(Time.current.iso8601).to eq("2019-09-11T09:30:00-05:00")
    end
    expect(Time.current.iso8601).to eq("2018-09-11T09:30:00-05:00")
  end
```

It fails on the last line because it returns back to the "real" now.
```
Failure/Error: expect(Time.current.iso8601).to eq("2018-09-11T09:30:00-05:00")

       expected: "2018-09-11T09:30:00-05:00"
            got: "2018-05-03T18:05:23-05:00"

       (compared using ==)
```

I'm not 100% sure how to integrate that test to see if it ever gets fixed.


jhanggi:
Yeah, ultimately, I'm not a huge fan of the global time-lock, but it's been around so long and so many tests rely on it that it would be a huge undertaking to remove. We've had the `travel_to_and_return` helpers for a little while. These new methods are really just a way of protecting us in the future.

I've updated the argument names to match Rails's names.

PR #1545, opened 2018-05-10
[#141665] Support both unix timestamps and iso8601 in calendar endpoint

jhanggi:
# Release Notes

Prevent "argument out of range" errors when browsers are still running old Fullcalendar code.

# Additional Context

For some reason, we’re continuing to get `reservations#index
(ArgumentError) "argument out of range”` errors caused by start/stop
parameters being in unix timestamp format. This format changed when we
upgraded Fullcalendar, so it seems like some browsers are continuing to
load the old code. That shouldn’t be happening because the library
should be fingerprinted.

This should stop the errors.

See: https://github.com/tablexi/nucore-open/pull/1478

Example:
https://nucore.northwestern.edu/facilities/imserc/instruments/nmr_a600/reservations.js?all=true&amp;with_details=true&start=1524373200&end=1524978000&_=1524755233886

Referrer: https://nucore.northwestern.edu/facilities/imserc/instruments/nmr_a600/schedule

### DIFF
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -399,9 +399,9 @@ class ReservationsController < ApplicationController
   def parse_time_param(value)
     return unless value
 
-    if value.to_s =~ /\A\d{10}\z/ # unix timestamp
+    if value.to_s.match?(/\A\d{10}\z/) # unix timestamp
       Time.zone.at(value.to_i)
-    elsif value.to_s =~ /\A\d{13}\z/ # unix timestamp with milliseconds
+    elsif value.to_s.match?(/\A\d{13}\z/) # unix timestamp with milliseconds
       Time.zone.at(value.to_i / 1000)
     else
       Time.zone.parse(value)
--- a/app/controllers/reservations_controller.rb
+++ b/app/controllers/reservations_controller.rb
@@ -27,8 +27,8 @@ class ReservationsController < ApplicationController
     @facility = Facility.find_by!(url_name: params[:facility_id])
     @instrument = @facility.instruments.find_by!(url_name: params[:instrument_id])
 
-    @start_at = params[:start] ? Time.zone.parse(params[:start]) : Time.zone.now
-    @end_at = params[:end] ? Time.zone.parse(params[:end]) : @start_at.end_of_day
+    @start_at = parse_time_param(params[:start]) || Time.zone.now
+    @end_at = parse_time_param(params[:end]) || @start_at.end_of_day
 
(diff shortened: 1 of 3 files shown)


app/controllers/reservations_controller.rb
```diff
@@ -394,6 +394,20 @@ def upcoming_today?(reservation)
...
 
+  # Some browsers are not updating their cached JS and have an out-of-date calendar
+  # JS library which uses unix timestamps for the parameters. This allows us to handle both.
+  def parse_time_param(value)
+    return unless value
+
+    if value.to_s.match?(/\A\d{10}\z/) # unix timestamp
+      Time.zone.at(value.to_i)
+    elsif value.to_s.match?(/\A\d{13}\z/) # unix timestamp with milliseconds
+      Time.zone.at(value.to_i / 1000)
```
vandrijevik: I believe this branch is unnecessary, because both `start` and `end` come in as UNIX timestamps with second precision (1524373200 and 1524978000). The only param that comes in with millisecond precision is `_` (1524755233886), which we don’t attempt to process in any way.

app/controllers/reservations_controller.rb
```diff
@@ -394,6 +394,20 @@ def upcoming_today?(reservation)
...
+    else
+      Time.zone.parse(value)
+    end
+  end
```
vandrijevik: I think that instead of checking regular expressions here and trying to guess how to process the timestamp, it may be better to do something like this:

```
def parse_time_param(string_value)
  return unless string_value
  Time.zone.parse(string_value)
rescue ArgumentError
  Time.at(string_value.to_i)
end
```

That way, we’re only falling back to `Time.at` if `Time.parse` fails, accomodating UNIX timestamps as well, without making assumptions about the format of the string.

app/models/user.rb
```diff
@@ -138,6 +138,7 @@ def account_price_groups
   # purchasing that product
   def accounts_for_product(product)
     acts = accounts.active.for_facility(product.facility).to_a
+    puts acts.map { |acct| [acct, acct.validate_against_product(product, self)] }
```
vandrijevik: Seems like this snuck through and it shouldn’t be here, no? 😄

app/controllers/reservations_controller.rb
jhanggi: Yeah, I thought about that, and was all "don't use exceptions for flow control", but in this case I think it probably makes more sense.

app/models/user.rb
jhanggi: Oops.

PR #1549, opened 2018-05-11
[#136765] Remove create tabs for CCs or POs for account managers

jhanggi:
# Release Notes

Remove create tabs for CCs or POs for account managers. Previously, the tabs were available, but would 404.

# Screenshot

Optional.

# Additional Context

This changes the behavior a bit, though the old behavior was unexpected. Now, it will fall back to the first available type, which is what the controller makes it look like it _should_ be doing anyway.

The spec might make it a little funky to backport to 4.2, so we should decide what we want to do about that.

### DIFF
--- a/app/models/account_config.rb
+++ b/app/models/account_config.rb
@@ -69,8 +69,8 @@ class AccountConfig
   # and the NullObject always returns `true` for cross_facility?.
   def account_types_for_facility(facility, action)
     types = account_types
-    types = account_types.select { |type| type.constantize.cross_facility? } if facility.try(:cross_facility?)
-    types = account_types - creation_disabled_types if action == :create
+    types = types.select { |type| type.constantize.cross_facility? } if facility.try(:cross_facility?)
+    types = types - creation_disabled_types if action == :create
     types
   end
 
--- a/app/controllers/facility_accounts_controller.rb
+++ b/app/controllers/facility_accounts_controller.rb
@@ -39,14 +39,10 @@ class FacilityAccountsController < ApplicationController
   end
 
   # GET /facilities/:facility_id/accounts/new
-  def new
-    @available_account_types = available_account_types
-  end
+  def new; end
 
   # POST /facilities/:facility_id/accounts
   def create
-    @available_account_types = available_account_types
-
     # The builder might add some errors to base. If those exist,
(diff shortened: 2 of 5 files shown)


app/models/account_config.rb
```diff
@@ -69,8 +69,8 @@ def account_type_to_route(class_string)
   # and the NullObject always returns `true` for cross_facility?.
   def account_types_for_facility(facility, action)
     types = account_types
-    types = account_types.select { |type| type.constantize.cross_facility? } if facility.try(:cross_facility?)
-    types = account_types - creation_disabled_types if action == :create
+    types = types.select { |type| type.constantize.cross_facility? } if facility.try(:cross_facility?)
```
meara: Possibly this is nit-picky, but couldn't you just keep the old line 72 and get rid of line 71? Was there a reason you chose to keep the separate lines?

jhanggi: We need 71 because each of the subsequent lines have a conditional, so we need the initial setup. If 72 wasn't conditional, then you would be right. Am I understanding what you're asking correctly?

meara: Oh oops, my eyes just slid past the conditional in 72. That makes sense then.

jhanggi:
No screenshot. Removed that section from the description.

What I mean by the fallback behavior is when the query param isn't available/allowed. Here's what the URL looks like: http://localhost:3000/facilities/example/accounts/new?account_type=CreditCardAccount&owner_user_id=16

With this change, if `CreditCardAccount` isn't available, it'll use `NufsAccount` as the `account_type`. Does that make sense?

meara:
So the tabs for CC/POs shouldn't show up now because they've been removed from the collection for account managers, right? So they could only go to a link like http://localhost:3000/facilities/example/accounts/new?account_type=CreditCardAccount&owner_user_id=16 if they manually hacked the url or were following an old link from somewhere else, right? Are there links in the app that will still point to those locations for account managers?

The fallback behavior essentially should never happen, right? We're just being extra safe?

jhanggi:
The only places those links are available should be in those tabs. So you are correct in the only ways for an account manager to get to those URLs would be manually hacking or a bookmark/external link. Yes, the fallback behavior is just a safety precaution.

PR #1553, opened 2018-05-14
[#141837] Remove ability for account managers to edit user details

jhanggi:
# Release Notes

Remove ability for account managers to edit user details


vandrijevik:
Makes sense to me, but it looks like the test needs a tweak? Good to go on 🍏!

PR #1554, opened 2018-05-14
[#139381] Gray out edit/remove price policy links for staff/senior staff

jhanggi:
# Release Notes

Gray out edit/remove item/service price policy links for staff/senior staff.

# Additional Context

The links already 403 correctly; this fixes the links. This is already restricted on instrument price policy views.

### DIFF
--- a/app/views/price_policies/_table.html.haml
+++ b/app/views/price_policies/_table.html.haml
@@ -9,23 +9,23 @@
       %th.currency= price_policy.class.human_attribute_name(:unit_adjustment)
       %th.currency= price_policy.class.human_attribute_name(:unit_net_cost)
   %tbody
-    - price_policies_to_show = price_policies.select { |pp| pp.can_purchase? }
+    - price_policies_to_show = price_policies.select(&:can_purchase?)
     - price_policies_to_show.each do |price_policy|
       %tr
         - if price_policies_to_show.first == price_policy
           %td.centered{ :rowspan => price_policies_to_show.length }
-            - unless price_policies.all?{|pp| pp.editable?}
-              %p.muted= t("shared.edit")
-              %p.muted= t("shared.remove")
-            - else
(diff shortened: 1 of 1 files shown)


app/views/price_policies/_table.html.haml
```diff
@@ -9,23 +9,23 @@
...
+    - price_policies_to_show = price_policies.select(&:can_purchase?)
     - price_policies_to_show.each do |price_policy|
       %tr
         - if price_policies_to_show.first == price_policy
           %td.centered{ :rowspan => price_policies_to_show.length }
-            - unless price_policies.all?{|pp| pp.editable?}
-              %p.muted= t("shared.edit")
-              %p.muted= t("shared.remove")
-            - else
+            - if price_policies.all?(&:editable?) && can?(:edit, PricePolicy)
```
vandrijevik: This is a bit confusing to me: why are we checking that _all_ price policies are editable for every edit link? Should we perhaps change this to `if price_policy.editable? && can?(:edit, price_policy)` so we’re checking just the policy we’re currently processing in the loop?

jhanggi: This is indeed a bit confusing that stems from how the data and controllers are modeled. `PricePolicy` is a price/subsidy for a price group for a product. These are grouped up by their `start_date` with each set having their own edit/remove links. The edit view is actually editing a group of `PricePolicy`s. (e.g. `/facilities/example/services/example-service/price_policies/2018-05-15/edit`)

Once a price policy has been assigned to an order, it becomes locked and is no longer `editable?`. So we want to block editing if _any_ of the group has become uneditable.

Maybe a screenshot will help clarify:

screen shot 2018-05-15 at 12 13 14 pm

Does that make sense? I've wanted for a very long time to add another model that represents the group (start/end date, product would get moved to the new model), but that would be a very large undertaking that would touch a lot of rarely touched pieces of the code.

PR #1559, opened 2018-05-18
Allow admin-defined logo in header

robstolarz:
# Release Notes
This change allows administrators to add a logo to their header. The new logo image should be called "logo-header.png" and it should be placed in the "app/assets/images/" folder. The changes are not visible immediately, only after the asset pipeline is rebuilt (which just involves restarting Rails).

The diff is a little interesting because the header was redone in Flexbox. The new version is pixel-for-pixel the same, though, and it shouldn't affect user overrides as far as I can tell.

# Screenshot
## Before
screen shot 2018-05-18 at 4 46 16 pm

## After
screen shot 2018-05-18 at 4 46 24 pm
(this is px-for-px the same as "Before")

## After, with logo added
screen shot 2018-05-18 at 4 46 31 pm

# Additional Context
This is part of a larger set of changes to the design.

### DIFF
--- a/app/assets/stylesheets/app/header.scss
+++ b/app/assets/stylesheets/app/header.scss
@@ -12,7 +12,7 @@ header {
       font-size: 27px;
       color: $headerColor;
 
-      padding: 11px 20px 10px;
+      padding: 17px 20px 16px;
 
       img {
         height: 1em;
--- a/app/assets/stylesheets/app/header.scss
+++ b/app/assets/stylesheets/app/header.scss
@@ -14,7 +14,7 @@ header {
 
       padding: 17px 20px 16px;
 
-      img {
+      .img__brand {
         height: 1em;
       }
 
--- a/app/assets/stylesheets/app/header.scss
+++ b/app/assets/stylesheets/app/header.scss
@@ -11,12 +11,34 @@ header {
       text-shadow: none;
       font-size: 27px;
       color: $headerColor;
+
+      padding: 11px 20px 10px;
+
+      img {
+        height: 1em;
+      }
+
+      span {
+        vertical-align: middle;
+        line-height: 1;
+      }
     }
   }
 
   .navbar-inner, .dropdown-toggle {
-    padding: 10px 0;
     @include gradient-vertical($headerBackgroundColorHighLight, $headerBackgroundColor);
+  }
+
+  .container {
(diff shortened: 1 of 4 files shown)


robstolarz:
My apologies for failing to notice the mobile layout. This has been rectified with these new commits.

Full nav now looks like this with a logo: 
screen shot 2018-05-25 at 12 49 17 pm

And the hamburger menu:
screen shot 2018-05-25 at 12 49 35 pm

And it looks like this when opened:
screen shot 2018-05-25 at 12 49 40 pm

The new CSS class has been added and the logo's location is now configurable. I left the unfortunately complex `ActionController::Base.helpers.resolve_asset_path` check in, as it prevents empty image tags from being created when Rails' asset pipeline has not been re-run. Let me know if I should remove that or not.

The commits are separate now for your convenience in reviewing them. Let me know when to squash them down.

jhanggi:
I probably won't get a chance to test out everything locally until Tuesday, but screenshots look much better on mobile. 

I would prefer to take the `resolve_asset_path` out. Asset compilation will happen on deployment so we shouldn't need to worry about it in stage/prod, and development should handle ok as long as the file is there. What do you mean by "when Rails' asset pipeline has not been re-run"?

Don't worry about squashing, we'll do that as part of the merge. Thanks!

robstolarz:
It's interesting, when I add the file and an `override.yml` entry in development, the image will not load until I restart the Rails server. It will produce an image tag but the image tag will lead to a 404. I added that check to prevent this, as I was very puzzled as to why the image was not showing up and then somehow restarting the Rails server made everything work.

So that's why the `resolve_asset_path` check is there, to make sure that the asset has actually made it through the pipeline before trying to display it.

jhanggi:
I just tested it out and it seemed to pick up the asses for me right away. I do know if you're adding new directories, sometimes those aren't watched by the asset pipeline for new files. And a heads up that many initializers rely on settings.yml. It seems to get picked up immediately for this scenario, but a lot of time if you change `settings.yml`, you'll need to restart the server.

Let's take that extra condition out. If this were something that were constantly changing the added check might be worth it, but I'm guessing the logo won't change too often so restarting the server in development isn't a huge hassle here.

robstolarz:
Took out the check.

PR #1561, opened 2018-05-21
[#134365] Improve error messaging if an instrument has no schedule rules or price policies

vandrijevik:
# Release Notes

Improved error messages if an instrument doesn’t have a schedule rules, and if a product does not have pricing rules.

# Screenshot

screen shot 2018-05-21 at 15 44 22
screen shot 2018-05-21 at 15 43 48

# Additional Context

All `Product`s have price policies, and only `Instrument`s have schedules. Therefore, the message about scheduling can appear only for instruments, and the message for pricing can appear for any type of product.

To make managing these checks easier, I consolidated all of the purchasability checks into `ProductForCart` and `InstrumentForCart` (which contains additional, instrument-only checks), along with model-level tests. As a result, the product controller specs now only test the interactions with these models, and the model specs test the actual error messages and paths.


app/models/instrument_for_cart.rb
```diff
@@ -0,0 +1,18 @@
+class InstrumentForCart < ProductForCart
+
+  def purchasable_by?(acting_user, session_user)
+    super.tap do |success|
```
jhanggi: I feel like the `super.tap` is a weird construct here. It'll always return the value of `super`, so even if we add an error inside the block, the method will still return `true` if it was successful in the superclass.

app/models/instrument_for_cart.rb
```diff
@@ -0,0 +1,18 @@
...
+      if acting_user.blank?
+        @error_path = url_helpers.new_user_session_path
+      elsif product.schedule_rules.none?
+        @error_message = text(".schedule_not_available", i18n_params)
+      end
+
+      # If an instrument is not purchasable, we always redirect. If ProductForCart
+      # didn’t set an error path due to some specific error, we should redirect
```
jhanggi: I have the codeclimate browser plugin and it's complaining about using only ascii in comments. I usually wouldn't care if it were a specific reason (an accented character or whatnot), but this looks like it's just a curly apostrophe.

screen shot 2018-05-21 at 1 28 43 pm

app/models/product_for_cart.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  def initialize(product)
+    @product = product
+  end
+
+  def purchasable_by?(acting_user, session_user)
+    raise NUCore::PermissionDenied unless product.is_accessible_to_user?(session_user)
+    return false if acting_user.blank?
+
+    if !product.available_for_purchase?
```
jhanggi: This method is super complex. Is there any way we can break it up?

I don't know if something like this would be any better?

```ruby
[
available_for_purchase,
associatated_price_policies,
...
].none?

def available_for_purchase
  if product.available_for_purchase?
    true
  else
    @error_message = text(".not_available", i18n_params)
    false
  end
end
```

spec/controller_spec_helper.rb
```diff
@@ -257,3 +257,80 @@ def split_date_to_params(key, date)
     :"#{key}_meridian" => date.strftime("%p"),
   }
 end
+
+def it_should_support_purchasing
```
jhanggi: I would lean towards using a shared example for this rather than a big method. And then I would put that in `spec/support/shared_examples`. I generally dislike most of the methods defined in this helper. They're mainly an older style of testing that we try to avoid doing anymore.

jhanggi: Also, we're tending towards moving to more active wording in our specs to go along with `should` -> `expect`, so this would be something like "it supports purchasing". Looks like you're doing it in the actual spec descriptions, so we should probably do it for these descriptions as well.

spec/models/instrument_for_cart_spec.rb
```diff
@@ -0,0 +1,33 @@
+require "rails_helper"
+
+RSpec.describe InstrumentForCart do
+
+  let(:facility) { FactoryBot.create(:facility) }
+  let(:facility_account) { facility.facility_accounts.create(FactoryBot.attributes_for(:facility_account)) }
+
+  let(:instrument) do
+    FactoryBot.create(:instrument, facility: facility, facility_account: facility_account, no_relay: true).tap do |instrument|
```
jhanggi: I'm pretty sure you don't need the `tap` here. FactoryBot's `create` method should take a block yielding the object being created.

app/models/instrument_for_cart.rb
vandrijevik: I think that’s a terrible rubocop check, so I always mark it as invalid. Would you like me to change the apostrophe here?

app/models/product_for_cart.rb
vandrijevik: Yeah, that’s fair. I felt the same way, but after spending so much time refactoring this code I needed a stopping point at which I could implement the new checks.

Do you think it merits a few more cycles to extract these checks into separate methods? I like that direction, for sure.

spec/controller_spec_helper.rb
vandrijevik: I didn’t know what the Rspec-sanctioned way to do these is, so I’ll change them! (With Minitest, I would have just created a module, but that doesn’t quite work here)

app/models/instrument_for_cart.rb
vandrijevik: With the refactoring of `ProductForCart#purchasable_by?` into separate methods that I did in 166039e37, I think I addressed this as well. What do you think?

app/models/instrument_for_cart.rb
```diff
@@ -0,0 +1,26 @@
+class InstrumentForCart < ProductForCart
+
+  def purchasable_by?(acting_user, session_user)
+    # If an instrument is not purchasable, we always redirect. Hence, if super
+    # doesn’t set an error path due to some specific error, we set the error path
+    # to the facility path.
+    super.tap do |is_purchasable|
```
jhanggi: This tap still feels confusing. Usually I only like tap if we're doing builder-y things.

```ruby
is_purchasable = super
@error_path ||= url_helpers.facility_path(product.facility) unless is_purchasable
is_purchasable
```

app/models/product_for_cart.rb
```diff
@@ -0,0 +1,118 @@
...
+    "models.#{self.class.name.underscore}"
+  end
+
+  private
+
+  attr_accessor :product
+
+  def checks(acting_user, session_user)
+    [
+      product_is_accessible?(session_user),
```
jhanggi: I like how this setup reads, but the fact that each method needs to return a proc/lambda feels a little funky, but it's not ultimately bad.

app/models/product_for_cart.rb
```diff
@@ -0,0 +1,118 @@
...
+  attr_accessor :error_path
+
+  def initialize(product)
+    @product = product
+  end
+
+  def purchasable_by?(acting_user, session_user)
+    error = checks(acting_user, session_user).detect do |check|
+      result = check.call
+      [error_path, error_message].any?(&:present?) || result == false
```
jhanggi: I think that I would prefer if each of the `?` methods  (or their procs) returned a proper boolean as the `?` convention usually indicates. Then we wouldn't need the checks for the attr_readers.

app/models/product_for_cart.rb
jhanggi: What if we did something like:

```ruby
[
  [:rpoduct_is_accessible?, session_user],
  [:user_is_present?, acting_user],
  ...
]
```

Then the loop could possibly be:
```ruby
checks(acting_user, session_user).all? do |check, *args|
  method(check).call(*args)
end
```

(I haven't actually verified this, especially with the splats). But that could clean up the need for the procs. Just an idea.

app/models/product_for_cart.rb
```diff
@@ -0,0 +1,118 @@
...
+
+  attr_accessor :error_message
+  attr_accessor :error_path
+
+  def initialize(product)
+    @product = product
+  end
+
+  def purchasable_by?(acting_user, session_user)
+    error = checks(acting_user, session_user).detect do |check|
```
jhanggi: Could this be `all?` instead of `detect`? Then you shouldn't need the `error.nil?` check at the end.

jhanggi:
Also, looks like we've got a couple test failures now.

app/models/product_for_cart.rb
vandrijevik: I had that before, but I ended up preferring the procs over like using `method` to extract a callable, and coming up with serialization of the name and arguments as an array.

Are you okay with keeping the procs? I think it’s more straightforward, but yeah — the fact that we want to stop after the first error we encounter makes this not ideal.

app/models/product_for_cart.rb
vandrijevik: `all?` will run all of the checks, and the current behavior (which we don’t want to change here) is to stop after the first encountered problem, so unfortunately, that’s why we can’t use it.

jhanggi: I'm pretty sure that `all?` short circuit on the first `false` in the block. I've run into that where I _do_ want to run all checks in an array but return the `all?`.

vandrijevik: I didn’t understand what Jason had meant originally here, and I asked about him directly. The goal was to get rid of the local variable, which  I refactored away in 11d0f18

jhanggi:
Just figured I'd reference this old, old branch I just noticed as I was trying to clean up some branches. I'll probably close it once this PR is in, but it looks similar in a lot of ways.

https://github.com/tablexi/nucore-open/compare/product-show-refactor?expand=1

PR #1562, opened 2018-05-21
[#137474] Handle invalid dates in price policies

jhanggi:
# Release Notes

Better handling of errors for invalid dates in price policy forms.

# Screenshot

screen shot 2018-05-21 at 12 11 23 pm

screen shot 2018-05-21 at 12 17 28 pm

# Additional Context

There was no validation on expire date before, but would hard error if the expire date was invalid/blank (`NoMethodError: undefined method `end_of_day' for nil:NilClass`). Adding that validation did cause some problems with specs that weren't setting it, but I think I got all those cleaned up.

This will now also include some JS validation as used elsewhere which does not actually block the submission. On submit, it will be treated as a null/blank field and on re-render will be blank.

### DIFF
--- a/app/controllers/price_policies_controller.rb
+++ b/app/controllers/price_policies_controller.rb
@@ -34,8 +34,7 @@ class PricePoliciesController < ApplicationController
     # If there are active policies, start tomorrow. If none, start today
     @start_date = Date.today + (active_policies? ? 1 : 0)
 
-    @price_policies = PricePolicyBuilder.get_new_policies_based_on_most_recent(@product, @start_date, @max_expire_date)
-
+    @price_policies = PricePolicyBuilder.get_new_policies_based_on_most_recent(@product, @start_date)
     raise ActiveRecord::RecordNotFound if @price_policies.blank?
   end
 
--- a/app/controllers/price_policies_controller.rb
+++ b/app/controllers/price_policies_controller.rb
@@ -31,8 +31,7 @@ class PricePoliciesController < ApplicationController
 
   # GET /facilities/:facility_id/{product_type}/:product_id/price_policies/new
   def new
-    # If there are active policies, start tomorrow. If none, start today
-    @start_date = Date.today + (active_policies? ? 1 : 0)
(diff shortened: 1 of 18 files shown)


app/controllers/price_policies_controller.rb
```diff
@@ -33,28 +31,36 @@ def index
 
   # GET /facilities/:facility_id/{product_type}/:product_id/price_policies/new
   def new
-    @start_date = new_start_date
-    @expire_date = set_max_expire_date
+    # If there are active policies, start tomorrow. If none, start today
+    @start_date = Date.today + (active_policies? ? 1 : 0)
```
vandrijevik: I think `active_policies? ? Date.tomorrow : Date.today` may be clearer here, and not even need the explanatory comment above.

app/controllers/price_policies_controller.rb
```diff
@@ -33,28 +31,36 @@ def index
...
+    else
+      flash.now[:error] = text("errors.save")
+      render :new
+    end
   end
 
   # PUT /facilities/:facility_id/{product_type}/:product_id/price_policies/:id
   def update
-    create_or_update(:edit)
+    if update_policies_from_params!
```
vandrijevik: It seems strange to use a conditional with a persistence method that ends with a `!` here (because the Rails methods with names like that raise exceptions). I see that the culprit is `PricePolicyUpdater.update_all!` (which calls `save` in its implementation, and not `save!`), but perhaps we can rename `update_policies_from_params!` to not have a `!`?

app/controllers/price_policies_controller.rb
jhanggi: Good call. That is definitely clearer. This was just an inlining of a separate method call, and I think that's probably even better than the method call.

app/controllers/price_policies_controller.rb
jhanggi: It calls yeah, you're right. I think the exclamations all the way down led me to use that here, but you're right that we don't need that. Will update.

PR #1570, opened 2018-05-25
Split facility sub-items to their own columns

ColinMcNeil:
# Release Notes

Instead of a single list of objects belonging to a facility, multiple columns representing each type of object are used.

# Screenshot

## Before
image

## After
image

### DIFF
--- a/app/assets/stylesheets/app/product_list.scss
+++ b/app/assets/stylesheets/app/product_list.scss
@@ -5,7 +5,9 @@
 .product_list_container{
     display: flex;
     justify-content: space-between;
+    flex-wrap:wrap;
 }
 .product_list{
     flex: 0 0 calc(25% - 2em);
+    min-width: 200px;
 }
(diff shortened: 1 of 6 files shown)


ColinMcNeil:
## Release Notes
We have placed this feature into its own feature-flag.

## Mobile Screenshot
image

As you can see, the columns wrap responsively.

ColinMcNeil:
This PR is in a state where we think is is ready to be merged with your requested changes and checks.

vandrijevik:
Hi @ColinMcNeil, @jhanggi is on vacation for a couple of weeks, so I just wanted to let you know that this is on our radar, but since he’s been involved with the review so far, it may be a little while before Jason gets back to you!

jhanggi:
I did a little stylistic cleanup on the CSS (use two spaces and add a space between the class name and the `{`, but otherwise is good. I'll merge this as soon as it passes CI.

PR #1578, opened 2018-05-31
Add gems to suport ed25519 keys in capistrano

vandrijevik:
# Release Notes

TECH TASK: Allow using ed25519 keys to deploy via capistrano

# Additional Context

Without these gems, capistrano fails with the following message:

```
$ bundle exec cap staging deploy
(Backtrace restricted to imported tasks)
cap aborted!
NotImplementedError: unsupported key type `ssh-ed25519'
net-ssh requires the following gems for ed25519 support:
 * rbnacl (>= 3.2, < 5.0)
 * rbnacl-libsodium, if your system doesn't have libsodium installed.
 * bcrypt_pbkdf (>= 1.0, < 2.0)
See https://github.com/net-ssh/net-ssh/issues/478 for more information
Gem::LoadError : "rbnacl is not part of the bundle. Add it to your Gemfile."

Tasks: TOP => rvm:hook
(See full trace by running task with --trace)
```

https://blog.g3rt.nl/upgrade-your-ssh-keys.html is a good overview of Ed25519 SSH keys, and why it’s a good idea to migrate towards using them.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -92,6 +92,13 @@ group :development, :deployment do
   gem "capistrano-rails",   require: false
   gem "capistrano-rvm",     require: false
   gem "capistrano-bundler", require: false
+
+  # net-ssh requires the gems below to support ed25519 keys
+  # for deploying via capistrano
+  # more info at https://github.com/net-ssh/net-ssh/issues/478
+  gem "bcrypt_pbkdf", ">= 1.0", "< 2.0"
+  gem "rbnacl", ">= 3.2", "< 5.0"
+  gem "rbnacl-libsodium"
 end
 
 group :development, :test do
--- a/Gemfile
+++ b/Gemfile
@@ -93,12 +93,12 @@ group :development, :deployment do
   gem "capistrano-rvm",     require: false
   gem "capistrano-bundler", require: false
 
-  # net-ssh requires the gems below to support ed25519 keys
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -92,6 +92,13 @@ group :development, :deployment do
   gem "capistrano-rails",   require: false
   gem "capistrano-rvm",     require: false
   gem "capistrano-bundler", require: false
+
+  # net-ssh requires the gems below to support ed25519 keys
+  # for deploying via capistrano
+  # more info at https://github.com/net-ssh/net-ssh/issues/478
+  gem "bcrypt_pbkdf", ">= 1.0", "< 2.0"
```
jhanggi: Can we get away with `require: false` in these? And it's still not possible to remove the upper constraint? Looks like that issue was over a year ago.

vandrijevik: I’ll try specifying `require: false` here to see if `net-ssh` will pick them up correctly. As for the upper version constraint, we can’t get rid of it because even `net-ssh` 5 (and our bundle is still at 4.2) specifies the same `< 2.0` limit: https://github.com/net-ssh/net-ssh/issues/565

vandrijevik: @jhanggi `require: false` worked, so I added it in 4e71fa5 (along with mentioning that the note is for 4.2 of `net-ssh`)

PR #1586, opened 2018-06-07
Ensure that test checking for accounts does not fail due to a lack of price groups

vandrijevik:
# Release Notes

TECH TASK: Ensure that test checking for accounts does not fail due to a lack of price groups

# Additional Context

Some of the school-specific forks contain engines that override `Product#can_purchase?` with rules beyond the default ones. As a result of this, the test which exercises the check that the acting user has appropriate accounts gets tripped up by the false return of `Product#can_purchase?`, which executes sooner. Since we test that path separately, in this context we stub it away to ensure that we get to the check we’re interested in testing.

### DIFF
--- a/spec/models/product_for_cart_spec.rb
+++ b/spec/models/product_for_cart_spec.rb
@@ -121,7 +121,10 @@ RSpec.describe ProductForCart do
     end
 
     context "when the acting user does not have any accounts that can purchase the product" do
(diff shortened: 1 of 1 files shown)


vandrijevik:
@danhodos Sounds good to me, added in 30fa4391e!

PR #1589, opened 2018-06-08
[#142184] Allow customizing error message for non-existent users signing in via SAML

vandrijevik:
# Release Notes

Allow customizing error message for non-existent users signing in via SAML

# Additional Context

The SAML authentication strategy for Devise currently uses the same i18n key for failed logins as other strategies (`devise.failure.invalid`). This PR makes SAML failed logins use `saml_invalid`, so we can customize the messaging in this flow separately from the email & password flow.

### DIFF
--- a/vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
+++ b/vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
@@ -16,7 +16,7 @@ module Devise
       # causes Devise to use the error message defined in devise.failure.invalid. That
       # error message is used by other strategies, including the email and password strategy
       # that we also use, so if we change the message we'll incorrectly affect the other
-      # strategies as well. Therefore, we override the m ethod responsible for failing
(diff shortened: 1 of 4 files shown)


vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
```diff
@@ -0,0 +1,40 @@
...
+
+    class SamlAuthenticatableWithCustomError < SamlAuthenticatable
+
+      private
+
+      # In the original SamlAuthenticatable, this method calls fail!(:invalid), which
+      # causes Devise to use the error message defined in devise.failure.invalid. That
+      # error message is used by other strategies, including the email and password strategy
+      # that we also use, so if we change the message we'll incorrectly affect the other
+      # strategies as well. Therefore, we override the m ethod responsible for failing
```
danhodos: Extra space in a word here

vandrijevik: 🤦‍♂️ Thanks! Fixed in cb82027a9

PR #1591, opened 2018-06-14
[#139111] Do not show canceled reservations on My Reservations screen

vandrijevik:
# Release Notes

Do not show canceled reservations on My Reservations screen

# Additional Context

The ticket mentions that problem reservations should not appear on this screen either; however, from what I can [see in the code](https://github.com/tablexi/nucore-open/blob/0be253c1abda99036b59acb249a4b8a644028e00/app/models/order_detail.rb#L896-L904) only completed reservations can be problematic. Since completed reservations do not appear on the My Reservations screen (which shows only upcoming and in-progress reservations), there was nothing to be done there.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -285,7 +285,12 @@ class OrderDetail < ApplicationRecord
   scope :pending, -> { joins(:order).where(state: %w(new inprocess)).purchased }
 
   scope :with_reservation, -> { purchased.joins(:reservation).order("reservations.reserve_start_at DESC") }
-  scope :with_upcoming_reservation, -> { with_reservation.where(reservations: { actual_start_at: nil }).merge(Reservation.ends_in_the_future) }
+  scope :with_upcoming_reservation, -> {
+    non_canceled.
+    with_reservation.
+      where(reservations: { actual_start_at: nil }).
+      merge(Reservation.ends_in_the_future)
+  }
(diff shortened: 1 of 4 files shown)


app/models/order_detail.rb
```diff
@@ -283,21 +283,15 @@ def can_be_viewed_by?(user)
...
 
-  scope :all_reservations, -> { confirmed_reservations.order("reservations.reserve_start_at DESC") }
+  scope :with_reservation, -> { purchased.joins(:reservation).order("reservations.reserve_start_at DESC") }
+  scope :with_upcoming_reservation, lambda {
+    non_canceled
+      .with_reservation
+      .where(reservations: { actual_start_at: nil })
+      .merge(Reservation.ends_in_the_future)
+  }
+  scope :with_in_progress_reservation, -> { non_canceled.with_reservation.merge(Reservation.relay_in_progress) }
```
meara: I can't remember off the top of my head--do these get ordered elsewhere in the code before they're displayed in the view? I just want to make sure the default ordering in the view(s) hasn't changed.

vandrijevik: The `with_reservation` scope just above, which these two scopes call, includes the order!

Previously, all three scopes included the exact same call to order, so I refactored it into the scope that is used as a base for these.

meara: Ah gotcha. Makes sense!

PR #1596, opened 2018-06-19
[#136143] Fix reservations drag and drop, and edit links for reservations without order

vandrijevik:
# Release Notes

Do not display edit links for reservations without an order, and fix drag and drop editing of reservations

### DIFF
--- a/app/views/reservations/_js_variables.html.haml
+++ b/app/views/reservations/_js_variables.html.haml
@@ -1,5 +1,6 @@
 :javascript
   var events_path     = "#{facility_instrument_reservations_path(@instrument.facility, @instrument, format: 'js', with_details: @instrument.show_details?)}";
+  var orders_path_base = "#{facility_orders_path(@instrument.facility)}";
   var minDate         = "#{@min_date}";
   var maxDate         = "#{@max_date}";
   var maxDaysFromNow  = #{@max_window};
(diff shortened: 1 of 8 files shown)


jhanggi:
I think this is actually ticket #136143. 

Also, the ticket is not entirely clear, but I think we should have an "Edit" link for Admin reservations as well. The path for those is `facilities/:facility/instruments/:instrument/reservations/:reservation_id/edit_admin`. Right now there is no link for those.

app/views/schedule_rules/index.html.haml
```diff
@@ -2,6 +2,7 @@
   = javascript_include_tag "fullcalendar.js", "schedule_rules.js"
   :javascript
     var events_path = "#{url_for(:format => 'js')}";
+    var orders_path_base = "#{facility_orders_path(current_facility)}";
```
jhanggi: I feel like the schedule rules shouldn't need this as that calendar view does not include any actual reservations.

vandrijevik:
@jhanggi I removed `orders_path_base` from the views where it’s unnecessary, and modified the guard clause in the render like we talked about. The ticket says that the edit link is not necessary for admin hold reservations, so I think this should be good now!

app/views/reservations/_js_variables.html.haml
```diff
@@ -1,5 +1,6 @@
 :javascript
   var events_path     = "#{facility_instrument_reservations_path(@instrument.facility, @instrument, format: 'js', with_details: @instrument.show_details?)}";
+  var orders_path_base = "#{facility_orders_path(@instrument.facility)}";
```
jhanggi: I think we don't want it here. The "Edit" link should not appear on the user facing views (new/edit reservation), as it will 404 for non-facility staff/admins. Where else is this partial shared?

vandrijevik: Gotcha! That’s used from `reservations/new` and `reservations/edit`, so I’ve removed it in e2989374c. `orders_path_base` is now defined in a couple of views under `facility_reservations`, and in `instruments/schedule`, which are all places that I think we want this, right?

PR #1600, opened 2018-06-20
[#139111] Do not allow ending a reservation if it is canceled

vandrijevik:
# Release Notes

Do not allow ending a reservation if it is canceled

# Additional Context

While the work in #1591 did not change the behavior of individual reservations listed on the `All` tab, the ticket flagged that a reservation that was canceled after it was started displays an `End Reservation` link that should not be there.


spec/models/reservation_spec.rb
```diff
@@ -1583,4 +1583,34 @@
       it { is_expected.to be_blank }
     end
   end
+
+  describe "#can_switch_instrument_off?" do
+    # let!(:account) { FactoryBot.create(:nufs_account, account_users_attributes: account_users_attributes_hash(user: user)) }
```
jhanggi: Can we remove this commented out code?

vandrijevik: What commented out code? _whistles, looks away in embarrassment_ 🤦‍♂️

PR #1608, opened 2018-06-26
[#140359] Fix SAML logout link error

jhanggi:
# Release Notes

Fix error on logout link when signed in via SSO.

# Additional Context

```
ActionView::Template::Error: undefined method `destroy_saml_authenticatable_with_custom_error_user_session_path
```

Broken after #1589 changed the strategy name to "saml_authenticatable_with_custom_error",
which causes a mismatch with the route. The old version is probably overly general. Within this controller, the winning strategy should always we `saml_authenticatable`.


jhanggi:
I need to get this out as it is causing problems for testing so I'm going to go ahead and merge.

vandrijevik:
Sorry for the breakage and thanks for the fix here, @jhanggi!

PR #1613, opened 2018-07-05
[#73295] Filter sub-statuses of Canceled to current facility

jhanggi:
# Release Notes

Filter sub-statuses of Canceled to current facility in the order detail modal.

# Additional Context

When an order is canceled, the order status dropdown was including all sub-statuses of Canceled, even across facilities. This filters to only the current facility.

### DIFF
--- a/app/controllers/order_management/order_details_controller.rb
+++ b/app/controllers/order_management/order_details_controller.rb
@@ -100,7 +100,7 @@ class OrderManagement::OrderDetailsController < ApplicationController
       @order_statuses = [OrderStatus.complete, OrderStatus.canceled]
       @order_statuses << OrderStatus.reconciled if @order_detail.can_reconcile?
     elsif @order_detail.order_status.root == OrderStatus.canceled
-      @order_statuses = OrderStatus.canceled.self_and_descendants
+      @order_statuses = OrderStatus.canceled.self_and_descendants.for_facility(current_facility)
     else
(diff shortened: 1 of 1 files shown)


app/controllers/order_management/order_details_controller.rb
```diff
@@ -100,7 +100,7 @@ def load_order_statuses
       @order_statuses = [OrderStatus.complete, OrderStatus.canceled]
       @order_statuses << OrderStatus.reconciled if @order_detail.can_reconcile?
     elsif @order_detail.order_status.root == OrderStatus.canceled
-      @order_statuses = OrderStatus.canceled.self_and_descendants
+      @order_statuses = OrderStatus.canceled.self_and_descendants.for_facility(current_facility)
```
vandrijevik: We have the [association-like method `Facility#order_statuses`](https://github.com/tablexi/nucore-open/blob/aec2e84d27cf7066a12e547ce79b044cbc23cd17/app/models/facility.rb#L68-L71), so maybe consider `current_facility.order_statuses.canceled.self_and_descendants`?

jhanggi: In this case, it does need to be this way. `OrderStatus.canceled` is not a scope; it resolves to a single record. `current_facility.order_statuses.canceled` results in that same single record, so when you call `self_and_decendants` on it, that scope is not filtered to a facility.

PR #1617, opened 2018-07-11
[#142184] Allow clickable links in failure message for SAML sign-in

vandrijevik:
# Release Notes

Allow clickable links in failure message for SAML sign-in

# Additional Context

If `fail!` is used with a symbol, Devise performs its own `I18n.t` lookup, which is not marked as `html_safe`. If the argument is a string, it only calls `to_s` on it, so we take advantage of that code path to pass in our own html-safe string.

### DIFF
--- a/vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
+++ b/vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
@@ -21,7 +21,7 @@ module Devise
       # failure message for authentication via SAML.
       def failed_auth(msg)
         DeviseSamlAuthenticatable::Logger.send(msg)
-        fail!(:saml_invalid)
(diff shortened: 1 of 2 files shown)


vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
```diff
@@ -21,7 +21,7 @@ class SamlAuthenticatableWithCustomError < SamlAuthenticatable
       # failure message for authentication via SAML.
       def failed_auth(msg)
         DeviseSamlAuthenticatable::Logger.send(msg)
-        fail!(:saml_invalid)
+        fail!(I18n.t("devise.failure.saml_invalid").html_safe)
```
jhanggi: It's probably a little overkill here, but we could use `text-helpers`'s `html` method and then the output could be in markdown.

PR #1618, opened 2018-07-12
Clean up factories for using FacilityAccounts

jhanggi:
# Release Notes

TECH TASK: Clean up factories to not use `attributes_for` so we can use `after(:build)` hooks for setting up account preconditions.

# Additional Context

In both NU, and now Dartmouth, accounts require validation against data in some other tables (like general ledger tables). We have a thing in NU's `NufsAccount` factory, which is admittedly weird.

```ruby
      sequence(:account_number) do |n|
        # 999-77XXXXX
        account = "901-77#{format('%05d', (n % 100_000))}" # fund3-dept7
        # Defined here as opposed to in an after_build hook
        # since some tests rely on it being defined on attributes_for
          define_gl066(account.account_number)
        account
      end
```

With this change, we can now use a `after(:build)` hook to do this.

However, this breaks the `facility.facility_accounts.create(attributes_for(:facility_account))` construct used throughout the specs. Instead, we switch to `FactoryBot.create`. I also moved some logic into the products' factories so that if they're given a facility which has a facility account already configured it'll use that by default. That allows us to not need to define the facility account in the spec at all in many cases.

We use a bunch of different styles throughout our test suite (`let` vs instance variables--we want to move towards the `let` version), so I tended to error on the side of maintaining consistency within a file and not get too bogged down in cleaning up everything.

This is extracted from https://github.com/tablexi/nucore-dartmouth/pull/206 (private repo).


spec/controllers/facility_order_details_controller_spec.rb
```diff
@@ -9,10 +9,8 @@
   before(:all) { create_users }
 
   before(:each) do
-    @authable = FactoryBot.create(:facility)
-    @facility_account = FactoryBot.create(:facility_account, facility: @authable)
+    @authable = FactoryBot.create(:setup_facility)
```
vandrijevik: Since we’re touching this here, I’d love to see the ivar renamed to `@facility` for better comprehension in the rest of this file, but that doesn’t have to happen in this PR.

spec/models/order_detail_spec.rb
```diff
@@ -14,9 +14,9 @@
   before(:each) do
     Settings.order_details.status_change_hooks = nil
     @facility = create(:facility)
-    @facility_account = @facility.facility_accounts.create(attributes_for(:facility_account))
-    @user     = create(:user)
-    @item     = @facility.items.create(attributes_for(:item, facility_account_id: @facility_account.id))
+    @facility_account = create(:facility_account, facility: @facility)
```
vandrijevik: In some places we use `FactoryBot.create`, and in others we just use `create`, which I’m assuming is available because of [this inclusion](https://github.com/tablexi/nucore-open/blob/d652e997b7c0aaece453926755965cfa9090c1c5/spec/rails_helper.rb#L34).

What do you think about switching these to `FactoryBot.create` so they’re clearer (and so we can maybe ultimately remove that inclusion because `create` is so overloaded in Rails world).

spec/controllers/facility_order_details_controller_spec.rb
jhanggi: The `controller_test_helper` methods like `it_should_allow` that are used throughout the codebase rely on it being called `@authable` (it does look like it allows the non-ivar version `authable` as well). I agree with your intuition, but that's a bigger change. Though we could call it `facility` throughout this file and have a `@authable = facility` for the sake of the helpers. Let's look into those kind of changes the next time we're in these specs.

spec/models/order_detail_spec.rb
jhanggi: I definitely agree with that sentiment. My general rule for this PR was to stay consistent within the each file to keep the number of lines changed smaller (though cleaning up some of the spacing obviously increased that...). I think that's another good thing for a future PR.

PR #1621, opened 2018-07-16
[#142643] Rake task for new price polices in the current fiscal year

jhanggi:
# Release Notes

Add a rake task for copying the all service price policies from the previous fiscal year into the current fiscal year.

# Additional Context

UIC has a facility with 200+ services. They requested that all of the
price policies be extended to 06/30/2019 (the end of the next fiscal
year). Since price policies are not allowed to span fiscal years in
NUcore, this makes new copies with the requested dates.

### DIFF
--- a/lib/tasks/price_policies.rake
+++ b/lib/tasks/price_policies.rake
@@ -1,7 +1,7 @@
 namespace :price_policies do
 
   desc "Duplicate the newest price policies for a facility's services to the current fiscal year"
-  task :duplicate_to_current_fiscal_year, [:url_name] => :environment do |t, args|
+  task :duplicate_to_current_fiscal_year, [:url_name] => :environment do |_t, args|
     facility = Facility.find_by!(url_name: args[:url_name])
 
     facility.services.active.each do |service|
--- a/lib/tasks/price_policies.rake
+++ b/lib/tasks/price_policies.rake
@@ -1,0 +1,19 @@
+namespace :price_policies do
+
+  desc "Duplicate the newest price policies for a facility's services to the current fiscal year"
+  task :duplicate_to_current_fiscal_year, [:url_name] => :environment do |t, args|
+    facility = Facility.find_by!(url_name: args[:url_name])
+
+    facility.services.active.each do |service|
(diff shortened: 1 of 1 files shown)


lib/tasks/price_policies.rake
```diff
@@ -0,0 +1,19 @@
+namespace :price_policies do
+
+  desc "Duplicate the newest price policies for a facility's services to the current fiscal year"
+  task :duplicate_to_current_fiscal_year, [:url_name] => :environment do |_t, args|
+    facility = Facility.find_by!(url_name: args[:url_name])
+
+    facility.services.active.each do |service|
+      service.price_policies.newest.each do |price_policy|
+        next unless price_policy.expired? # Don't duplicate a current policy
```
vandrijevik: The comment should maybe be `Don't duplicate an expired policy` here? 😄 Or am I missing something?

lib/tasks/price_policies.rake
```diff
@@ -0,0 +1,19 @@
...
+
+        new_price_policy = price_policy.dup
+        new_price_policy.start_date = SettingsHelper.fiscal_year_beginning
+        new_price_policy.expire_date = SettingsHelper.fiscal_year_end
+        new_price_policy.save!
```
vandrijevik: Maybe we can refactor these three lines to a single:

```
new_price_policy.update!(start_date: SettingsHelper.fiscal_year_beginning, expire_date: SettingsHelper.fiscal_year_end)
```

If you prefer the current form though, feel free to keep it as-is!

lib/tasks/price_policies.rake
jhanggi: The particular use case here is we're copying expired price policies into the current fiscal year. If the current fiscal year already has price policies, we don't want to copy them.

This is a fairly specific case, but figured putting it in the open branch would save us 15 minutes if we need to do again.

vandrijevik: Yep yep, makes sense—my brain parsed that `unless expired` backwards 🤦‍♂️

PR #1626, opened 2018-07-20
Upgrade to Circle 2.0

jhanggi:
# Release Notes

Switch to Circle 2.0 for CI.

# Additional Context

One thing I'm not sure of is what's the best thing to do with teaspoon. It's pretty fast, but right now it's running on each of the containers.

I have a similar branch in UIC that I'm working on. NU will be a tricky one because of all the oracle stuff. One thing I would love to figure out (though I worry about the build times) would be to use the workflows to test both mysql and oracle against open. Maybe it's something we only do against `master`?


.circleci/config.yml
```diff
@@ -0,0 +1,77 @@
+version: 2
+jobs:
+  build:
+    parallelism: 4
```
pturley: Does this help you?

jhanggi: What do you mean? It increases the parallelism so should split the specs over 4 containers to finish much faster. This project has an advantage in that it can use open source containers so we're not quite as constrained as we are on our private projects (where I believe we usually use 2 as our parallelism).

pturley: Ahhh I just didn’t see the circle test glob bit. :+1:

spec/rails_helper.rb
```diff
@@ -6,13 +6,6 @@
 require File.expand_path("../../config/environment", __FILE__)
 require "rspec/rails"
 require "shoulda/matchers"
-#
-# Check for engine factories. If they exist and the engine is in use load it up
-Dir[File.expand_path("vendor/engines/*", Rails.root)].each do |engine|
```
pturley: How does this change relate?

jhanggi: So, this only seems to affect UIC so far (which is where I was doing most of the work and then pulled the relevant changes up into this open-source repo). I assume it has to do with load order, likely caused by the difference between ubuntu that 1.0 ran on and the OS the `ruby` docker image this runs on. I was getting this error:

```
FactoryBot::DuplicateDefinitionError:
  Factory already registered: uic/external_account
# ./spec/rails_helper.rb:14:in `block in <top (required)>'
# ./spec/rails_helper.rb:11:in `each'
# ./spec/rails_helper.rb:11:in `<top (required)>'
# ./spec/app_support/end_reservation_only_spec.rb:1:in `<top (required)>'
```

Basically, it seemed like the `uic` engine (which is in
This change combined with a change in the `uic` engine (in their fork) equivalent to https://github.com/tablexi/nucore-open/pull/1626/files#diff-3cbe2913baf8300fd721845776e24710R46) now, inside an initializer, adds the engines' factory paths to FactoryBot's lookup paths and lets it handle the `require`s. We already were doing this in most of our newer engines. Some of our older engines like `c2po` and `uic` (and there are probably some in NU's engines we'll need to deal with downstream) didn't have this initializer setup.

Yes, it does not seem directly related to the Circle upgrade, but it wasn't working without it. And the initializer-based method of loading factories seems much more reliable.

Here's the circle build with the problem (it had some other ones, too related to yaml parsing of aliases, e.g. `&defaults`): https://circleci.com/gh/tablexi/nucore-uic/854 (restricted access).

pturley: Huh, super interesting. Thanks for the explanation

jhanggi: Yeah, we do some kind of funky stuff with engines in this project, largely to keep school-specific functionality isolated and also for feature flagging (simply by removing the line from the Gemfile). We've definitely learned a lot over the years so there are some kind of hacky things still lying around from the older ones.

PR #1632, opened 2018-07-25
Fix SecureRooms migration

jhanggi:
# Release Notes

Tech task: Fix SecureRooms migration.

# Additional Context

The superclass of SecureRooms::CardReader in the migration needs to match the
superclass of the app class. We had originally run this migration on NU's stage
before the change to ApplicationRecord. Now that we're trying to deploy to production,
the migration is failing.


jhanggi:
Vlad and I paired on this, so he said we're good to go.

PR #1607, opened 2018-06-22
#8 Create product listing quick purchase links

ColinMcNeil:
# Release Notes

Created a quick link next to facility sub-products to add each product to the cart. This link skips multiple clicks that would otherwise be necessary.

# Screenshot

image

## Not logged in
image
When a user is not logged in, the add to cart will take them to login. Note this will redirect back to facility page after successful login.

## No payment source (admin)
image
If the user has no payment source, the link will become text that reflects this.

## No permission

If for some reason none of the above apply, such as if the user doesn't have permission for a specific facility, it will say "Cannot purchase"

PR #1635, opened 2018-07-26
Both quicklink

ColinMcNeil:
# Release Notes

Quick links for the facility page to purchase and view items. 

We previously submitted a Pull Request #1607, which introduced our purchase quicklink. Our final design has moved the hyperlink to view facility into its own button, which goes next to the purchase quick link. Here is our final quicklink design (localized as well!)

# Screenshot
image

# Additional Context

The behavior of the "Add to Cart" is the same as in #1607

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -44,7 +44,6 @@ class FacilitiesController < ApplicationController
     @order_form = Order.new if acting_user && current_facility.accepts_multi_add?
     @active_tab = "home"
     set_column_class
-    set_quicklinks
     render layout: "application"
   end
 
@@ -255,8 +254,9 @@ class FacilitiesController < ApplicationController
     @columns = "columns" if SettingsHelper.feature_on?(:product_list_columns)
   end
 
-  def set_quicklinks
-    @quicklinks = SettingsHelper.feature_on?(:product_quicklinks)
+  def quicklinks?
+    SettingsHelper.feature_on?(:product_quicklinks)
   end
+  helper_method :quicklinks?
 
 end
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -44,6 +44,7 @@ class FacilitiesController < ApplicationController
     @order_form = Order.new if acting_user && current_facility.accepts_multi_add?
     @active_tab = "home"
     set_column_class
+    set_quicklinks
     render layout: "application"
   end
 
@@ -252,6 +253,10 @@ class FacilitiesController < ApplicationController
   def set_column_class
     @columns = ""
     @columns = "columns" if SettingsHelper.feature_on?(:product_list_columns)
+  end
+
+  def set_quicklinks
+    @quicklinks = SettingsHelper.feature_on?(:product_quicklinks)
   end
 
 end
--- a/app/views/facilities/_product_list.html.haml
+++ b/app/views/facilities/_product_list.html.haml
@@ -11,7 +11,7 @@
       - indent_for_icon = products.any? { |p| p.respond_to? :reservations }
       - products.sort.each do |product|
         .product_listing
-          %li{ class: product.class.model_name.to_s.downcase }
+          %li{ class: product.class.model_name.to_s.downcase + '_title' }
             = product.name
             - if local_assigns[:f]
               = f.fields_for :order_details do |builder|
@@ -29,7 +29,7 @@
                   No payment source
               - else
                 Cannot purchase
-            .view_product
+            %li{class: [product.class.model_name.to_s.downcase, 'view_product']}
               = public_calendar_link(product, indent: indent_for_icon)
               = link_to 'View' + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product)
               - if acting_user.present? && !product.can_be_used_by?(acting_user)
--- a/app/views/facilities/_product_list.html.haml
+++ b/app/views/facilities/_product_list.html.haml
@@ -10,17 +10,29 @@
     %ul
       - indent_for_icon = products.any? { |p| p.respond_to? :reservations }
       - products.sort.each do |product|
-        %li{ class: product.class.model_name.to_s.downcase }
-          - if local_assigns[:f]
-            = f.fields_for :order_details do |builder|
-              - if session_user.can_override_restrictions?(product) || product.can_be_used_by?(acting_user)
-                = builder.text_field :quantity, value: 0, class: "product_quantity", index: nil
-                = builder.hidden_field :product_id, value: product.id, index: nil
-          = public_calendar_link(product, indent: indent_for_icon)
-          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product)
-          - if acting_user.present? && !product.can_be_used_by?(acting_user)
-            %i.fa.fa-lock
-            = " (#{product.class.human_attribute_name(:requires_approval_show)})"
+        .product_listing
+          %li{ class: product.class.model_name.to_s.downcase }
+            - if local_assigns[:f]
+              = f.fields_for :order_details do |builder|
+                - if session_user.can_override_restrictions?(product) || product.can_be_used_by?(acting_user)
+                  = builder.text_field :quantity, value: 0, class: "product_quantity", index: nil
+                  = builder.hidden_field :product_id, value: product.id, index: nil
+                  
+            = public_calendar_link(product, indent: indent_for_icon)
+            = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product)
+            - if acting_user.present? && !product.can_be_used_by?(acting_user)
+              %i.fa.fa-lock
+              = " (#{product.class.human_attribute_name(:requires_approval_show)})"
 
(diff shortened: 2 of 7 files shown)


ColinMcNeil:
In case you were wondering about circleci failing, there are 0 failures and all the tests passed. It was simply a malformed xml file for the test results.

image

jhanggi:
I'll try to look at this later today. We've been seeing that error since we switched to circle 2.0. PhantomJS (a javascript/headless browser which we use for tests) is crashing while running javascript/teaspoon/jasmine tests. If you rebuild in CI likely it'll go away. It's happening frequently enough that we are planning to look into it.

ColinMcNeil:
Neither @robstolarz nor I have write access needed to rebuild :/

app/controllers/facilities_controller.rb
```diff
@@ -254,4 +255,8 @@ def set_column_class
     @columns = "columns" if SettingsHelper.feature_on?(:product_list_columns)
   end
 
+  def set_quicklinks
```
jhanggi: If you don't want to inline the `feature_on?` in the view, I think the view would read cleaner if this were:
```ruby
def quicklinks?
  SettingsHelper.feature_on?(:product_quicklinks)
end
helper_method :quicklinks?
```

Then you can do `if quicklinks?` inside the view.

app/views/facilities/_product_list.html.haml
```diff
@@ -10,17 +10,16 @@
...
-                = builder.hidden_field :product_id, value: product.id, index: nil
-          = public_calendar_link(product, indent: indent_for_icon)
-          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product)
-          - if acting_user.present? && !product.can_be_used_by?(acting_user)
-            %i.fa.fa-lock
-            = " (#{product.class.human_attribute_name(:requires_approval_show)})"
-
-          - if product.offline?
-            = tooltip_icon "fa fa-exclamation-triangle icon-large", t("instruments.offline.note")
+        .product_listing
```
jhanggi: Is this additional class necessary?

app/views/facilities/_quicklinks.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  .add_to_cart
+    - if not session_user
+      = link_to "#{t("admin.shared.add_to",model:t("pages.cart"))} (#{t("pages.login")})", new_user_session_path
+    - elsif session_user.can_override_restrictions?(product) || product.can_be_used_by?(acting_user)
+      - if not session_user.accounts.empty?
+        = link_to t("admin.shared.add_to",model:t("pages.cart")), add_order_path(acting_user.cart(session_user), order: {order_details: [{product_id: product.id, quantity: 1}] }), method: :put
+      - else
+        = t("facility_accounts.account_table.notice")
+    - else
+      = t("facility_account_orders.cannot_purchase")
```
jhanggi: I would recommend putting the translation at `views.facilities.quicklinks` and then you should be able to access it with `text("quicklinks")` because the scope will be inferred from the partial's filename.

app/views/facilities/_quicklinks.html.haml
```diff
@@ -0,0 +1,21 @@
+.quick_links 
+  .add_to_cart
+    - if not session_user
+      = link_to "#{t("admin.shared.add_to",model:t("pages.cart"))} (#{t("pages.login")})", new_user_session_path
+    - elsif session_user.can_override_restrictions?(product) || product.can_be_used_by?(acting_user)
+      - if not session_user.accounts.empty?
```
jhanggi: In general, we prefer not to double negative (counting `empty?` as a negative). So this I would probably do as `if session_user.accounts.any?`. If you really want to use `empty?`, then I would do `unless session_user.accounts.empty?`

app/views/facilities/_quicklinks.html.haml
jhanggi: What would you think of wrapping this in a in a `.label.label-important` class. (http://getbootstrap.com/2.3.2/components.html#labels-badges)

app/views/facilities/_quicklinks.html.haml
```diff
@@ -0,0 +1,21 @@
...
+  %li{class: [product.class.model_name.to_s.downcase, 'view_product']}
+    = public_calendar_link(product, indent: indent_for_icon)
+    = link_to 'View' + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product), :id => product.name
+    - if acting_user.present? && !product.can_be_used_by?(acting_user)
+      %i.fa.fa-lock
+      = " (#{product.class.human_attribute_name(:requires_approval_show)})"
+
+      - if product.offline?
+        = tooltip_icon "fa fa-exclamation-triangle icon-large", t("instruments.offline.note")
+
```
jhanggi: Can you enable your text editor to automatically add a blank newline to the end of files? In sublime the setting is `"ensure_newline_at_eof_on_save": true`,

app/views/facilities/_product_list.html.haml
```diff
@@ -10,17 +10,16 @@
...
+          %li{ class: product.class.model_name.to_s.downcase + '_title' }
```
jhanggi: This logic was already pretty gnarly, so adding more to it is going to make it even more complex. What would you think of extracting the default behavior out of this file. So we end up with something like this:
```
- products.sort.each do |product|
  - if quicklinks?
    = render "quicklinks", product: product
  - else
    = render "default", product: product
```
That would also allow easier side-by-side comparison of the two views if we can't get them DRYed up.

app/views/facilities/_quicklinks.html.haml
```diff
@@ -0,0 +1,21 @@
...
+    - else
+      = t("facility_account_orders.cannot_purchase")
+  %li{class: [product.class.model_name.to_s.downcase, 'view_product']}
+    = public_calendar_link(product, indent: indent_for_icon)
+    = link_to 'View' + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product), :id => product.name
+    - if acting_user.present? && !product.can_be_used_by?(acting_user)
+      %i.fa.fa-lock
+      = " (#{product.class.human_attribute_name(:requires_approval_show)})"
+
+      - if product.offline?
```
jhanggi: Not sure why, but this isn't rendering for me.

ColinMcNeil: How can we replicate it? I'm using the existing logic, but we don't have a locked facility to test with. If we could set it up the same way we can fix it.

jhanggi: You can mark an instrument offline by going into the products tab from manage facilities, choose an instrument. You should be taken to the "Reservations". One of the buttons should be "Mark Instrument Offline". Put in a note and a category and the instrument will be considered offline. This feature is meant for unplanned maintenance and issues.

Once you've marked it offline, on the product listing on the "use" side, you should see a red triangle with an exclamation mark.

ColinMcNeil:
These should be most of the changes. Thank you for the logic suggestions, it definitely is a big improvement.

app/views/facilities/_product_list.html.haml
```diff
@@ -10,17 +10,15 @@
...
-          = link_to product.name + (product.is_hidden? ? ' (hidden)' : ''), facility_product_path(current_facility || product.facility, product)
-          - if acting_user.present? && !product.can_be_used_by?(acting_user)
-            %i.fa.fa-lock
-            = " (#{product.class.human_attribute_name(:requires_approval_show)})"
-
-          - if product.offline?
-            = tooltip_icon "fa fa-exclamation-triangle icon-large", t("instruments.offline.note")
+        - if quicklinks?
+          %li{ class: product.class.model_name.to_s.downcase + '_title' }
+          = render :partial => "facilities/quicklinks", locals: {product: product, indent_for_icon: indent_for_icon}
```
jhanggi: You can also write this as `render "facilities/quicklinks", product: product, indent_for_icon: indent_for_icon`. I don't feel too strongly about that vs the with keys, but if you do leave it this way, can we do `partial: ` for the hash syntax and add spaces inside the `locals` hash: `= render partial: "facilities/quicklinks", locals: { product: product, indent_for_icon: indent_for_icon} `

app/views/facilities/_product_list.html.haml
```diff
@@ -10,17 +10,15 @@
...
+        - else
```
jhanggi: I think some of the original code got lost here. The `else` part of this should be the original code, correct? We've definitely lost all of the icons, and the order form input is now to the right of the name where previously it was to the left.

app/views/facilities/_quicklinks.html.haml
jhanggi: Looks like the indentation is wrong. This is nested in the `- if acting_user.present? && !product.can_be_used_by?(acting_user)`. It should be a separate piece of logic.

app/views/facilities/_quicklinks.html.haml
```diff
@@ -0,0 +1,20 @@
+.quick_links 
+  .add_to_cart
```
jhanggi: Looks like the product name got lost.

screen shot 2018-07-30 at 11 50 21 am

This is also happening on the recent products view on the homepage.

ColinMcNeil:
Dan has made the call that this is not a priority for us atm.

PR #1643, opened 2018-08-03
Changes to typography

robstolarz:
# Release Notes
As we discussed earlier, we'll start to make some changes to spacing and typography. I'm reluctant to start making massive changes to spacing until some of the UI elements begin to change, except where improper spacing makes the site unusable.

This pull request modifies the typography of the site to fit the specification we shared.

# Screenshots

### Before
facilities-nucore-upstream
### After
facilities-design-rework-typography

### Before
facilities_example-nucore-upstream
### After
facilities_example-design-rework-typography

# Additional Context
If any spacing looks egregiously incorrect, let me know and I will fix it. I'm just reluctant to make every change in a single PR.


app/assets/stylesheets/app/typography.scss
```diff
@@ -22,17 +22,36 @@ h1, h2, h3, h4, h5 {
...
+}
+
+h2 { /* "instrument title" */
+  font-size: 2.417em;
+}
+
+h3, th { /* "static title" */
   margin-bottom: 0;
+  font-weight: bold;
+  font-size: 1em;
```
jhanggi: I think it might be worth going around the "manage" section for various places we use `h3`s. There are a few places where this now doesn't seem to have enough visual weight. There are already `h1` (facility name) and `h2` (product name) on some pages, so `h3` does seem to be the correct semantically.

Examples:
screen shot 2018-08-10 at 7 17 43 pm
screen shot 2018-08-10 at 7 17 26 pm

app/assets/stylesheets/app/typography.scss
```diff
@@ -1,5 +1,5 @@
-a {
-  text-decoration: underline;
+a { /* "links" */
+  text-decoration: none;
```
jhanggi: I go back and forth where I usually like my links underlined so it's obvious what's clickable. Especially for people who might not be able to distinguish color as well. It's less of a thing in listings and menu bars. Definitely within a sentence/paragraph I think it's critical. Maybe we can add `p a` is underlined so that the link is underlined if it's within a paragraph?

app/assets/stylesheets/app/typography.scss
jhanggi: screen shot 2018-08-10 at 7 21 04 pm

robstolarz:
Sorry for not getting back to you on this, deployment stuff has taken precedence. I agree that some changes need to be made.

>One other question: I'm seeing flashes of blank content. I believe that's while the font is loading. Is that just a development environment thing or do you think we'll have something like that once we go to prod?

It might have something to do with the app server having to serve the files, and it not being the fastest thing. Additionally, there's a great chance the fonts aren't cached in dev. That shouldn't be an issue in the future, I don't think... but I haven't tested it in production mode yet.

> [the `p a` thing]

I think that sounds reasonable. Can you point me at some examples of the inline links?

> I've been trying to sit with it a little bit to overcome my initial reaction of "I don't like it simply because it's different"

I get that, really I do.

The design prescribes that the Instrument Reservations page look like this (although Preview has somehow managed to export the colors wrong)
2 handfuls of apple sauce page 34

As you can see, it's pretty obvious that the spacing is completely wrong in this branch. I ignored it temporarily, to make the typography changes easier to review. This worked from a code perspective, but it means the branch doesn't quite match the proposed design and that the screenshots will look wrong. I think that might be why you don't like it (but please do correct me if I'm wrong).

There are a lot of spacing changes to be made. Our spacing packet is just under fifty pages long. In this context, how should we try and assemble and PR our changes so that they're possible to review?

jhanggi:
I'm having trouble finding a good example `p a` thing--several of the places where we have it might not really be the right semantics (e.g. the XLS/CSV links for journals). The place I was primarily thinking about it is in facility and product descriptions where the facilities can add their own links through the wysiwyg editor.

I'm not that worried about the flash of blank content just yet. We can see how it looks once it's deployed.

I'm ok if some of the spacing is off for a little bit: I know y'all have put a lot of thought into that and it'll take a little while to get perfect. I was initially concerned with the header sizes on the manage side thinking it was an accidental byproduct, but if that's the direction you're going, I support it. What does your design for the order detail modal look like? That header in particular feels like it should be bigger (it might just be we want to change it to an `h2`).

oliviacrosby:
manage billing order detail v2

ColinMcNeil:
Should be okay now.
screen shot 2018-09-25 at 10 32 33 am

ColinMcNeil:
Switched to using our colors:
screen shot 2018-09-25 at 10 55 05 am

jhanggi:
I just pulled the latest from master into `tablexi/uconn-design-updates`. Possibly rebase off that branch?

app/assets/stylesheets/nucore-overrides.scss
```diff
@@ -9,3 +9,21 @@ Example of this to include here:
...
 */
+
+// Shared
+$darkGrey: #3F3B26;
+$grey: #9B9B9B;
+$lightGrey: #D4D1D1;
+$lighterGrey: #F3F3F3;
+
+// Manage Mode
+$orange: #FF7126;
```
jhanggi: Looks like the use/manage variables aren't used yet. Just want to make sure on that.

app/assets/stylesheets/nucore-overrides.scss
```diff
@@ -9,3 +9,21 @@ Example of this to include here:
   $headerBackgroundColor: #571963;
   $headerColor: white;
 */
+
+// Shared
```
jhanggi: I feel like these variables could probably move to `nucore-variables.scss`. All of these seem like more generic variables for colors, so there's no harm in keeping them around. Usually, we like to reserve this file for changes in the fork and leave this empty in `nucore-open`.

One thing to think about while use/manage mode is if there's a way to define and use variables like `$manageColor`, `$manageLightColor` (or ideally better names). Then you would have something like `$manageColor: $orange`, but other schools could use a different color scheme but still be able to use the use/manage feature.

app/assets/stylesheets/nucore-overrides.scss
```diff
@@ -9,3 +9,21 @@ Example of this to include here:
...
+$darkGrey: #3F3B26;
+$grey: #9B9B9B;
+$lightGrey: #D4D1D1;
+$lighterGrey: #F3F3F3;
+
+// Manage Mode
+$orange: #FF7126;
+$orange: #F89662;
```
jhanggi: Looks like you're defining `$orange` twice. Maybe one is supposed to be `$darkOrange`?

app/assets/stylesheets/app/font.scss
```diff
@@ -0,0 +1,25 @@
+/* define fonts */
+@font-face {
```
jhanggi: Can you provide a link to where the font comes from? (I believe https://rsms.me/inter/). I had a fear that maybe it had some sort of commercial license. Looks like we're all good there, though.

ColinMcNeil: @robstolarz might have to answer this one.

app/assets/stylesheets/nucore-overrides.scss
ColinMcNeil: This is correct. Currently they are being used on our use/manage CSS branch, but we want to minimize the redundancy and therefore think it is best to include them here. Assuming this branch is the first of our design changes to get merged, it is the best candidate for this commit.

app/assets/stylesheets/nucore-variables.scss
```diff
@@ -20,3 +20,35 @@ $sidebarActiveBackgroundColor: $linkColor;
 $sidebarActiveBackgroundColorHighlight: darken($sidebarActiveBackgroundColor, 5%);
 
 $sidebarActiveColor: $white;
+
+/*
```
jhanggi: Can we move this comment back to `nucore-overrides.scss`. The rest of this looks good.

ColinMcNeil:
Added the license and re-added the comment.

PR #1665, opened 2018-08-21
Rubocop: Frozen string literal

jhanggi:
# Release Notes

Tech task: Enable frozen string literal cop and apply.

# Tech notes

I'll rebase this off of master once #1663 is in since the `.rubocop_todo.yml` would conflict otherwise, but the last two commits are the main ones.


.rubocop_todo.yml
```diff
@@ -585,12 +581,24 @@ Style/FormatStringToken:
...
 
-# Offense count: 1343
+# Offense count: 12
 # Cop supports --auto-correct.
 # Configuration parameters: EnforcedStyle.
 # SupportedStyles: when_needed, always, never
 Style/FrozenStringLiteralComment:
-  Enabled: false
+  Exclude:
+    - 'public/files/000/000/001/original/alpha_survey.rb'
```
jhanggi: I think these are some local files that I had uploaded as order forms or something. These probably shouldn't be here.

PR #1666, opened 2018-08-23
[#136050] Send email notification to facility on cancellation

jhanggi:
# Release Notes

Send email notifications when a user cancels their reservation. This is optional per product. Create a new "Notifications" tab for the product to contain notification preferences (moved from the Manage/Edit product page).

# Screenshot

screen shot 2018-08-23 at 4 54 18 pm


db/schema.rb
```diff
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20180406211501) do
+ActiveRecord::Schema.define(version: 20180822191538) do
```
jhanggi: @vandrijevik can you try running the migration on your machine and see what sort of output you get? ran migrations from scratch to get this output because my primary database had even more noise (every table now got a `COLLATE=utf8_unicode_ci` added into the `options`). And weirdly, I needed to run `rake db:migrate` twice to get both the main app migrations and the engine migrations.

config/locales/en.models.yml
```diff
@@ -252,6 +252,7 @@ en:
         user_notes_field_mode: Users may submit a note when purchasing?
         user_notes_label: Label for Notes Field
         order_notification_recipient: Order Notification Recipient
+        cancellation_notification_contacts: Cancellation Notification Contacts
```
vandrijevik: I think we should call this `Recipient`, to be consistent with the other fields on this page, and we should probably change it to be plural (`Recipients`) on all fields?

config/routes.rb
```diff
@@ -62,6 +62,10 @@
     resources :products, only: [:index] do
       resources :product_accessories, only: [:index, :create, :destroy], path: "accessories"
       resources :training_requests, only: [:new, :create] if SettingsHelper.feature_on?(:training_requests)
+
+      get "notifications", to: "product_notifications#index", as: "notifications"
+      get "notifications/edit", to: "product_notifications#edit", as: "edit_notifications"
+      post "notifications", to: "product_notifications#update"
     end
```
vandrijevik: How about we use resource-based routes for these? It’s a brand-new resource 😄

db/schema.rb
vandrijevik: Running `rake db:migrate` for me results in a diff for `db/schema.rb` on this branch, which includes a bunch of lines that would add `COLLATE=utf8_unicode_ci"`. I am running MySQL 5.7.23 locally.
 
Running `rake db:migrate` a second time for me did not do anything additional compared to the first run. I don’t see any migrations in `vendor/engines` in this PR, though, so I am not sure which engine migrations you are referring to?

config/locales/en.models.yml
jhanggi: I wasn't sure which way to go. We have "Training Request Contacts" and "Order Notification Recipient". Making this change definitely highlighted the fact that those two are inconsistent functionally: training requests allows a comma separated list of emails while order notification only allows one. I felt changing that behavior was outside the scope of this.

I think I do prefer Recipients over Contacts. I'll make that change, but probably leave "Order Notification Recipient" singular.

config/routes.rb
jhanggi: I started that way and it wasn't working the way I wanted it to. `edit`/`update` need IDs. Conceptually, I could see each of these fields being their own "resource", but we're bulk editing them on a single page.

I could use `new`/`create` to get the routes working, but that didn't feel right.

What about 
```ruby
     resources :product_notifications, only: :index, path: "notifications", as: "notifications" do
        get :edit, on: :collection
        post :update, on: :collection
      end
```

What do you think?

vandrijevik: I was thinking that we could use a singleton resource here, which doesn’t need ids for edit and update. Given that a product has one set of notifications it may send, it seems that the following would fit the bill:

```
resource :product_notifications, only: [:show, :edit, :update]
```

Since we’re editing all the notifications on a single page, if we treat “the set of notifications tied to this product” as the resource, I think this gets us exactly what we want.

jhanggi: Ah, good call. I forgot about the singleton `resource` option.

PR #1670, opened 2018-08-28
[#142448] Cross facility bulk email

jhanggi:
# Release Notes

Enable cross-facility bulk email for global admins. This also enables several of the Users tab features such as Payment Sources, and Secure Room card numbers. Access List remains disabled in cross-facility mode. 

# Screenshot

screen shot 2018-08-28 at 11 30 24 am

# Additional Context

I decided to spend the extra time to get these other features to work as well since their tab gets opened up when I enable Bulk Email for global admins. Before, they had the global User tab disabled unless they were also a billing admin. Access List seemed outside the scope of this work to ensure it works properly with scheduling groups and it would probably get too unwieldy with every single product on the page.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -97,6 +97,9 @@ class UsersController < ApplicationController
 
   # GET /facilities/:facility_id/users/:user_id/access_list
   def access_list
+    # Unsupported in cross-facility mode
+    raise ActiveRecord::RecordNotFound if current_facility.cross_facility?
+
     @facility = current_facility
     @products_by_type = Product.for_facility(@facility).requiring_approval_by_type
     @training_requested_product_ids = @user.training_requests.pluck(:product_id)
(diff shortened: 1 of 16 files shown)


app/controllers/users_controller.rb
```diff
@@ -97,13 +97,19 @@ def show
 
   # GET /facilities/:facility_id/users/:user_id/access_list
   def access_list
+    # Unsupported in cross-facility mode
+    raise ActiveRecord::RecordNotFound if current_facility.cross_facility?
```
jhanggi: I chose to do this in the controller rather than the Ability since I want it entirely disabled rather than just disabled for some roles.

app/models/serializable_facility.rb
```diff
@@ -0,0 +1,17 @@
+class SerializableFacility < SimpleDelegator
```
jhanggi: This seemed like a fairly clean way of making the cross-facility (i.e. Facility with a nil ID) serializable by ActiveJob. I'm definitely open to other ways. I considered just passing IDs, but that would involve extra logic in the Job deciding how to look it up anyways.

PR #1683, opened 2018-09-06
Implement Use/Manage Separation

ColinMcNeil:
# Release Notes

Implementation of our Use/Manage design separation. Currently, "Use" will bring the user back to the facility list home page unless they are managing a facility. If they are an admin and managing a facility, then they will be taken to the facility-specific homepage with the product list.

Manage mode will take the user to the manageable facility list, however it could also be a dropdown such as in our original design.

# Screenshot
screen shot 2018-09-06 at 9 47 16 am

### DIFF
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -43,9 +43,9 @@ class NavTab::LinkCollection
   end
 
   def home_button
-    if SettingsHelper.feature_on?(:use_manage) 
+    if SettingsHelper.feature_on?(:use_manage)
       use_button
-    else 
+    else
       home
     end
   end
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -27,7 +27,7 @@ class NavTab::LinkCollection
   end
 
   def admin
-    use_button + admin_only
+    [use_button] + admin_only
   end
 
   def customer
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -27,7 +27,11 @@ class NavTab::LinkCollection
   end
 
   def admin
-    [home_button] + [manage] + admin_only
+    if SettingsHelper.feature_on?(:use_manage)
+      [home_button] + [manage] + admin_only
+    else
+      [home_button] + admin_only
+    end
   end
 
   def customer
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -145,8 +145,6 @@ class NavTab::LinkCollection
   end
 
   def manage
-    # url = facility ? dashboard_facility_path(facility) : root_path
-    # model = facility
     NavTab::Link.new(text: I18n.t("pages.manage", model: Facility.model_name.human(count: 2)), url: list_facilities_url)
   end
 
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -27,11 +27,11 @@ class NavTab::LinkCollection
   end
 
   def admin
-    default + admin_only
+    use_button + admin_only
   end
 
   def customer
-    default + [orders, reservations, payment_sources, files]
+    default
   end
 
   def default
@@ -39,7 +39,12 @@ class NavTab::LinkCollection
   end
 
   def use_button
-    [use]
+    [use] + [manage]
+  end
+
+  def home_button
+    return use_button if SettingsHelper.feature_on?(:use_manage)
+    return [home]
   end
 
   private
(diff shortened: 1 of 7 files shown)


app/presenters/nav_tab/link_collection.rb
```diff
@@ -27,15 +27,24 @@ def self.tab_methods
...
-    [home]
+    [orders, reservations, payment_sources, files]
+  end
+
+  def use_button
+    [use] + [manage]
+  end
+
+  def home_button
+    return use_button if SettingsHelper.feature_on?(:use_manage)
```
jhanggi: Can we do this as an if/else? If we do the guard clause here then the `return` on the following line is not necessary.

app/views/shared/nav/_nav_links.html.haml
```diff
@@ -2,6 +2,9 @@
   - if responsive?
     -# .hidden-with-nav is hidden > 979px
     %li.hidden-with-nav= link_to t('pages.cart') + " (#{current_cart.order_details.count})", :cart
+    %li{class: [home_button[0].tab_class(controller), "useButton"], id: home_button[0].tab_id }
```
jhanggi: Is it possible to avoid needing to reference this by index?

app/views/shared/nav/_nav_links.html.haml
```diff
@@ -14,13 +17,14 @@
...
+  
 
 -# .visible-with-nav is visible > 979px
 = form_tag global_search_path, class: "navbar-search pull-right visible-with-nav" do
   = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2"
 
 -# trailing "<" enables use of .nav-collapse .nav:empty in boostrap-responsive-overrides.scss
 %ul.nav.pull-right<
-  = render "shared/nav/manage_facilities"
+
```
jhanggi: Can we use the feature flag to leave this where it was? I think moving it will make sense once some of the other design changes are in (making use and manage pills and the rest tabs), but for now, I think it would confuse people if it just moves (we know why because we know what's coming later, but the users would probably consider it arbitrary).

app/presenters/nav_tab/link_collection.rb
```diff
@@ -27,15 +27,25 @@ def self.tab_methods
   end
 
   def admin
-    default + admin_only
+    use_button + admin_only
```
jhanggi: Since `use_button` is now returning just a single element, you can't `+` it. We're getting a bunch of test failures.

app/presenters/nav_tab/link_collection.rb
```diff
@@ -27,15 +27,25 @@ def self.tab_methods
...
-    [home]
+    [orders, reservations, payment_sources, files]
+  end
+
+  def use_button
+    use
+  end
+
+  def home_button
+    if SettingsHelper.feature_on?(:use_manage) then use_button
```
jhanggi: It's nitpicky, but I'm not really a fan of this style. We don't tend to use the `then` construct.

Can we either do:

```ruby
if SettingsHelper.feature_on?(:use_manage)
  use_button
else
  home
end
```

or

```ruby
SettingsHelper.feature_on?(:use_manage) ? use_button : home
```

jhanggi:
Also, can you run `rubocop -a` over everything? I'll look into getting you access to codeclimate so you can see the problems.

ColinMcNeil:
It passes! 🎉

ColinMcNeil:
This should be everything!

app/presenters/nav_tab/link_collection.rb
```diff
@@ -131,8 +139,15 @@ def general_reports
...
 
-  def home
-    NavTab::Link.new(tab: :home, url: root_path)
+  def use
+    url = facility ? facility_path(facility) : root_path
+    NavTab::Link.new(tab: :use, url: url)
+  end
+
+  def manage
+    # url = facility ? dashboard_facility_path(facility) : root_path
```
jhanggi: Can these comments be removed?

ColinMcNeil:
The commented out code is what it will look like if we use a dropdown. I removed it.

app/views/shared/nav/_nav_links.html.haml
```diff
@@ -14,13 +19,15 @@
...
+  
 
 -# .visible-with-nav is visible > 979px
 = form_tag global_search_path, class: "navbar-search pull-right visible-with-nav" do
   = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2"
 
 -# trailing "<" enables use of .nav-collapse .nav:empty in boostrap-responsive-overrides.scss
 %ul.nav.pull-right<
-  = render "shared/nav/manage_facilities"
+  - if not SettingsHelper.feature_on?(:use_manage)
```
jhanggi: Not a big deal, but we tend to use `!` over `not`.

app/views/shared/nav/_nav_links.html.haml
```diff
@@ -2,6 +2,11 @@
   - if responsive?
     -# .hidden-with-nav is hidden > 979px
     %li.hidden-with-nav= link_to t('pages.cart') + " (#{current_cart.order_details.count})", :cart
+    %li{class: [home_button.tab_class(controller), "useButton"], id: home_button.tab_id }  
+      = home_button.to_html
+    - if SettingsHelper.feature_on?(:use_manage)
```
jhanggi: I believe the extra "Manage Facilities" link is coming from the `navigation_links` list.

ColinMcNeil:
That's very strange. This is my view when feature is off:


screen shot 2018-09-18 at 8 04 42 am

ColinMcNeil:
Oh, I see it. I keep forgetting the admin only links appear once you're managing a facility. Fixed.

PR #1685, opened 2018-09-06
[#137154] Email user when an order goes into problem queue

jhanggi:
# Release Notes

Notify the user via email if their reservation/occupancy enters into the problem queue. 

# Additional Context

We do not trigger the email for missing price policies as that is something the facility needs to address.

For reservations, this only applies to timer and relay based instruments.

* Reservation never started, 12 hours after it was scheduled to end
* Reservation started, but never ended, 12 hours after it was scheduled to end
* Reservation started, but never ended and another reservation is started. Applies across a shared scheduled.
* Logged out via the Auto Logout setting (relays only)
* Reservation started and not yet ended and instrument is marked offline. Applies across a shared schedule.

For occupancies, this only applies to normal users. Facility staff does not get orders associated with their occupancies so they will never go to the problem queue.

* Scan into a room, do not scan out, 12 hours after the scan in
* Scan into a room, scan in again. First scan's order goes into the problem queue
* Scan out without an associated scan in

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -32,7 +32,7 @@ class Reservation < ApplicationRecord
 
   # Used when we want to force the order to complete even if it doesn't meet the
   # requirements of order_completeable?, e.g. the reservation time isn't over yet.
-  attr_accessor :force_completable
+  attr_accessor :force_completion
 
   # Delegations
   #####
@@ -151,7 +151,7 @@ class Reservation < ApplicationRecord
 
   # Is there enough information to move an associated order to complete/problem?
   def order_completable?
-    force_completable || actual_end_at || reserve_end_at < Time.current
+    force_completion || actual_end_at || reserve_end_at < Time.current
   end
 
   def start_reservation!
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -155,8 +155,11 @@ class Reservation < ApplicationRecord
   end
 
   def start_reservation!
-    # Mark running reservations as complete, which will move them to problem orders
-    product.schedule.products.flat_map(&:started_reservations).each(&:complete!)
+    # If there are any reservations running over their time on the shared schedule,
+    # kick them over to the problem queue.
+    product.schedule.products.flat_map(&:started_reservations).each do |reservation|
+      MoveToProblemQueue.move!(reservation.order_detail)
+    end
     update!(actual_start_at: Time.current)
   end
 
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -30,6 +30,10 @@ class Reservation < ApplicationRecord
   # used for overriding certain restrictions
   attr_accessor :reserved_by_admin
 
+  # Used when we want to force the order to complete even if it doesn't meet the
+  # requirements of order_completeable?, e.g. the reservation time isn't over yet.
+  attr_accessor :force_completable
+
   # Delegations
   #####
   delegate :note, :note=, :ordered_on_behalf_of?, :complete?, :account, :order,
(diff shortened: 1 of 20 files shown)


spec/mailers/previews/problem_order_mailer_preview.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class ProblemOrderMailerPreview < ActionMailer::Preview
+
+  def notify_user
+    order_detail = NUCore::Database.random(Reservation.user.joins(:order_detail).merge(OrderDetail.complete)).order_detail
```
vandrijevik: Instead of `random`, why don’t we move these previews to use `first` and get rid of the method entirely? That way they’d show a predictable email for previewing, too.

app/models/reservation.rb
```diff
@@ -147,12 +151,15 @@ def end_at_required?
...
+    force_completable || actual_end_at || reserve_end_at < Time.current
   end
 
   def start_reservation!
-    # Mark running reservations as complete, which will move them to problem orders
-    product.schedule.products.flat_map(&:started_reservations).each(&:complete!)
+    # If there are any reservations running over their time on the shared schedule,
+    # kick them over to the problem queue.
+    product.schedule.products.flat_map(&:started_reservations).each do |reservation|
+      MoveToProblemQueue.move!(reservation.order_detail)
```
vandrijevik: Should we set `force_complete` here as well, to ensure that we can kick them over?

app/services/move_to_problem_queue.rb
```diff
@@ -0,0 +1,20 @@
...
+  def self.move!(order_detail)
+    new(order_detail).move!
+  end
+
+  def initialize(order_detail)
+    @order_detail = order_detail
+  end
+
+  def move!
+    @order_detail.complete!
```
vandrijevik: I kinda think that inside this class we should set `@order_detail.force_completion = true` and call `@order_detail.complete!`. This would ensure that `move!` can succeed, and the call sites of `MoveToProblemQueue` do not need to know about the internal implementation detail `force_completion` of a problem order—that’s encapsulated by this class.

app/models/reservation.rb
```diff
@@ -30,6 +30,10 @@ class Reservation < ApplicationRecord
   # used for overriding certain restrictions
   attr_accessor :reserved_by_admin
 
+  # Used when we want to force the order to complete even if it doesn't meet the
+  # requirements of order_completeable?, e.g. the reservation time isn't over yet.
+  attr_accessor :force_completable
```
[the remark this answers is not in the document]
jhanggi: Oh, yeah, definitely a better name.

spec/mailers/previews/problem_order_mailer_preview.rb
jhanggi: I personally like the non-predictability here. It's saved me from missing some null pointers and whatnot. Not a big deal if you feel strongly about it.

app/models/reservation.rb
jhanggi: Probably a good idea.

app/services/move_to_problem_queue.rb
jhanggi: Word. Good call.

spec/mailers/previews/problem_order_mailer_preview.rb
vandrijevik: I don’t care very much about it, so fine by me! (I would tend towards getting rid of the method on `NUCore::Database` myself, but this is okay, too)

app/models/reservation.rb
jhanggi: Hmm, the test failure appears related to this seems legit. I'm thinking this one should not be `force_complete`. I'll need to dig into it a little more.

"ReservationsController needs now reservation switch_instrument on before the reservation start (in grace period) in the grace period, but there is another reservation still running does not start the reservation - spec.controllers.reservations_controller_spec"

PR #1689, opened 2018-09-10
[#143484] Deny card scan if a matching user is not found

vandrijevik:
# Release Notes

Deny card scan if a matching user is not found


vendor/engines/secure_rooms/db/migrate/20180907115057_allow_user_id_to_be_null_on_secure_rooms_events.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AllowUserIdToBeNullOnSecureRoomsEvents < ActiveRecord::Migration
+
+  def change
+    change_column :secure_rooms_events, :user_id, :integer, null: true
```
jhanggi: Is this reversible?

vandrijevik: I have made it so in e7e7ba3!

PR #1691, opened 2018-09-10
TECH TASK: Fix remove_foreign_key call for non-standard column

vandrijevik:
# Release Notes

TECH TASK: Fix remove_foreign_key call for non-standard column

# Additional Context

Running `rake db:migrate` on an empty database currently fails on this call to `remove_foreign_key`, because Rails can’t find the appropriate column. When the foreign key [is added](https://github.com/tablexi/nucore-open/blob/de02a553304d013dc089ea654ba74003cdcf807a/db/migrate/20100101000000_initial_schema.rb#L127-L128), it is added against the `products` table but it is on the `instrument_id` column, so the `remove_foreign_key` needs additional information to be able to find the appropriate foreign key.


joshea0:
I know this was closed a while ago, but I just ran into this issue while creating a new staging environment so I just wanted to say thanks for the fix 👍

PR #1692, opened 2018-09-10
TECH TASK: Rename cancellation_notification_recipients to cancellation_email_recipients

vandrijevik:
# Release Notes

TECH TASK: Rename cancellation_notification_recipients to cancellation_email_recipients

# Additional Context

On Oracle, and identifier can’t be longer than 30 characters, and the column name `cancellation_notification_recipients` is 36 characters. Hence, running the migration which adds this column results in the error:

```
OCIError: ORA-00972: identifier is too long: ALTER TABLE "PRODUCTS" ADD "CANCELLATION_NOTIFICATION_RECIPIENTS" CLOB
```

so the feature from #1666 will not work on Oracle as-is.

### DIFF
--- a/app/controllers/product_notifications_controller.rb
+++ b/app/controllers/product_notifications_controller.rb
@@ -34,7 +34,7 @@ class ProductNotificationsController < ApplicationController
     params.require(:product).permit(
       :training_request_contacts,
       :order_notification_recipient,
-      :cancellation_notification_recipients,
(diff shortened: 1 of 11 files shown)


jhanggi:
One thing that I've bee thinking of that might be nice, but maybe too much effort, would be to configure circle to run against oracle in parallel. That could also make all of our builds way too slow.

PR #1693, opened 2018-09-10
[#143399] Add another column to current rooms dashboard

jhanggi:
# Release Notes

Adds another "Current Occupancies" column to the secure rooms dashboard so that a high occupancy (more than 10) does not overflow the screen.

# Screenshot

screen shot 2018-09-10 at 4 09 11 pm

# Additional Context

I chatted with Aaron and showed him a couple options. This maintains the current behavior of always showing the "Current Occupancies" header and table header of the first column for empty rooms. We hide the second column header if there is nothing there.

The facilities occupancies shows all of the rooms for a facility, but shared the same code with the public dashboard so this change applies to both places.

### DIFF
--- a/vendor/engines/secure_rooms/app/views/secure_rooms/shared/_secure_rooms_dashboard.html.haml
+++ b/vendor/engines/secure_rooms/app/views/secure_rooms/shared/_secure_rooms_dashboard.html.haml
@@ -1,27 +1,9 @@
 %label= text(:time_label)
 .refreshed_at= format_usa_datetime(Time.current)
+- current_occupancies = room.occupancies.current
 .row
-  .span6
-    %h4= text(:current)
-    %table.table.table-striped.table-hover.current_occupants
-      %thead
-        %tr
-          %th= User.human_attribute_name(:full_name)
-          %th= SecureRooms::Occupancy.human_attribute_name(:entry_at)
-      %tbody
-        - room.occupancies.current.each do |occupancy|
-          %tr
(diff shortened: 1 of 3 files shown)


vendor/engines/secure_rooms/app/views/secure_rooms/shared/_secure_rooms_dashboard.html.haml
```diff
@@ -1,27 +1,9 @@
...
-          %th= User.human_attribute_name(:full_name)
-          %th= SecureRooms::Occupancy.human_attribute_name(:exit_at)
-      %tbody
-        - room.occupancies.recent.each do |occupancy|
-          %tr
-            %td= occupancy.user.full_name(suspended_label: false)
-            %td= format_usa_datetime(occupancy.exit_at)
+  .span4= render "secure_rooms/shared/dashboard_column", occupancies: current_occupancies.limit(10), header: text(:current), timestamp: :entry_at
+  - second_column_occupancies = current_occupancies.offset(10)
+  .span4= render "secure_rooms/shared/dashboard_column", occupancies: second_column_occupancies, header: text(:current_continued), timestamp: :entry_at, hide: second_column_occupancies.none?
```
vandrijevik: Can we get rid of the `local_assigns[:hide]` here and only call `render` as necessary? It seems strange to put the local variable here and call render, only to do nothing in the partial.

jhanggi: Good call.

PR #1696, opened 2018-09-11
Update where_ids_in query for Rails 5

jhanggi:
# Release Notes

Tech task: Update `where_ids_in` scope for ActiveRecord `or` support.

# Additional Context

Pulled up from https://github.com/tablexi/nucore-nu/pull/408

I ended up using the dumb string version because I couldn't get an Arel version
working on Rails 5/`where-or`. This now works on both 4.2 and 5.0.

### DIFF
--- a/lib/nucore.rb
+++ b/lib/nucore.rb
@@ -60,14 +60,14 @@ module NUCore
         # Oracle has a limit of 1000 items in a WHERE IN clause. Use this in
         # place of `where(id: ids)` when there might be more than 1000 ids.
         # Example: facility.order_details.complete.where_ids_in(ids)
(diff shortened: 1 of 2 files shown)


jhanggi:
`where-or` is essentially pulling out the `or` functionality from Rails 5 so it can be used in 4.2. I couldn't find a solution that worked consistently in Rails 5. The first things I did worked in simple cases but broke when the relation included more complex associations. 

This was the only thing I found that would work in rails 5, and it just happened to work in 4.2 as well.

vandrijevik:
Ah, gotcha! Thanks for the explanation.

PR #1703, opened 2018-09-18
[#133770] Facility Notification Banner

jhanggi:
# Release Notes

Adds a per-facility optional banner notification that can be configured by the facility. The banner appears on the facility's home page (the instrument listing).

# Screenshot

screen shot 2018-09-17 at 7 06 28 pm

In Dartmouth when combined with the SSO warning notice (hopefully will be going away in the next little bit).

screen shot 2018-09-17 at 7 19 57 pm

Combined with the order on behalf of banner. This doesn't look great, but that banner could definitely use some love in the future. We could also hide the notification banner if the order on behalf banner.

# Additional Context

I gave this a feature flag because I could see NU not being interested.

### DIFF
--- a/app/assets/javascripts/ckeditor/editor.js
+++ b/app/assets/javascripts/ckeditor/editor.js
@@ -1,8 +1,8 @@
 $(document).ready(function() {
   if (! CKEDITOR) return;
   $('textarea.editor').each(function(){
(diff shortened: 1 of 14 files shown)


jhanggi:
Note to self: disable the feature flag for NU for now.

PR #1707, opened 2018-09-20
Refactor price policy new/edit views

jhanggi:
# Release Notes

Tech task: Refactor the price policy new/edit views to use `fields_for`. It saves us from needing to refer to `"price_policy_#{price_group.id}[unit_cost]"` everywhere.

# Additional Context

I'm adding to the instrument price policy page for and I did this refactor to make those changes easier and cleaner. I'm separating this out so those changes don't get lost in this noise.


app/views/instrument_price_policies/_adjustment_row.html.haml
```diff
@@ -1 +1 @@
-= render "time_based_price_policies/adjustment_row", price_group: price_group, price_policy: price_policy, cancellation: true, minimum_cost: true
+= render "time_based_price_policies/adjustment_row", price_group: price_group, price_policy: price_policy, cancellation: true, minimum_cost: true, pp: pp
```
vandrijevik: It seems a bit strange to me to have these one-line partials which only render another partial. What do you think about updating the calls sites which render these partials, like [here](https://github.com/tablexi/nucore-open/pull/1707/files#diff-5ea7effd71575e1289a5e209029135beR32), to just do these render calls directly and eliminate the indirection?

app/views/price_policies/_price_policy_fields.html.haml
```diff
@@ -15,31 +15,28 @@
...
-          %td.centered
-            %span.unit_cost
-            = hidden_field_tag "price_policy_#{price_group.id}[unit_cost]",
-              number_to_currency(price_policy.unit_cost, unit: "", delimiter: ""),
-              class: "unit_cost"
-          %td.centered
-            = text_field_tag "price_policy_#{price_group.id}[unit_subsidy]",
-              number_to_currency(price_policy.unit_subsidy, unit: "", delimiter: ""),
-              size: 8
+      = fields_for "price_policy_#{price_group.id}", price_policy do |pp|
```
vandrijevik: 👏

PR #1708, opened 2018-09-20
[#136151] Ability to charge full reservation cost for cancellation

jhanggi:
# Release Notes

Adds the ability to charge the full reservation cost (based on reservation times) on cancellation. This take precedence over the "Reservation/Cancellation Cost"

# Screenshot

screen shot 2018-09-20 at 4 44 42 pm

# Additional Context

This could have compliance issues for NU, so that's why it's being feature flagged. Also, Dartmouth overrides the `cancellation_cost` locale from "Reservation Cost" to "Cancellation Cost" so we'll want to adjust the local for this new field to match for them.

### DIFF
--- a/app/assets/javascripts/price_policy.js.coffee
+++ b/app/assets/javascripts/price_policy.js.coffee
@@ -9,67 +9,50 @@ $(document).ready ->
     $inputElement.val(if isDisabled then "" else $inputElement.data("original-value"))
     $inputElement.prop "disabled", isDisabled
 
-  toggleGroupFields = ($checkbox) ->
+  # Triggered by "Can purchase?"
+  toggleFieldsInSameRow = ($checkbox) ->
     $cells = $checkbox.parents("tr").find("td")
     isDisabled = !$checkbox.prop "checked"
     $cells.toggleClass "disabled", isDisabled
     $cells.find("input[type=text], input[type=hidden], input[type=checkbox]").not($checkbox).each (_i, elem) ->
       hardToggleField($(elem), isDisabled)
     # If we are enabling the row, make sure the cancellation cost field gets the correct state
-    $cells.find(".js--full_cancellation_cost").trigger("change") unless isDisabled
+    $cells.find(".js--fullCancellationCost").trigger("change") unless isDisabled
 
-  getMasterUsageRate = ->
-    rate = parseFloat $("input.master_usage_cost.usage_rate").val()
-    if isFiniteAndPositive(rate) then rate else 0
+  updateAdjustmentFields = ($sourceElement) ->
+    rate = parseFloat($sourceElement.val())
+    rate = if isFiniteAndPositive(rate) then rate else 0
 
-  getUsageAdjustment = (usageAdjustmentElement) ->
-    usageAdjustment = parseFloat usageAdjustmentElement.value
-    if isFiniteAndPositive(usageAdjustment) then usageAdjustment else 0
+    $targets = $(".js--adjustmentRow").find($sourceElement.data("target"))
+    $targets.filter(":input").val(rate)
+    $targets.filter("span").html(rate)
 
-  setUsageSubsidy = (usageAdjustmentElement, usageSubsidy) ->
-    $(usageAdjustmentElement).parents("tr").find("span.minimum_cost").data("usageSubsidy", usageSubsidy)
+  toggleFullCancellationCostInCurrentCell = ($checkbox) ->
+    $container = $checkbox.parents(".js--cancellationCostContainer")
+    hardToggleField($container.find("input[type=text]"), $checkbox.is(":checked"))
 
-  refreshCosts = ->
-    $(".master_minimum_cost").each (index, element) -> setInternalCost element
-    $("input[type=hidden].usage_cost").val(getMasterUsageRate())
+  toggleFullCancellationInAdjustmentRows = (isChecked) ->
+    $adjustmentFields = $(".js--adjustmentRow .js--fullCancellationCost")
+    $adjustmentFields.val(if isChecked then "1" else "0")
+    hardToggleField($adjustmentFields.siblings(".js--cancellationCost").filter(":input"), isChecked)
+    # Show/hide the pricing spans
+    $adjustmentFields.siblings(".js--cancellationCost").filter("span").toggle(!isChecked)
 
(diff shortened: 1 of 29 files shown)


jhanggi:
Marking this WIP because I think the full reservation cost checkbox is not getting applied to adjustment rows.

jhanggi:
Fixed up the adjustment issue. I also did some refactoring of the JS. Not sure if I went too far. I definitely could have gone farther.

app/views/time_based_price_policies/_adjustment_row.html.haml
```diff
@@ -1,22 +1,28 @@
...
-  = t("price_policies.adjustment")
-  %br
-  = text_field_tag "price_policy_#{price_group.id}[usage_subsidy]",
-    display_usage_subsidy(price_group, price_policy), size: 8, class: "usage_adjustment"
+  = pp.hidden_field :usage_rate, value: display_usage_rate(price_group, price_policy),
+    class: "js--usageRate"
+  = pp.label :usage_subsidy, t("price_policies.adjustment"), class: "normal-weight"
+  = pp.text_field :usage_subsidy, value: display_usage_subsidy(price_group, price_policy),
+    size: 8,
+    class: "usage_adjustment"
```
jhanggi: This class, along with `usage_rate` and `master_usage_cost` are still needed by a different file that I haven't touched which does the per-minute rate calculations.

app/assets/javascripts/price_policy.js.coffee
```diff
@@ -1,65 +1,57 @@
...
-  $interval.change ->
-    int_value = $(this).val()
-    interval = "#{int_value} minute"
-    interval += "s" if int_value > 1
-    $("span.interval_replace").each -> $(this).html(interval)
+  isFiniteAndPositive = (number) -> isFinite(number) && number > 0
 
-  $interval.trigger('change')
+  # In addition to disabling the field, also hide its value. But still store it
+  # so we can display it again if it gets renabled
```
vandrijevik: Do you know why this is necessary? It seems to me like it would be enough (and simpler) to just disable the field…

app/helpers/bootstrap_helper.rb
```diff
@@ -25,6 +25,10 @@ def status_badge(order_detail)
     content_tag :span, order_detail.order_status, class: classes
   end
 
+  def tooltip_hint_icon(text)
+    tooltip_icon "fa fa-question-circle-o", text
+  end
```
vandrijevik: It looks like this is called in just one place—what do you think about moving the `tooltip_icon` call directly in the view, and avoiding this helper method entirely?

db/migrate/20180919211201_add_full_cost_cancellation_to_price_policies.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddFullCostCancellationToPricePolicies < ActiveRecord::Migration[5.0]
+
+  def change
+    add_column :price_policies, :full_cancellation_cost, :boolean, default: false, null: false
```
vandrijevik: This column name took my brain a few reads to grok fully… I am not sure if any of my suggestions are better, but what do you think about something like:

- `charge_fully_on_cancellation`
- `full_price_cancellation`
- `cancellation_cost_equals_full_cost`

app/assets/javascripts/price_policy.js.coffee
jhanggi: I'm guessing our reasoning was that if you just set `disabled` on the the field, you still see the value that was there before it got disabled, though the field is grayed out. To me, it's confusing to see a value when it's not going to be used. This is especially true in with the new toggle for full cancellation.

app/helpers/bootstrap_helper.rb
jhanggi: That is true: it's only in one place right now. I couldn't find a pattern for hint tooltips like this in the existing app, and figured putting it in a helper establishes it as a standardized pattern for the future. But I could be convinced YAGNI.

db/migrate/20180919211201_add_full_cost_cancellation_to_price_policies.rb
jhanggi: Hmm, that's fair that the name isn't clear. I struggled a bit with that at first, because the name could imply an amount rather than a boolean, but ultimately had trouble coming up with something better. Maybe `charge_full_price_on_cancellation`?

app/helpers/bootstrap_helper.rb
vandrijevik: My vote would be YAGNI, and if we find ourselves repeating this for the third time in the future, we can refactor to a DRY method then 😄I don’t feel too strongly about it, though!

app/assets/javascripts/price_policy.js.coffee
vandrijevik: Gotcha, thanks!

db/migrate/20180919211201_add_full_cost_cancellation_to_price_policies.rb
vandrijevik: Oooh, I like that! `charge_full_price_on_cancellation` is good.

jhanggi:
@vandrijevik this if finally ready for another review. When I got into the confirmation display, I decided to get the two places we do it all into the `CancellationCostCalculator`. Once I had that, it made sense to me to try to refactor the actual cancellation fee logic to use that calculator. That turned out trickier than anticipated, but I think I got it all sorted out. (the code climate warning is about the complexity of the calculator, while `OrderDetail#cancellation_fee` used to be equally complex, so it's trading off a little complexity).

It's probably worth having you try to test out a few scenarios locally, both with a normal cancellation fee as well as this new full cancellation fee.

* Canceling as a user more than 24 hours in advance
* Canceling as a user less than 24 hours in advance
* Canceling as an admin from the user's reservation tab (http://localhost:3000/facilities/example/users/3/reservations)
* Canceling as an admin from the order detail popup
  * With "apply cancellation fee" checked
  * With it unchecked
  * With it checked more than 24 hours ahead of the reservation (this one was interesting because even on master, it doesn't apply the fee since you're more than X hours out). I would say we continue to maintain that behavior.
* Canceling as an admin from the order detail popup after the order was Complete. Make sure we can still apply the fee.

PR #1718, opened 2018-10-01
[#142011] Allow global admins to unreconcile orders

vandrijevik:
# Release Notes

Allow global admins to unreconcile orders

# Screenshot

screen shot 2018-10-01 at 15 04 02

### DIFF
--- a/app/controllers/reconcilliations_controller.rb
+++ b/app/controllers/reconcilliations_controller.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class ReconcilliationsController < ApplicationController
+
+  load_resource :order
+  load_resource :order_detail, through: :order
+
+  def destroy
+    if current_user.administrator?
+      @order_detail.update!(state: "complete", order_status: OrderStatus.complete, reconciled_at: nil)
+      redirect_to facility_transactions_path(current_facility)
+    else
+      raise CanCan::AccessDenied
+    end
+  end
+
+end
--- a/spec/controllers/reconcilliations_controller_spec.rb
+++ b/spec/controllers/reconcilliations_controller_spec.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+require "controller_spec_helper"
+
+RSpec.describe ReconcilliationsController do
+  before(:all) { create_users }
+
+  let(:user) { FactoryBot.create(:user) }
+  let(:facility) { FactoryBot.create(:setup_facility) }
+  let(:item) { FactoryBot.create(:item, facility: facility) }
+  let(:account) do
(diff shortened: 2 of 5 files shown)


spec/controllers/reconcilliations_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+require "controller_spec_helper"
+
+RSpec.describe ReconcilliationsController do
+  before(:all) { create_users }
+
+  let(:user) { FactoryBot.create(:user) }
+  let(:facility) { FactoryBot.create(:setup_facility) }
+  let(:item) { FactoryBot.create(:item, facility: facility) }
+  let(:account) do
+    FactoryBot.create(:nufs_account, account_users_attributes: [FactoryBot.attributes_for(:account_user, user: user)]).tap do |account|
```
jhanggi: We shouldn't need this full setup anymore. If you use `:setup_account`, that should do everything you need.

spec/controllers/reconcilliations_controller_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+  let(:order_detail) { place_and_complete_item_order(user, facility, account) }
+
+  before :each do
+    order_detail.change_status! OrderStatus.reconciled
+  end
+
+  describe "DELETE to :destroy" do
+    context "for an administrator" do
+      before :each do
+        sign_in @admin
```
jhanggi: Can we use `let`s for defining the users rather than the `@` variables created by `before(:all) { create_users }`? It's an older pattern that I think only really makes sense with the `it_should_all` pattern (which is one of those things I'd like to get rid of).

app/controllers/reconcilliations_controller.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+class ReconcilliationsController < ApplicationController
+
+  load_resource :order
+  load_resource :order_detail, through: :order
+
+  def destroy
+    if current_user.administrator?
```
jhanggi: Usually, I would say we should strive for using the `Ability` for this kind of thing, but we don't really have a "real" model for this and this is just global managers, so I'm ok with it.

spec/controllers/reconcilliations_controller_spec.rb
vandrijevik: Awesome! Changed in 21dff8c57

spec/controllers/reconcilliations_controller_spec.rb
vandrijevik: For sure, thanks for the context here! 21dff8c57 does this (and gets rid of the `create_users` call entirely)

PR #1719, opened 2018-10-02
Fixes for Rails 5 issues discovered in testing by NU

vandrijevik:
# Release Notes

Fixes for Rails 5 issues discovered in testing

# Additional Context

1. Due to the datetime column changes, a `TO_TIMESTAMP` type cast is necessary to make the query in Oracle comparisons
2. Rails 5 automatically disables form submit buttons when they are clicked. On the bulk email form, this is undesirable for the Export button, since clicking the button doesn’t leave the page (the file is just downloaded).

### DIFF
--- a/app/services/admin_reservation_expirer.rb
+++ b/app/services/admin_reservation_expirer.rb
@@ -10,7 +10,7 @@ class AdminReservationExpirer
 
   def expired_reservations
     if NUCore::Database.oracle?
-      AdminReservation.where("reserve_start_at - expires_mins_before / (60*24) < ?", Time.current)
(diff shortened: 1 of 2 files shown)


vandrijevik:
@jhanggi I had the same thought, but I think we can tackle those if they come up—the new default behavior is good, and would prevent accidental double-submissions. It’s just on this page that it doesn’t quite fit because the `Export` link leaves the browser on the page.

If it turns out there are many other places where we want to flip the behavior, we can then think about setting `config.action_view.automatically_disable_submit_tag` globally.

jhanggi:
Sounds good. 

I know there are a few places that we use `disable_with` that already did the disabling previously. Maybe we can remove those as we come across them. I didn't realize this was a new default, but I support it.

PR #1721, opened 2018-10-03
[#137154] Fix problem order email issues

jhanggi:
# Release Notes

Fixes issues with emailing users when an order goes into the problem queue (#1685):

* Duplicate emails were being sent out
* Hard error when there was a problem order that had been canceled
* Email was not being sent when scanning out of a room without an entry scan
* Email was not being sent when double-scanning in

### DIFF
--- a/app/services/move_to_problem_queue.rb
+++ b/app/services/move_to_problem_queue.rb
@@ -14,7 +14,7 @@ class MoveToProblemQueue
   def move!
     # Some scopes may accidentally try send already-complete orders to the queue.
     # This protects against sending duplicate emails to things already in the queue.
-    return unless OrderDetail.pending.include?(@order_detail)
+    return unless @order_detail.pending?
 
     @order_detail.time_data.force_completion = @force
     @order_detail.complete!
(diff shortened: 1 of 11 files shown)


app/services/move_to_problem_queue.rb
```diff
@@ -12,6 +12,10 @@ def initialize(order_detail, force: false)
   end
 
   def move!
+    # Some scopes may accidentally try send already-complete orders to the queue.
+    # This protects against sending duplicate emails to things already in the queue.
+    return unless OrderDetail.pending.include?(@order_detail)
```
vandrijevik: Rather than querying a scope on the class when we have the object handy, can we introduce `OrderDetail#pending?` here, along the lines of:

```
def pending?
  order.purchased? && state.in?(["new", "inprocess"])
end
```

That way we can call `return unless @order_detail.pending?` which seems more natural to me.

jhanggi: Yeah, definitely more natural. I think I was a little annoyed at how much duplication there was in these scope, that adding a predicate seemed duplicative as well. I've now gotten rid of the `pending` method, which is identical to `new_or_inprocess`.

vandrijevik:
Awesome! I’m still a 👍

PR #1726, opened 2018-10-08
[#142635] Keep date navigation on timeline pages always visible

vandrijevik:
# Release Notes

Keep date navigation on timeline pages always visible

# Additional Context

This makes the date navigation always visible on the public timeline, and manage timeline pages for a facility. I tried to make the datepicker scroll along with the date navigation, but after a few attempts, including a fix [like this one](https://stackoverflow.com/a/22038827), I was unable to get this working. Since closing and reopening the date picker positions it correctly, I don’t think this is worth pursuing further.

A note about the instrument ids: because the date navigation is always visible, the JS code takes its height into account when determining which instrument is the first visible one. Therefore, the URL fragment when clicking next and previous gets set to the element which is partially covered by the date navigation, thus maintaining the first visible element between clicks. This doesn’t maintain the exact vertical pixel position, which would be more complicated since we’d have to scroll the viewport in JS (which I don’t think we should do), but I think that it is good enough for maintaining position on a long page.

### DIFF
--- a/app/assets/javascripts/timeline.js
+++ b/app/assets/javascripts/timeline.js
@@ -20,7 +20,10 @@ $(function() {
     showOn: "button",
     buttonText: "<i class='fa fa-calendar icon-large'>",
   }).change(function() {
-    $(this).parents("form").submit();
+    var form = $(this).parents('form');
+    var formUrl = form.attr('action');
+    form.attr('action', formUrl + '#' + lastHiddenInstrumentId());
+    form.submit();
   });
 
   //Get the Current Hour, create a class and add it the time div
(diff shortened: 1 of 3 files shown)


jhanggi:
Is there a way that instead of using the first visible id as target, use the last hidden one? Right now, if I click the next/previous, the first visible ends up getting about halfway hidden (behind the datepicker). I think it would be less jarring if the first visible remains visible, even if it moves down a bit. What do you think?

And when we pull this in, we should make sure the ticket includes any of the caveats (like the open calendar not staying fixed).

> For now, selecting a date from the calendar will not maintain the scroll position. If we want this, we'll address it as a separate ticket

I know we had said this, but while clicking around, it feels a little weird to me to not address it as part of this feature. It's no worse than the current behavior, but how much extra effort do you think that would be?

vandrijevik:
@jhanggi I think this is ready for a re-review: I made the JS use the last hidden element, and I got the datepicker to maintain the page position as well!

PR #1727, opened 2018-10-09
Fix columns in order form

ColinMcNeil:
# Release Notes

Resolves https://github.com/SquaredLabs/nucore-uconn/issues/31

Placed the columns on the order form, and gave the button its own row.

# Screenshot

screen shot 2018-10-09 at 9 12 15 am


ColinMcNeil:
Gave the columns a flex-grow property which seems to be working.

PR #1728, opened 2018-10-09
Localize member roles for facility accounts

ColinMcNeil:
# Release Notes

Resolves https://github.com/SquaredLabs/nucore-uconn/issues/37

Added localization to member roles under facility accounts.

# Screenshot

screen shot 2018-10-09 at 9 42 00 am

# Additional Context

Because the roles are hard coded and in the database, there is no way to elegantly allow them to be localized. 
See: `account_user.rb`
```ruby
  ACCOUNT_PURCHASER = "Purchaser"
  ACCOUNT_OWNER = "Owner"
  ACCOUNT_ADMINISTRATOR = "Business Administrator"
```
Business administrator has a space, so it cannot be used to query i18n. My best workaround is to convert it in the view:
```haml
%td= t(".role.#{au.user_role.downcase.parameterize.underscore}")
```

### DIFF
--- a/app/presenters/account_user_presenter.rb
+++ b/app/presenters/account_user_presenter.rb
@@ -1,15 +1,17 @@
+# frozen_string_literal: true
 
 class AccountUserPresenter < SimpleDelegator
+
   def self.localized_role(role)
-    I18n.t(role.downcase.parameterize.underscore, scope: "accounts.role") # Move the en.yml location to be more generic than `member_table`
+    I18n.t(role.downcase.parameterize.underscore, scope: "accounts.role")
   end
 
   def self.selectable_user_roles(granting_user = nil, facility = nil)
-    # Nothing other than the collection uses `selectable_user_roles`, so it could even be pulled up into this class
-    AccountUser.selectable_user_roles(granting_user, facility).map {  |role| [localized_role(role), role] }
+    AccountUser.selectable_user_roles(granting_user, facility).map { |role| [localized_role(role), role] }
   end
 
   def localized_role
     self.class.localized_role(user_role)
   end
+
 end
--- a/app/presenters/account_user_presenter.rb
+++ b/app/presenters/account_user_presenter.rb
@@ -1,0 +1,15 @@
+
+class AccountUserPresenter < SimpleDelegator
+  def self.localized_role(role)
+    I18n.t(role.downcase.parameterize.underscore, scope: "accounts.role") # Move the en.yml location to be more generic than `member_table`
+  end
+
+  def self.selectable_user_roles(granting_user = nil, facility = nil)
+    # Nothing other than the collection uses `selectable_user_roles`, so it could even be pulled up into this class
+    AccountUser.selectable_user_roles(granting_user, facility).map {  |role| [localized_role(role), role] }
+  end
+
+  def localized_role
+    self.class.localized_role(user_role)
+  end
+end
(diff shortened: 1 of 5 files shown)


app/views/accounts/_member_table.html.haml
```diff
@@ -29,5 +29,5 @@
 
         %td= Users::NamePresenter.new(au.user, username_label: true).full_name
         %td= au.user.email
-        %td= au.user_role
+        %td= t(".role.#{au.user_role.downcase.parameterize.underscore}")
```
jhanggi: It's not ideal, but it's probably the most expedient rather that trying to fully refactor it into something better. Maybe an interim step would be to pull this up into an `AccountUser#user_role_key` that holds the `user_role.downcase.parameterize.underscore`. That could make future refactoring easier.

ColinMcNeil:
Not happy with the solution to the dropdown localization. Only alternative I see is a major refactor of AccountUser or moving the logic to the controller.

app/views/facility_account_users/new.html.haml
```diff
@@ -26,7 +26,7 @@
   = f.input :user_id, as: :hidden, input_html: { value: @user.id, name: "user_id" }
 
   = f.input :user_role,
-    collection: AccountUser.selectable_user_roles(current_user, current_facility),
+    collection: AccountUser.selectable_user_roles(current_user, current_facility).map {|role| [text("accounts.member_table.role.#{role.downcase.parameterize.underscore}"), role]},
```
jhanggi: Yeah, this isn't great. What would you think about using a presenter?

```ruby
class AccountUserPresenter < SimpleDelegator
  def self.localized_role(role)
    I18n.t(role.downcase.parameterize.underscore, scope: "accounts.role") # Move the en.yml location to be more generic than `member_table`
  end

  def self.selectable_user_roles(granting_user = nil, facility = nil)
    # Nothing other than the collection uses `selectable_user_roles`, so it could even be pulled up into this class
    AccountUser.selectable_user_roles(granting_user, facility).map {  { |role| [localized_role(role), role] }
  end

  def localized_role
    self.class.localize_role(user_role)
  end
end
```

In the `member_table`:

`= AccountUserPresenter.new(au).localized_role`

In the form:
`AccountUserPresenter.selectable_user_roles(current_user, current_facility)`

There's also `account_users/new` that uses `selectable_user_roles`. That's for account owners to manage access to their own accounts.

* Note: I haven't actually tested that code

ColinMcNeil: Wow, besides one semicolon this code worked perfect! Look good now?

app/presenters/account_user_presenter.rb
```diff
@@ -0,0 +1,15 @@
+
+class AccountUserPresenter < SimpleDelegator
+  def self.localized_role(role)
+    I18n.t(role.downcase.parameterize.underscore, scope: "accounts.role") # Move the en.yml location to be more generic than `member_table`
```
jhanggi: I believe this comment can go away.

app/presenters/account_user_presenter.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def self.selectable_user_roles(granting_user = nil, facility = nil)
+    # Nothing other than the collection uses `selectable_user_roles`, so it could even be pulled up into this class
```
jhanggi: I don't think we need his comment either.

app/presenters/account_user_presenter.rb
```diff
@@ -0,0 +1,15 @@
+
+class AccountUserPresenter < SimpleDelegator
```
jhanggi: Can you run `rubocop -a app/presenters/account_user_presenter.rb`?

PR #1729, opened 2018-10-09
Update LDAP documentation

jhanggi:
# Release Notes

Update documentation for developing against LDAP

# Additional Context

I did this while starting to work on the tool for auto-suspending users if they are removed from the LDAP directory.

Removes older instructions for setting up an LDAP server on your Mac OS system. We now recommend using Docker.

This also adds the ability to use YAML aliases in the configuration.

# Questions

When we get to the story about removing if they're not in LDAP, do we think it might be worth seeding an LDAP server on Circle to test against? I feel that would be nice, but would make running a local test suite harder.

### DIFF
--- a/config/ldap.yml.template
+++ b/config/ldap.yml.template
@@ -11,4 +11,4 @@ development:
   base: dc=example,dc=org
   attribute: uid
   admin_user: "cn=admin,dc=example,dc=org"
-  admin_password: secret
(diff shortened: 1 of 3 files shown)


jhanggi:
For running ldap in docker, I don't think it's worth the effort right now. If we were doing serious development against it, I'd be more interested.

PR #1730, opened 2018-10-10
Update Oracle instructions for running official Oracle 12.2 image

vandrijevik:
# Release Notes

TECH TASK: Update Oracle instructions for running official Oracle 12.2 image

# Additional Context

Using these updated instructions, I was able to get the offical Oracle 12.2 image running locally, create the NU databases, and run tests successfully.

I have not yet been able to test that the restore from backup section works with the updated setup, so if that’s not the case, we can update it as necessary. The new `database.yml` template also makes `rails dbconsole` work (which the current template does not for me).


doc/HOWTO_oracle.md
```diff
@@ -57,92 +76,79 @@ sudo ln -s libocci.dylib.12.1   libocci.dylib
...
-* Start, setup, and seed the oracle container:
-
-```
-docker/oracle/setup.sh
-# demo:seed is optional
-rake demo:seed
-```
+1. Make sure you have the `activerecord-oracle_enhanced-adapter` and `ruby-oci8` gems enabled in your `Gemfile`.
+1. Copy the Oracle template for `database.yml` by running `cp config/database.yml.oracle.template config/database.yml`.
+1. Run `bundle exec rake db:setup`.
```
jhanggi: When I got to this step, it asked me for the oracle system password. We should probably note that in the instructions.

doc/HOWTO_oracle.md
```diff
@@ -57,92 +76,79 @@ sudo ln -s libocci.dylib.12.1   libocci.dylib
...
+      oracle \
+      bash
+    ```
+
+1. Import the dump, replacing DUMPFILE with the name of your dump file, and REMAP_SCHEMA with your database’s username if necessary:
 
-Run this (replacing the DUMPFILE filename and the second part of REMAP_SCHEMA with your database's username):
 ```
-impdp system/oracle@//localhost:1521/xe DIRECTORY=data_pump_dir DUMPFILE=expdp_schema_COR1PRD_201708191913.dmp REMAP_SCHEMA=bc_nucore:nucore_nu_development
+impdp \
```
jhanggi: This step didn't work for me. It's telling me `TNS:listener does not currently know of service requested in connect descriptor`

doc/HOWTO_oracle.md
vandrijevik: Did you set the env variable as mentioned on line 86, which talks about this?

jhanggi: Ah, makes sense. That feels like kind of an odd thing to me to add into your dotfiles, though.

doc/HOWTO_oracle.md
```diff
@@ -57,92 +76,79 @@ sudo ln -s libocci.dylib.12.1   libocci.dylib
...
 
-##### Add Environment Variables to profile
+### Set Up Environment Variables
 
-* Add to `~/.profile` (bash) or `~/.zprofile` (zsh)
+1. Add to `~/.profile` (bash) or `~/.zprofile` (zsh)
 
 ```
-export NLS_LANG="AMERICAN_AMERICA.AL32UTF8"
-export ORACLE_HOME=/opt/oracle/instantclient
```
jhanggi: Are you sure we don't need `ORACLE_HOME` anymore? I think we need that in order to install the OCI gem.

vandrijevik: I think I forgot to update the instructions for installing instant client via `brew`—when I did that, everything gets put into `/usr/local`, like other brew-managed packages, and there is no need to specify these environment variables.

I’ll revisit the installation instructions for instant client.

doc/HOWTO_oracle.md
vandrijevik: You don’t have to—you can specify it when you invoke `rake db:setup` or `rake db:drop` or any other similar tasks, but I found it’s easiest to put this in a dotfile, kinda like `NLS_LANG`.

doc/HOWTO_oracle.md
vandrijevik: f6a5913 should fix this, along with subsequent permissions errors because even though the image mounts at `/ORCL`, the server’s `DATA_PUMP_DIR` expects files in `/u01` 🤦‍♂️

doc/HOWTO_oracle.md
```diff
@@ -4,145 +4,145 @@ Follow these instructions to set up the Oracle database to run in Docker, but th
...
-sudo ln -s libclntsh.dylib.12.1 libclntsh.dylib
-sudo ln -s libocci.dylib.12.1   libocci.dylib
+### Install Oracle Instant Client
 
-sudo ln -s /opt/oracle/instantclient/sqlplus /usr/local/bin/sqlplus
-```
+1. Enable the homebrew tap for Oracle Instant Client by running `brew tap InstantClientTap/instantclient`.
+1. Run each of these commands, following the displayed instructions for how to download the appropriate .zip files from Oracle’s website. After you download the files and place them in Homebrew’s cache directory, brew will take care of the rest:
+    ```
+    brew install instantclient-basic
```
jhanggi: I didn't know you could use `brew` for this. This makes things much easier. 🎉

vandrijevik: Yeah, you still have to download the files manually and put them in the right directory (which will be displayed in the instructions), but it’s been pretty nice!

jhanggi: Makes sense. I was wondering how they were getting around the distribution restrictions.

doc/HOWTO_oracle.md
```diff
@@ -4,145 +4,145 @@ Follow these instructions to set up the Oracle database to run in Docker, but th
...
+## Install Oracle SQL Developer
 
 * Download from: `http://www.oracle.com/technetwork/developer-tools/sql-developer/overview/index.html`
 
 * Install into `/Applications`
 
-#### Restore from a backup
+## Restore From Backup
 
 Run `bundle exec rake db:oracle_drop_severe`. This will ensure that your database
```
jhanggi: We'll probably address this separately, but when loading from a dmp, while I haven't gotten it successfully running, I do believe that `db:reset` or `db:setup` won't work because when you try to load the file, it will complain about the tables already existing. So I think these instruction should probably include `db:drop db:create`. And hopefully get rid of `db:oracle_drop_severe`.

vandrijevik: Good call! I just did this in 98402df and subsequently removed the now-unneeded `db:oracle_drop*` tasks—the regular `oracle_drop` isn’t needed since `db:drop` now works properly.

PR #1731, opened 2018-10-10
[#143771] Prevent error on invalid dates on Reservation#update

jhanggi:
# Release Notes

Prevents an error on the reservation edit form when start/end at is blank because of invalid data.

# Additional Context

Here's the error: "NoMethodError: undefined method `<' for nil:NilClass" on the `if duration_is < duration_was` line. 

The error that we originally got notified of was super weird: it was missing the `reserve_start_mins` and `reserve_end_mins` parameters. I could recreate that if I deleted the dropdown from within my browser's dev tools. But I was able to replicate it as well by putting in an invalid date like "10/0/2018" into the date fields.

### DIFF
--- a/app/support/reservations/duration_change_validations.rb
+++ b/app/support/reservations/duration_change_validations.rb
@@ -10,7 +10,6 @@ class Reservations::DurationChangeValidations
   include ActiveModel::Validations::Callbacks
 
   attr_reader :reservation
-
   validate :start_time_not_changed, unless: "reservation.in_cart?"
   validate :duration_not_shortened, unless: "reservation.in_cart?"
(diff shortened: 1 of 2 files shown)


app/support/reservations/duration_change_validations.rb
```diff
@@ -37,6 +36,8 @@ def start_time_not_changed
   end
 
   def duration_not_shortened
+    return unless reservation.reserve_start_at && reservation.reserve_end_at
```
vandrijevik: You can use `Reservation#has_reserved_times?` here for a slightly-nicer read, if you want

PR #1732, opened 2018-10-12
[#144271] Fix bulk email in IE

jhanggi:
# Release Notes

Fix issue in IE where "Compose Mail" button was triggering CSV download rather than taking you to the compose form

# Additional Context

The bulk email JS code uses `Array.includes` which is not available in IE 11 and below (it is available in Edge). This adds a polyfill for that functionality.

The email form is not particularly pretty in IE (the email list and greeting line inputs are not a good height), but I believe that is outside the scope of this PR.

screen shot 2018-10-12 at 6 37 38 pm

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -16,3 +16,4 @@
 //= require clockpunch
 //= require fine-uploader/jquery.fine-uploader
 //= require moment
+//= require polyfills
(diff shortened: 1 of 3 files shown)


jhanggi:
I'm planning to mark the codeclimate issues as wontfix since they're on a vendored file that we didn't write.

vandrijevik:
> I'm planning to mark the codeclimate issues as wontfix since they're on a vendored file that we didn't write.

I approved them in Code Climate with a comment about the exact same thing before reading your comment 😄

PR #1734, opened 2018-10-15
[#143568] Use Roboto font for PDFs

jhanggi:
# Release Notes

Use Roboto font when generating PDFs in order to support UTF-8 characters.

# Additional Context

This will prevent `ActionView::Template::Error: Your document includes text that's not compatible with the Windows-1252 character set. If you need full UTF-8 support, use TTF fonts instead of PDF's built-in fonts.`

I went with Roboto as it's a free and widely used font. It seemed a reasonable default replacing the Helvetica that was the PDF's default.

I went back and forth on where the configuration of the font should live. I eventually settled on putting it in `settings.yml` rather than the actual `StatementPdf` class. That allows this single change to apply downstream without needing to change each school's custom class. I could be convinced otherwise, though.

I also pulled out `StatementPdfDefault` from `StatementPdf` to try to make it a little clearer what's shared across all the schools vs what's the implementation of the default statement.


app/support/statement_pdf.rb
```diff
@@ -40,77 +36,11 @@ def filename
                                       invoice_number: @statement.invoice_number)
   end
 
-  def render
```
jhanggi: Looks like I do need this method. I think it gets called when creating the PDF for attaching to an email.

app/support/pdf_font_helper.rb
```diff
@@ -0,0 +1,37 @@
+class PdfFontHelper
+
+  class << self
+
+    delegate :set_fonts, to: :new
+
+  end
```
vandrijevik: I think this construct is a bit weird for a helper—there is no state at all being manipulated in `set_fonts`, so why even create an object? Can we `def self.set_fonts` and eliminate this indirection in the implementation?

app/support/statement_pdf_default.rb
```diff
@@ -0,0 +1,71 @@
+class StatementPdfDefault < StatementPdf
```
vandrijevik: Is there a reason to introduce a class hierarchy with one subclass at this point? Having a base class, and then a subclass named `*Default` seems a bit weird here.

I think the inheritance hierarchy adds complexity here, and it may be YAGNI—if we find that we need to override the behavior down the road, can we figure out how to best achieve it at that point? I’m not sure that inheritance would be the best solution even if we do need to vary behavior between schools later…

jhanggi: Right now, each of the schools has a class that in one of their engines that inherits from `StatementPdf` and overrides `generate`. So while the class hierarchy does not exist in open, it already does exist in the forks. With this, I was trying to break apart the base class that the schools' custom statements rely on and the sample implementation that's included in open. Maybe `Sample` is a better name than `Default`?

vandrijevik: > So while the class hierarchy does not exist in open, it already does exist in the forks.

Oh, this is the piece I was missing, thanks for the context 😄

> Maybe `Sample` is a better name than `Default`?

How about `Open` to mirror the repository? Or do you think that’s too ambiguous/confusing? In either case, I like `Sample` better than `Default` so go with your ❤️!

app/support/pdf_font_helper.rb
jhanggi: Oh, right. The reason I did this is because making `set_fonts` a class method means the private methods need to be class methods as well. I'm open to doing something else, but this feels like a reasonable balance between usable API and readability.

Options: 
* Change the API to `PdfHelper.new.set_fonts(pdf)` or `PdfHelper.new(pdf).set_fonts`. Little more verbose, but not bad.
* Make all the methods (including the private methods) class methods. Little more verbose internally, and makes the internal helper methods public.

What do you think?

vandrijevik: What do you think about something like this?

```ruby
class PdfFontHelper

  FONT_DIR = Rails.root.join("app/assets/fonts")

  def self.set_fonts
    font_name = Settings.statement_pdf.font_name
    return if font_name.blank?

    pdf.font_families.update(
      font_name => {
        bold: FONT_DIR.join("#{font_name}-Bold.ttf"),
        italic: FONT_DIR.join("#{font_name}-Italic.ttf"),
        normal: FONT_DIR.join("#{font_name}-Regular.ttf"),
      }
    )
    pdf.font(font_name)
  end

end
```

I’m not sure there’s a need for maintaining `mappings` only to use it in `file_mappings` to create the new hash, so I’d get rid of the private methods entirely.

If you don’t like it and want to go with your original implementation, I’m okay with that too.

jhanggi: I like that implementation. I had originally had the `mappings` method pulled out because I was thinking of allowing defining other variants (e.g. BoldItalic) through the settings, but then I decided that was overkill, but still left the method. I agree that with just the 3, it's fine to leave them inline and not doing the `reduce`.

PR #1740, opened 2018-10-24
[#136151] Rename column to full_price_cancellation so it fits into Oracle <= 30 limit

vandrijevik:
# Release Notes

TECH TASK: Rename column to full_price_cancellation so it fits into Oracle <= 30 limit

# Additional Context

When deploying to NU, whose DB server runs Oracle 12.1, the migration fails:

```
== 20180919211201 AddFullCostCancellationToPricePolicies: migrating ===========
-- add_column(:price_policies, :charge_full_price_on_cancellation, :boolean, {:default=>false, :null=>false})
rake stderr: rake aborted!
StandardError: An error has occurred, all later migrations canceled:

OCIError: ORA-00972: identifier is too long: ALTER TABLE "PRICE_POLICIES" ADD "CHARGE_FULL_PRICE_ON_CANCELLATION" NUMBER(1) DEFAULT 0 NOT NULL
stmt.c:243:in oci8lib_240.so
```

Therefore, I am renaming the DB column only, from `charge_full_price_on_cancellation` to `full_price_cancellation`, so it is less than 30 characters and can be applied. Earlier, we thought that the database server runs Oracle 12.2, but that was only the version of client tools that NU provides on the servers (which is different than the server version).

### DIFF
--- a/app/models/instrument_price_policy.rb
+++ b/app/models/instrument_price_policy.rb
@@ -22,7 +22,7 @@ class InstrumentPricePolicy < PricePolicy
   end
 
   def charge_full_price_on_cancellation?
-    SettingsHelper.feature_on?(:charge_full_price_on_cancellation) && self[:charge_full_price_on_cancellation]
+    SettingsHelper.feature_on?(:charge_full_price_on_cancellation) && full_price_cancellation?
   end
 
   private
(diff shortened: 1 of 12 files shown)


vandrijevik:
> This feature is already out on txi's stage (nucore.stage.tablexi.com), so it might have problems when we deploy it there. How should we deal with that?

Hmm… Since the migration has been run there already, I will just rename the column manually after this is deployed, and restart the unicorns!

PR #1749, opened 2018-11-02
Refactor scopes on Schedule

vandrijevik:
# Release Notes

TECH TASK: Refactor scopes on Schedule

# Additional Context

I ran into these and noticed they can be simpler by using ActiveRecord built-ins (and the existing query was unnecessarily taxing by including a GROUP BY clause when no aggregation is needed), so I refactored them!


vandrijevik:
I ran into this code while browsing around for schedule-related things (for the partial availability story), and it just stuck out!

PR #1750, opened 2018-11-02
[#140549] Disable users who were removed from LDAP

jhanggi:
# Release Notes

Disable users who were removed from LDAP

# Additional Context

There's a separate NU-based API for a queue of netid de-provisioning that we'll want to look at, but this handles the model setup and a one-time task to clean up all of the current users. I'll probably try running this against the production LDAP just to see how it performs. If it's performant enough that we could run once a week or so, it might not be worth expending the effort on dealing with the queue API.


config/locales/views/admin/en.users.yml
```diff
@@ -12,6 +12,8 @@ en:
         subhead: You may use this to edit external users. The email address and username must match.
         update: Update
         internal: Internal Pricing
+      show:
+        unexpire: Un-expire
```
vandrijevik: TIL that there’s no verb unexpire, and that unexpired is just an adjective (I went looking after this sounded a bit off to my ears). Maybe consider `remove expiration` as the action here, since it seems like that’s what we’re doing?

If you prefer unexpire as a verb, that’s cool too, but Apple’s Dictionary says there’s no dash in it 😄

config/locales/views/admin/en.users.yml
```diff
@@ -23,3 +25,9 @@ en:
       update:
         error: There was an error updating the user. %{message}
         success: The user has been updated successfully.
+      suspend:
+        success: User suspended
+      unsuspend:
+        success: User re-activated
+      unexpire:
+        success: User un-expired
```
vandrijevik: I think we need to remove the hyphens here, reactivate and unexpired do not have them.

config/routes.rb
```diff
@@ -146,6 +146,7 @@
         end
         patch "suspend", on: :member
         patch "unsuspend", on: :member
+        patch "unexpire", on: :member
```
vandrijevik: I don’t love routes like these (I’d prefer to have a resource called Expiration, and DELETE it to unexpire in this instance), but given that it fits the existing pattern, I think I’m fine with it here.

db/migrate/20181102173232_add_expired_to_users.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddExpiredToUsers < ActiveRecord::Migration[5.0]
+
+  def change
+    add_column :users, :expired_at, :datetime
```
vandrijevik: We probably want an index on this column, since we use the scope `unexpired` in the `ldap_authentication` rake tasks

app/presenters/users/name_presenter.rb
```diff
@@ -21,7 +21,13 @@ def last_first_name
     def render(name)
       parts = Array(name)
       parts << "(#{username})" if username_label?
-      parts << "(#{user.class.human_attribute_name(:suspended)})" if suspended_label? && suspended?
+      if suspended_label?
+        if suspended?
```
jhanggi: I just realized that this may not be the behavior we want if the user was suspended by an admin and then later gets removed from LDAP. This would have them showing the `SUSPENDED` label. I've asked the client for a preference of precedence.

config/locales/views/admin/en.users.yml
jhanggi: I think that's probably why I put the dash in there because it's not really a word. And possibly influenced by "re-activated". 

I like "remove expiration".

PR #1753, opened 2018-11-03
[#132414] Allow instruments to be marked as partially available

vandrijevik:
# Release Notes

Allow instruments to be marked as partially available

# Screenshot

## Buttons and Form

screen shot 2018-11-03 at 23 07 16

screen shot 2018-11-03 at 23 07 34

screen shot 2018-11-03 at 22 58 40

## Sample Display of Warnings

screen shot 2018-11-03 at 22 59 00

screen shot 2018-11-03 at 23 06 50

# Additional Context

The note is currently limited to 256 characters, which I figured should be enough to be descriptive, but not too much to fit in a tooltip.

The warning triangle is displayed everywhere that the currently offline alert triang is displayed.

### DIFF
--- a/app/assets/stylesheets/app/reservations.scss
+++ b/app/assets/stylesheets/app/reservations.scss
@@ -10,6 +10,10 @@
   color: red;
 }
 
+.partially-available-warning {
+  color: #faa732;
+}
+
 .icon-large {
   font-size: 120%;
 }
--- a/app/assets/stylesheets/app/reservations.scss
+++ b/app/assets/stylesheets/app/reservations.scss
@@ -11,7 +11,7 @@
 }
 
 .partially-available-warning {
-  color: #faa732;
+  color: $orange;
 }
 
 .icon-large {
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -43,7 +43,7 @@ module ApplicationHelper
     if instrument.offline?
       tooltip_icon "fa fa-exclamation-triangle icon-large", t("instruments.offline.note")
     elsif instrument.has_alert?
-      tooltip_icon "fa fa-exclamation-triangle partially-available-warning icon-large", instrument.partial_availability.note
+      tooltip_icon "fa fa-exclamation-triangle partially-available-warning icon-large", instrument.alert.note
     end
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -39,6 +39,14 @@ module ApplicationHelper
     send(method, facility, product)
   end
 
+  def warning_if_instrument_is_offline_or_partially_available(instrument)
+    if instrument.offline?
+      tooltip_icon "fa fa-exclamation-triangle icon-large", t("instruments.offline.note")
+    elsif instrument.partially_available?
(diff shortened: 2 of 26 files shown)


app/helpers/application_helper.rb
```diff
@@ -39,6 +39,14 @@ def facility_product_path(facility, product)
     send(method, facility, product)
   end
 
+  def warning_if_instrument_is_offline_or_partially_available(instrument)
```
jhanggi: Good call on pulling this up into a helper. I don't have a strong feeling, but I wonder if `ApplicationHelper` is the right place for this to live. Should we make a PartialAvailabilityHelper?

app/views/partial_availabilities/new.html.haml
```diff
@@ -0,0 +1,22 @@
+= content_for :h1 do
```
jhanggi: It feels a little weird that it loses the tabs once you click into this view, but that's also what we do for offline, so I guess I'm ok with it.

app/assets/stylesheets/app/reservations.scss
```diff
@@ -10,6 +10,10 @@
   color: red;
 }
 
+.partially-available-warning {
+  color: #faa732;
```
jhanggi: Is there a variable we can use from bootstrap for this value? Or should we wrap it in a SASS variable?

app/views/instruments/schedule.html.haml
```diff
@@ -25,25 +25,31 @@
...
 
   = render_view_hook "after_offline_toggle"
 
 - if @admin_reservations.empty?
   %p.notice= t(".no_upcoming_admin_reservations")
-
 - else
   - if @product.offline?
     %p.alert.alert-danger= t(".instrument_is_offline")
+  - elsif @product.partially_available?
```
jhanggi: This gets hidden when there are no upcoming reservations. I feel like we should put that out of the surrounding if/else so it and the offline banners are always visible.

screen shot 2018-11-05 at 12 27 04 pm

app/helpers/application_helper.rb
vandrijevik: I like `ApplicationHelper` since this method is shared between views of different controllers, so it seems like the textbook place to put! I am not a fan of introducing another module just for this one method here—if, in the future we end up having several helpers related solely to this funcitonality, I could see grouping them into a module then (probably after we have 3 such methods or so), so I’ll leave it as is.

app/views/partial_availabilities/new.html.haml
vandrijevik: Yeah, I think there are improvements to be made here, but I was just mimicking the established pattern in this area.

app/views/instruments/schedule.html.haml
vandrijevik: The trick here is that if there’s an offline reservation, that is an admin hold, so it would appear. I’ll move out the instrument alert notice, though!

app/assets/stylesheets/app/reservations.scss
vandrijevik: Good call — I’ve switched this to `$orange`.

vandrijevik:
> I feel like the warning alert should be on reservations/new so that users see. We already have a spot there for offline.

Good call!

PR #1767, opened 2018-11-14
[#144294] Normalize whitespace in statement PDFs

jhanggi:
# Release Notes

Fix tab characters display on statement PDFs.

# Screenshot

sans serif

# Additional Context

This started happening when we switched to using a TTF for the statements. Tab characters were displaying funkily. This should also handle if non-breaking spaces and its ilk get pasted in.

### DIFF
--- a/app/services/whitespace_normalizer.rb
+++ b/app/services/whitespace_normalizer.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+class WhitespaceNormalizer
+
+  def self.normalize(text)
+    new(text).normalize
+  end
+
+  def initialize(text)
+    @text = text
+  end
+
+  def normalize
+    result = normalize_spaces(@text)
+    result = normalize_new_lines(result)
+    result
+  end
+
+  private
+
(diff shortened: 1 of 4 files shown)


app/services/whitespace_normalizer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class WhitespaceNormalizer
```
vandrijevik: Given that there is presently only one place where this class is used, and we never create objects with it, how about we just fold the method `WhitespaceNormalizer.normalize` into `StatementPdf#normalize_whitespace`?

jhanggi: I initially broke it out into its own class because it was longer (I was able to shorten it up to just a couple lines) and also it made testing a little clearer. It felt like its own concern. I don't feel too strongly about it.

vandrijevik: I leave it up to you—I would definitely get rid of it, but I think that’s just a “here’s how I’d do it” thought on me, not a problem 😄

PR #1770, opened 2018-11-15
[#140549] Retry LDAP queries up to 3 times

jhanggi:
# Release Notes

Tech task: Retry, up to three times, LDAP queries that time out.

# Additional Context

This should help with the LDAP deactivation job getting occasional timeouts which
was causing the entire job to fail. It'll still end up crashing if there's a serious issue with LDAP, but hopefully, this will handle most of the cases.

### DIFF
--- a/vendor/engines/ldap_authentication/lib/ldap_authentication/user_entry.rb
+++ b/vendor/engines/ldap_authentication/lib/ldap_authentication/user_entry.rb
@@ -10,7 +10,7 @@ module LdapAuthentication
 
       ldap_entries = nil
       ActiveSupport::Notifications.instrument "search.ldap_authentication" do |payload|
-        ldap_entries = admin_ldap.search(filter: Net::LDAP::Filter.eq(LdapAuthentication.attribute_field, uid))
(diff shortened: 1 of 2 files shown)


vendor/engines/ldap_authentication/lib/ldap_authentication/user_entry.rb
```diff
@@ -52,6 +52,16 @@ def to_user
       UserConverter.new(self).to_user
     end
 
+    def self.with_retry(errors = [Net::LDAP::Error], max_attempts: 3)
+      tries = 0
+      begin
+        yield
+      rescue *errors => e
```
vandrijevik: Since we’re in a method here, we don’t need an explicit `begin` and we can just move the rescue up to the method level, right? Also, given that there is only one caller of `with_retry`, what do you think about just`rescue Net::LDAP::Error` here and simplify the method signature to `(max_attempts=3)` instead of introducing an options hash unnecessarily?

jhanggi: We do need the `begin` because otherwise it would retry the whole method and keep resetting `tries` to zero.

I'll clean up the signature.

PR #1773, opened 2018-11-19
[#144530] Allow requiring a note when creating price policies

jhanggi:
# Release Notes

Add a note field to price policies and optionally make it required.

# Additional Context

Dartmouth wants to make it required. NU is fine adding the field, but does not want to require it being filled out.

### DIFF
--- a/db/migrate/20181119211456_add_note_to_price_policies.rb
+++ b/db/migrate/20181119211456_add_note_to_price_policies.rb
@@ -3,7 +3,7 @@
 class AddNoteToPricePolicies < ActiveRecord::Migration[5.0]
   def change
     change_table :price_policies do |t|
-      t.text :note
+      t.string :note, limit: 256
       t.references :created_by, index: true, foreign_key: { to_table: :users }
     end
   end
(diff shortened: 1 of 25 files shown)


jhanggi:
@vandrijevik updated. I definitely could have gone further with refactoring, but based on the fact that the form is not too complicated (though some of the building and parameters are non-standard), it probably wouldn't be worth it.

I also added a `created_by` foreign key. It seemed like something that would be useful for the auditing and was fairly straightforward. If you edit/update a price policy, that'll get updated. Since a policy becomes locked for editing once somebody has purchased on it, it felt like it was ok to to be the last person to update it.

db/migrate/20181119211456_add_note_to_price_policies.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddNoteToPricePolicies < ActiveRecord::Migration[5.0]
+  def change
+    change_table :price_policies do |t|
+      t.text :note
```
jhanggi: @vandrijevik I went with a `text` field before our conversation today. Do you think I should switch it to `string`/`varchar`? I kind of like being more flexible in terms of giving the user enough space.

vandrijevik: Hmm… I’d probably put this as a `t.string :note, limit: 256`, but I don’t feel too strongly about that 256. I do think we should specify a hard limit on the column size, though!

jhanggi:
@vandrijevik I made a few changes. Could you give this another look?

vandrijevik:
Yep yep, I’m still a ✅ on this one!

PR #1776, opened 2018-11-21
[#144074] Add debugging statements to secure_rooms engine

vandrijevik:
# Release Notes

Add debugging statements to secure_rooms engine

# Additional Context

Some occupancies of secure rooms are not updating the related order detail records when a user exits the room. However, looking at these occupancies and order details does not reveal why this happens — calling the code that the controller would call, processes these order details correctly. This PR is to add some tracing calls to see what in the processing chain may be getting skipped.

### DIFF
--- a/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
+++ b/vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
@@ -17,22 +17,30 @@ module SecureRooms
       end
 
       def process
+        Rails.logger.info("[SecureRooms] Entered SecureRooms::AccessHandlers::EventHandler#process")
+
         return unless event.success?
 
         if current_occupant? && exiting?
+          Rails.logger.info("[SecureRooms] Processing branch current_occupant? && exiting?")
           existing_occupancy.associate_exit!(event)
         elsif new_occupant? && entering?
+          Rails.logger.info("[SecureRooms] Processing branch new_occupant? && entering?")
           new_occupancy.associate_exit!(event) if entry_only?
           new_occupancy.associate_entry!(event)
         elsif new_occupant? && exiting?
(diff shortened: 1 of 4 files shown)


vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -17,22 +17,30 @@ def initialize(event)
       end
 
       def process
+        Rails.logger.info("[SecureRooms] Entered SecureRooms::AccessHandlers::EventHandler#process")
```
jhanggi: Do we want to add any IDs in there? I know we should be able to trace through the log by the request id, but should we include order detail ids and occupancy ids?

vendor/engines/secure_rooms/app/services/secure_rooms/access_handlers/occupancy_handler.rb
```diff
@@ -17,22 +17,30 @@ def initialize(event)
...
           new_occupancy.associate_exit!(event)
         elsif current_occupant? && entering?
+          Rails.logger.info("[SecureRooms] Processing branch current_occupant? && entering? (existing occupancy id: #{existing_occupancy.id})")
           complete_existing_occupancy!
           new_occupancy.associate_entry!(event)
         else
           raise NotImplementedError, "Encountered unexpected scan context"
         end
+
+        Rails.logger.info("[SecureRooms] Exiting SecureRooms::AccessHandlers::EventHandler#process")
```
jhanggi: I'm pretty sure this `process` method needs to return an `occupancy`. It's probably not clear, but the `Occupancy#associate_entry!` methods returns `self`.

vandrijevik: Yeah, that was very surprising! I have a fix incoming for this, just read it after fixing the problem 😄

PR #1779, opened 2018-11-27
Use-Manage Table

ColinMcNeil:
# Release Notes

The UCONN redesign for tables.

# Screenshot

screen shot 2018-11-27 at 9 01 56 am

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -597,6 +597,12 @@ li:hover {
   color: $darkGrey;
 }
 
+#use_tab.active > a:hover {
+  background-color: transparent;
+  box-shadow: none;
+  color: $darkGrey;
+}
+
 .manage-mode #use_tab.active a {
   background-color: $manageColorDark;
 }
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -634,3 +634,44 @@ li:hover {
 .nav.pull-right {
   letter-spacing: $fontSpacing;
 }
+
+
+/*
+ UCONN Redesign Table View
+*/
+td, tr, {
+  transition: ease 0.2s all;
+  border-bottom: solid  $grey 1px;
+}
+.table-striped tbody > tr:nth-child(2n+1) {
+  td {
+    background-color: $lighterGrey;
+  }
+  th {
+    background-color: $lighterGrey;
+  }
+} 
+
+.table {
+  td {
+    background-color: $headerBackgroundColor;
+  }
+
+  th {
+    background-color: $headerBackgroundColor;
+  }
+}
+
+.table tbody > tr:hover, table-striped tbody > tr:hover {
+  td,th {
+    background-color: $lightGrey;
+    border:  none;
+    border-bottom: solid  $grey 1.5px;
+  }
+}
+
+thead {
+  color: $grey;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/nucore.scss
```diff
@@ -628,3 +634,44 @@ li:hover {
...
+    background-color: $lighterGrey;
+  }
+  th {
+    background-color: $lighterGrey;
+  }
+} 
+
+.table {
+  td {
+    background-color: $headerBackgroundColor;
```
jhanggi: I don't think this is a good option. I think it should probably be one of the grays. In the other schools we override this variable. In NU it's purple and Dartmouth it's green.

app/assets/stylesheets/nucore.scss
```diff
@@ -628,3 +634,44 @@ li:hover {
...
+  }
+
+  th {
+    background-color: $headerBackgroundColor;
+  }
+}
+
+.table tbody > tr:hover, table-striped tbody > tr:hover {
+  td,th {
+    background-color: $lightGrey;
```
jhanggi: I'm a little concerned that even this is a little too dark. There doesn't feel like there's enough contrast between the background color and the black text (or the blue links). It's possibly an accessibility concern.

app/assets/stylesheets/nucore.scss
```diff
@@ -628,3 +634,44 @@ li:hover {
...
+    background-color: $headerBackgroundColor;
+  }
+
+  th {
+    background-color: $headerBackgroundColor;
+  }
+}
+
+.table tbody > tr:hover, table-striped tbody > tr:hover {
+  td,th {
```
jhanggi: Can we put a space after the comma?

app/assets/stylesheets/nucore.scss
```diff
@@ -628,3 +634,44 @@ li:hover {
 .nav.pull-right {
   letter-spacing: $fontSpacing;
 }
+
+
+/*
+ UCONN Redesign Table View
+*/
+td, tr, {
```
jhanggi: Remove the trailing comma

app/assets/stylesheets/nucore.scss
```diff
@@ -628,3 +634,44 @@ li:hover {
...
+  transition: ease 0.2s all;
+  border-bottom: solid  $grey 1px;
+}
+.table-striped tbody > tr:nth-child(2n+1) {
+  td {
```
jhanggi: I think we should probably be consistent and use `th, td` rather than two identical blocks.

app/assets/stylesheets/nucore.scss
```diff
@@ -628,3 +634,44 @@ li:hover {
...
+
+
+/*
+ UCONN Redesign Table View
+*/
+td, tr, {
+  transition: ease 0.2s all;
+  border-bottom: solid  $grey 1px;
+}
+.table-striped tbody > tr:nth-child(2n+1) {
```
jhanggi: Can we add a space between these blocks?

jhanggi:
My bad on the comment about the billing tab. Ignore that comment. That's because those orders are 90+ days past fulfillment.

ColinMcNeil:
I ended up giving the table rows their own variables. I'm going to re-evaluate the background gradient and color contrast with Rob and Olivia.

PR #1780, opened 2018-11-27
Revert "Handle survey urls that already have a question mark in them"

vandrijevik:
Reverts tablexi/nucore-open#1771 because it seems to be causing https://rollbar.com/TableXI/nucore-nu/items/341/ and https://rollbar.com/TableXI/nucore-nu/items/342/


vandrijevik:
@jhanggi This is resulting in errors when certain orders are accessed for NU, so I am going to merge the revert without waiting for approval, in order to fix the problem on production.

jhanggi:
Ok, sounds good. I'll look into this further to try to figure out why it was causing problems.

PR #1786, opened 2018-12-04
Use-Manage Buttons

ColinMcNeil:
# Release Notes

Restyle buttons with the UCONN design. Buttons colored based on use/manage mode.

# Screenshot
## Add Facility in use mode
screen shot 2018-12-04 at 8 29 32 am

## Add instrument in manage mode
screen shot 2018-12-04 at 8 29 48 am

## Hover state
Simply adding a 15% alpha layer
screen shot 2018-12-04 at 8 29 54 am

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -673,7 +673,9 @@ thead {
   color: $grey;
   text-transform: uppercase;
   font-weight: bold;
-.btn-add {
+}
+
+.btn, .btn-add {
   border: solid $useColorDark 1px;
   outline: none;
   background: transparent;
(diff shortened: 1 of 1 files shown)


jhanggi:
Is there anything in your styles for `btn-primary`? I feel like there should be something that calls makes those stand out a little more, especially when there is something like a "Save" and a "Cancel".

ColinMcNeil:
Nothing in the design spec, but I agree so I gave them `font-weight: bold;` and `border-width: 1.5px;`. Stands out much better. Also gave them a different hover state.

app/assets/stylesheets/nucore.scss
```diff
@@ -632,21 +632,33 @@ li:hover {
...
   text-shadow: none;
 }
 
+.btn-primary {
+  font-weight: bold;
+  border-width: 1.5px;
+}
+
 .btn:hover, .btn-add:hover {
+  transition: ease 0.2s all;
```
jhanggi: Do we need this? The `.btn` style feels like it should cover it.

jhanggi:
This one also needs some merge conflict resolution.

app/assets/stylesheets/nucore.scss
ColinMcNeil: Typically I wouldn't, but bootstrap is applying a hover transition so I had to override that.

jhanggi: Gotcha. If you can clean up the conflicts, I'll merge it!

ColinMcNeil:
Should be good

PR #1788, opened 2018-12-04
Add support for multiple favicons

ColinMcNeil:
# Release Notes

Similar to the brand image, we have added the support for multiple favicon resolutions.

# Screenshot

screen shot 2018-12-04 at 11 19 03 am
As you can see, they are added via paths in `settings.yml`. They are only rendered in the head if a path is specified.

##
screen shot 2018-12-04 at 11 20 00 am

`favicon_link_tag` looks in `assets/images` for favicons, so we just added them to their own folder.


.gitignore
```diff
@@ -32,6 +32,8 @@ config/secrets.yml
 
 public/images/bg-search-form.pxm
 
+app/assets/images/favicons/*
```
jhanggi: I don't think these should be gitignored. Otherwise, they'll be left out of any deployments.

app/views/shared/_favicons.html.haml
```diff
@@ -0,0 +1,14 @@
+- if Settings.favicon_paths.x16.present?
```
jhanggi: Here's what we did for dartmouth: https://github.com/tablexi/nucore-dartmouth/blob/master/vendor/engines/dartmouth_radar/lib/dartmouth_radar/engine.rb and the associated icon file: https://github.com/tablexi/nucore-dartmouth/blob/master/vendor/engines/dartmouth_radar/app/views/dartmouth_radar/_icons.html.erb

We have a lot more icons including for windows. If we do want to bake this in, I wonder if there's a way to make the settings a little more structured and also allow more flexibility.

```yaml
favicons:
  - path: radar/icons/apple-touch-icon-57x57.png
    sizes: 57x57
    rel: apple-touch-icon-precomposed
  - ...
  - path: radar/icons/favicon-196x196.png
    sizes: 196x196
    rel: icon
```

db/schema.rb
```diff
@@ -780,10 +780,10 @@
   end
 
   create_table "vestal_versions", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
-    t.integer  "versioned_id"
     t.string   "versioned_type"
-    t.integer  "user_id"
+    t.integer  "versioned_id"
```
jhanggi: These are just moving ordering around of the columns, but can we exclude these changes?

app/views/shared/_favicons.html.haml
ColinMcNeil: If you think it's a good idea, maybe I could write something that loads the directory and automatically renders them based on file name.

jhanggi: That's not a bad idea. If it's something that can be done at application launch and cached then I think it's a good idea. I wouldn't want to need to do a disk read on every single request. We haven't really done anything with caching in this application, so I don't believe we have Rails's caching set up.

ColinMcNeil: I'm going to try myself a bit, but that sounds beneficial.

ColinMcNeil:
Alright, I'm now loading favicons from the directory. Right now the logic is in application_controller. I could move it to its own helper, or alternatively try a different approach entirely.

ColinMcNeil:
Moved it to a helper. I would love to fix those codeclimate issues, but I don't have access to see it.

app/helpers/favicons_helper.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+module FaviconsHelper
```
jhanggi: Overall, I like the idea here. I would maybe put this in `presenters/favicon_presenter.rb` and not even worry about the helper methods. I would also consider memoizing the results. If done in a class method, it means it would only be loaded into memory once. It would require you restarting the server if you add new icons, but would boost the performance of each page because it's not doing the extra file reads.

```ruby
class FaviconPresenter
  def self.favicons
    return @favicons if defined?(@favicons)
    pngs = Dir.glob("app/assets/images/favicons/*.png").map do |png|
      rel = png.include?("apple-touch-icon") ? "apple-touch-icon" : "shortcut icon"
      { path: "favicons/" + File.basename(png), rel: rel, type: "image/png" }
    end
    svgs = Dir.glob("app/assets/images/favicons/*.svg").map  do |svg|
      { path: "favicons/" + File.basename(svg), rel: 'shortcut icon', type: "image/svg" }
    end
    @favicons = pngs + svgs
  end
end

# in view
```haml
- FaviconPresenter.favicons.each do |favicon|
  = favicon_link_tag favicon[:path], rel: favicon[:rel], type: favicon[:type]
```

jhanggi:
Can you rebase this off of `nucore-open/master` so it only includes the commits specific to this feature. Right now, it's based off of `uconn-design-changes` so it's including all those (which we're not quite ready to merge into master.

ColinMcNeil:
Should be all set. If you would like to show me a better way to do it, I am still happy to pair.

jhanggi:
Changes look 👍 

I'd like this to go straight into `master` rather than sitting in the design branch while we wrap that up. Do you think you could get the PR so that it's only the favicon changes? It's still including a bunch of the design changes.

ColinMcNeil:
Okay, dropped the redesign commits. Seems to still work fine.

config/settings.yml
```diff
@@ -1,5 +1,15 @@
 header_logo_path: ~
 
+favicon_paths:
```
jhanggi: We don't need these now, do we?

ColinMcNeil:
Got that taken care of

PR #1787, opened 2018-12-04
Use-Manage Form

ColinMcNeil:
# Release Notes

Give forms and inputs UCONN design.

# Screenshot

## Basic text inputs
screen shot 2018-12-04 at 10 27 58 am

## Select Dropdown
screen shot 2018-12-04 at 10 28 31 am

## Control Groups
screen shot 2018-12-04 at 10 43 09 am

## Order Form
screen shot 2018-12-04 at 10 43 15 am

PR #1795, opened 2018-12-11
Allow JS capybara finders to find upcased labels

jhanggi:
# Release Notes

Tech Task: Support finding labels that have been upcased via CSS in specs.

# Additional Context

UCONN is applying a `text-transform` to `label`s. Capybara uses exact matches, so in JS mode where the styles get applied, it can't find them. This switches to using regexes to allow case insensitivity. See #1787.

### DIFF
--- a/spec/features/placing_an_order_on_behalf_of_spec.rb
+++ b/spec/features/placing_an_order_on_behalf_of_spec.rb
@@ -47,14 +47,14 @@ RSpec.describe "Placing an item order" do
     choose account.to_s
     click_button "Continue"
 
-    find("label", text: "More options").click
+    find("label", text: /More options/i).click
     fill_in "Order date", with: I18n.l(2.days.ago.to_date, format: :usa)
     select "Complete", from: "Order Status"
 
     click_button "Purchase"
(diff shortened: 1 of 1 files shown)


ColinMcNeil:
Not sure why its failing codeclimate, otherwise it works for me.

spec/features/placing_an_order_on_behalf_of_spec.rb
```diff
@@ -47,14 +47,14 @@
...
-    find("label", text: "More options").click
+    find("label", text: /More options/i).click
     fill_in "Order date", with: I18n.l(2.days.ago.to_date, format: :usa)
     select "Complete", from: "Order Status"
 
     click_button "Purchase"
 
     expect(page).to have_content "Order Receipt"
-    expect(page).to have_content "Ordered For\n#{user.full_name}"
+    expect(page).to have_content /Ordered For\n#{user.full_name}/i
```
jhanggi:
```suggestion
    expect(page).to have_content(/Ordered For\n#{user.full_name}/i)
```

PR #1804, opened 2018-12-13
Fixes for settings not being reset properly in tests

jhanggi:
# Release Notes

Tech task: Fix tests that don't properly clean up `Settings` after themselves. This also makes it explicit which tests need to reload their routes, which should speed up overall suite speed.

# Additional Context

The critical ones that caused https://github.com/tablexi/nucore-dartmouth/pull/244 to fail are related to the fiscal year. Using `allow(Settings.financial).to receive(: fiscal_year_begins)` makes it so rspec will un-stub it after the test is over so changes don't bleed across specs.

I could be convinced the `reload_routes` option isn't worth it.

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -31,16 +31,16 @@ RSpec.configure do |config|
   config.include Devise::Test::ControllerHelpers, type: :controller
   config.include FactoryBot::Syntax::Methods
 
-  config.around(:each, :feature_setting) do |example|
-    example.metadata[:feature_setting].each do |feature, value|
-      SettingsHelper.enable_feature(feature, value)
+  config.before(:each, :feature_setting) do |example|
+    example.metadata[:feature_setting].except(:routes).each do |feature, value|
+      allow(Settings.feature).to receive("#{feature}_on").and_return(value)
     end
-    Nucore::Application.reload_routes!
+    # Necessary since some routes are enabled/disabled
+    Nucore::Application.reload_routes! if example.metadata[:feature_setting][:reload_routes]
+  end
 
-    example.call
-
-    Settings.reload!
-    Nucore::Application.reload_routes!
+  config.after(:each, :feature_setting) do |example|
+    Nucore::Application.reload_routes! if example.metadata[:feature_setting][:reload_routes]
   end
 
   config.around(:each, :billing_review_period) do |example|
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -33,7 +33,7 @@ RSpec.configure do |config|
 
   config.around(:each, :feature_setting) do |example|
     example.metadata[:feature_setting].except(:reload_routes).each do |feature, value|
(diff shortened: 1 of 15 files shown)


spec/rails_helper.rb
```diff
@@ -32,15 +32,16 @@
   config.include FactoryBot::Syntax::Methods
 
   config.around(:each, :feature_setting) do |example|
-    example.metadata[:feature_setting].each do |feature, value|
-      SettingsHelper.enable_feature(feature, value)
+    example.metadata[:feature_setting].except(:reload_routes).each do |feature, value|
+      Settings.feature.public_send("#{feature}_on=", value)
```
vandrijevik: I am not entirely sure what value this Rspec-ism brings over writing `Settings.feature. product_specific_contacts_on = true` explicitly in the spec context where it is needed. I prefer that approach over `public_send`, because `public_send` makes grepping the code base impossible. It also seems more direct to me, and we don’t have to metaprogram our way to support this.

Just my 2¢—if you strongly prefer this type of approach, that’s fine by me.

jhanggi: The downside of putting `Settings.feature.product_specific_contacts_on = true` is it doesn't get cleaned up after the spec is done running and can bleed into other tests. The individual examples/groups that don't need routes that rely on the settings can get away with `allow(Settings.feature).to receive(:product_specific_contacts).and_return(true)`, but the ones that need the routes need some extra logic for reloading them. 

Using metadata tags seemed a reasonable way of standardizing/reusing the logic. Before we implemented this, it was too easy to forget to clean up `Settings` after yourself.

As far as the metaprogramming goes, I would normally agree with you, but this is a pretty narrow use case. Maybe it was clearer when it was hidden inside `SettingsHelper.enable_feature`? This was the only place that method was used.

I'm open to other ideas, though.

As far as `reload_routes`, most of the tests will pass without it here in open, but UIC has a bunch of features (`create_users`, `edit_accounts`, etc) disabled (as do NU and DC) so the tests would fail there.

vandrijevik: Gotcha, thanks for explaining about reload routes and the metadata tags, that makes sense!

As for the `public_send`, I think I’d be more okay with it if it didn’t add `_on` to the tag used, but instead just invoked the appropriate setter. That way, if you search for the key that is specified in `settings.yml` (for example, `daily_view_on`), you’ll get matches for all of the specs tagged with it, rather than nothing if the specs are just tagged with `daily_view`. What do you think about that?

As an aside, I’d be all for renaming the settings so they’re just called `daily_view`, rather than `something_on = true` because I find the `_on` in the names redundant, but that’s a bigger bite for another day.

jhanggi: 👍 to renaming the settings to get rid of the `_on`. I think the reason we went with appending it in the getter/setters (see also `SettingsHelper#feature_on?`) was to make the application API more fluent until we did the renames.

If we want to change this, I think I would rather spend the effort truncating the settings than finding all the places in the app and specs and appending `_on`.

PR #1809, opened 2018-12-18
Remove `_on` suffix from settings

jhanggi:
# Release Notes

Tech task: Remove `_on` suffix from all settings names.

# Additional Context

This is into https://github.com/tablexi/nucore-open/pull/1804/files

It will also be slightly tedious to clean up the merge conflicts when this goes downstream.

### DIFF
--- a/lib/settings_helper.rb
+++ b/lib/settings_helper.rb
@@ -32,10 +32,10 @@ module SettingsHelper
   #
   # Used to query the +Settings+ under feature:
   # [_feature_]
-  #   If you want to check setting 'feature.password_update_on'
-  #   then this parameter would be :password_update
+  #   If you want to check setting 'feature.password_update' then this parameter
+  #.  would be :password_update. Will raise an error if the setting does not exist.
   def self.feature_on?(feature)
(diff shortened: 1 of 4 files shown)


lib/settings_helper.rb
```diff
@@ -35,7 +35,7 @@ def self.relays_enabled_for_reservation?
   #   If you want to check setting 'feature.password_update_on'
   #   then this parameter would be :password_update
   def self.feature_on?(feature)
-    !!Settings.feature["#{feature}_on"]
+    !!Settings.feature[feature]
```
vandrijevik: Since we’re now just checking the hash key, what do you think about using `Settings.fetch(feature)` here, so we get an exception if we somehow try to check a setting that has not been configured? That way, it will be more obvious then getting back an unexpected result…

P.S. I think the comment above this method should be removed, since it’s no longer true 😄

jhanggi: e5e29239-a786-431d-b285-3e4640f6c20b

Actually, using `fetch` is a great idea.

PR #1812, opened 2018-12-20
[#137479] Allow configuring a support email for top-level link

vandrijevik:
# Release Notes

Allow configuring a support email for top-level link

# Additional Context

Shows a Support link in the top-level navigation if a support email has been configured. This way, we can configure per-school support emails, and this link is only displayed if such an email is configured.


app/views/shared/_support.html.haml
```diff
@@ -0,0 +1,3 @@
+- if Settings.support_email.present?
+  %li.visible-with-nav= mail_to Settings.support_email, t('pages.support')
```
jhanggi: Can we add a placeholder in `settings.yml` that can be null by default?

PR #1829, opened 2019-01-07
[#144419] Make Account-Facility relationship Many-to-Many

jhanggi:
# Release Notes

Tech task: Change Account-Facility relationship to be many-to-many.

# Additional Context

This leaves the existing `accounts.facility_id` in place, but uses the new relationship. That column will go away in a future PR. Interface will also come later.


app/models/account.rb
```diff
@@ -80,12 +99,13 @@ def self.using_journal?
...
-        facility: facility,
-      )
+      left_outer_joins(:facilities)
+        .where(
+          "accounts.type in (:global_account_types) or (accounts.type in (:facility_account_types) and account_facility_joins.facility_id = :facility)",
+          global_account_types: config.global_account_types,
+          facility_account_types: config.facility_account_types,
+          facility: facility,
+        )
     else
```
vandrijevik: I think this `where` can be simplified by extracting a scope for global accounts and using Rails’s `or` support. How about something like this?

```
scope :global, -> { where(type: config.global_account_types) }
scope :per_facility -> { where(type: config.facility_account_types) }

def self.for_facility(facility)
  if facility.single_facility?
    global.
    or(per_facility.left_ourter_joins(:facilities).where(account_facility_joins: { facility_id: facility.id }))
  else
```

app/models/account_facility_join.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AccountFacilityJoin < ApplicationRecord
```
vandrijevik: I really wish we could use `FacilityAccount`… This name is killing me here 😞

jhanggi: I would love to rename `FacilityAccount` to `RechargeAccount` (it would be more accurate and avoid some of the confusion in controllers--I'm looking at you `FacilityFacilityAccountsController`), but I feel that's a huge endeavor.

app/models/account.rb
jhanggi: I totally dig that if it works. I'll give it a shot. 

Also, I'm probably going to rename `single_facility?` and `cross_facility?` to something like `per_facility?` (if we think that's a good name) and `global?` since the old names won't really be accurate anymore.

vandrijevik: I like those names! My code may not work exactly as-is (the `.where` call may need a little more help but I think it’s worth it to use the hash over a raw string), but I think you get the idea.

PR #1830, opened 2019-01-08
[#137738] Total cost on pages

vandrijevik:
# Release Notes

Add total cost on unpaginated lists of transactions

# Screenshot

screen shot 2019-01-08 at 21 48 09

# Additional Context

I am not a fan of the way I keep track of how many columns this table has, so that I can position the `Total` correctly. I’m open to suggestions on how to make this better!

### DIFF
--- a/app/views/shared/transactions/_table.html.haml
+++ b/app/views/shared/transactions/_table.html.haml
@@ -1,4 +1,3 @@
-- order_details = local_assigns.key?(:order_details) ? order_details : @order_details
 #table_billing
   - if @export_enabled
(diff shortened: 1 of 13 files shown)


app/views/shared/transactions/_table.html.haml
```diff
@@ -1,4 +1,3 @@
-- order_details = local_assigns.key?(:order_details) ? order_details : @order_details
```
jhanggi: I fully support your cleanup of this!

jhanggi:
And let's hold off off on merging this until we can get the NU release out.

PR #1834, opened 2019-01-08
Add Docker local development environment

rdunlop:
# Release Notes

This creates a docker-compose environment which includes the rails app, and a mysql 5.7 database.

It doesn't include other database types, nor delayed_job (both of which could be useful for developers in this project). As I feel that those could be added in subsequent work.

The goal is to make spinning up a development environment faster/easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ spec/examples.txt
 .tags
 .tags1
 erd.*
+/ssh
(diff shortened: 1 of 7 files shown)


jhanggi:
Should we add a `delayed_job` container? There's also a bunch of cron jobs, but I don't know if that's worth trying to deal with just yet.

jhanggi:
Superseded by #2084

PR #1836, opened 2019-01-10
Account facility rename single cross methods

jhanggi:
# Release Notes

Tech task: Rename Account `single_facility?` and `cross_facility?` methods to better describe their new behavior.

# Additional Context

Extends from #1829

### DIFF
--- a/app/views/shared/_statements_table.html.haml
+++ b/app/views/shared/_statements_table.html.haml
@@ -9,7 +9,7 @@
         %th= Statement.human_attribute_name(:sent_to)
         - unless @account
           %th= Account.model_name.human
-        - unless current_facility.try(:single_facility?)
+        - if current_facility&.cross_facility?
           %th= Facility.model_name.human
         %th.currency # of #{Order.model_name.human.pluralize}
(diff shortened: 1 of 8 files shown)


app/views/shared/_statements_table.html.haml
```diff
@@ -9,7 +9,7 @@
         %th= Statement.human_attribute_name(:sent_to)
         - unless @account
           %th= Account.model_name.human
-        - unless current_facility.try(:single_facility?)
+        - if current_facility&.cross_facility?
```
vandrijevik: I think this should now be `&.global?`, no? Or am I confused by `current_facility` (which is not an account)?

jhanggi: `Facility#cross_facility?` and `Facility#single_facility?` are different from account. And they make sense (at least to me). `Facility#cross_facility?` returns true for the facility null object (when you're in the admin cross facility context). `single_facility?` returns true when you're in a single facility's context.

This was one of the first ones I caught going through the diff after a global find/replace and realized I replaced a bunch of `cross_facility?`s on Facilities that shouldn't had gotten applied.

PR #1841, opened 2019-01-16
[#144419] Remove Direct Account -> Facility references

jhanggi:
# Release Notes

Tech task: Remove direct references from Account to facilities. We should always be using the many-to-many relationship now. The `accounts.facility_id` field is still populated.

# Additional Context

The `accounts.facility_id` will be completely removed in #1863. The `facility_id` can still be used in the meantime by BI, but will only reference the original facility it was created under.


jhanggi:
Normally, I would agree with the PR basing you suggested. My only concern is I'd like to hold off on merging this one for a little bit. I'd like to leave the duplicate references (`accounts.facility_id` and the join table) so that NU's BI team has a chance to update any of their queries. That's part of why I featured flagged #1840. As long as the interface isn't used, there should be no problems with the DB relationships. Does that make sense? Or do you think we should go ahead and pull this in?

vandrijevik:
Oh, that makes more sense to me now (when you said hold off on merging, I thought you were talking about #1840 and #1841 together, but this makes way more sense)!

Agreed.

jhanggi:
@vandrijevik I made a few more updates to this PR and made sure it's up to date with master. I realized this PR didn't actually include the migration to remove the foreign key column, so there's nothing really blocking this anymore.

There's one edge case: Create an account in Facility A. Use the new form to change to Facility B (removing A). NUcore should treat it as Facility B, but the database will still reference Facility A.

vandrijevik:
> There's one edge case: Create an account in Facility A. Use the new form to change to Facility B (removing A). NUcore should treat it as Facility B, but the database will still reference Facility A.

🤔 I don’t think this is worth fixing, but it makes me think: if NU turns on this feature, should we merge this at the same time as the feature and eliminate the window during which we’d have this edge case?

jhanggi:
I'm cool with not turning on the feature in production until this is merged. I just hope they can get the BI updates in place quickly.

vandrijevik:
> Also, this handles the issue Aaron reported with the 403s when visiting an account that was created under a different facility.

@jhanggi How easy is it to pull this fix only into a separate PR, so we can put it on NU’s staging for Aaron to take a look at? I think it would be good to demonstrate the fix before we ask them to turn the feature on?

jhanggi:
I'm hoping it's easy. Let me try.

jhanggi:
@vandrijevik Do you think we can close this PR and just move forward with #1863. If I remember correctly (and this was now a year ago at this point) that we would maybe merge this PR while we wait for BI to make their updates and then we'd merge the other one. I think we're getting close with their team that we can just merge the other one once we get the go-ahead. Does that sound right?

vandrijevik:
@jhanggi Yes, I believe that’s correct! Once we have the final go-ahead from the BI team, I think we can merge and deploy https://github.com/tablexi/nucore-open/pull/1863, so this can be closed.

PR #1845, opened 2019-01-23
Remove unused and unusable InstrumentsController#instrument_status

vandrijevik:
# Release Notes

TECH TASK: Remove unused and unusable InstrumentsController#instrument_status

# Additional Context

I was scratching my head with the `InstrumentStatus` records while working on `#139335`, and this action is unused and unusable (because there is no local variable or method called `instrument`, it always raises `ActiveRecord::RecordNotFound`, which results in a 404).

### DIFF
--- a/app/controllers/instruments_controller.rb
+++ b/app/controllers/instruments_controller.rb
@@ -55,22 +55,6 @@ class InstrumentsController < ProductsCommonController
     end
   end
 
-  # GET /facilities/:facility_id/instruments/:instrument_id/status
-  def instrument_status
-    begin
-      @relay = @product.relay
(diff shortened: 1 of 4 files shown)


vandrijevik:
For now, I returned the template because it is [rendered in the `#switch` action](https://github.com/tablexi/nucore-open/blob/c69e1b755d154e306af52d3770cd92e9c809a757/app/controllers/instruments_controller.rb#L105-L106), but I think that it can go away too — I think that action is only ever triggered via JSON. I will verify that separately, and if so, I will do that cleanup in another PR.

vandrijevik:
@jhanggi Please hold on the review, there is a bit more cleanup for me to do here.

vandrijevik:
This is ready for a look now.

PR #1848, opened 2019-01-25
[#139335] Tool for turning all relays for a core on or off, at once

vandrijevik:
# Release Notes

Tool for turning all relays for a core on or off, at once

# Screenshot

screen shot 2019-01-25 at 13 51 11

# Additional Context

Since these actions call activate/deactivate on all relays regardless of their current status, I am not creating InstrumentStatus records on the instruments to signify that there was a change (because perhaps there was not one). This should be okay, because https://github.com/tablexi/nucore-open/blob/master/app/controllers/instruments_controller.rb#L72-L77 reconciles a relay’s actual status by creating (or skipping) an InstrumentStatus record on view.

### DIFF
--- a/app/controllers/relays_activations_controller.rb
+++ b/app/controllers/relays_activations_controller.rb
@@ -7,21 +7,23 @@ class RelaysActivationsController < ApplicationController
   authorize_resource :facility
 
   def create
-    current_facility.instruments.includes(:relay).select(&:has_real_relay?).each do |instrument|
-      instrument.relay.activate
-      instrument.instrument_statuses.create(is_on: true)
-    end
-
+    relays.each(&:activate)
     redirect_to facility_instruments_path(current_facility), flash: { notice: text("turned_on") }
+  rescue NetBooter::Error
+    redirect_to facility_instruments_path(current_facility), flash: { notice: text("connection_error") }
   end
 
   def destroy
-    current_facility.instruments.includes(:relay).select(&:has_real_relay?).each do |instrument|
-      instrument.relay.deactivate
-      instrument.instrument_statuses.create(is_on: false)
-    end
-
+    relays.each(&:deactivate)
     redirect_to facility_instruments_path(current_facility), flash: { notice: text("turned_off") }
+  rescue NetBooter::Error
+    redirect_to facility_instruments_path(current_facility), flash: { notice: text("connection_error") }
+  end
+
+  private
+
+  def relays
+    current_facility.instruments.includes(:relay).select(&:has_real_relay?).map(&:relay)
   end
 
 end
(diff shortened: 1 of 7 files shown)


app/views/admin/products/index.html.haml
```diff
@@ -12,7 +12,11 @@
 %p= text("views.admin.#{plural_object_name}.index.description")
 
 - if current_user.manager_of?(current_facility)
```
jhanggi: Since we're in here, I feel like we should switch to using an ability check. `can?(:create, Product)`

app/controllers/relays_activations_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class RelaysActivationsController < ApplicationController
+
+  before_action :authenticate_user!
+  before_action :init_current_facility
+  authorize_resource :facility
```
jhanggi: Can you double check this permissioning and possibly add a spec for it? I suspect this will check for `:create`/`:destroy` on `Facility`, which will only be available to global admins. I believe we want to allow facility directors (and probably facility administrators since they pretty much share permissions).

app/controllers/relays_activations_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  def create
+    relays.each(&:activate)
```
jhanggi: This is probably ok, but I know there are some facilities with lots of relays so this could be very slow (some facilities take 12-15 seconds to collect all the statuses per skylight). Would it be worth fanning out a bunch of background jobs rather than doing each one inline? Just a thought.

vandrijevik: One thing we’re saving on here is that we’re _not_ checking the status after doing an activate/deactivate, so I’d like to move to background jobs only if this starts to be problematic, if that’s okay!

app/controllers/relays_activations_controller.rb
vandrijevik: I thought I had checked it as a Facility Director locally, but I will double-check!

app/views/admin/products/index.html.haml
vandrijevik: Oh, what’s the benefit of that? I thought `current_user.manager_of?` reads very nicely, but perhaps there are some drawbacks I don’t get?

app/controllers/relays_activations_controller.rb
vandrijevik: I checked this in the UI by signing in as ddi123@example.com, and the permissions were correct. I also changed the existing spec to [use a more restrictive Facility Director user](https://github.com/tablexi/nucore-open/pull/1848/files#diff-c24558c22a835cfca78251df8c28fcf0R7), so we’re testing the permissions as well as the behavior.

app/views/admin/products/index.html.haml
vandrijevik: Changed in d53a621!

vandrijevik:
Thanks for the feedback, @jhanggi! I made some changes, and I think this is ready for another look.

PR #1860, opened 2019-02-01
[#147756] Fix PO page when multi_facility_accounts is turned off

vandrijevik:
# Release Notes

Fix PO page when multi_facility_accounts is turned off

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -17,6 +17,10 @@ module AccountsHelper
     end
   end
 
+  def show_account_facilities_tab?(ability, account)
+    SettingsHelper.feature_on?(:multi_facility_accounts) && account.per_facility? && ability.can?(:edit, AccountFacilityJoinsForm.new(account: account))
+  end
+
   private
 
   def available_accounts_array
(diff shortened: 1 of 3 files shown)


app/helpers/accounts_helper.rb
```diff
@@ -17,6 +17,10 @@ def payment_source_link_or_text(account)
     end
   end
 
+  def show_account_facilities_tab?(ability, account)
```
jhanggi: I’m pretty sure `current_ability` is a view helper, so I think it would be available inside this method. I don’t think it needs to be passed in as an argument.

vandrijevik: I know, but I’m not a fan of relying on shared state in helper methods (because that state is outside of the module), so I try to make them take all the arguments they need; are you okay with it like this, or do you think consistency (and using `current_ability`) is preferable?

jhanggi: I’m good with it. Your logic is sound. :shipit: (damn. The custom emoji don’t work on the mobile version)

jhanggi: Oh, they do! They just don’t auto suggest

vandrijevik: lol 😂

PR #1863, opened 2019-02-01
[#144419] Drop accounts.facility_id column

jhanggi:
# Release Notes

Tech task: Drop `accounts.facility_id` column.

# Additional Context

As of #1841, this column is no longer used. We should be able to merge that PR, and leave this one as the blocked story until NU gets BI updated.

Clean diff: https://github.com/tablexi/nucore-open/compare/account_facility_many_to_many_remove_foreign_key...account_facility_many_to_many_remove_foreign_key_drop_column?expand=1


db/migrate/20190201222556_remove_facility_id_from_accounts.rb
```diff
@@ -0,0 +1,27 @@
...
+                  index: { name: "fk_account_facility_id" }
+
+    # If there happen to be more than one join, then :shrug:, only one will get set.
+    execute <<~SQL
+      UPDATE accounts
+      SET accounts.facility_id = (
+        SELECT facility_id
+        FROM account_facility_joins
+        WHERE accounts.id = account_facility_joins.account_id
+        LIMIT 1
```
vandrijevik: Since we know that the oldest row in `account_facility_joins` came from the migration we did to populate that table, can we add an `ORDER BY created_at DESC` here, and we’ll be certain that we’re restoring the data correctly? That way, we don’t need the 🤷‍♂️ comment, and we’re not restoring a random associated facility but the one where the account was created originally.

jhanggi: Except where they removed the original facility and then re-added it. And we're still technically losing data, so I think a comment is probably still warranted. But yeah, the `ORDER BY` sounds like a good plan to cover everything but a super weird case (and we're hopefully not going to need the `down` anyways).

PR #1866, opened 2019-02-05
Upgrade Ruby to 2.5

vandrijevik:
# Release Notes

Upgrade Ruby to 2.5.5

# Additional Context

This PR **should not** be merged into `master` until all of the universities are running Ruby 2.5.5. I have opened it and marked it as Blocked, so that I can pull this branch into each university-specific fork to do releases that only upgrade Ruby, and after I’ve completed that everywhere, I will merge this PR.


app/models/reports/instrument_utilization_report.rb
```diff
@@ -65,7 +65,7 @@ def /(other)
     end
 
     def safe_divide(a, b)
-      return 0 if b == 0
+      return 0.0 if b.zero?
```
jhanggi: Fun times when dividing by zero.

vandrijevik:
Given that ruby 2.5.5, which includes security vulnerability fixes, has come out in the meantime, I’ve updated this PR (and will update UIC and Dartmouth, which run on 2.5 already, during the week).

PR #1873, opened 2019-02-19
[#142504] Add “Ordered For” Filter on Billing Pages and remove Facilities Filter

rolfrussell:
# Release Notes

Add a filter for the Ordered For field on the following Billing pages. Also, hide the Facilities filter in the same contexts

# Screenshot

screen shot 2019-02-19 at 11 09 13 am

# Additional Context

As this is my first story I'll put details on how I did the story to make sure I covered everything ...

This is the story request:

> Add a filter for the Ordered For field on the following Billing pages. Also, hide the Facilities filter in the same contexts:# 
> 
> - All Transactions
> - Send Notifications
> - Orders In Review
> - Move Transactions
> - Disputed Orders
> - Create Journal
> - Create Statements
> 

It seems like the filters displayed depend on what is sent to `TransactionSearch::Searcher.new`
So first I went to find all the uses of `TransactionSearch:Searcher` to see what might be impacted by this change.
Here are the controller methods that use the default searchers (send nothing to `.new`)
 - `FacilitiesController.transactions`  (GET /facilities/:facility_id/transactions)
 - `FacilityNotificationsController.index`   (GET /facilities/notifications)
 - `FacilitiesController.disputed_orders` (GET /facilities/:facility_id/disputed_orders)
 - `FacilitiesController.movable_transactions` (GET /facilities/:facility_id/movable_transactions)
 - `FacilityJournalsController.new`  (GET /facilities/journals/new)
 - `FacilityNotificationsController.in_review`  (GET /facilities/notifications/in_review)
 - `FacilityStatementsController.new`  (GET /facilities/:facility_id/statements/new)

These seem to correspond rather cleanly to the list in the story.  
So I believe I can just change the default searchers in `TransactionSearch::Searcher` and the story is done.

In a lucky break there was already a `OrderedForSearcher`.  So I just used that.

To be complete here are the other places that `TransactionSearch:Searcher` is used.
For each of these a specific list of Searchers is passed to `TransactionSearch:Searcher.new` so changing the defaults should have no affect on these.
 - `FacilityAccountsReconciliationController.index`
 - `TransactionsController.index`
 - `TransactionsController.in_review`
 - `NewInprocessController.index`
 - `NewInprocessController.index`
 - `ProblemOrderDetailsController.show_problems`
 - `SecureRooms::FacilityOccupanciesController.index`

I checked a few of these to confirm no impact

I also searched nucore-nu or nucore-dartmouth for uses of `TransactionSearch::Searcher`, but didn't find anything different from nucore-open.  I didn't search in any other repos.

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -36,7 +36,7 @@ class TransactionsController < ApplicationController
       },
     )
 
-    @search = TransactionSearch::Searcher.search(searchers, order_details, @search_form)
+    @search = TransactionSearch::Searcher.new(searchers).search(order_details, @search_form)
     @date_range_field = @search_form.date_params[:field]
     @order_details = @search.order_details
 
@@ -51,7 +51,7 @@ class TransactionsController < ApplicationController
     order_details = current_user.administered_order_details.in_review
 
     @search_form = TransactionSearch::SearchForm.new(params[:search])
-    @search = TransactionSearch::Searcher.search(searchers, order_details, @search_form)
+    @search = TransactionSearch::Searcher.new(searchers).search(order_details, @search_form)
     @date_range_field = @search_form.date_params[:field]
     @order_details = @search.order_details
 
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -8,6 +8,18 @@ class TransactionsController < ApplicationController
 
   include OrderDetailsCsvExport
 
+  cattr_accessor(:searchers) do
+    [
+      TransactionSearch::FacilitySearcher,
+      TransactionSearch::AccountSearcher,
+      TransactionSearch::ProductSearcher,
+      TransactionSearch::DateRangeSearcher,
+      TransactionSearch::OrderStatusSearcher,
+      TransactionSearch::AccountOwnerSearcher,
+      TransactionSearch::OrderedForSearcher,
+    ]
+  end
+
   def initialize
     @active_tab = "accounts"
     super
@@ -24,13 +36,7 @@ class TransactionsController < ApplicationController
       },
     )
 
-    @search = TransactionSearch::Searcher.new(TransactionSearch::FacilitySearcher,
-                                              TransactionSearch::AccountSearcher,
-                                              TransactionSearch::ProductSearcher,
-                                              TransactionSearch::DateRangeSearcher,
-                                              TransactionSearch::OrderStatusSearcher,
-                                              TransactionSearch::AccountOwnerSearcher,
-                                              TransactionSearch::OrderedForSearcher).search(order_details, @search_form)
+    @search = TransactionSearch::Searcher.search(searchers, order_details, @search_form)
     @date_range_field = @search_form.date_params[:field]
     @order_details = @search.order_details
 
@@ -45,13 +51,7 @@ class TransactionsController < ApplicationController
     order_details = current_user.administered_order_details.in_review
 
     @search_form = TransactionSearch::SearchForm.new(params[:search])
-    @search = TransactionSearch::Searcher.new(TransactionSearch::FacilitySearcher,
-                                              TransactionSearch::AccountSearcher,
-                                              TransactionSearch::ProductSearcher,
-                                              TransactionSearch::DateRangeSearcher,
-                                              TransactionSearch::OrderStatusSearcher,
-                                              TransactionSearch::AccountOwnerSearcher,
-                                              TransactionSearch::OrderedForSearcher).search(order_details, @search_form)
+    @search = TransactionSearch::Searcher.search(searchers, order_details, @search_form)
     @date_range_field = @search_form.date_params[:field]
     @order_details = @search.order_details
 
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -8,18 +8,6 @@ class TransactionsController < ApplicationController
 
   include OrderDetailsCsvExport
 
-  cattr_accessor(:searchers) do
-    [
-      TransactionSearch::FacilitySearcher,
-      TransactionSearch::AccountSearcher,
-      TransactionSearch::ProductSearcher,
-      TransactionSearch::DateRangeSearcher,
-      TransactionSearch::OrderStatusSearcher,
-      TransactionSearch::AccountOwnerSearcher,
-      TransactionSearch::OrderedForSearcher,
-    ]
-  end
-
   def initialize
     @active_tab = "accounts"
     super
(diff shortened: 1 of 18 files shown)


jhanggi:
The place this doesn't quite work is the cross-facility billing tab. I think we still want the facility filter there. It seemed just a bit too easy, didn't it :)

You can access it as `bba123@example.com` (if you're running from the seeds). That's a billing admin. http://localhost:3000/facilities/all/transactions

jhanggi:
I'm not sure what the right solution is. You could do something like this, but you would need to do that in every place (and expose `attr_reader :searchers`.

```
    @search = TransactionSearch::Searcher.new.tap do |searcher|
      searcher.searchers.unshift(TransactionSearch::FacilitySearcher) if current_facility.cross_facility?
    end.search(order_details, @search_form)
```

We could also do away with the idea of "default" searchers. Maintaining a consolidated list probably still makes sense, but maybe something that can take something about the context in. And it might be better to think about the list as the "billing" or "billing" tab searchers.

Sorry if that's rambling, I just don't have the right solution in my head. Maybe you and vlad can think of something better.

jhanggi:
One other thought: it might be nice to try to collapse the `optimized` method as part of this feature. It's a remnant of a previous refactor where we had two ways of doing the search (the old way had a bunch of metaprogramming in the controllers, but still used a similar pattern with the `Searcher`s) side by side as we converted the pages one by one. The old way should be all gone at this point.

See https://github.com/tablexi/nucore-open/blob/master/app/services/transaction_search/searcher.rb#L50

rolfrussell:
@jhanggi  I made the change to handle all facilities.  I'm not very excited about how it looks but no solutions seemed clean and concise.

rolfrussell:
> One other thought: it might be nice to try to collapse the `optimized` method as part of this feature. It's a remnant of a previous refactor where we had two ways of doing the search (the old way had a bunch of metaprogramming in the controllers, but still used a similar pattern with the `Searcher`s) side by side as we converted the pages one by one. The old way should be all gone at this point.
> 
> See https://github.com/tablexi/nucore-open/blob/master/app/services/transaction_search/searcher.rb#L50

@jhanggi I didn't have more than 2 minutes to look at the optimization stuff.  Normally I'd look more before asking but given our timezones I better ask - can you give me an overview of what it is doing?

app/services/transaction_search/searcher.rb
```diff
@@ -26,22 +26,20 @@ class Searcher
       ]
     end
 
-    # Shorthand method if you only want the default searchers
-    def self.search(order_details, params)
-      new.search(order_details, params)
+    # Searchers used on the billing page
+    def self.billing_searchers(include_facilities = false)
```
jhanggi: Your changes aren't bad, but I don't love the change in API. What if we did:

```
def self.billing_search(order_details, params, facility:)
  searchers = default_searchers.dup
  searchers.unshift(TransactionSearch::FacilitySearcher) if facility.cross_facility?
  new(searchers).search(order_details, params)
end
```
Or use a `include_facility:` keyword to keep that logic out of this.

app/services/transaction_search/searcher.rb
```diff
@@ -26,22 +26,20 @@ class Searcher
...
     end
 
-    # Expects an array of `TransactionSearch::BaseSearcher`s
-    def initialize(*searchers)
-      searchers = self.class.default_searchers if searchers.blank?
-
-      @searchers = Array(searchers)
+    # Shorthand method
+    def self.search(searchers, order_details, params)
+      new.search(searchers, order_details, params)
```
jhanggi: You could do this like this and not need to change the instance method API
```
new(searchers).search(order_details, params)`
```

spec/services/transaction_search/searcher_spec.rb
```diff
@@ -9,31 +9,31 @@
     let(:order) { create(:purchased_order, product: item) }
     let(:order_detail) { order.order_details.first }
     let(:account) { order.account }
-    let(:searcher) { described_class.new(TransactionSearch::AccountSearcher, TransactionSearch::DateRangeSearcher) }
+    let(:searchers) { [TransactionSearch::AccountSearcher, TransactionSearch::DateRangeSearcher] }
```
jhanggi: In general, I think I prefer the instance to be initialized with the list of searchers, then you can search on that.

jhanggi:
> Also how would I test something like `facility_occupancies_controller.rb`

Check out https://github.com/tablexi/nucore-open/blob/master/vendor/engines/secure_rooms/spec/features/facility_occupancies_search_spec.rb

app/services/transaction_search/searcher.rb
```diff
@@ -9,10 +9,10 @@ class Searcher
     # to SearchForm) that `register` handles.
     cattr_accessor(:default_searchers) do
       [
-        TransactionSearch::FacilitySearcher,
         TransactionSearch::AccountSearcher,
         TransactionSearch::ProductSearcher,
         TransactionSearch::AccountOwnerSearcher,
+        # TransactionSearch::OrderedForSearcher,
```
jhanggi: Is this intentionally left commented out?

app/services/transaction_search/searcher.rb
```diff
@@ -31,10 +31,16 @@ def self.search(order_details, params)
       new.search(order_details, params)
     end
 
+    # Shorthand method for searchers used on the billing page
+    def self.billing_search(order_details, params, include_facilities = false)
```
jhanggi: I might consider making `include_facilities` a keyword argument. After about 3 arguments, I like to start naming them, especially optional ones.

app/controllers/transactions_controller.rb
```diff
@@ -45,13 +51,7 @@ def in_review
...
 
     @search_form = TransactionSearch::SearchForm.new(params[:search])
-    @search = TransactionSearch::Searcher.new(TransactionSearch::FacilitySearcher,
-                                              TransactionSearch::AccountSearcher,
-                                              TransactionSearch::ProductSearcher,
-                                              TransactionSearch::DateRangeSearcher,
-                                              TransactionSearch::OrderStatusSearcher,
-                                              TransactionSearch::AccountOwnerSearcher,
-                                              TransactionSearch::OrderedForSearcher).search(order_details, @search_form)
+    @search = TransactionSearch::Searcher.new(searchers).search(order_details, @search_form)
```
jhanggi: This might need to be `TransactionSearch::Searcher.new(
*searchers)`, but maybe the initializer is smart enough to deal with either type of input.

app/services/transaction_search/searcher.rb
rolfrussell: 👍

spec/services/transaction_search/searcher_spec.rb
rolfrussell: 👍

app/services/transaction_search/searcher.rb
rolfrussell: yeah I had an error that I hadn't gotten to yet

app/services/transaction_search/searcher.rb
```diff
@@ -32,7 +32,7 @@ def self.search(order_details, params)
     end
 
     # Shorthand method for searchers used on the billing page
-    def self.billing_search(order_details, params, include_facilities = false)
+    def self.billing_search(order_details, params, include_facilities:)
```
jhanggi: I don't have a strong feeling either way, but this could be optional with a default of `false`.

PR #1879, opened 2019-02-26
Only execute multi_facility test if the c2po engine is enabled

vandrijevik:
# Release Notes

Only execute multi_facility test if the c2po engine is enabled

# Additional Context

The c2po engine adds the ability to accept credit cards and purchase orders. When this engine is not turned on (like the case is with UIC), the test modified in this PR can’t run because the code it tests is not loaded. Hence, this PR adds a guard for executing the test only when the engine is turned on.

### DIFF
--- a/spec/controllers/facility_accounts_controller_spec.rb
+++ b/spec/controllers/facility_accounts_controller_spec.rb
@@ -79,17 +79,19 @@ RSpec.describe FacilityAccountsController, feature_setting: { edit_accounts: tru
       end
     end
 
-    context "when the multi_facility_accounts feature is turned off", feature_setting: { multi_facility_accounts: false, reload_routes: true } do
-      let(:purchase_order) { FactoryBot.create(:purchase_order_account, :with_account_owner, facility: facility) }
+    if EngineManager.engine_loaded?(:c2po)
+      context "when the multi_facility_accounts feature is turned off", feature_setting: { multi_facility_accounts: false, reload_routes: true } do
+        let(:purchase_order) { FactoryBot.create(:purchase_order_account, :with_account_owner, facility: facility) }
 
-      before do
-        sign_in admin
-      end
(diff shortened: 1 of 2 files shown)


spec/controllers/facility_accounts_controller_spec.rb
```diff
@@ -79,17 +79,19 @@
       end
     end
 
-    context "when the multi_facility_accounts feature is turned off", feature_setting: { multi_facility_accounts: false, reload_routes: true } do
-      let(:purchase_order) { FactoryBot.create(:purchase_order_account, :with_account_owner, facility: facility) }
+    if EngineManager.engine_loaded?(:c2po)
+      context "when the multi_facility_accounts feature is turned off", feature_setting: { multi_facility_accounts: false, reload_routes: true } do
+        let(:purchase_order) { FactoryBot.create(:purchase_order_account, :with_account_owner, facility: facility) }
```
jhanggi: I think I see that this spec is failing when the engine isn't loaded because it's referencing a PO. Is it possible to have it reference a plain account? I don't love using the status of engines to decide what specs to run. But it's not the end of the world.

vandrijevik: The multi-facility feature is only for POs and credit cards, so I don’t think we can use just a regular account here… Perhaps the spec belongs in the engine’s set of specs!

vandrijevik:
@jhanggi I moved the spec into the engine’s spec for the same controller (which already existed), so we don’t have to check whether the engine is loaded from outside it. Since it’s 9:42pm where you are, and you already approved the previous approach (which was clunkier), I am going to merge this so I can do the UIC release. If you disagree, let me know and I can make changes afterwards!

jhanggi:
I haven’t seen the actual change, but the move sounds good. :+1:

PR #1881, opened 2019-02-27
[#144828] Toggle training requests per product

jhanggi:
# Release Notes

Adds the ability to toggle the training requests feature per product.

# Additional Context

One of the things I've done here is bring the instrument show and product show behavior closer together. Before, if there was an error such as missing schedule rules, or user not having accounts that can be used for the product, for instruments it would redirect back to the facility homepage and display an error, while for items and services it would show the product description with the error message. NU gave us the thumbs up for this change of behavior.

There's also some improvements to the admin interface to still do. The client would like a dropdown, which I agree would probably be a more efficient and intuitive interface, but it probably warrants a form object, which makes that page more complicated. Thoughts on that?

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -135,6 +135,10 @@ li.nav-spacer {
   @extend .alert-info;
 }
 
+.alert p:last-child {
+  margin-bottom: 0;
+}
+
 /*
   Product List
 */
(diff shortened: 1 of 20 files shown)


jhanggi:
I'm not sure we'll want to merge this until we have a solution for the admin interface, but this should be everything for the front-end side, so it's definitely ready for a review.

jhanggi:
Sounds like NU is cool with using a simple JS toggle (which we already have the code for, we just need to add a couple `data` attributes), so I'll get that up shortly.

jhanggi:
@vandrijevik I've pushed up a change to the interface. I also made sure to run all the tests with the training requests feature off (https://circleci.com/gh/tablexi/nucore-open/7316?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), so we should still be good when we merge down to UIC.

Issue #1884, opened 2019-02-27
Provide documentation that setup with sqlite database is possible


fwolfst:
Although probably not a good idea, setting up nucore with a SQLite database is possible, however it lacks documentation.

vandrijevik:
Closing per https://github.com/tablexi/nucore-open/pull/1885#issuecomment-468291757.

PR #1885, opened 2019-02-27
[#1884] Provide instructions for SQLite but no hope

fwolfst:
# Release Notes

Added a file under doc/ with instructions on how to use nucore with a sqlite-database. For some developers (me ...) this might be the prefered way over using a MySQL or Oracle setup. However it has to be made clear that this is necessarily a good idea for production use.

To be fair I should note that I did not run tests and used ruby 2.5.1 .

# Screenshot

Optional.

# Additional Context

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Open source version of Northwestern University Core Facility Management Software
 Welcome to NUcore! This guide will help you get a development environment up and running. It makes a few assumptions:
 
 1. You write code on a Mac.
-2. You have a running Oracle or MySQL instance with two brand new databases. (Oracle setup instructions [here](doc/HOWTO_oracle.txt).)
+2. You have a running Oracle or MySQL instance with two brand new databases. (Oracle setup instructions [here](doc/HOWTO_oracle.txt).) . Installation with a SQLite database is [possible](doc/HOWTO_sqlite.md) but not supported.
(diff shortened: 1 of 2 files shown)


vandrijevik:
Hi @fwolfst,

Thank you very much for your support of `nucore-open` and this contribution!

Unfortunately, I don’t think it is best to merge this documentation, since we do not support SQLite in any way, and we should not be implying that we do (which we fear we would be even if there are clear warnings in the documentation).

We are currently experimenting with using Docker in some capacity for this project, which would make installing MySQL even easier than it is today, so we recommend that you use MySQL for development to avoid spending time tracking down bugs or differences that may arise if you use SQLite.

Once again, thank you for your work and this contribution!

Best,
Vlad

fwolfst:
Understandable. I personally prefer quick Docker-less, SQLite-setups for hacking and evaluating, but thank god that tastes differ.

I do not agree with your risk assessment, but also I don't have to deal with support and issues myself :)
Btw, if there would have been a demo-instance running somewhere, I wouldnt have had the wish to trick around.

TL;DR: All good, I understand if you dont want to merge.

vandrijevik:
Thanks for your understanding, @fwolfst, and I certainly agree that different tastes are a good thing 😄

PR #1887, opened 2019-02-28
Make product "access control" field a dropdown

jhanggi:
# Release Notes

Make the access control option for products a dropdown rather than two checkboxes.

# Additional Context

Into #1881, but I wanted to keep it as a separate PR to get feedback on this approach.

### DIFF
--- a/app/controllers/products_common_controller.rb
+++ b/app/controllers/products_common_controller.rb
@@ -12,7 +12,7 @@ class ProductsCommonController < ApplicationController
 
   include TranslationHelper
 
-  load_resource except: [:show, :manage, :index], instance_name: :product
+  # load_resource except: [:show, :manage, :index], instance_name: :product
   authorize_resource except: [:show, :manage], instance_name: :product
 
   layout "two_column"
@@ -57,9 +57,9 @@ class ProductsCommonController < ApplicationController
 
   # POST /services
   def create
-    @product = current_facility_products.new(resource_params)
-    @product.initial_order_status_id = OrderStatus.default_order_status.id
+    @product = current_facility_products.new(initial_order_status: OrderStatus.default_order_status)
     @product_form = EditProductForm.for(@product)
+    @product_form.assign_attributes(resource_params)
     if @product_form.save
       flash[:notice] = "#{@product.model_name.human} was successfully created."
       redirect_to([:manage, current_facility, @product])
--- a/app/controllers/products_common_controller.rb
+++ b/app/controllers/products_common_controller.rb
@@ -52,15 +52,16 @@ class ProductsCommonController < ApplicationController
   # GET /services/new
   def new
     @product = current_facility_products.new(account: Settings.accounts.product_default)
+    @product_form = EditProductForm.for(@product)
   end
 
   # POST /services
   def create
     @product = current_facility_products.new(resource_params)
     @product.initial_order_status_id = OrderStatus.default_order_status.id
-
-    if @product.save
-      flash[:notice] = "#{@product.class.name} was successfully created."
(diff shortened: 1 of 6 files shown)


app/forms/edit_product_form.rb
```diff
@@ -0,0 +1,34 @@
+class EditProductForm < TranslatableBaseForm
```
jhanggi: Using a form object felt more right that putting some of these methods directly onto Product. It did mean I had to do some extra work to get the form itself to get translations properly.

app/forms/translatable_base_form.rb
```diff
@@ -0,0 +1,30 @@
+class TranslatableBaseForm < SimpleDelegator
+
+  extend ActiveModel::Translation
+
+  def self.for(instance)
+    instance_class = instance.class
+    # Dynamically create a subclass where the model_name is based off the original
+    # instance's type so that `human_attribute_name` can work off the same translations
+    # as the original object.
+    Class.new(self) do
```
jhanggi: Please let me know if I've gone too meta here. It's definitely a little tricksy, but makes for an object that is virtually indistinguishable from the object itself in regards to the view and its translations.

I went kind of meta because we need a class-level `model_name` method in order for `human_attribute_name` to work, which is what simple_form uses under the hood for its labels. We usually don't need something quite so tricky because our form objects are a one-to-one mapping from a form to a single class. This one should support any of the subclasses of product.

app/controllers/products_common_controller.rb
```diff
@@ -12,7 +12,7 @@ class ProductsCommonController < ApplicationController
 
   include TranslationHelper
 
-  load_resource except: [:show, :manage, :index], instance_name: :product
+  # load_resource except: [:show, :manage, :index], instance_name: :product
```
vandrijevik: Let’s ✂️this instead of just commenting it 😄

app/controllers/products_common_controller.rb
```diff
@@ -52,15 +52,16 @@ def show
...
   def create
-    @product = current_facility_products.new(resource_params)
-    @product.initial_order_status_id = OrderStatus.default_order_status.id
-
-    if @product.save
-      flash[:notice] = "#{@product.class.name} was successfully created."
+    @product = current_facility_products.new(initial_order_status: OrderStatus.default_order_status)
+    @product_form = EditProductForm.for(@product)
+    @product_form.assign_attributes(resource_params)
+    if @product_form.save
```
vandrijevik: Maybe fold the `assign_attributes` and `save` into one `update(resource_params)` call for brevity?

app/forms/translatable_base_form.rb
vandrijevik: > Please let me know if I've gone too meta here.
😆 

I definitely am not a fan of this level of trickery and working around what SimpleForm and I18n expect in order to be able to trick them, as it’s usually a smell that there’s probably a better way to do this. Is there another way to organize this code that doesn’t require all this hiding and injecting ourselves in the method chain?

vandrijevik: As an aside, since this is the first time we’re doing this work, can we maybe not introduce a class hierarchy from the get-go? The inheritance and base class here seem a bit YAGNI to me.

What if it’s just `
class EditProductForm` and all the methods are there implemented only for the scope we currently have? Does that simplify things? ¢2

jhanggi:
Yeah, I agree it's probably a bit too far. It's all in my quest for the Perfect Form Object® that interfaces nicely with simple_form. One day I'll find something that works well, is easy to use, and doesn't break brains.

Client is cool with a different interface, so I'll close this.

PR #1912, opened 2019-03-25
[#148570] Remove N+1 queries on public and facility timeline pages

vandrijevik:
# Release Notes

Remove N+1 queries on public and facility timeline pages

# Additional Context

Should speed up the facility and public timelines, which per Skyline spend most of their time in the database. For facilities with a large number of products, this results in very slow requests, as we are currently executing 5 queries for each instrument that is shown on the timeline. Those queries look like this:

```
  OfflineReservation Load (0.3ms)  SELECT `reservations`.* FROM `reservations` WHERE `reservations`.`deleted_at` IS NULL AND `reservations`.`type` IN ('OfflineReservation') AND `reservations`.`product_id` = 14 AND `reservations`.`type` IN ('OfflineReservation') AND `reservations`.`reserve_end_at` IS NULL AND (reserve_start_at < '2019-03-25 17:14:32.245104')
  Reservation Load (0.4ms)  SELECT `reservations`.* FROM `reservations` INNER JOIN `order_details` ON `order_details`.`id` = `reservations`.`order_detail_id` INNER JOIN `orders` ON `orders`.`id` = `order_details`.`order_id` WHERE `reservations`.`deleted_at` IS NULL AND `reservations`.`product_id` = 14 AND `orders`.`state` = 'purchased' AND (reserve_end_at >= '2019-03-25 05:00:00') AND (reserve_start_at < '2019-03-26 04:59:59.999999') ORDER BY `reservations`.`reserve_start_at` ASC
  AdminReservation Load (0.2ms)  SELECT `reservations`.* FROM `reservations` WHERE `reservations`.`deleted_at` IS NULL AND `reservations`.`type` IN ('AdminReservation') AND `reservations`.`product_id` = 14 AND (reserve_end_at >= '2019-03-25 05:00:00') AND (reserve_start_at < '2019-03-26 04:59:59.999999')
  OfflineReservation Load (0.2ms)  SELECT `reservations`.* FROM `reservations` WHERE `reservations`.`deleted_at` IS NULL AND `reservations`.`type` IN ('OfflineReservation') AND `reservations`.`product_id` = 14 AND (reserve_end_at >= '2019-03-25 05:00:00') AND (reserve_start_at < '2019-03-26 04:59:59.999999')
  ScheduleRule Load (0.1ms)  SELECT `schedule_rules`.* FROM `schedule_rules` WHERE `schedule_rules`.`product_id` = 14
```

After this PR, we don’t execute any per-instrument queries, and we instead eager load all the data used in the instrument and reservation partials in the controller. To make this easier to grok, and to avoid conditionals in shared partials, I untangled the partials used in rendering, since they look differently in the different contexts.


app/views/facility_reservations/_instrument.html.haml
```diff
@@ -1,25 +1,19 @@
 .timeline_instrument[instrument]
   %h4
-    - if public_timeline?
-      = link_to instrument,
-        facility_instrument_path(current_facility, instrument)
-    - else
-      = link_to instrument,
-        facility_instrument_schedule_path(current_facility, instrument)
-
+    = link_to instrument, facility_instrument_schedule_path(current_facility, instrument)
```
meara: Are public/admin-only timelines not a thing anymore?

vandrijevik: They are! The facility timelines are `FacilityReservations#timeline`, and the public timelines are `Reservations#public_timeline`. This partial is `facility_reservations/_instrument.html.haml`, so the conditionals are gone because we know exactly which context we’re rendering from.

I find this to be much better than putting partials in `shared` and then using ivars like `@public_timeline = true` to set the context — that suggests that the partial is not really shared, since its content depends on the context where it is rendered, so it’s simpler to just split the content into unconditioned partials 😄 Does that make sense?

PR #1913, opened 2019-03-28
[#148570] Add missing indexes to make timeline pages faster

vandrijevik:
# Release Notes

[#148570] Add missing indexes to make timeline pages faster

# Additional Context

Some of the queries on the timeline pages (which we do fewer of after #1912, but we still do them) result in very expensive query plans due to the lack of suitable indexes that they can use, given the heavy filtering we do on these pages (based on types of reservations, their dates, whether the products are archived, etc.)

### DIFF
--- a/app/models/schedule.rb
+++ b/app/models/schedule.rb
@@ -17,7 +17,7 @@ class Schedule < ApplicationRecord
 
   validates_presence_of :facility
 
-  scope :active, -> { where(id: Product.not_archived.with_schedule.select(:schedule_id)).order(:name) }
+  scope :active, -> { where(id: Instrument.not_archived.with_schedule.select(:schedule_id)).order(:name) }
 
   def shared?
     products.count > 1
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190301022930) do
+ActiveRecord::Schema.define(version: 20190328133218) do
 
   create_table "account_facility_joins", force: :cascade, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8" do |t|
     t.integer  "facility_id", null: false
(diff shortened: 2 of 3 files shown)


app/models/schedule.rb
```diff
@@ -17,7 +17,7 @@ class Schedule < ApplicationRecord
 
   validates_presence_of :facility
 
-  scope :active, -> { where(id: Product.not_archived.with_schedule.select(:schedule_id)).order(:name) }
+  scope :active, -> { where(id: Instrument.not_archived.with_schedule.select(:schedule_id)).order(:name) }
```
meara: This is definitely supposed to be Instrument only? I think that makes sense with the schedule page, just checking though since it's been a minute since I did much in this code base.

db/schema.rb
```diff
@@ -817,6 +819,8 @@
 
   add_foreign_key "account_facility_joins", "accounts"
   add_foreign_key "account_facility_joins", "facilities"
+  add_foreign_key "account_users", "accounts", name: "fk_accounts"
```
meara: These don't seem to have been added by your migration? I'm fine with them being there, just wondering where they came from.

app/models/schedule.rb
vandrijevik: Absolutely — I also double checked, and there are 0 products that have a `schedule_id` and are not `Instruments` (what a complicated way to say the only products with schedules, are instruments)!

db/schema.rb
vandrijevik: 🤦‍♂️ You are right! `account_users` to `users came in https://github.com/tablexi/nucore-open/blob/4095962f73d481e67307f55f396cd6d287fb880a/db/migrate/20170508150135_add_foreign_keys.rb#L12-L13 and I believe that `fk_accounts` came as a result of https://github.com/tablexi/nucore-open/blob/4095962f73d481e67307f55f396cd6d287fb880a/db/migrate/20100528212342_add_account_users_table.rb#L7-L8 creating foreign keys by default in Rails, but the schema never got updated along the way.

vandrijevik: I will take them out of here, though, and fix `schema.rb` separately. Good catch!

PR #1947, opened 2019-05-08
[#148949] Fix handling of nested parameters for sanger samples

vandrijevik:
# Release Notes

Fix handling of nested parameters for sanger samples

# Additional Context

When saving a sanger sequencing submission form (which is also what the ACGT form targets), we had some very complicated indirection going on about how we handle the nested parameters for samples. This indirection, which created invalid samples (via `create_samples!`) and used AJAX polling to request data for pre-filling customer sample ids, caused problems when editing and saving some empty rows (because `clean_samples` would reject those rows instead of invalidating them, which is the desired behavior).

This PR eliminates all of that code and indirection, for standard handling of nested parameters which correctly handles creation and updates, with validations that apply in both cases (instead of us precreating invalid samples, and then only validating on update). We now create pre-filled values by using `Time.now.to_i` instead of relying on the database’s primary key sequence just to get an example 4-digit number as a string.


vandrijevik:
@rolfrussell I don’t think so — the ones I removed were for behavior that now comes standard courtesy of nested parameter handling, and for the use-database-ids-for-4-digit-numbers behavior that we don’t need.

PR #2000, opened 2019-06-11
[#142338] Add billable minutes to reservations and automatically set them

vandrijevik:
# Release Notes

Add billable minutes to reservations and automatically set them

# Additional Context

Adds the column and code for automatically calculating and setting it.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20190416205054) do
+ActiveRecord::Schema.define(version: 20190610112539) do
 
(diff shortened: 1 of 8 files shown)


db/schema.rb
```diff
@@ -820,6 +821,8 @@
 
   add_foreign_key "account_facility_joins", "accounts"
   add_foreign_key "account_facility_joins", "facilities"
+  add_foreign_key "account_users", "accounts", name: "fk_accounts"
```
rolfrussell: Are these two lines right?  I don't see a migration

vandrijevik: They alwas creep up because I think our schema.rb is not correct in all cases. I’ll take them out and address them separately, good catch!

PR #2012, opened 2019-06-19
[#148102] Allow choosing financial notes from dropdowns

vandrijevik:
# Release Notes

Allow choosing financial notes from dropdowns

# Additional Context

If dropdown options are configured, this PR allows selecting them from a dropdown for a price policy or when changing an order’s price. Otherwise, the current behavior of presenting these as text fields is preserved. This is a UI-only enhancement, and the note still gets saved in the existing database columns.


vandrijevik:
@rolfrussell Sure! Done in 8e2ae4b1a.

PR #2045, opened 2019-08-20
[#150534] Fix reference to LogEventSearcher, it is not in LogEvent

vandrijevik:
# Release Notes

Fix reference to LogEventSearcher, it is not in LogEvent

### DIFF
--- a/app/views/log_events/index.html.haml
+++ b/app/views/log_events/index.html.haml
@@ -20,7 +20,7 @@
   .row
     .span3
       = label_tag(text(:event_filter))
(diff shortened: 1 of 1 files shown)


vandrijevik:
@rolfrussell This is currently broken for NU and Dartmouth both, and I’ve verified this fix. Since it’s so trivial, and would help with another feature, I’m going to merge it. Please let me know if you’d rather that the code did something different!

rolfrussell:
@vandrijevik sounds good.

PR #2057, opened 2019-08-30
WIP statement history filtering

meara:
It is entirely possible that you will not want to use any of this at all, but in case this will save you time, it's here.

I abandoned this because I'm guessing we may want to go a different direction vs following the pattern established for transaction filtering, but this is here for posterity just in case.


jhanggi:
Closing in favor of #2067

PR #2058, opened 2019-08-30
[#141584] move transactions in review into nav

meara:
# Release Notes

Move a link to Transactions In Review into nav under My Payment Sources

### DIFF
--- a/app/presenters/nav_tab/link_collection.rb
+++ b/app/presenters/nav_tab/link_collection.rb
@@ -5,13 +5,14 @@ class NavTab::LinkCollection
   include Rails.application.routes.url_helpers
   include TranslationHelper
 
-  attr_reader :ability, :facility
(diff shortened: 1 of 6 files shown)


app/presenters/nav_tab/link_collection.rb
```diff
@@ -5,13 +5,14 @@ class NavTab::LinkCollection
   include Rails.application.routes.url_helpers
   include TranslationHelper
 
-  attr_reader :ability, :facility
+  attr_reader :ability, :facility, :user
 
   delegate :single_facility?, to: :facility
 
-  def initialize(facility, ability)
+  def initialize(facility, ability, user)
```
jhanggi: Would it be worth making these keyword args?

(I won't lie, it took me a few minutes to remember the term "keyword argument". This is what 6 months away will do 😜)

PR #2061, opened 2019-08-30
[#133036] do not validate custom subject/message presence for offline emails

meara:
# Release Notes

Do not validate presence of custom subject or message for Instrument Offline emails, as the default messaging is sufficient.


vendor/engines/bulk_email/app/models/bulk_email/delivery_form.rb
```diff
@@ -37,6 +39,20 @@ def deliver_all
...
+      return if custom_subject.present?
+
+      errors.add(:custom_subject, "can't be blank")
+    end
+
+    def custom_message_present_if_required
+      return if product&.offline?
+      return if custom_message.present?
+
+      errors.add(:custom_message, "can't be blank")
```
jhanggi: I think this can be `errors.add(:custom_message, :blank)` and it'll take the default I18n of "can't be blank".

jhanggi:
Failing test should be resolved by #2066

PR #2062, opened 2019-08-30
[#144741] add link for instruments from training requests

meara:
# Release Notes

Add a link to the instrument reservations page from the training request list if the product is in instrument.


app/views/training_requests/index.html.haml
```diff
@@ -23,7 +23,10 @@
         %tr
           %td= link_to user.to_s, facility_user_access_list_path(current_facility, user)
           %td= mail_to user.email
-          %td= product.name
+          - if product.is_a?(Instrument)
```
jhanggi: Would it be worth keeping this congruent with `admin/products/index.html`?

```haml
- if product.is_a?(Instrument)
  = link_to product.name, [current_facility, product, :schedule]
- else
  = link_to product.name, [:manage, current_facility, product]
```

meara: That's probably a good idea. Not what NU explicitly asked for, but might be what they're expecting.

PR #2067, opened 2019-09-04
[#149187] Add filtering to Statement History

jhanggi:
# Release Notes

Add account, "sent to", and date range filtering to the Statement History view.

# Additional Context

This also supports filtering by facility while in cross-facility mode (e.g. as a billing admin)


app/models/statement.rb
```diff
@@ -14,6 +14,28 @@ class Statement < ApplicationRecord
...
+
+  scope :created_between, lambda { |start_at, end_at|
+    if start_at
+      where(created_at: start_at..(end_at || DateTime::Infinity.new))
+    elsif end_at
+      where(arel_table[:created_at].lt(end_at))
+    end
+  }
+
+  RECONCILED_SQL = OrderDetail.unreconciled.where("order_details.statement_id = statements.id").select(1).to_sql
```
jhanggi: @vandrijevik Alternate queries:

```ruby
scope :unreconciled, -> { where(id: OrderDetail.unreconciled.where.not(statement_id: nil).select(:statement_id)) }
  scope :reconciled, -> { where.not(id: OrderDetail.unreconciled.where.not(statement_id: nil).select(:statement_id)) }
```
I'm not sure if there's a really good way to dry that up.

PR #2069, opened 2019-09-05
Add jhanggi back into codeowners

jhanggi:
# Release Notes

Tech Task: Add jhanggi back into CODEOWNERS

# Additional Context

This hopefully will work with on user per row. I think this should make future diffs cleaner.


vandrijevik:
Welcome back 😄

PR #2073, opened 2019-09-10
[#149187] Statement filtering feedback

jhanggi:
# Release Notes

Fix statement searching CSV export and adjust the form layout to be more consistent with transaction search.

# Additional Context

Based on client feedback to #2067 

> * Move the "Statement History" header to the space above left-hand navigation box
> * Move the "Status" drop down to the top right corner of the filtering selection area (to the right of the "Payment Sources" selection box)
> * Relabel the Date Range fields to "Start Date" and "End Date" to be consistent with other date filters

Rollbar error of the CSV export: https://rollbar.com/TableXI/nucore-nu/items/471/

```
NameError: undefined local variable or method `search_params' for #<FacilityStatementsController:0x000000000d0d2c00> (Most recent call first)
File /home/nucore/nucore-staging.northwestern.edu/releases/20190909125311/app/controllers/facility_statements_controller.rb line 32 in block (3 levels) in index
```

### DIFF
--- a/config/locales/forms/en.statement_search_form.yml
+++ b/config/locales/forms/en.statement_search_form.yml
@@ -2,5 +2,5 @@ en:
   activemodel:
     attributes:
       statement_search_form:
-        date_range_start: Created At Start
-        date_range_end: Created At End
+        date_range_start: Start Date
+        date_range_end: End Date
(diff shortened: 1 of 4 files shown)


config/locales/forms/en.statement_search_form.yml
```diff
@@ -2,5 +2,5 @@ en:
   activemodel:
     attributes:
       statement_search_form:
-        date_range_start: Created At Start
-        date_range_end: Created At End
+        date_range_start: Start Date
```
vandrijevik: This change broke the test, but other than that looks good!

PR #2089, opened 2019-10-14
[#148527] Add support for Form.IO order forms

vandrijevik:
# Release Notes

Add support for Form.IO order forms

# Additional Context

Form.IO’s forms have a prominently featured “Embed URL”. For each form, I wanted this to be the URL that the users can copy and enter into NUcore. However, the integration is done via JavaScript (instead of us redirecting to a different domain and then back to NUcore), and we’re serving up a NUcore page with JS that embeds the Form.IO form.

In order to make this all work, I’ve extended the existing classes to check whether the order form’s URL is a Form.IO one, and we do some special behavior. This is a short-term solution, because NU will be looking to replace Qualtrics and Surveyor forms with Form.IO-backed ones, at which point we can clean up all of this code and simplify it.

I’m still working on tests for the new controller, but I wanted to get this PR out because all of the behavior has been tested and accepted on staging from this branch, so this is ready for production.

### DIFF
--- a/app/controllers/formio/submissions_controller.rb
+++ b/app/controllers/formio/submissions_controller.rb
@@ -46,7 +46,7 @@ module Formio
         orderedForEmail: order_detail.user.email,
         orderedForName: order_detail.user.full_name,
         orderedForUsernname: order_detail.user.username,
-        paymentSourceAccountNumber: order_detail.account.account_number
+        paymentSourceAccountNumber: order_detail.account.account_number,
       }.merge(extra_prefill_data)
     end
 
--- a/app/controllers/formio/submissions_controller.rb
+++ b/app/controllers/formio/submissions_controller.rb
@@ -8,6 +8,7 @@ module Formio
 
     def new
       @formio_url = params[:formio_url]
+      @prefill_data = prefill_data
       @redirect_url = redirect_url
     end
 
@@ -33,6 +34,18 @@ module Formio
         success_url_params = Rack::Utils.parse_query(success_url.query)
         success_url.query = success_url_params.merge(referer: referer_url.to_s).to_query
       end.to_s
+    end
+
+    def prefill_data
+      order_detail = OrderDetail.find(params[:receiver_id])
+      {
+        accountOwnerEmail: order_detail.account.owner_user.email,
+        accountOwnerName: order_detail.account.owner_user.full_name,
+        nucoreOrderNumber: order_detail.order_number,
+        orderedAtDate: order_detail.created_at.to_date.to_s,
+        orderedForEmail: order_detail.user.email,
+        orderedForName: order_detail.user.email,
+      }
     end
 
   end
--- a/app/controllers/formio/submissions_controller.rb
+++ b/app/controllers/formio/submissions_controller.rb
@@ -35,8 +35,11 @@ module Formio
       end.to_s
     end
 
+    def order_detail
+      @order_detail ||= OrderDetail.find(params[:receiver_id])
+    end
+
     def prefill_data
-      order_detail = OrderDetail.find(params[:receiver_id])
       {
         accountOwnerEmail: order_detail.account.owner_user.email,
         accountOwnerName: order_detail.account.owner_user.full_name,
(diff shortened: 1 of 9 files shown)


app/controllers/formio/submissions_controller.rb
```diff
@@ -0,0 +1,63 @@
...
+        success_url_params = Rack::Utils.parse_query(success_url.query)
+        success_url.query = success_url_params.merge(referer: referer_url.to_s).to_query
+      end.to_s
+    end
+
+    def order_detail
+      @order_detail ||= OrderDetail.find(params[:receiver_id])
+    end
+
+    def prefill_data
```
danhodos: When you do decide to refactor, you might consider a decorator for `OrderDetail` that adds this conversion method.

app/models/external_service/external_service_receiver.rb
```diff
@@ -8,15 +8,29 @@ class ExternalServiceReceiver < ApplicationRecord
   validates_presence_of :external_service_id, :receiver_id, :response_data
 
   def show_url
-    parsed_response_data[:show_url]
+    url = parsed_response_data[:show_url]
+    if formio_submission?(url)
```
danhodos: Excited to read in your PR description that a lot of these conditionals will be going away. They're icky for sure, but if they're short-lived... :D

app/models/external_service/url_service.rb
```diff
@@ -23,6 +27,10 @@ def edit_url(receiver, request = nil)
 
   private
 
+  def formio_form?
+    URI(location).host.try(:ends_with?, "form.io")
```
danhodos: Is there a reason you use `include?` above and `ends_with?` here?

app/views/formio/submissions/edit.html.haml
```diff
@@ -0,0 +1,11 @@
+:javascript
+  window.onload = function() {
+    Formio
+      .createForm(document.getElementById('formio'), '#{@formio_url}')
+      .then(function(form) {
+        form.on('submitDone', function(submission) {
+          var surveyUrl = '#{@formio_url}';
```
danhodos: I notice that you use `encodeURIComponent` below and not here; just wanted to make sure that was intentional.

app/controllers/formio/submissions_controller.rb
vandrijevik: That’s a great idea! Maybe I’ll just add it as a method on `OrderDetail` called `#formio_prefill_data`? This is `nucore-open` and this code is always available…

app/models/external_service/url_service.rb
vandrijevik: Not a good one! I think I like the tighter `ends_with?` here, so I’ll change the above invocation. The URLs here are things like `https://nucore-development-12ea0e74.form.io/labitsupport`, so the host should always end in `form.io`.

app/views/formio/submissions/edit.html.haml
vandrijevik: Good call! In this case, things work because I’m not munging `@formio_url`, but I’ll change for consistency and to avoid this question in the future.

app/views/layouts/formio.html.haml
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+%html
+  %head
+    %meta{ "http-equiv" => "content-type", "content" => "text/html; charset=UTF-8" }
+    :plain
```
jhanggi: This mixing of haml and straight html is kind of nasty. Maybe in a future refactor we can either convert this to either full haml or this layout is basic enough I'd be fine with plain `html.erb`.

app/models/external_service/external_service_receiver.rb
jhanggi: Heads up that I'm pretty sure the sanger sequencing engine actually hooks into this. It probably wasn't worth the extra isolation to use this hook as opposed to something more direct, but that's what we did. So even if we get rid of surveyor/qualtrics/imserc at NU, dartmouth still relies on it.

PR #2095, opened 2019-10-29
Remove gem for now-unused Code Climate

vandrijevik:
# Release Notes

Remove gem for now-unused Code Climate


jhanggi:
I think there are a couple more things to do. I think we can remove this line `- run: bundle exec report_coverage` from the circle config. And then there's a chunk at the top of spec_helper about simplecov that can probably go away.

vandrijevik:
Yes, thank you @jhanggi! fde145c should take care of that.

PR #2098, opened 2019-10-31
Add health_check gem

jhanggi:
# Release Notes

Adds a health check endpoint at `/healthz`.

# Additional Context

We're adding this for UMass, but I believe we had been thinking of this for NU as well so their load balancers are hitting something a little more lightweight than the homepage.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -390,5 +390,6 @@ Nucore::Application.routes.draw do
     resource :submission, only: [:new, :show, :edit]
   end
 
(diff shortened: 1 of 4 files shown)


config/routes.rb
```diff
@@ -389,4 +389,6 @@
   namespace :formio do
     resource :submission, only: [:new, :show, :edit]
   end
+
+  health_check_routes
```
jhanggi:
```suggestion
  # See config/initializers/health_check.rb for more information
  health_check_routes
```

PR #2101, opened 2019-11-06
Add additional SAML logging

jhanggi:
# Release Notes

Add more logging to the SAML engine for debugging purposes

### DIFF
--- a/vendor/engines/saml_authentication/lib/saml_authentication/saml_attributes.rb
+++ b/vendor/engines/saml_authentication/lib/saml_authentication/saml_attributes.rb
@@ -16,6 +16,11 @@ module SamlAuthentication
       end.with_indifferent_access
     end
 
+    # Useful for debugging to see all the raw values we received
+    def to_raw_h
+      @saml_response.raw_response.attributes.to_h
+    end
+
   end
 
 end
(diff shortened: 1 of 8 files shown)


vendor/engines/saml_authentication/lib/saml_authentication/saml_attributes.rb
```diff
@@ -16,6 +16,11 @@ def to_h
       end.with_indifferent_access
     end
 
+    # Useful for debugging to see all the raw values we received
+    def to_raw_h
+      @saml_response.raw_response.attributes.to_h
+    end
```
vandrijevik: Hmm… I’m not a huge fan of adding this to the code here just for debugging purposes — can we do it https://github.com/tablexi/nucore-open/pull/2101/files#diff-76b4492a33c632e52ded6ee5707951deR13 where we construct the instrumentation payload, since we have `saml_response` there, and we wouldn’t need to go indirectly via this class?

jhanggi: Yeah, I hear you. The reason I put it here was because I originally created this class as a wrapper around the `DeviseSamlAuthenticatable::SamlResponse` and `MappedAttributes` API--which I consider to be kind of gross--to make it look more like a hash (i.e. the `to_h` method) and make the classes that use this easier to read.

When I wanted to see a different view of the the `@saml_response` for debugging, I knew it was going to be a somewhat complex traversing of the API. By putting it here, all of the knowledge about the internals of the SamlResponse api are encapsulated in one place.

I don't feel super strongly about it, but to me, keeping the grossness in one place seemed to outweigh my aversion to having code here just for debugging.

PR #2107, opened 2019-11-12
Move timezone configuration to settings

arnamak:
I'm assuming that by changing `application.rb` here, we only need to add the changes to the `settings.yml` file downstream in order for these changes to take effect. Let me know if I'm mistaken.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -149,3 +149,5 @@ paperclip:
 
 price_policy_note_options: ~
 order_detail_price_change_reason_options: ~
+
+time_zone: "Central Time (US & Canada)"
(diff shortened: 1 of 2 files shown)


vandrijevik:
This PR seems to include a lot of non-time-zone-related stuff, @arnamak — can you please merge `master` or rework the commits so it’s only that change?

config/settings.yml
```diff
@@ -146,3 +146,5 @@ paperclip:
 
 price_policy_note_options: ~
 order_detail_price_change_reason_options: ~
+
+time_zone: "Central Time (US & Canada)"
```
jhanggi: Not a big deal, but I would maybe move this higher up in the file. It won't get changed often, but it's moderately important to the behavior of the app.

PR #2116, opened 2019-11-19
[#147850] Move ordered_at field from Order to OrderDetail

jhanggi:
# Release Notes

Tech task: Move `Order#ordered_at` onto `OrderDetail`.

# Additional Context

This is in preparation for a feature where adding to an existing order will take the current time rather than the original cart's timestamp.

This will affect the BI queries. I completely removed the field from `orders` to verify I got all the references. I could be convinced that we should leave the original field temporarily for BI.

### DIFF
--- a/app/controllers/concerns/new_inprocess_controller.rb
+++ b/app/controllers/concerns/new_inprocess_controller.rb
@@ -29,10 +29,10 @@ module NewInprocessController
   def sort_lookup_hash
     {
       "order_number" => ["order_details.order_id", "order_details.id"],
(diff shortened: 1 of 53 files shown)


vandrijevik:
> This will affect the BI queries. I completely removed the field from orders to verify I got all the references. I could be convinced that we should leave the original field temporarily for BI.

I think this is a question for NU, and that unless they explicitly say otherwise, we have to leave things as they are for now so we don’t break their reporting as soon as we deploy and migrate.

jhanggi:
I'll have one more commit once I've updated the migration for Oracle. It's a weird beast when dealing with `UPDATE` combined with `JOIN`s.

PR #2119, opened 2019-11-27
[#146482] Allow users to fix problem reservations

jhanggi:
# Release Notes

Allow users to fix their own problem reservations when they forget to log out.

# Additional Context

This feature intentionally does not address problem occupancies.

# TODO

- [X] Capture when the user fixes their own times (new fields are not populated if an admin fixes it)
- [X] Make feature toggle-able per instrument
- [X] Send different emails based on whether the user can fix it or not
- [x] Add accessories

I'm going to continue to work on adding the accessories feature, but I figured I could go ahead and get some feedback.


app/controllers/problem_reservations_controller.rb
```diff
@@ -0,0 +1,57 @@
+class ProblemReservationsController < ApplicationController
```
jhanggi: I considered calling this `ProblemOrderDetailsController` and basing the `:id` on the order detail. That would allow more flexibility for if we ever get to resolving issues with occupancies. However, I ran into an issue because we have a shared module called `ProblemOrderDetailsController` that's shared between the order details and reservations and occupancies for doing the problem order tab.

I could be convinced this should be deeply nested under `/orders/:order_id/order_details/:order_detail_id/problem_reservations/:id` or something like that. Instead I just put it under `/problem_reservations/:id` because it was just simpler.

I do feel like this should be a different controller than the existing ones.

app/support/reservations/date_support.rb
```diff
@@ -172,25 +172,29 @@ def set_all_split_times
...
-    # take precedence
-    self.reserve_start_at = nil
-    self.reserve_end_at   = nil
     reserve_attrs = params.slice(:reserve_start_date, :reserve_start_hour, :reserve_start_min, :reserve_start_meridian,
                                  :duration_mins,
                                  :reserve_start_at, :reserve_end_at)
+    # need to be reset to nil so the individual pieces will
+    # take precedence, but only reset them if we're going to overwrite them
+
+    self.reserve_start_at = nil if reserve_attrs.keys.any? { |k| k.to_s.start_with?("reserve_start") }
```
jhanggi: I was not a huge fan of making changes to this part of the code, but it seemed the most expedient thing to do. What was happening was since we were only submitting the `actual_end_*` fields, the reservation timestamps as well as `actual_start_at` was being cleared out at first.

app/assets/stylesheets/app/no_modal.scss
```diff
@@ -0,0 +1,5 @@
+.not-in-modal {
```
jhanggi: This felt a little awkward, but it was easier to fix in CSS than in the structure of the partials.

app/controllers/accessories_controller.rb
```diff
@@ -8,12 +8,15 @@ class AccessoriesController < ApplicationController
   before_action :authorize_order_detail
   before_action :load_product
 
+  customer_tab :all
+  layout -> { modal? ? false : "application" }
```
jhanggi: What I did for this feature was to fix the accessories "modal" to work outside of a modal--which it mostly did except the JS and some styling. After fixing the reservation timestamps, we then redirect here.

I wanted to put the popup on the problem reservation edit page, but in order for the manual and auto scaled accessories to have the right defaults (matching the duration of the reservation) the reservation changes need to be persisted first.

I also tried some stuff with auto-popup of the modal on the receipt page, but that started to get hacky.

The main reason we've used popup modals in the past is to use the same feature, but to retain the current context. At the end of this flow, you're back on the my reservations page, which I think is fine.

spec/controllers/problem_reservations_controller_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      expect(response).to be_not_found
+    end
+  end
+
+  describe "as a facility director" do
+    before { sign_in create(:user, :facility_director, facility: facility) }
+
+    it "is a 404" do
+      get :edit, params: { id: problem_reservation }
+      expect(response).to be_not_found
```
jhanggi: The cancan ability is kind of messy was was really hard to restrict, so I went with scoping to the user when doing the `find` which will `404` rather than `403`.

jhanggi:
This is now ready for a full review

PR #2124, opened 2019-12-02
[#169636849] Accessibility tweaks

arnamak:
Adds several tweaks to make nucore slightly more accessible. To identify the issues we have within nucore I used a browser extention called Axe. To find the most suitable contrast ratios for our colors I used a contrast tool found here: https://webaim.org/resources/contrastchecker/.

Outside of the changes introduced here there were a number of additional warnings that I either was not able to fix, or did not find it necessary to fix.

1. "IDs of active elements must be unique" - I was unable to fix this due to how we adjust the navbar based on screen size. Within the code, we dynamically assign IDs depending on the link type, which means we generate the same ID for the same link within the dropdown nav and the static nav on larger screens. I tinkered with it for a while, but was unable to find a solution that didn't involve completely changing our existing logic. Ultimately, since one nav is hidden and the other is visible at any given time, I thought it was best to skip any potentially involved work here.

2. "Headings must not be empty" - Skipped due to the fact that it is not a violation, just a missed opportunity at a "Best Practice". We populate an h1 as a result of a block yeild, it's empty if the block does not return anything.

3. "Document must have one main landmark" - Also skipped due to not being a violation.

Other notes:
- If running tests locally, whatever tool you use may throw errors related to color contrast. Many of these errors are due to a tool's inability to determine the color contrast rather than an actual violation. I manually inspected each potential violation and adjusted the colors accordingly. A tool may not be able to discern contrast due to usage of gradiants, or overlapping divs.

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -140,32 +140,6 @@ li.nav-spacer {
 }
 
 /*
-  Product List
-*/
-
-.product_list {
-  ul {
-    @extend .unstyled;
-  }
-  input {
-    width: 2em;
-    padding: 0 3px;
-    margin-top: 5px;
-    margin-bottom: 5px;
-  }
-  a.fa-calendar {
-    text-decoration: none;
-    color: $textColor;
-    &.available {
-      color: $successText;
-    }
-    &.in-use {
-      color: $errorText;
-    }
-  }
-}
-
-/*
   Product View
 */
 
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -139,10 +139,6 @@ li.nav-spacer {
   margin-bottom: 0;
 }
 
-.alert {
-  color: #24506B; // minimum lightness for WCAG AAA compliance (accessibility).
-}
-
 /*
   Product List
 */
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -85,7 +85,7 @@ li.nav-spacer {
 
 .footer__copyright {
   font-size: 85%;
-  color: #919191;
+  color: #595959; // maximum lightness for WCAG AAA compliance (accessibility).
   text-align: right;
   float: right;
 }
@@ -137,6 +137,10 @@ li.nav-spacer {
 
 .alert p:last-child {
   margin-bottom: 0;
+}
+
+.alert {
+  color: #24506B; // minimum lightness for WCAG AAA compliance (accessibility).
 }
 
 /*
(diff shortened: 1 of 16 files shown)


app/views/shared/_header.html.haml
```diff
@@ -40,7 +40,8 @@
               = render "shared/message_summary"
               %li= link_to t('pages.logout'), sign_out_user_path
               - if responsive? && !acting_as? && current_user
-                = form_tag global_search_path, class: "navbar-search pull-right hidden-with-nav" do
-                  = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2"
+                %li
+                  = form_tag global_search_path, class: "navbar-search pull-right hidden-with-nav" do
+                    = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2", "aria-label": "Search"
```
jhanggi: Can we use the I18n value for the `aria-label`?

app/views/shared/nav/_nav_links.html.haml
```diff
@@ -23,7 +23,7 @@
 
 -# .visible-with-nav is visible > 979px
 = form_tag global_search_path, class: "navbar-search pull-right visible-with-nav" do
-  = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2"
+  = text_field_tag :search, nil, placeholder: t("global_search.search"), class: "search-query span2", "aria-label": "navSearch"
```
jhanggi: Should the label match the `placeholder`, or is this a reference to another element?

app/assets/stylesheets/nucore.scss
```diff
@@ -139,6 +139,10 @@ li.nav-spacer {
   margin-bottom: 0;
 }
 
+.alert {
+  color: #24506B; // minimum lightness for WCAG AAA compliance (accessibility).
```
jhanggi: I think we may want a slightly different color between `alert-info` and `alert-danger`. We might even want to think of only using the alerts for flashes. I think we use it a couple other ways including the welcome text on the homepage.

app/views/facilities/_facilities_list.html.haml
```diff
@@ -1,9 +1,9 @@
 - if @facilities.any?
-  %h4.facility_listing.js--facility_listing.all_header= text(".all")
+  %h2.facility_listing.js--facility_listing.all_header= text(".all")
   - @azlist.map.with_index do |(letter, matched_facilities), index|
     - matched_facilities.each do |facility|
       .facility_listing.js--facility_listing{ class: az_classname_for_facility(index, letter) }
-        %h4.header-tight= link_to facility, facility_path(facility)
+        %h3.header-tight= link_to facility, facility_path(facility)
```
jhanggi: I'm not sure if it's just because it's different from what it was before, but the size feels a bit too much if this is an `h3`. Perhaps a header is not the right element for this element.

jhanggi:
Did you happen to see the contract violation of the normal link color? 

```
Element has insufficient color contrast of 3.89 (foreground color: #0088cc, background color: #ffffff, font size: 11.3pt (15px), font weight: bold). Expected contrast ratio of 4.5:1
```

PR #2133, opened 2019-12-16
Move research safety engine into main app

jhanggi:
# Release Notes

Move the research safety engine code into the main app.

# Additional Context

Follow up to #2131 

Next step is a feature flag


vandrijevik:
> Next step is a feature flag

@jhanggi A thought I just had: does this need to be behind a feature flag? I know that Dartmouth doesn’t use it, but what would the harm be if the code just existed all the time — it wouldn’t prevent anything from working, right?

jhanggi:
> what would the harm be if the code just existed all the time — it wouldn’t prevent anything from working, right?

@vandrijevik I think you have a fair point. As long as no certificates have been added to a product, everyone should be able to buy everything as normal. And we will be adding this feature to Dartmouth, likely in early 2020. 

My only concern is confusing users in the meantime with useless links/options.

I'm thinking of not going too crazy with the feature flag: just show/hiding the tabs. Disabling the routes and controller seems a little overkill for this. And then possibly a short-circuit in `certified?` so it always returns true if the feature is off.

Another option would be to leave the global settings tab on and in the user/products, check if there are any `Certificate`s configured by the admins. If there aren't, hide the user/product tabs.

What do you think?

vandrijevik:
> Another option would be to leave the global settings tab on and in the user/products, check if there are any Certificates configured by the admins. If there aren't, hide the user/product tabs.

I like this idea a lot! We wouldn’t be adding a feature flag that we already have a lot of, global admins will always be able to see that this is a supported feature, and the majority of users will only see it if/when applicable to them.

My vote is to go down this route.

PR #2144, opened 2020-01-03
Upgrade to Rails 5.1

jhanggi:
# Release Notes

Upgrade to Rails 5.1.

# Additional Context

I did an upgarde to 5.1 and pulled a bunch of the changes from Vlad's 5.2 branch. This was super helpful as 5.1 gives helpful deprecation warnings for some of the Dirty API changes. I also left the making `belongs_to` `required` by default off since that's what the app was largely built around. I like the idea of changing to the newer default, but it feels risky and I think many of the relationships should be thought about whether they really should be required or not.

I will likely continue off of this into 5.2. I think once we're there, we can think about changing some of the configuration.


jhanggi:
I'll probably end up rebasing this once we get some of the other research safety PRs merged in, especially as this should be it's own release.

PR #2152, opened 2020-01-08
Upgrade I18n to 1.7.1

jhanggi:
# Release Notes

Tech task: Bump I18n gem to 1.7.1.

# Additional Context

This change in I18n https://github.com/ruby-i18n/i18n/pull/503 breaks our lookup for one/many when there is also a nested hash.

`Account.model_name.human` was returning the hash `{:one=>"Payment Source", :other=>"Payment Sources", :owner=>"Owner", :statuses=>{:active=>"Active", :suspended=>"Suspended"}}`.

Cleaning this up seemed simpler than trying to dig in to I18n.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -331,7 +331,7 @@ GEM
       sysexits (~> 1.1)
     health_check (3.0.0)
       railties (>= 5.0)
-    i18n (1.7.0)
+    i18n (1.7.1)
       concurrent-ruby (~> 1.0)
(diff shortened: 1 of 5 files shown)


jhanggi:
Looks like this issue was found and possibly fixed upstream. https://github.com/ruby-i18n/i18n/issues/510

I still think the cleanup is good; I'm not sure those locales should be nested under `activerecord` anyways.

PR #2155, opened 2020-01-11
[#139340] Add ability to export account search results to CSV

jhanggi:
# Release Notes

Add ability to export account search results to CSV.

# Additional Context

I chose not to send this asynchronously via email as I suspect this won't be slow enough to have a problem. I've asked the client. We can always switch it over later.

I extracted the `AccountSearcher` to help clean up the controller, but with an eye towards Dartmouth's upcoming feature of being able to search by more fields for accounts. I chose not to do any further refactoring there in favor of doing the refactoring when we get to that story.


jhanggi:
I'll address the test failure on Monday.

app/services/account_searcher.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  include SearchHelper
+
+  def initialize(original_scope, query)
+    @original_scope = original_scope
+    @query = query
+  end
+
+  def valid?
+    @query.to_s.length >= 3
```
arnamak: Could we extract the magic number to a constant?

app/services/account_searcher.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  def results
+    owner_where_clause = <<-end_of_where
```
arnamak: What do you think about defining `owner_where_clause` as its own method?

app/services/account_searcher.rb
```diff
@@ -0,0 +1,46 @@
...
+        OR LOWER(CONCAT(users.first_name, users.last_name)) LIKE :term
+      )
+      AND account_users.user_role = :acceptable_role
+      AND account_users.deleted_at IS NULL
+    end_of_where
+
+    term = generate_multipart_like_search_term(@query)
+
+    # retrieve accounts matched on user for this facility
+    results = @original_scope.joins(account_users: :user).where(
```
arnamak: Similar to my comment above, I'd consider extracting this query logic and the query logic below into their own methods and just using the `results` method to deliver the results.

So you wind up with something like:
```
def results
    results = accounts_matched_on_user_query(owner_where_clause) += accounts_matched_on_number_query

    results.uniq
end
```

jhanggi: I like these ideas.For this particular PR I opted to just extract the class and not really change the underlying code too much.  I've started doing refactoring on this class as I've begun work on Dartmouth's story around expanding which fields we're searching. I'm planning on extracting methods so the `results` method is something simple and readable like `owner_matches.or(account_number_matches).or(description_matches)` (this may not be exactly what it ends up as, but it's in the vein).

jhanggi:
@arnamak I'd like to address the additional refactors in the PR for Dartmouth's search expansion feature.

PR #2160, opened 2020-01-13
[#170607701] Add ability to customize IP Port

arnamak:
# Release Notes

Full disclosure, I forgot if this was UMass-specific or if it was meant to go in open. 

Renames the existing `port` column on `Relay` to `outlet`
Adds `port` column on `Relay` (referring to IP Port)

I may advocate for renaming the `port` column as `ip_port` to avoid confusion for existing users, though on the form it reads as "Relay Port" now so it might not be that bad.


app/models/power_relay.rb
```diff
@@ -1,15 +1,18 @@
...
   extend ActiveSupport::Concern
 
+  MAXIMUM_NUMBER_OF_OUTLETS = 1..16.freeze
+
   included do
     ## validations
-    validates_presence_of :ip, :port, :username, :password
+    validates_presence_of :ip, :outlet, :username, :password
     validates :auto_logout_minutes, presence: { if: :auto_logout }
+    validates :outlet, inclusion: MAXIMUM_NUMBER_OF_OUTLETS
```
jhanggi: What would you think of using a `numericality` validation? Since the db field is an integer, that feels like it might be more consistent.

And this is kind of a personal preference, but I lean towards putting all the validations on a field together.

```
validates :outlet, presence: true, numericality: { only_integer: true, greater_than: 0, less_than_or_equal_to: 16 }
```

app/models/power_relay.rb
```diff
@@ -1,15 +1,18 @@
...
+    validates :port, numericality: { only_integer: true, allow_nil: true }
```
jhanggi: Can we add a `greater_than: 0` in here?

spec/factories/relays.rb
```diff
@@ -4,7 +4,7 @@
   factory :relay do
     type { "RelaySynaccessRevA" }
     ip { "192.168.1.1" }
-    sequence(:port) { |p| p }
+    sequence(:outlet) { |o| o }
```
jhanggi: We didn't have a max value validation on this field before, so we didn't see any problems, but now if we create more than 16 of these from this factory, they'll start becoming invalid (I think that's what the majority of current test failures are). Maybe we could add a modulo in there: `sequence(:outlet) { |o| o % PowerRelay::MAXIMIUM_OUTLETS }`

spec/models/power_relay_spec.rb
```diff
@@ -11,11 +11,26 @@ class SomeRelay < Relay
 RSpec.describe SomeRelay do
 
   it { is_expected.to validate_presence_of :ip }
-  it { is_expected.to validate_presence_of :port }
+  it { is_expected.to validate_presence_of :outlet }
   it { is_expected.to validate_presence_of :username }
   it { is_expected.to validate_presence_of :password }
   it { is_expected.not_to validate_presence_of :auto_logout_minutes }
 
+  it "should allow a range of 1-16 outlets" do
```
jhanggi: We still have a bunch of old specs following this pattern, but just as we're switching from `should` to `expect` rspec syntax, we're trying to move away from "should" in our descriptions. So something like "it 'allows a range of 1-16'"

spec/models/power_relay_spec.rb
```diff
@@ -11,11 +11,26 @@ class SomeRelay < Relay
...
+    relay = SomeRelay.new(ip: "123", username: "nucore", password: "password")
+
+    expect(relay.update(outlet: 16)).to be true
```
jhanggi: This gets the job done, but I don't feel like it's very descriptive of what it's checking. Also, the `false` case could easily lead to a false positive where the object is invalid, but for the wrong reason.

It's a little more verbose, but I would consider one of these options. And since you're not hitting the database in the setup, I would lean towards one expectation per example.

```ruby
relay.outlet = 16
expect(relay).to be_valid

relay.outlet = 17
expect(relay).to be_invalid
# A couple options
expect(relay.errors).to include(:outlet)
expect(relay.errors[:outlet]).to include("is not included in the list") # I don't love this one because it relies on matching I18n
expect(relay.errors[:outlet]).to include(/included/)

# I like the clarity of this one, but can get annoying in that you need to include any interpolation arguments if the error includes them
expect(relay.errors).to be_added(:outlet, :inclusion)

File: app/models/power_relay.rb (hunk 1 above)
Comment (arnamak) [reply app/models/power_relay.rb]: sure, works for me

File: spec/models/power_relay_spec.rb (hunk 5 above)
Comment (arnamak) [reply spec/models/power_relay_spec.rb]: sure, I might argue that this route it becomes slightly more tedious should the maximum ever change in the future. but I have to admit the amount of clarity is rather nice and not something I'm used to seeing.

File: app/views/instruments/_instrument_fields.html.haml [hunk 6]
```diff
@@ -28,7 +28,8 @@
   = f.simple_fields_for :relay, @product.relay || @product.build_relay do |rf|
     = rf.input :type, collection: options_for_relay, include_blank: false
     = rf.input :ip
-    = rf.input :port
+    = rf.input :ip_port
```
jhanggi: Just a thought: do you think it would be worth adding a hint for this. Something like "leave blank unless your device is on a non-standard port"?

And maybe for `outlet`, "e.g. 1 or 2" (or something similar if you have a better suggestion).

I'm not committed to this, but thinking it may be a good idea

arnamak: yeah, I'm good with that.

PR #2164, opened 2020-01-15
[#139339] Add ability to bulk import timed services

jhanggi:
# Release Notes

Add the ability to bulk import Timed Services. 

# Additional Context

Note that the quantity field should be whole minutes rather than "H:MM" format in order to avoid issues with Excel exporting. There are already enough issues with date formats in the import. This restriction has been approved by NU.


app/support/order_row_importer.rb
```diff
@@ -100,7 +100,11 @@ def account_number
   def add_product_to_order
     ActiveRecord::Base.transaction do
       begin
-        @order_details = order.add(product, quantity, note: note)
+        if product.quantity_as_time?
+          @order_details = order.add(product, 1, duration: quantity, note: note)
```
arnamak: what’s the significance of this `1`?

jhanggi: In most cases, when you add `N` TimedServices to an order, it'll generate `N` order details. 

The ways you can do this are 
1. the facility order form (see https://nucore-staging.northwestern.edu/facilities/test-facility) where you can choose how many of each thing you want.
2. Add to an existing order (go to an order as an administrator and you'll see a form at the bottom). When doing this for a timed service, we add a new field that lets you enter the duration as `H:MM`.

So this is kind of a special case where we want to only add **one** line item, but use the "Quantity" field as the duration. It's kind of a quirk of the `Order#add` api.

In my mind there's some usability issues with it: I don't remember exactly how it works, but if I add 2 of a `Service` through the order form, I get two line items with quantity of 1, but if I add two of an `Item` I get one line item with a quantity of 2. But changing this behavior is a whole other can of worms that it more effort than I think it's worth.

jhanggi: Does this make sense, or do you want to discuss further?

arnamak: It does make sense, I guess I was wondering if we could clarify that a bit in the code but it seems like it's just quirky enough to the point where it isn't worth it to do so.

jhanggi: I just added a comment. Hopefully that'll be enough to cover why it's an idiosyncrasy.

PR #2167, opened 2020-01-16
[#151885] Add ability to add to an existing order in bulk import

jhanggi:
# Release Notes

Allow users to add line items to an existing order during bulk import. There is a new column in the import; fill this in with the order number (first part of `#XXX-YYY`). Note that the order must be part of the same facility, must be purchased, and must match the original user.

# Additional Context

I did a good amount of rearrangement of methods trying to group related things together. Let me know if you have any questions about what to focus on.


app/support/order_row_importer.rb
```diff
@@ -110,23 +118,33 @@ def add_product_to_order
...
-    @fulfillment_date ||= parse_usa_import_date(fulfillment_date_field)
+    @fulfillment_date ||= parse_usa_import_date(field(:fulfillment_date))
   end
 
-  def fulfillment_date_field
-    @fulfillment_date_field ||= @row[HEADERS[:fulfillment_date]].try(:strip)
+  def existing_order
+    return @existing_order if defined?(@existing_order)
+
+    @existing_order = Order.find_by(id: field(:order_number))
```
arnamak: was it intentional to use `find_by`here instead of `find`? do we not care to trigger an exception here if the record isn’t found?

jhanggi: The value we're searching for comes straight from user input, so it could easily be not found. In that case, I do want to return `nil` rather than raise an error so I can check for that case (see the first check of `validate_existing_order`) and render an error to present to the user.

What happens when you're bulk uploading is if there are errors we spit out another CSV that should be identical to the uploaded one but with an "Errors" column. This allows the user to fix anything in that file and then upload the modified file.

PR #2173, opened 2020-01-20
Add version numbers to ActiveRecord Migrations

arnamak:
# Release Notes

As of [Rails 5.1](https://guides.rubyonrails.org/5_1_release_notes.html) (section 9.3), an [error](https://github.com/rails/rails/commit/249f71a22ab21c03915da5606a063d321f04d4d3) is raised when inheriting from `ActiveRecord::Migration` directly, a version number must be specified.


jhanggi:
Just a note, but we'll probably need to update the individual schools as well since they have some custom migrations in their engines.

PR #2177, opened 2020-01-20
Add parallel_tests gem

arnamak:
# Release Notes

Adds the [parallel_tests](https://github.com/grosser/parallel_tests) gem to speed up running tests locally. Usage is optional, but it can reduce the amount of time your test suite takes to run dramatically. Personally, `rspec spec` will take about 21 minutes to finish; `rake parallel:spec` reduces that to 4 minutes.

Caveat: if you're expecting a lot of errors, the output can be a little annoying to parse through. These sorts of things can be tweaked with additions to an optional `.rspec_parallel` file.

See the updated README for instructions on how to run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -165,6 +165,11 @@ NUcore uses [Rspec](http://rspec.info) to run tests. Try any of the following fr
 * Create additional databases:
     rake parallel:create
 
+* Run migrations (only needed if building from scratch):
+    rake parallel:create
+  OR
+    rake parallel:load_schema
+
 * Copy development schema (repeat after migrations):
     rake parallel:prepare
 
--- a/README.md
+++ b/README.md
@@ -162,6 +162,8 @@ NUcore uses [Rspec](http://rspec.info) to run tests. Try any of the following fr
 
 #### Parallel Tests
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -160,6 +160,28 @@ NUcore uses [Rspec](http://rspec.info) to run tests. Try any of the following fr
 * To run just the controller tests
     rake spec:controllers
 
+#### Parallel Tests
```
jhanggi: Can we add a link to the parallel_tests readme?

README.md
```diff
@@ -160,6 +160,28 @@ NUcore uses [Rspec](http://rspec.info) to run tests. Try any of the following fr
...
+
+* Create additional databases:
+    rake parallel:create
```
jhanggi: Can we add the `<%= ENV['TEST_ENV_NUMBER'] %>` to the `database.yml` template?

arnamak:
@jhanggi I did not. It seems that if you have an `.rspec_parallel` file in your directory that you won't need to remove those options from `spec/spec.opts`. I didn't commit it since I didn't configure any options but apparently that's enough to have it not fall back to `spec.opts`

PR #2184, opened 2020-01-22
Clean up formatting in readme

jhanggi:
# Release Notes

Clean up formatting of commands in README.

# Additional Context

The commands weren't being formatted as code when parsed by Github's markdown parser. I noticed this because of some merge conflicts when bringing this downstream (we have some other school-specific info in the same spot of the file--easy enough to keep both).


jhanggi:
@arnamak Thanks for catching those. I didn't even notice they were misformatted. Updated.

PR #2193, opened 2020-02-03
[#149740] Add reference_id to OrderDetail

arnamak:
# Release Notes

Adds a field to `OrderDetail` and updates relevant forms.

This still needs a little work. Some of the context lost on me has likely left holes in areas that require updating.


arnamak:
👍 thanks for the changes @jhanggi; looks good

PR #2194, opened 2020-02-03
[#144526] Add link to download results as zip file

arnamak:
# Release Notes

Adds a link to "Download all as zip" from the file selector modal.


jhanggi:
Looks pretty good. I did a little extra I18n-ing and linting, but this will be ready to go as soon as I see a green build.

PR #2195, opened 2020-02-03
[#136393] Clone account memberships

arnamak:
# Release Notes

Adds an option to clone account memberships from one user to another.

# Screenshots

Screen Shot 2020-02-21 at 11 00 39 AM

Screen Shot 2020-02-21 at 11 01 02 AM


app/assets/javascripts/app/clone_account_memberships.coffee
```diff
@@ -0,0 +1,17 @@
+$(document).ready ->
+
+  # Select all for checkboxes
+  selectAllSelector = $("a#selectAllClone")
```
jhanggi: I think we have some common logic that we can use for this. I'll have to find it.

app/controllers/clone_account_memberships_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    cloner = account_membership_cloner
+
+    if cloner.perform
+      flash[:notice] = t(:success, scope: "forms.clone_account_membership")
+
+      redirect_to facility_user_accounts_path
+    else
+      flash[:error] = t(
+        :error,
+        scope: "forms.clone_account_membership",
```
jhanggi: We can probably consider using the `text` helper which gives us some some implicit scoping. Not a big deal, though.

app/controllers/clone_account_memberships_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def account_users
+    @account_users = AccountUser.where(user: user_to_clone).where.not(account: init_user.accounts)
+  end
+
+  def clone_account_membership_params
+    params.permit(:user_to_clone_id, :account_user_ids =>[])
+  end
+
+  def account_membership_cloner
```
jhanggi: I could go either way, but I'm not sure this method warrants extraction. I'm thinking it could just be left inline in the main action.

app/services/account_membership_cloner.rb
```diff
@@ -0,0 +1,43 @@
+class AccountMembershipCloner
+
+  PROTECTED_USER_ROLE = "Owner".freeze
```
jhanggi: We may want to have this reference other constants, i.e. `PROTECTED_USER_ROLE = AccountUser::ACCOUNT_OWNER`. If that causes circular loading or something, we could use the `AccountUser` constants and not worry about having them here. I don't feel super strongly about it, though.

app/views/clone_account_memberships/index.html.haml
```diff
@@ -0,0 +1,18 @@
+= content_for :tabnav do
+  = render partial: "admin/shared/tabnav_users", locals: {secondary_tab: "accounts"}
+= content_for :sidebar do
+  = render :partial => 'admin/shared/sidenav_users', locals: { sidenav_tab: 'users' }
```
jhanggi: This is probably copy/paste, but we should try to maintain stylistic consistency within the same files, e.g. using the newer hash syntax and double quotes. `partial: "admin/shared/sidenav_users"`

app/views/clone_account_memberships/index.html.haml
```diff
@@ -0,0 +1,18 @@
...
+
+= content_for :h1 do
+  = current_facility
+
+%header
+  %h2= "Clone Account Memberships to #{@user.full_name}"
```
jhanggi: Can we I18n this?

app/views/clone_account_memberships/index.html.haml
```diff
@@ -0,0 +1,18 @@
...
+
+= form_tag url_for(action: :search), id: "ajax_form", method: :get do
+  = label_tag :search_term, "Search by name or email" # t(".label.search")
```
jhanggi: This one should definitely be I18n-ed because this is a shared form and I believe we override it in NU and other schools to reference netids.

app/views/clone_account_memberships/new.html.haml
```diff
@@ -0,0 +1,28 @@
+= content_for :tabnav do
+  = render partial: "admin/shared/tabnav_users", locals: {secondary_tab: "accounts"}
+= content_for :sidebar do
+  = render :partial => 'admin/shared/sidenav_users', locals: { sidenav_tab: 'users' }
+
+= content_for :h1 do
+  = current_facility
+
+= form_with url: facility_user_clone_account_memberships_path(user_to_clone_id: @user_to_clone), html: { id: "cloneAccountForm" } do |f|
```
jhanggi: There's a lot of legacy code so it's easy to take advantage of an older style. We've been trying more and more to modernize and one of those is using `js--` as our class names for JS-enhanced functionality as opposed to IDs.

app/views/clone_account_memberships/search.html.haml
```diff
@@ -0,0 +1,16 @@
+%table.table.table-striped.table-hover
+  %thead
+    %tr
+      %th
+      %th Username
```
jhanggi: We should I18n this.

app/views/clone_account_memberships/new.html.haml
```diff
@@ -0,0 +1,28 @@
...
+  - if @account_users.any?
+    %h2= t(:header, scope: "forms.clone_account_membership", first_name: @user_to_clone.first_name, last_name: @user_to_clone.last_name)
+    %table.table.table-striped.table-hover
+      %thead
+        %tr
+          %th= link_to "Select All", 'javascript:;', id: "selectAllClone"
+          %th Account
```
jhanggi: I think we should probably include the role in this table.

app/controllers/clone_account_memberships_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+    AccountMembershipCloner.new(
+      account_users_to_clone: account_users_to_clone,
+      clone_to_user: init_user
```
jhanggi: I think we need to pass in the `created_by` to be `current_user`.

app/services/account_membership_cloner.rb
```diff
@@ -0,0 +1,43 @@
...
+  def perform!
+    ActiveRecord::Base.transaction do
+      @account_users_to_clone.each do |au|
+        create_account_user(au.attributes)
+      end
+    end
+  end
+
+  def create_account_user(attributes)
+    AccountUser.create!(
```
jhanggi: We need to create a `LogEvent` when we do this as well.

jhanggi: I'm unsure if we need to send an email to the account owners. I feel like we will want to do that, but we'll need to think about it and if there's a way to roll up the emails.

arnamak:
@jhanggi thanks for leaving comments describing the changes.

I noticed one (sort of two) small bug(s).

1. It seems we'll display `AccountUser` objects now even if a user already belongs to that payment source. I'm fine with that but it's a little weird that you can submit the form and receive a successful response with no items selected in this case.

2. Semi-related to that, even if all the checkbox fields are disabled a user is still able to "select all". It doesn't actually collect the params and send them along with the response, but the user is still able to click the clone button and the text for the select all mechanism changes from "select all" to "select none".

I propose we disable the "clone" button if no checkboxes are selected.

Outside of that :+1:

app/views/clone_account_memberships/new.html.haml
```diff
@@ -0,0 +1,37 @@
...
+= form_with url: facility_user_clone_account_memberships_path(user_to_clone_id: @clone_from_user) do |f|
+  - if @account_users.present?
+    %h2= text("header", name: @clone_from_user.full_name)
+    %p= text("instructions")
+    %table.table.table-striped.table-hover
+      %thead
+        %tr
+          %th= select_all_link
+          %th= Account.model_name.human
+          %th
```
arnamak: @jhanggi was `role` intentionally left blank?

jhanggi:
@arnamak I've updated the JS so that it enables/disables the submit button based on if anything is checked. I started to try to rewrite it in vanilla javascript, but decided it was going to take me too long so I wasn't sure it was worth it.

Rather than hiding the accounts the user is already a part of, I'm partial to showing the account, but disabling the checkbox. I chose disabling the checkbox rather than hiding it so I could put a `title` tooltip on it. The other thing I considered was a second table for those accounts.

arnamak:
Looks good @jhanggi! Just give it an an approval and feel free to merge it yourself (I realize it’s nitpicky but I prefer not merging an unapproved PR if I can help it).

PR #2205, opened 2020-02-10
[#148698] Facility instruments dashboard

jhanggi:
# Release Notes

Add a dashboard to see who is currently using which instruments in a given facility. This dashboard is available for the facility staff and also includes an optional public dashboard which can be used to display on a screen similar to the Occupancies dashboard.

# Screenshot

Screen Shot 2020-02-10 at 3 03 37 PM


app/controllers/instruments_dashboard_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  layout "plain", only: :public_dashboard
+
+  admin_tab :all
+  before_action :authenticate_user!, except: :public_dashboard
+  before_action :authenticate_token, only: :public_dashboard
+  before_action :init_reservations
+  # This sets "Reservations" as the active tab
+  before_action { @active_tab = "admin_reservations" }
+
+  def dashboard
```
jhanggi: I originally went with `index` here but it was causing problems with the active tab logic (and I didn't really feel like refactoring that was worth it in this story).

app/controllers/instruments_dashboard_controller.rb
```diff
@@ -0,0 +1,35 @@
+class InstrumentsDashboardController < ApplicationController
```
jhanggi: I could be convinced that `ReservationsDashboardController` or something like that is a better name. I originally used this name because I was looping over the all the instruments and finding the active one, but since they want to only show active reservations, it became easier to loop over reservations. I'd also be open to other names.

app/models/reservation.rb
```diff
@@ -89,6 +89,15 @@ def self.not_this_reservation(reservation)
 
   scope :ongoing, -> { not_ended.where("actual_start_at <= ?", Time.current) }
 
+  scope :current_in_use, lambda {
```
jhanggi: There's already an `active` scope, but that includes everything that has been purchased but restricts it to only user reservations (i.e. it excludes admin and offline reservations). I wouldn't mind a better name if someone has an idea.

app/controllers/instruments_dashboard_controller.rb
arnamak: I have no real opinion here, personally I think `InstrumentsDashboardController` makes sense.

PR #2210, opened 2020-02-13
[#140079] Improve product search for Oracle

jhanggi:
Adds on to #2209 to support Oracle

This is blocked on PRs into the oracle adapter. Right now, the schema dumper fails if there is a fulltext index and a regular index. It also does not dump the `sync: "ON COMMIT"`, which we would need in order to ensure the index is being updated properly.


jhanggi:
PRs https://github.com/rsim/oracle-enhanced/pull/1988 and https://github.com/rsim/oracle-enhanced/pull/1986 have been merged into `master`. Unsure of the status of backporting.

PR #2211, opened 2020-02-13
[#137160] Reorganize Billing Sidebar and remove AR module

jhanggi:
# Release Notes

Reorganize the billing sidebar into a more linear progression of steps. Remove the accounts receivable module in favor of improved functionality in "Statement History".

# Additional Context

From NU:

> There are multiple issues with the AR tab, and now that there's all this great functionality in "Statement History" it really doesn't serve a useful purpose. 
>
> The Show/Hide link in AR never seemed to work, and AR for internal chart strings isn't super helpful if the cores are billing regularly in any case. Statement History is a much better AR module.

This also removes the `manage_payment_sources_with_users` setting. Way back when, the "Payment Sources" tab was in the billing tab. We moved it to the Users tab for NU, but left it where it was for UIC. Now it will be in Users for everyone.

# Screenshot

### Before

Screen Shot 2020-02-13 at 1 41 51 PM

### After

Screen Shot 2020-02-13 at 1 41 37 PM


config/locales/views/admin/en.sidenav_billing.yml
```diff
@@ -0,0 +1,15 @@
+en:
+  views:
+    admin:
+      shared:
+        sidenav_billing:
+          transactions_header: Transactions
+          all_transactions: All Transactions
+          move_transactions: "!facilities.movable_transactions.head!"
```
arnamak: What does this do?

jhanggi: It's a feature of [text_helpers](https://github.com/ahorner/text-helpers#suggested-use) that allows you to reference other locale entries. I like to use it sometimes in menus to reference the page title it points to so we're not having the string "Move Transactions" in two places.

PR #2214, opened 2020-02-15
Switch from Poltergeist/PhantomJS to Chrome headless and Rails system tests

jhanggi:
# Release Notes

Tech tasks: Switch to using Rails system specs. Switch from using Poltergeist/PhantomJS to Chrome Headless as the former is no longer maintained.

# Additional Context

I found this as I was working on some newer JS which was breaking in PhantomJS. Switching to system specs is a good thing to do anyways now that we're at Rails 5.2 since they claim to be faster.

A lot of the changes were simply renaming  files from `spec/features` to `spec/system`.


spec/system/admin/facility_reservations_search_spec.rb
```diff
@@ -28,7 +28,7 @@
       login_as director
       visit facility_reservations_path(facility)
 
-      select instrument, from: "Products"
+      select instrument.name, from: "Products"
```
jhanggi: This gets rid of a capybara warning that you're only supposed to target strings and integers, not objects.

spec/system/admin/item_price_policies_controller_spec.rb
```diff
@@ -69,6 +69,6 @@
   end
 
   def table_row(*columns)
-    columns.join("\t")
+    columns.join(" ")
```
jhanggi: It seems phantomjs and chrome treat the content to text of tables slightly different.

PR #2227, opened 2020-02-24
Use Uglifier with harmony in order to allow ES6 syntax.

jhanggi:
# Release Notes

Tech task: Allow ES6 syntax in JS.

# Additional Context

This error came up on open stage: `Uglifier::Error: Unexpected token: keyword (const). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true)`


jhanggi:
Separately, we should assess if Uglifier has fixed this bug: https://github.com/mishoo/UglifyJS2/issues/3245

We're currently locked down to an older version.

PR #2246, opened 2020-03-09
[#149739] Fix sorting when deciding most recent order

jhanggi:
# Release Notes

Fix "Add to Order" issue where the defaults may be coming from the wrong order detail.

# Additional Context

See #2207 for the original implementation

In Oracle, it's possible that the IDs of the rows in `order_details` (or really in any table) may not be assigned in the order of the sequence. This is due to the way that Oracle caches values in a sequence: it preallocates them into a cache, but different nodes get their own cache. So you can end up with an `ORDER_BY id` being different than `ORDER BY created_at`.

In the future, we should consider how to ensure proper ordering (see the references), but that is out of scope of the current ticket.

A couple references:
https://asktom.oracle.com/pls/apex/f?p=100:11:::NO::P11_QUESTION_ID:9539374000346423410
https://stackoverflow.com/questions/13513936/oracle-sequence-value-are-not-ordered
https://stackoverflow.com/questions/4866959/oracle-rac-and-sequences

### DIFF
--- a/app/forms/add_to_order_form.rb
+++ b/app/forms/add_to_order_form.rb
@@ -66,6 +66,24 @@ class AddToOrderForm
     AvailableAccountsFinder.new(original_order.user, current_facility)
   end
 
+  # We're using the ordered_at of the order details to determine if additional OrderDetails
+  # have been added to the order. We find the most recently added order that does not
+  # match the original order's ordered_at and use that for our basis for defaults.
+  # There is an edge case: 1) Upload an order via bulk upload. 2) Add on to that order
+  # here. 3) Upload again via bulk upload matching the ordered_at of the original.
+  # In that situation, we will return the results of #2.
+  def previously_added_order_detail
+    return @previously_added_order_detail if defined?(@previously_added_order_detail)
+
+    # We are ordering by created_at rather than ID because it is possible in Oracle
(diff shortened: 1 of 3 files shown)


app/forms/add_to_order_form.rb
```diff
@@ -66,6 +66,24 @@ def available_accounts
     AvailableAccountsFinder.new(original_order.user, current_facility)
   end
 
+  # We're using the ordered_at of the order details to determine if additional OrderDetails
+  # have been added to the order. We find the most recently added order that does not
+  # match the original order's ordered_at and use that for our basis for defaults.
+  # There is an edge case: 1) Upload an order via bulk upload. 2) Add on to that order
+  # here. 3) Upload again via bulk upload matching the ordered_at of the original.
+  # In that situation, we will return the results of #2.
+  def previously_added_order_detail
```
jhanggi: I felt a little weird making this `public` just for testing purposes, but it made my confidence in the tests much stronger.

spec/factories/order_details.rb
```diff
@@ -4,17 +4,21 @@
   factory :order_detail do
     quantity { 1 }
     created_by { 0 }
+    order_status { OrderStatus.new_status }
```
jhanggi: It's a little weird to me that these weren't here before. My guess is we don't use this factory all that often.

PR #2257, opened 2020-03-19
[#152869] Add event logging to user suspension/unsuspension

cardoce:
# Release Notes

Log the following event: Who/when someone suspends a user


app/services/log_event_searcher.rb
```diff
@@ -28,7 +28,7 @@ def search
     result = result.merge(filter_date) if start_date || end_date
     result = result.merge(filter_event) if events.present?
     result = result.merge(filter_query) if query.present?
-    result
+    result.reverse_chronological
```
jhanggi: This ordering is now happening in the controller from #2257. It's also likely the cause of the spec failures.

```suggestion
    result
```

app/services/log_event_searcher.rb
```diff
@@ -4,7 +4,7 @@ class LogEventSearcher
 
   ALLOWED_EVENTS = ["account.create", "account.update",
                     "account_user.create", "account_user.delete",
-                    "user.create"].freeze
+                    "user.create", "user.suspended", "user.unsuspended"].freeze
```
jhanggi: I'm guessing this will conflict with #2257 if that goes in first, but we can manage that conflict when it arises.

jhanggi:
@cardoce can you rebase this off master (or merge in master)? There are some things that are not necessarily conflicts, but add some extra noise to this PR.

app/services/log_event_searcher.rb
cardoce: Yes, I was counting on this, and as I keep moving forward in this ticket Im guessing we'll encounter this situation over and over

PR #2260, opened 2020-03-27
[#152869] Log creation of statement 

cardoce:
# Release Notes

Log the following event: Who/when someone created a statement


app/models/statement.rb
```diff
@@ -53,6 +53,11 @@ def self.find_by_invoice_number(query)
     find_by(id: id, account_id: account_id)
   end
 
+  def self.where_invoice_number(query)
```
jhanggi: Can we refactor `find_by_invoice_number` to use this method? I think something like this will do the trick?

```ruby
def self.find_by_invoice_number(query)
  where_invoice_number(query)&.first
end
```

app/models/statement.rb
```diff
@@ -78,4 +83,8 @@ def rows_for_order_detail(order_detail)
     statement_rows.where(order_detail_id: order_detail.id)
   end
 
+  def to_s
```
jhanggi: I'm going back and forth on how we should do this. I don't love making this the `to_s`, which could be used in a broader context and this format wouldn't necessarily make sense.

We could do what we do for `AccountUser` where it's in a case statement. That might be fine to continue for now.

Another idea would be something like `to_log_s` and we could do a `respond_to?(:to_log_s)` check. 

Yet another idea would be to create `LogEvents::StatementPresenter` and then a `LogEvents.const_get("#{object.class}Presenter")` with a rescue which falls back.

I'm just brainstorming right now. What are your thoughts?

app/services/statement_creator.rb
```diff
@@ -17,8 +17,8 @@ def create
       set_order_details_to_statement
       setup_statement_from_details
       raise ActiveRecord::Rollback if errors.any?
+      LogEvent.log(@statement, :create, session_user)
```
jhanggi: It's possible for this statement creator to create multiple statements. It creates one for each account. I think we need to either put this in the loop where the statement gets created or we might be able to loop over `@account_statements.values` (that variable appears to be a mapping of account => statement.

app/models/statement.rb
cardoce: Sure I'll change it, and make sure that the test swipe runs

app/services/log_event_searcher.rb
```diff
@@ -5,7 +5,8 @@ class LogEventSearcher
   ALLOWED_EVENTS = ["account.create", "account.update",
                     "account_user.create", "account_user.delete",
                     "user.create", "user.suspended", "user.unsuspended",
-                     "account.suspended", "account.unsuspended"].freeze
+                    "account.suspended", "account.unsuspended", 
+                    "statement.create"].freeze
```
jhanggi: Whenever the next thing goes in, can we move `].freeze` onto its own line. That should make the diffs a little cleaner (as long as we always put a comma after the last item in the list).

app/models/statement.rb
cardoce: I like the `to_log_s` idea, if we implement it we should change this method tho
```
def loggable_to_s
    case loggable
    when AccountUser
      "#{loggable.account} / #{loggable.user}"
    else
      loggable.to_s
    end
  end
```
Change the `loggable.to_s` to `loggable.to_log_s` and refactor the ones we are logging right now.
Would this be ok?

app/models/log_event.rb
```diff
@@ -27,7 +27,11 @@ def loggable_to_s
     when AccountUser
       "#{loggable.account} / #{loggable.user}"
     else
-      loggable.to_s
+      if loggable.respond_to?(:to_log_s)
+        loggable.to_log_s
+      else
+        loggable.to_s
+      end
```
jhanggi: Can we pull the `AccountUser` case into `AccoutnUser#to_log_s` and then get rid of the `case` completely?

cardoce: Sure I'm on it

PR #2265, opened 2020-03-30
[#152869] Log creation of journal

cardoce:
# Release Notes

Log the following event: Who/when someone created a journal


cardoce:
This PR is going to have conflicts ones #2260 is merge

app/models/log_event.rb
```diff
@@ -27,7 +27,11 @@ def loggable_to_s
     when AccountUser
       "#{loggable.account} / #{loggable.user}"
     else
-      loggable.to_s
+      if loggable.respond_to?(:to_log_s)
```
jhanggi: I noted this over in another PR. It doesn't matter which PR it happens, but I'd like to pull up the `AccountUser` code. The main thing for me is I don't like this nested conditional.

cardoce: Done changes where fixed in #2260

PR #2266, opened 2020-03-30
Upgrade Rails to 5.2.4.2

jhanggi:
# Release Notes

Tech task: Upgrade Rails to 5.2.4.2. This is primarily a security update.

# Additional Context

See https://github.com/advisories/GHSA-65cv-r6x7-79hv for a description of the vulnerability.


jhanggi:
Failing test is due to https://github.com/rails/rails/pull/36671

Before, the `Occupancy` was being validated as an association of `OrderDetail`. After this change, it is no longer being validated. I was able to `git bisect` through the rails commits to the change.

I believe the issue is that because nothing is changing on the Occupancy, the validations are not running (`record.changed_for_autosave?`).

Will look into how we can force this association to get validated, possibly by manually marking it dirty.

PR #2269, opened 2020-04-03
[#152869] Log resolve disputed order

cardoce:
# Release Notes

Log the following event: Who/when someone disputes and resolve an order

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -895,6 +895,10 @@ class OrderDetail < ApplicationRecord
     where(id: id, order_id: order_id)
   end
 
+  def to_log_s
+    "Order # #{order_id}-#{id}"
+  end
+
   private
 
   # Is there enough information to move an associated order to complete/problem?
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -890,6 +890,11 @@ class OrderDetail < ApplicationRecord
     end
   end
 
+  def self.where_order_number_and_id(query)
+    return nil unless /\A(?<order_id>\d+)-(?<id>\d+)\z/ =~ query
+    where(id: id, order_id: order_id)
+  end
+
   private
 
   # Is there enough information to move an associated order to complete/problem?
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -896,7 +896,7 @@ class OrderDetail < ApplicationRecord
   end
 
   def to_log_s
-    "Order # #{order_id}-#{id}"
+    "Order ##{order_number}"
   end
 
   private
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -890,7 +890,7 @@ class OrderDetail < ApplicationRecord
     end
   end
 
-  def self.where_order_number_and_id(query)
+  def self.where_order_number(query)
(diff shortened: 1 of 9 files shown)


app/models/order_detail.rb
```diff
@@ -890,6 +890,15 @@ def notify_purchaser_of_order_status
     end
   end
 
+  def self.where_order_number_and_id(query)
+    return nil unless /\A(?<order_id>\d+)-(?<id>\d+)\z/ =~ query
+    where(id: id, order_id: order_id)
```
jhanggi, line 895: I wonder if we could maybe now use this in the `GlobalSearch::OrderSearcher`. It's fine for now, though.

app/models/order_detail.rb
```diff
@@ -890,6 +890,15 @@ def notify_purchaser_of_order_status
...
+  end
+
+  def to_log_s
+    "Order # #{order_id}-#{id}"
```
jhanggi, line 899: Can we use `OrderDetail#order_number`? And this is kind of picky, but everywhere we display an order number there's no space between the `#` and the number.

app/services/order_details/dispute_resolved_notifier.rb
```diff
@@ -4,12 +4,20 @@ module OrderDetails
 
   class DisputeResolvedNotifier < SimpleDelegator
 
+    attr_accessor :current_user
+
+    def initialize(order_detail, current_user='')
```
jhanggi, line 9: I think this is a good place for a keyword arg. And also, using a a blank string as the default seems a little weird here. Maybe just a nil.

```suggestion
    def initialize(order_detail, current_user: nil)
```

config/locales/views/admin/en.log_events.yml
```diff
@@ -13,7 +13,10 @@ en:
         journal:
           create: Journal created
         statement:
-          create: Statement created
+          create: Statement added
```
jhanggi, line 16: Why did you change this? I don't have a strong preference for "created" vs "added", but I think we should be consistent with journals.

app/models/order_detail.rb
cardoce: Sure

config/locales/views/admin/en.log_events.yml
cardoce: Sorry I'll change this back, probably I miss it when I merge master and fixed conflicts.

app/services/order_details/param_updater.rb
```diff
@@ -79,7 +79,7 @@ def update_attributes(params)
   end
 
   def trigger_notifications
-    OrderDetails::DisputeResolvedNotifier.new(@order_detail).notify
+    OrderDetails::DisputeResolvedNotifier.new(@order_detail, @editing_user).notify
```
jhanggi, line 82: Looks like we need to update this to use the keyword arg now

config/locales/views/admin/en.log_events.yml
```diff
@@ -14,6 +14,9 @@ en:
           create: Journal created
         statement:
           create: Statement created
+        order_detail:
+          dispute: Order disputed
+          resolve: Order resolved
```
jhanggi, line 19: Let's say `Order dispute resolved` to be a little more specific.

PR #2270, opened 2020-04-03
[#152869] Log send billing notification

cardoce:
# Release Notes

Log the following event: Who/when someone sends the billing notifications (record each order)

### DIFF
--- a/app/controllers/facility_notifications_controller.rb
+++ b/app/controllers/facility_notifications_controller.rb
@@ -46,6 +46,10 @@ class FacilityNotificationsController < ApplicationController
 
     if sender.perform
       flash[:notice] = send_notification_success_message(sender)
+      order_details = OrderDetail.where(id: params[:order_detail_ids])
+      order_details.each do |order_detail|
+        LogEvent.log(order_detail, :notify, current_user)
+      end
     else
       flash[:error] = I18n.t("controllers.facility_notifications.errors_html", errors: sender.errors.join("<br/>")).html_safe
     end
--- a/app/controllers/facility_notifications_controller.rb
+++ b/app/controllers/facility_notifications_controller.rb
@@ -43,10 +43,10 @@ class FacilityNotificationsController < ApplicationController
     end
 
     sender = NotificationSender.new(current_facility, params[:order_detail_ids])
+    order_details = OrderDetail.where_ids_in(params[:order_detail_ids])
 
     if sender.perform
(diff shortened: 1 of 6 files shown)


cardoce:
Should be merge after #2269 to fix conflicts, also the tests for the query search are in #2269

app/controllers/facility_notifications_controller.rb
```diff
@@ -46,6 +46,10 @@ def send_notifications
 
     if sender.perform
       flash[:notice] = send_notification_success_message(sender)
+      order_details = OrderDetail.where(id: params[:order_detail_ids])
```
jhanggi, line 49: You probably haven't run into it before, but something to be aware of is Oracle has a 1000 item limit in `WHERE IN` statements. So this query would fail if we're sending notifications to more than 1000 order details. We have a special scope we've created to deal with this `where_ids_in` that we would want to use.

In order to completely avoid the issue, can we get the list of order details out of the `sender` rather than doing an additional query? It looks like `NotificationSender#order_details` is private, but I wouldn't have a problem with making it public.

In that sender, there's a `@orders_notified` variable that I don't think is used. If it's not used, can we get rid of it.

app/models/order_detail.rb
```diff
@@ -890,6 +890,15 @@ def notify_purchaser_of_order_status
     end
   end
 
+  def self.where_order_number_and_id(query)
```
jhanggi, line 893: Maybe just `where_order_number`? This method name feels a little unclear.

app/controllers/facility_notifications_controller.rb
cardoce: I'll take a look and change it

app/controllers/facility_notifications_controller.rb
```diff
@@ -43,9 +43,13 @@ def send_notifications
     end
 
     sender = NotificationSender.new(current_facility, params[:order_detail_ids])
+    order_details = OrderDetail.where_ids_in(params[:order_detail_ids])
 
     if sender.perform
       flash[:notice] = send_notification_success_message(sender)
+      order_details.each do |order_detail|
```
jhanggi, line 50: Sorry if I confused you. I wanted to bring up the Oracle thing  so you were aware of it and this seemed like a good place to bring it up. I think we can avoid the issue entirely by using `sender.order_details.each` now that that method is public. It also avoids an extra lookup and loading of objects into memory.

PR #2272, opened 2020-04-03
[#152869] Log when a price is manually change

cardoce:
# Release Notes

Log the following event: Who/when someone manually reprices an Order

### DIFF
--- a/app/services/order_details/param_updater.rb
+++ b/app/services/order_details/param_updater.rb
@@ -62,15 +62,14 @@ class OrderDetails::ParamUpdater
     assign_attributes(params)
 
     @order_detail.manually_priced! # don't auto-reassign price
-    assign_price_changed_by_user
 
     @order_detail.transaction do
+      assign_price_changed_by_user
       @order_detail.reservation.save_as_user(@editing_user) if @order_detail.reservation
       if order_status_id && order_status_id.to_i != @order_detail.order_status_id
         change_order_status(order_status_id, @options[:cancel_fee]) || raise(ActiveRecord::Rollback)
       else
         @order_detail.save_as_user(@editing_user) || raise(ActiveRecord::Rollback)
-        LogEvent.log(@order_detail, :price_change, @editing_user) unless @order_detail.price_changed_by_user == nil
       end
     end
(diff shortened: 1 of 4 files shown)


app/services/order_details/param_updater.rb
```diff
@@ -90,6 +90,7 @@ def assign_price_changed_by_user
       @order_detail.price_changed_by_user = nil
     elsif %w[actual_cost actual_subsidy price_change_reason].any? { |a| @order_detail.changed.include?(a) }
       @order_detail.price_changed_by_user = @editing_user
+      LogEvent.log(@order_detail, :price_change, @editing_user)
```
jhanggi, line 93: I'm about 95% sure it will get rolled back, but can we add a spec where we hit this code path, but something else causes the order detail to be invalid and not save. We should make sure this doesn't get created in the situation where the update doesn't happen.

cardoce: Sure I'll add it

app/services/order_details/param_updater.rb
```diff
@@ -70,6 +70,7 @@ def update_attributes(params)
         change_order_status(order_status_id, @options[:cancel_fee]) || raise(ActiveRecord::Rollback)
       else
         @order_detail.save_as_user(@editing_user) || raise(ActiveRecord::Rollback)
+        LogEvent.log(@order_detail, :price_change, @editing_user) unless @order_detail.price_changed_by_user == nil
```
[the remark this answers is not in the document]
jhanggi: Will we get duplicate logs if: 
* User 1 makes a price change. This should log as expected.
* User 2 makes a non-price change

Will this second action also end up with a log because the price changed by user is user 1?

Maybe we could use something in [ActiveRecord::Dirty](https://api.rubyonrails.org/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-saved_change_to_attribute-3F) like `saved_change_to_price_change_reason?`

Or maybe even simpler: move the log back into `assign_price_changed_by_user`, but move the call to that method _inside_ the transaction block. That way if the save on order detail fails, it raises the rollback and the log event creation should get rolled back as well.

PR #2280, opened 2020-04-24
[152869] Log order move to problem queue

cardoce:
# Release Notes

Log the following event: When an order moves to the problem queue
Add metadata to Log Event


vendor/engines/secure_rooms/app/support/secure_rooms/auto_orphan_occupancy.rb
```diff
@@ -25,7 +25,7 @@ def long_running_occupancies
     def orphan_occupancy(occupancy)
       occupancy.mark_orphaned!
       SecureRooms::AccessHandlers::OrderHandler.process(occupancy)
-      MoveToProblemQueue.move!(occupancy.order_detail) if occupancy.order_detail
+      MoveToProblemQueue.move!(occupancy.order_detail, cause: :orphaned) if occupancy.order_detail
```
jhanggi, line 28: I think this is probably fine, though I could see NU asking for to distinguish between these two orphan cases. I think this one looks like it's similar to the `auto_expire`, so we could all this one `auto_orphan` if we need to change it. The other one I feel like could have multiple context and figuring those out would be a pain.

app/controllers/offline_reservations_controller.rb
```diff
@@ -61,7 +61,7 @@ def move_ongoing_reservations_to_problem_queue
       # We need to force completion because a default completion requirement is that
       # the current time is after the reserve_end_at. In this offline case, the reservation
       # might not be over yet.
-      MoveToProblemQueue.move!(order_detail, force: true)
+      MoveToProblemQueue.move!(order_detail, force: true, user: :current_user, cause: :new_offline_reservation)
```
jhanggi, line 64: I think this needs to be the method `current_user` not the symbol.

PR #2286, opened 2020-05-06
Dispute resolution logging

cardoce:
# Release Notes

Fix issue with logging and notifying orders with accessories


cardoce:
We now have the issue of triggering an email if the resolve changes.

PR #2287, opened 2020-05-08
Fix Oracle error test involving metadata

cardoce:
# Release Notes

Fix error involving metadata in WHERE clause in tests. 

# Additional Context
In Oracle, we can't put a CLOB in the WHERE clause. Because Large objects (LOBs) are not supported in comparison conditions.

### DIFF
--- a/spec/controllers/offline_reservations_controller_spec.rb
+++ b/spec/controllers/offline_reservations_controller_spec.rb
@@ -45,7 +45,7 @@ RSpec.describe OfflineReservationsController do
         post :create, params: params
         log_event = LogEvent.find_by(loggable: reservation.order_detail, event_type: :problem_queue)
         expect(log_event).to be_present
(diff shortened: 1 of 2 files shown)


spec/controllers/offline_reservations_controller_spec.rb
```diff
@@ -43,8 +43,9 @@
 
       it "logs the problem reservation" do
         post :create, params: params
-        log_event = LogEvent.find_by(loggable: reservation.order_detail, event_type: :problem_queue, metadata: {"cause"=>"new_offline_reservation"})
+        log_event = LogEvent.find_by(loggable: reservation.order_detail, event_type: :problem_queue)
         expect(log_event).to be_present
+        expect(log_event.metadata).to eq({"cause"=>"new_offline_reservation"})
```
jhanggi, line 48: We can leave the curly braces off since they'd be implied within the parenthese.

PR #2290, opened 2020-05-11
[#152869] Log resolve order problem queue

cardoce:
# Release Notes

Log the following event: Who/when someone resolves an order in the problem queue

# Additional Context

We are only logging when a user resolves the problem, and we are not logging or tracking when a facility resolves the problem

### DIFF
--- a/config/locales/views/admin/en.log_events.yml
+++ b/config/locales/views/admin/en.log_events.yml
@@ -21,6 +21,7 @@ en:
           review: Order reviewed
           problem_queue: Order added to problem queue %{cause}
           price_change: Order's price was manually change
+          resolve_from_problem_queue: Order remove from problem queue
         end_date: End Date
         event_filter: Event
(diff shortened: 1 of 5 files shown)


config/locales/views/admin/en.log_events.yml
```diff
@@ -21,6 +21,7 @@ en:
           review: Order reviewed
           problem_queue: Order added to problem queue %{cause}
           price_change: Order's price was manually change
+          resolve_from_problem_queue: Order remove from problem queue
```
jhanggi, line 24: Let's shorten this call this "Problem order resolved".

While we're in here, can we change the `price_change` to "Order's price was manually changed" (add a "d" to the end)?

jhanggi:
Looks like there's a test failure

cardoce:
> Looks like there's a test failure

PR #2292, opened 2020-05-12
[#152604] Prevent abandoned reservation carts

jhanggi:
# Release Notes

Prevent abandoned carts created when clicking an instrument name. This should also help prevent order/detail IDs from growing too fast.

# TODO

- [x] Refactor so there is less duplicated code between the old ReservationsController and the new controller
- [x]  Try to get rid of `@add_to_order` in the old controller
- [x] Further testing/QA

# Additional Context

Before, if you click an instrument, a new order and order detail are created. Users are often doing this to check the schedule for the instrument. If the user closes the tab comes back to the instrument, yet another order detail is created. We were deleting the cart if the user clicks "Cancel", but the order and order detail IDs were still eaten up.

Now, for normal single-reservation purchases, we go to a parallel controller where we do not create the order/detail prior to the user filling out the form. I did need to rework some of the code to avoid persisting and then rolling back the order/detail. Without these changes, the abandoned carts would not be created, but the IDs would still be used up.

For the order form (purchasing more than one or with other things) or part of a bundle, as well as edit/update, we are continuing to use the old controller.

Some offline discussion for context:

From Aaron:

> What we highly suspect is that users are clicking on the instrument links to check out the calendar, and then closing the tab. I assume that if a user clicks on an instrument name bringing them to a "Create Reservation" page, and then clicks "cancel" then this will not generate an abandoned cart? Clicking cancel from the one-off "Create Reservation" page generates the banner: "the product has been removed." I assume the Order Number has still been used up, but that this wouldn't create an "abandoned cart"?

Jason:

> You're right about the cancel button. I did a quick test: if you click "cancel" we do remove the order detail (so that number is used up), but we keep the order. That order gets re-used. So clicking an instrument creates, say, "100-100". Click cancel. Click the instrument again. You get "100-101". This won't count as an abandoned cart. However, if you click the browser's back button or close the tab, it leaves both of them hanging around. So "100-100", click on the instrument again, then you get "101-101". "100-100" is an abandoned cart.

### DIFF
--- a/app/controllers/single_reservations_controller.rb
+++ b/app/controllers/single_reservations_controller.rb
@@ -1,0 +1,82 @@
+class SingleReservationsController < ApplicationController
+
+  customer_tab  :all
+  before_action :authenticate_user!
+  load_resource :facility, find_by: :url_name
+  load_resource :instrument, through: :facility, find_by: :url_name
+  before_action :build_order
+  before_action { @submit_action = facility_instrument_single_reservations_path }
+
+  def new
+    options = current_user.can_override_restrictions?(@instrument) ? {} : { user: acting_user }
+    next_available = @instrument.next_available_reservation(after: 1.minute.from_now, duration: default_reservation_mins.minutes, options: options)
+    @reservation = next_available || default_reservation
+    @reservation.order_detail = @order_detail
+
+    authorize! :new, @reservation
+
+    @reservation.round_reservation_times
+    unless @instrument.can_be_used_by?(acting_user)
+      flash[:notice] = text(".acting_as_not_on_approval_list")
+    end
+    set_windows
+    render "reservations/new"
+  end
+
+  def create
+    creator = ReservationCreator.new(@order, @order_detail, params)
+    @reservation = creator.reservation
+    authorize! :create, @reservation
+    if creator.save(session_user)
+      flash[:notice] = I18n.t "controllers.reservations.create.success"
+      flash[:error] = I18n.t("controllers.reservations.create.admin_hold_warning") if creator.reservation.conflicting_admin_reservation?
+      redirect_to purchase_order_path(@order, params.permit(:send_notification))
+    else
+      flash.now[:error] = creator.error.html_safe
+      set_windows
+      render "reservations/new"
+    end
+  end
+
+  private
+
+  def ability_resource
+    @reservation
+  end
+
+  def build_order
+    @order = Order.new(user: current_user,
+                       facility: current_facility,
+                       created_by: session_user.id,
+                       )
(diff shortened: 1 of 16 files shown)


app/controllers/single_reservations_controller.rb
```diff
@@ -0,0 +1,84 @@
+class SingleReservationsController < ApplicationController
```
jhanggi, line 1: I'm not in love with the name--especially in the route/url--but I haven't come up with anything better.

app/controllers/single_reservations_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  customer_tab  :all
+  before_action :authenticate_user!
+  load_resource :facility, find_by: :url_name
+  load_resource :instrument, through: :facility, find_by: :url_name
+  before_action :build_order
+  before_action { @submit_action = facility_instrument_single_reservations_path }
```
jhanggi, line 8: This seemed to be the simplest way of still using the same view partial as the old reservations form. I'm open to other ideas.

app/controllers/single_reservations_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+
+    @reservation.round_reservation_times
+    unless @instrument.can_be_used_by?(acting_user)
+      flash[:notice] = text(".acting_as_not_on_approval_list")
+    end
+    set_windows
+    render "reservations/new"
+  end
+
+  def create
```
jhanggi, line 26: I copied this from `ReservationsController` and cleaned out some of the unneeded pieces (like merge orders). I still want to do some further refactors to remove some of the duplication.

app/controllers/single_reservations_controller.rb
```diff
@@ -0,0 +1,84 @@
...
+    Reservation.new(instrument: @instrument,
+                    reserve_start_at: Time.zone.now,
+                    reserve_end_at: default_reservation_mins.minutes.from_now)
+  end
+
+  def default_reservation_mins
+    @instrument.min_reserve_mins.to_i > 0 ? @instrument.min_reserve_mins : 30
+  end
+
+  def set_windows
```
jhanggi, line 71: This and some of the other helper methods are ripe for refactoring.

app/models/reservation.rb
```diff
@@ -206,6 +206,16 @@ def assign_actuals_off_reserve
     self.actual_end_at   ||= reserve_end_at
   end
 
+  def valid_as_user?(user)
```
jhanggi, line 209: This is essentially a duplicate of `save_as_user`, but with `valid?`. I might try again to DRY it up, but my initial attempts didn't work.

PR #2293, opened 2020-05-12
Fix issue in Log event index dropdown

cardoce:
# Release Notes
Fix issues with dropdown for event type in the log event index
# Additional Content
Error Reported: Dropdown contain "Order added to problem queue %{cause}"


cardoce:
Sure I'll fix the Release Notes

PR #2294, opened 2020-05-13
[#152869] Log edit flag user

cardoce:
# Release Notes

Log the following event: Who/when someone edits the internal/external flag on a user


app/controllers/users_controller.rb
```diff
@@ -118,8 +118,13 @@ def edit
 
   # PUT /facilities/:facility_id/users/:id
   def update
+    previous = @user.price_groups.to_a
     @user_form = user_form_class.new(@user)
     if @user_form.update_attributes(edit_user_params) && @user.update_price_group(price_group_params)
+      if previous != @user.price_groups.reload.to_a
+        new_price_group = @user.price_groups.to_a - previous
+        LogEvent.log(@user, :internal_changed, current_user, metadata: {price_group_rate: new_price_group.first.name})
```
jhanggi, line 126: I think our normal style is to use spaces inside the hash braces. Not a big deal, though.

```suggestion
        LogEvent.log(@user, :internal_changed, current_user, metadata: { price_group_rate: new_price_group.first.name })
```

app/controllers/users_controller.rb
```diff
@@ -118,8 +118,13 @@ def edit
 
   # PUT /facilities/:facility_id/users/:id
   def update
+    previous = @user.price_groups.to_a
     @user_form = user_form_class.new(@user)
     if @user_form.update_attributes(edit_user_params) && @user.update_price_group(price_group_params)
+      if previous != @user.price_groups.reload.to_a
+        new_price_group = @user.price_groups.to_a - previous
```
jhanggi, line 125: I'm thinking this variable should be plural since it's an array. `new_price_groups`.

app/controllers/users_controller.rb
```diff
@@ -118,8 +118,13 @@ def edit
 
   # PUT /facilities/:facility_id/users/:id
   def update
+    previous = @user.price_groups.to_a
```
jhanggi, line 121: I think we can be a little more explicit in our variable name here: `previous_price_groups`.

app/services/log_event_searcher.rb
```diff
@@ -5,8 +5,8 @@ class LogEventSearcher
   ALLOWED_EVENTS = ["account.create", "account.update",
                     "account_user.create", "account_user.delete",
                     "user.create", "user.suspended", "user.unsuspended",
-                    "account.suspended", "account.unsuspended",
-                    "journal.create","statement.create",
+                    "user.internal_changed", "account.suspended",
```
jhanggi, line 8: This is kind of picky, but can we leave the two account events on the same line since they're related to each other?

config/locales/views/admin/en.log_events.yml
```diff
@@ -5,6 +5,8 @@ en:
         dropdown_titles:
           order_detail:
             problem_queue: Order added to problem queue
+          user:
+            internal_changed: User price group changed
```
jhanggi, line 9: I'm not sure I like `internal_changed`. It seems like it's not quite descriptive enough. Maybe `default_price_group_changed`, or `base_price_group_changed`?

spec/controllers/users_controller_spec.rb
```diff
@@ -77,6 +77,13 @@
       expect(user.username).to eq("user234")
       expect(response).to redirect_to facility_user_path(facility, user)
     end
+
+    it_should_allow(:admin, "to log the change of the internal flag and log the new rate as metadata") do
```
jhanggi, line 81: Can we also include a test that specifies that it does not log if the price group does not change?

Also, it's possible we have the feature setting overridden in the spec, but can you make sure this test still works when the setting `user_based_price_groups` is set to false (as it is at Dartmouth)?

spec/controllers/users_controller_spec.rb
```diff
@@ -77,6 +77,13 @@
...
+      log_event = LogEvent.find_by(loggable: user, event_type: :internal_changed)
+      expect(log_event).to be_present
+      expect(log_event.metadata).to eq("price_group_rate"=>"Northwestern Base Rate")
```
jhanggi, line 84: Can we include spaces in the hash? 

```suggestion
      expect(log_event.metadata).to eq("price_group_rate" => "Northwestern Base Rate")
```

app/controllers/users_controller.rb
cardoce: Sorry, I wrote with the intention to change it once it was running properly and it slip my mind

config/locales/views/admin/en.log_events.yml
```diff
@@ -38,6 +40,7 @@ en:
           create: User added
           suspended: User suspended
           unsuspended: User unsuspended
+          default_price_group_changed: User price group changed to %{price_group_rate}
```
jhanggi, line 43: I think we can just call the metadata key `price_group`.

PR #2300, opened 2020-05-19
[#152869] Log creation of price group

cardoce:
# Release Notes

Log the following event: Price group gets created or deleted

# Additional Context

Adds soft delete to PriceGroup and also adds an irreversible migration to avoid issues with duplication

### DIFF
--- a/spec/models/price_group_spec.rb
+++ b/spec/models/price_group_spec.rb
@@ -59,6 +59,7 @@ RSpec.describe PriceGroup do
       expect(@price_group).to be_can_delete
       @price_group.destroy
       expect { PriceGroup.find(@price_group.id) }.to raise_error(ActiveRecord::RecordNotFound)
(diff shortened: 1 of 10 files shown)


spec/models/price_group_spec.rb
```diff
@@ -71,7 +71,7 @@
       it "should be deletable if no orders on policy" do
         expect(@price_group).to be_can_delete
         @price_group.destroy
-        expect(@price_group).to be_destroyed
+        expect { PriceGroup.find(@price_group.id) }.to raise_error(ActiveRecord::RecordNotFound)
```
jhanggi, line 74: I think it might be a good idea to add a test that verifies that it is still there.

`expect(PriceGroup.with_deleted.find(@price_group.id)).to be_present`

jhanggi:
@cardoce Let's hold on merging this for right now. I think we should try to keep master clean for now until we can get the next NU release out.

PR #2301, opened 2020-05-19
[#153552] Fix issue where relays were getting deleted

jhanggi:
# Release Notes

Fix situation where when creating a new product group containing an instrument with a relay the instrument was getting set to "Reservation Only"

# Additional Context

This seems to be happening only when creating a new product group. Adding to an existing group does not trigger the bug.

I'm actually a little surprised this hasn't bitten us before in some other case like bundles.

What's happening is that the validations on `Product`/`Instrument` are being triggered as part of the `ProductDisplayGroup`'s validation. I can't explain why it's only being triggered in this single case. But because we're triggering the validations, there's some code in there that should only be triggered during product edit that deletes and rebuilds the relay. However, since we're only running the validations and not the save, the new relay is not persisted.

I ended up doing two things to fix it:

* At the base level, only run the destroy/create relay callback during the product edit/update process. This should avoid us getting bitten by this in other associations we may create at some point.

* Don't validate the products from the display groups. In this case, we should be safe to assume the products are already valid.

### DIFF
--- a/app/models/concerns/products/relay_support.rb
+++ b/app/models/concerns/products/relay_support.rb
@@ -64,8 +64,13 @@ module Products::RelaySupport
   # causes too much trouble. Just get rid of the old relay and
   # setup from scratch.
   def destroy_and_init_relay
+    # This instance variable is only present when using the instrument edit form.
+    # We're adding this check to avoid any autosave or association validations
+    # from triggering this callback.
+    return unless @control_mechanism
+
     attrs = relay.try(:attributes) || {}
-    relay.try :destroy
+    relay&.destroy
 
     case control_mechanism
     when Relay::CONTROL_MECHANISMS[:timer]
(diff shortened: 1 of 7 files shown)


app/models/concerns/products/relay_support.rb
```diff
@@ -64,8 +64,13 @@ def check_relay_with_right_type
...
   # setup from scratch.
   def destroy_and_init_relay
+    # This instance variable is only present when using the instrument edit form.
+    # We're adding this check to avoid any autosave or association validations
+    # from triggering this callback.
+    return unless @control_mechanism
+
     attrs = relay.try(:attributes) || {}
-    relay.try :destroy
+    relay&.destroy
```
cardoce, line 73: Why are we changing the `try` for this other safe navigation operator, does it make a particular difference? Not asking to change it, just wondering why.

jhanggi: `try` is kind of a funky method in that it's more of a `object.send(:method_name) if object.respond_to?(:method_name)` than a nil check. It will still respond with a `nil` if the the object is present, but doesn't not have that method. Or at least that's how it worked at some point.

`try` existed in Rails way before the safe navigation operator and this is a really old method (nucore was on ruby 1.8.7 at one point). And  I personally think the safe operator is clearer in that it implies you're doing a nil check.

PR #2302, opened 2020-05-19
#[152869] Log when a user gets added or removed from a price group

cardoce:
# Release Notes

Log the following event: Who/when someone added a user to a price group

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,11 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-<<<<<<< HEAD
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -434,7 +437,10 @@
     t.integer "display_order", null: false
     t.boolean "is_internal", null: false
     t.boolean "admin_editable", default: true, null: false
-    t.index ["facility_id", "name"], name: "index_price_groups_on_facility_id_and_name", unique: true
+    t.datetime "created_at"
```
jhanggi, line 440: I think these might have snuck in from another PR. Not a big deal and we can clean this up when we're merging.

PR #2303, opened 2020-05-21
[#152869] Log relay tool use

cardoce:
# Release Notes

Log the following event: Who/when the Relay Tool is used

# Additional Context

Add Rollbar report to ActiveSupport::Notifications

### DIFF
--- a/app/controllers/relays_activations_controller.rb
+++ b/app/controllers/relays_activations_controller.rb
@@ -7,17 +7,21 @@ class RelaysActivationsController < ApplicationController
   authorize_resource :facility
 
   def create
+    LogEvent.log(current_facility, :activate, current_user)
     relays.each(&:activate)
     redirect_to facility_instruments_path(current_facility), flash: { notice: text("turned_on") }
   rescue NetBooter::Error
     redirect_to facility_instruments_path(current_facility), flash: { alert: text("connection_error") }
+    raise ActiveRecord::Rollback
   end
 
   def destroy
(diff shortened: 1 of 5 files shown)


config/locales/views/admin/en.log_events.yml
```diff
@@ -27,6 +27,9 @@ en:
           problem_queue: Order added to problem queue %{cause}
           price_change: Order's price was manually changed
           resolve_from_problem_queue: Problem order resolved
+        facility:
+          activate: Relays turn on
```
jhanggi, line 31: I'm trying to think of a better phrase. Not sure what's best

- "Turned on all relays"
- "Global relay activated"
- "All relays activated"

jhanggi: And if we change this, we should update the deactivate to match.

app/controllers/relays_activations_controller.rb
```diff
@@ -7,16 +7,26 @@ class RelaysActivationsController < ApplicationController
...
     redirect_to facility_instruments_path(current_facility), flash: { alert: text("connection_error") }
   end
 
   def destroy
+    LogEvent.log(current_facility, :deactivate, current_user)
     relays.each(&:deactivate)
     redirect_to facility_instruments_path(current_facility), flash: { notice: text("turned_off") }
-  rescue NetBooter::Error
+  rescue NetBooter::Error => e
+    ActiveSupport::Notifications.instrument(
```
jhanggi, line 26: If this fits on one line without being too long, can we do that?

PR #2316, opened 2020-06-17
[#153781] Handle special characters and reserved words in oracle search

jhanggi:
# Release Notes

Fixes an issue where Oracle would error when parsing reserved words like "and", "or", or "about" when doing a search for products.

# Additional Context

See https://docs.oracle.com/cd/B10501_01/text.920/a96518/cqspcl.htm#20741 for a list of the reserved words.

See https://github.com/tablexi/nucore-nu/pull/573 (private) for verification this passes CI.

### DIFF
--- a/app/models/full_text_search/oracle_searcher.rb
+++ b/app/models/full_text_search/oracle_searcher.rb
@@ -5,7 +5,14 @@ module FullTextSearch
     def search(fields, query)
       return @scope.none if query.blank?
 
-      formatted_query = query.to_s.split.join(",")
+      query_elements = query.to_s.split
+      # By wrapping each portion in {}, it escapes what's inside so it handles special
+      # characters and other reserved words
+      # https://docs.oracle.com/cd/B10501_01/text.920/a96518/cqspcl.htm#20741
+      formatted_query = query_elements
+        .map { |s| s.gsub(/[{}]/, "") } # remove any existing curly braces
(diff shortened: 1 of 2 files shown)


app/models/full_text_search/oracle_searcher.rb
```diff
@@ -5,7 +5,14 @@ class OracleSearcher < BaseSearcher
...
       return @scope.none if query.blank?
 
-      formatted_query = query.to_s.split.join(",")
+      query_elements = query.to_s.split
+      # By wrapping each portion in {}, it escapes what's inside so it handles special
+      # characters and other reserved words
+      # https://docs.oracle.com/cd/B10501_01/text.920/a96518/cqspcl.htm#20741
+      formatted_query = query_elements
+        .map { |s| s.gsub(/[{}]/, "") } # remove any existing curly braces
+        .map { |s| "{#{s}}" }
```
giladshanan, line 14: I'm guessing there aren't enough `query_elements` for this to make a performance impact, but you could do this all in one `map` like:
```
.map do |query_element| 
  query_string = query_element.gsub(/[{}]/, "") # remove any existing curly braces
  "{#{query_string}}"
end
```

jhanggi: I think I prefer the multiple maps from a readability perspective; it's only a slight preference. If I were worried about the performance impact, I would definitely make the switch, though.

PR #2320, opened 2020-06-23
[#148167] Add order for button

giladshanan:
# Release Notes

[Additional Order For Button](https://pm.tablexi.com/issues/148167)

It looks to me like it is possible to have a case where user cannot `:edit`, but they can `:switch_to`.  This is my attempt to handle both of those cases while also adding the new button inline with other buttons if they exist.

### DIFF
--- a/app/views/users/index.html.haml
+++ b/app/views/users/index.html.haml
@@ -14,7 +14,6 @@
 
   - if @users.any?
     = render "search/results_table",
-      order_for: current_facility.try(:single_facility?),
       users: @users
 
     = will_paginate(@users)
--- a/app/views/users/show.html.haml
+++ b/app/views/users/show.html.haml
@@ -23,15 +23,15 @@
   = render_view_hook "additional_user_fields", f: f
 
 %ul.inline
-  - if current_ability.can?(:switch_to, @user) && @user.active? && !current_facility.cross_facility?
-    %li= link_to text("search.results_table.order_for"), facility_user_switch_to_path(current_facility, @user), class: "btn btn-primary"
   - if can?(:edit, @user)
-    %li= link_to text("shared.edit"), edit_facility_user_path(current_facility, @user), class: "btn btn-primary"
+    %li= link_to text("shared.edit"), edit_facility_user_path(current_facility, @user), class: "btn"
     - if can?(:unexpire, @user) && @user.expired?
       %li= link_to text("unexpire"), unexpire_facility_user_path(current_facility, @user), class: "btn", method: :patch
     - elsif can?(:deactivate, @user) && @user.suspended?
       = link_to t("shared.activate"), facility_user_suspension_path(current_facility, @user), class: "btn", method: :delete
-- if can?(:edit, @user) && can?(:deactivate, @user) && @user.active?
+  - if current_ability.can?(:switch_to, @user) && @user.active? && current_facility.single_facility?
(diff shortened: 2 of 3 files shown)


app/views/users/show.html.haml
```diff
@@ -22,19 +22,17 @@
 
   = render_view_hook "additional_user_fields", f: f
 
-- if can? :edit, @user
-  %ul.inline
+%ul.inline
+  - if current_ability.can?(:switch_to, @user) && @user.active? && !current_facility.cross_facility?
+    %li= link_to text("search.results_table.order_for"), facility_user_switch_to_path(current_facility, @user), class: "btn btn-primary"
```
jhanggi, line 27: I kind of feel like `btn-primary` is not the right styling. I'm not quite sure what is.

https://bootstrapdocs.com/v2.3.2/docs/base-css.html#buttons

jhanggi: Also, can we put this after the "Edit" button?

jhanggi: Maybe the edit button should just be a plain button? For the "edit" button for products we don't use `btn-primary`. What do you think?

app/views/users/show.html.haml
```diff
@@ -22,19 +22,17 @@
 
   = render_view_hook "additional_user_fields", f: f
 
-- if can? :edit, @user
-  %ul.inline
+%ul.inline
+  - if current_ability.can?(:switch_to, @user) && @user.active? && !current_facility.cross_facility?
```
jhanggi, line 26: We also have `Facility#single_facility?` which is the inverse of `cross_facility?`.

app/views/users/index.html.haml
```diff
@@ -14,7 +14,6 @@
 
   - if @users.any?
     = render "search/results_table",
```
jhanggi, line 16: Should we dig into this view and disable the order for link when you're in cross facility mode?

app/views/users/show.html.haml
giladshanan: Yup, that makes sense to me - edit button will just be `btn`, and order for will be `btn-primary`, and come after the edit button.

app/views/users/index.html.haml
giladshanan: I'm not sure - I thought Jeff said they were going to think on it.  Should be an easy change to make if the want it done.  Should I check in with them about it?

jhanggi: You're right that it's probably an easy change so we can wait for them to give us the go ahead to change that.

app/views/users/show.html.haml
```diff
@@ -22,19 +22,17 @@
...
       = link_to t("shared.activate"), facility_user_suspension_path(current_facility, @user), class: "btn", method: :delete
-  - if can?(:deactivate, @user) && !@user.suspended? && !@user.expired?
-    %fieldset.well
-      = simple_form_for @user, url: facility_user_suspension_path(current_facility, @user), method: :post do |f|
-        = f.input :suspension_note
-        = f.submit t("shared.suspend"), class: "btn btn-danger"
-
-
-
+- if can?(:edit, @user) && can?(:deactivate, @user) && @user.active?
```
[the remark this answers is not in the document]
giladshanan: I was trying to replicate the existing logic.  I wrote a little script to help check permissions, and it looks like :edit and :deactivate always return the same (at least for the seed users).

PR #2323, opened 2020-06-26
[#152604] Fix cancel button behavior on new reservation page

jhanggi:
# Release Notes

Fixes the "Cancel" button behavior on new reservation page.

# Additional Context

This does modify behavior a little bit, but the behavior before felt unexpected.

There are a couple ways to get to this page:

* Clicking on an instrument from the facility homepage. This is the new order-less reservation flow from #2292. 
  * Before: The cancel button was missing
  * After: The cancel button will return you the facility homepage.
* The reservation was in the cart because it was a bundle.
  * Before: The cancel button brought you back to the facility homepage, but left everything in your cart
  * After: The cancel button brings you back to your cart
* The reservation was in a cart because it was added with the order form
  * Before: The order detail was removed from your cart (this seemed wrong) and brought you to the facility homepage
  * After: You are returned to your cart and everything is left alone. If you want to remove the line from your cart, you can use the "Remove" button
* Adding to an existing order
  * Before: You are brought back to the existing order's page and the line item still requires resolution.
  * After: Same behavior

### DIFF
--- a/spec/system/purchasing_a_reservation_on_behalf_of_spec.rb
+++ b/spec/system/purchasing_a_reservation_on_behalf_of_spec.rb
@@ -72,6 +72,20 @@ RSpec.describe "Purchasing a reservation on behalf of another user" do
     end
   end
 
+  describe "when the user is not authorized" do
+    before do
+      instrument.update!(requires_approval: true)
+    end
+
+    it "can purchase" do
+      click_link instrument.name
+      expect(page).to have_content("The user you are ordering for is not on the authorized list for this instrument.")
+      select user.accounts.first.description, from: "Payment Source"
+
+      click_button "Create"
+    end
+  end
+
   describe "creating multiple reservations" do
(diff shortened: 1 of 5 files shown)


spec/system/purchasing_a_reservation_on_behalf_of_spec.rb
```diff
@@ -72,6 +72,20 @@
...
+    before do
+      instrument.update!(requires_approval: true)
+    end
+
+    it "can purchase" do
+      click_link instrument.name
+      expect(page).to have_content("The user you are ordering for is not on the authorized list for this instrument.")
+      select user.accounts.first.description, from: "Payment Source"
+
+      click_button "Create"
```
giladshanan, line 85: Is there another expectation after this?

jhanggi: Oops. There should be.

PR #2337, opened 2020-07-07
[#80435] Add 'Send $0 emails?' checkbox

giladshanan:
# Release Notes

Adds a "Send $0 emails?" checkbox, unchecked by default, to the Send Notifications page.

https://pm.tablexi.com/issues/80435

# Screenshot
Screen Shot 2020-07-07 at 4 57 56 PM


app/views/facility_notifications/index.html.haml
```diff
@@ -10,7 +10,22 @@
   = render "shared/transactions/top", tab: "notifications"
 
 - if @order_details.any?
-  = render "shared/transactions/table", order_details: @order_details
+  #table_billing
+    = form_tag url_for(controller: params[:controller], action: :send_notifications), id: "journals_create_form", name: "journals_create_form", method: :post, class: "form-horizontal" do
```
jhanggi, line 14: Now that this is pulled up and only being used in one place, can we make it less generic? I'm thinking use `facility_notififications_send_notifications_path` rather than `url_for`. 

Also, can we change the name and id to something that fits the context (or remove them altogether)? If they're being used for CSS/JS targeting, I'll accept that it might be too much effort to be worth fixing, but if it's straightforward can we clean it up?

app/views/facility_notifications/index.html.haml
```diff
@@ -10,7 +10,22 @@
...
+      .row
```
jhanggi, line 15: It doesn't look like this view is using `:action_instructions`. Can we get rid of this here?

app/views/facility_notifications/index.html.haml
```diff
@@ -10,7 +10,22 @@
...
+        .span12= yield :action_instructions
+      .row.table-actions
+        .span1.select_all_none= select_all_link
+        #zero_dollar_orders.span2.offset8
+          = label_tag :notify_zero_dollar_orders, class: "checkbox" do
+            = check_box_tag :notify_zero_dollar_orders
+            = text("send_zero_dollar_notifications", scope: "admin.transaction_search.actions")
+        .span1
+          .submit
+            = submit_tag text(:send_notifications, scope: "admin.transaction_search.actions"), class: ['btn', 'btn-primary']
```
jhanggi, line 25: Can we be consistent in our use of symbols vs strings in the `text` method? It looks like we tend to favor strings. Also, let's use double quotes in the class (we're not always perfectly consistent right now, but that's the style we're aiming for).

config/locales/views/admin/en.transaction_search.yml
```diff
@@ -9,6 +9,7 @@ en:
       actions:
         reassign_chart_strings: Reassign Chart Strings
         send_notifications: Send Notifications
+        send_zero_dollar_notifications: Send $0 emails?
```
jhanggi, line 12: Can we double check with the client and make sure they're good with this wording? I definitely prefer this positive phrasing rather than the negative "suppress", but let's make sure NU is on board with it.

spec/system/admin/review_period_spec.rb
```diff
@@ -65,7 +98,7 @@
 
     within(".old-table tfoot") do
       expect(page).to have_selector("th", text: "Total")
-      expect(page).to have_selector("td", text: "$2.0")
+      expect(page).to have_selector("td", text: "$1.0")
```
jhanggi, line 101: I'm curious why this only had the single decimal point before. Does the test pass if we ask for `"$1.00"`?

app/support/notification_sender.rb
```diff
@@ -2,36 +2,31 @@
 
 class NotificationSender
```
jhanggi, line 3: While we're in here, can we try moving this from the `app/support` folder into the `app/services` folder? It's not a pressing need, but I have a desire to eventually be rid of the `app/support` directory. It manifests itself in the confusing `spec/support` (supporting rspec files like configuration and shared contexts) and `spec/app_support` (where we test the `app/support` classes.

So then if this moves, the affiliated spec should move as well.

app/views/facility_notifications/index.html.haml
giladshanan: Good call, I don't think we need the id/name in this context.

app/views/facility_notifications/index.html.haml
giladshanan: Good catch, this used to be `text(@order_detail_action...`, which is how I ended up with the symbol there.  I meant to clean it up but missed it.

config/locales/views/admin/en.transaction_search.yml
giladshanan: Good call, I'll check.

spec/system/admin/review_period_spec.rb
giladshanan: Yea, it passes, and so does "$1".  Looks like `have_selector` looks for elements that _contain_ the text passed in.  You can also pass in `exact: true` if you want an exact match.

app/support/notification_sender.rb
giladshanan: Oh fun - I 💯 support this

jhanggi: I forget how we ended up in this scenario, but I'm pretty sure it's my fault--at least for the spec folder. There's also a lot of stuff in `lib` that probably belongs in `app`, but it's been kind of a move it as we touch it kind of thing.

app/assets/stylesheets/app/transaction_search.scss
```diff
@@ -35,3 +35,13 @@
     text-align: right;
   }
 }
+
+#zero_dollar_orders {
+  align-items: center;
+  display: flex;
+
+  .checkbox {
```
jhanggi, line 43: I'm not sure how to make it better, but this feels overly specific, especially now that we're in full control of the partial. Is there any way we can clean it up? I would be ok if we did something with flexbox and abandoned the bootstrap grid here (flexbox wasn't a thing when nucore was first designed).

PR #2339, opened 2020-07-11
[#153177] Fix incorrect order numbers when adding forms to an existing order

giladshanan:
# Release Notes

Includes the merge_with_order_id in the order number when adding to an order with form.io or IMSERC forms.
Also used when adding to an order with ACGT/Sanger forms, but that flow isn't allowed with current permissions.

## Additional Context
I tested ACGT forms by adjusting [permissions](https://github.com/tablexi/nucore-open/blob/master/vendor/engines/sanger_sequencing/app/models/sanger_sequencing/ability.rb#L12), removing the `user: user` scoping.  With this change, it is possible to add ACGT forms to an existing order, but the user is redirected to an invalid route afterwards (`/orders/:id`).  Leaving this alone for now, assuming that this flow isn't needed.

### DIFF
--- a/app/controllers/formio/submissions_controller.rb
+++ b/app/controllers/formio/submissions_controller.rb
@@ -25,8 +25,8 @@ module Formio
     # NUcore’s SurveysController expects a referer param, which contains the page
     # to which NUcore should redirect after it processes a form submission. However,
     # the success_url that we get from UrlService does not include this param. Instead,
-    # we get it as a separate pram, so we construct the full redirect_url here to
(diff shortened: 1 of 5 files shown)


jhanggi:
@giladshanan when you merge this down, can you update the title to be something a little more client-friendly? Like "Fixes incorrect order number when completing an order form while adding to an existing order" (or maybe you can come up with something a little more concise for the title and use something like that for the main release notes).

I think the part about ACGT/Sanger probably belongs more in the "Additional context" section.

PR #2342, opened 2020-07-13
[#153876] Event log improvements

giladshanan:
# Release Notes

Includes the facility name for event logs where possible.  Also includes the price group name for soft deleted price groups.

# Screenshot

Screen Shot 2020-07-13 at 10 24 50 AM

### DIFF
--- a/app/models/log_event.rb
+++ b/app/models/log_event.rb
@@ -19,9 +19,21 @@ class LogEvent < ApplicationRecord
       start_date: start_date, end_date: end_date, events: events, query: query).search
   end
 
-# The reason that we are doing this is because in some case we can't add a default value to a text column 
+# The reason that we are doing this is because in some case we can't add a default value to a text column
   def metadata
     self[:metadata] || {}
+  end
+
+  def facility
+    if loggable.respond_to?(:facility)
+      loggable.facility
+    elsif loggable.respond_to?(:price_group_with_deleted)
+      loggable.price_group_with_deleted.facility
+    elsif loggable.respond_to?(:facilities)
+      loggable.facilities.first if loggable.facilities.count == 1
+    elsif loggable.respond_to?(:user)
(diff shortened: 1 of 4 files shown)


app/models/log_event.rb
```diff
@@ -19,11 +19,23 @@ def self.search(start_date: nil, end_date: nil, events: [], query: nil)
...
   def metadata
     self[:metadata] || {}
   end
 
+  def facility
+    if loggable.respond_to?(:facility)
+      loggable.facility
+    elsif loggable.respond_to?(:price_group_with_deleted)
+      loggable.price_group_with_deleted.facility
+    elsif loggable.respond_to?(:facilities)
```
jhanggi, line 32: I'm not really sure I like all these special cases.

I feel like there are two better options: 

- Drop this column and deal with the facility in the `to_log_s` methods (See `UserRole` for an example) 
- Delegation. I could see `PriceGroupMember` `delegate :facility, to: :price_group_with_deleted` (or it might even work as `belongs_to :facility, through: :price_group_with_deleted`

Happy to talk through this.

app/views/log_events/index.html.haml
```diff
@@ -46,6 +46,7 @@
       %th= text("event_time")
       %th= text("event")
       %th= text("object")
+      %th= "Facility"
```
jhanggi, line 49: We should use `Facility.model_name.human` here. Dartmouth always uses "Resource" instead.

app/models/log_event.rb
giladshanan: I don't like the special cases either.  Taking another look at the ticket, it looks like the request is to know the facility for price groups and journals, so I think we can simply add the delegation to `PriceGroupMember` for now.  

The models that use the special case branches are `User`, `AccountUser`, and the various `Account` subclasses. These are not mentioned in the ticket - do you think it's worth asking the client about the importance of knowing the facility for those?

PR #2360, opened 2020-07-29
[#153876] Fix error when viewing logs for account price groups

giladshanan:
# Release Notes

I think this has actually been broken since https://github.com/tablexi/nucore-open/pull/2302.

Run `rake cleanup:log_events:metadata` after merging

### DIFF
--- a/app/controllers/concerns/price_group_members_controller.rb
+++ b/app/controllers/concerns/price_group_members_controller.rb
@@ -28,7 +28,7 @@ module PriceGroupMembersController
     raise NUCore::PermissionDenied unless @price_group.can_manage_price_group_members?
 
     if price_group_member.save
-      LogEvent.log(price_group_member, :create, current_user)
+      LogEvent.log(price_group_member, :create, current_user, metadata: { member_type: price_group.member_type })
       set_flash(:notice, :create, create_flash_arguments)
     else
       set_flash(:error, :create, create_flash_arguments)
@@ -40,7 +40,7 @@ module PriceGroupMembersController
     raise NUCore::PermissionDenied unless @price_group.can_manage_price_group_members?
 
     if destroy_price_group_member!
-      LogEvent.log(price_group_member, :delete, current_user)
+      LogEvent.log(price_group_member, :delete, current_user, metadata: { member_type: price_group.member_type })
       set_flash(:notice, :destroy)
     else
       set_flash(:error, :destroy)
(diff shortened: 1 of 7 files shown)


lib/tasks/cleanup.rake
```diff
@@ -31,4 +31,15 @@ namespace :cleanup do
       puts "Price group with id: #{pg.id}, name: #{cc_name} has been destroyed."
     end
   end
+
+  namespace :log_events do
+    desc "update metadata for PriceGroupMembers"
+    task metadata: :environment do
+      events = LogEvent.where("loggable_type LIKE ?", "PriceGroupMember").where(metadata: nil)
```
jhanggi, line 38: Why not just `loggable_type: "PriceGroupMember"`?

app/controllers/concerns/price_group_members_controller.rb
```diff
@@ -28,7 +28,7 @@ def create
     raise NUCore::PermissionDenied unless @price_group.can_manage_price_group_members?
 
     if price_group_member.save
-      LogEvent.log(price_group_member, :create, current_user)
+      LogEvent.log(price_group_member, :create, current_user, metadata: { member_type: price_group.member_type })
```
jhanggi, line 31: Looks like you're trying to access `member_type` on the wrong object

```suggestion
      LogEvent.log(price_group_member, :create, current_user, metadata: { member_type: price_group_member.member_type })
```

app/controllers/concerns/price_group_members_controller.rb
```diff
@@ -40,7 +40,7 @@ def destroy
     raise NUCore::PermissionDenied unless @price_group.can_manage_price_group_members?
 
     if destroy_price_group_member!
-      LogEvent.log(price_group_member, :delete, current_user)
+      LogEvent.log(price_group_member, :delete, current_user, metadata: { member_type: price_group.member_type })
```
jhanggi, line 43:
```suggestion
      LogEvent.log(price_group_member, :delete, current_user, metadata: { member_type: price_group_member.member_type })
```

lib/tasks/cleanup.rake
```diff
@@ -31,4 +31,15 @@ namespace :cleanup do
...
+      puts "Updating #{events.size} LogEvents..."
+      events.each do |event|
```
jhanggi, line 40: Not sure how many we'll have on production, but we might want to do `find_each`

lib/tasks/cleanup.rake
giladshanan: I started out thinking there were loggable types for the subclasses, and forgot to update.  Good catch.

PR #2363, opened 2020-07-31
Add a changelog

jhanggi:
# Release Notes

Add a changelog for keeping track of breaking changes and things to be kept aware of when merging changes into the downstream forks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,48 @@
+# Changelog
+
+Because we use squash and merge, you should be able to see the changes by looking
+at the [commit log](commits/master). However, we have begun eeping track of breaking changes
+or optional rake tasks.
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -0,0 +1,48 @@
+# Changelog
+
+Because we use squash and merge, you should be able to see the changes by looking
+at the [commit log](commits/master). However, we have begun eeping track of breaking changes
+or optional rake tasks.
+
+### Welcome email refactoring/cleanup (#2362)
```
jhanggi, line 7: We could also consider adding the merge date to make it easier for downstream developers to keep track of what's new. The only problem with that is then the change would need to be a separate commit made right before merging.

I would recommend that whoever makes the PR includes notes here as part of it.

PR #2365, opened 2020-07-31
Restructure uploaded file paths and improve instructions for migrating to S3

jhanggi:
# Release Notes

Tech task: Update the file paths for uploaded files and improve instructions and helper rake takes to aid in migrating to S3.

# Breaking changes

The setting for where uploaded files are stored locally has changed from `public/files` to `public/system/:class`. You either need to change it back to its previous value or run the rake task as part of your deployment (CHANGELOG.md).

In addition, if you are using S3 already, the path for the keys should now just be `paperclip.aws_access_key` and `paperclip.aws_secret_access_key`.

# Notes

We'll need to update the access keys for NU, but their file paths in S3 already have the `:class` interpolation variable, so nothing should need to move for them.


config/settings.yml
```diff
@@ -125,12 +125,14 @@ split_accounts:
...
-# be sure to check there for configuration
+# To use S3 or another Paperclip backend, you can override the configuration in the
+# appropriate environment, e.g. config/settings/production.yml.
+# If you wish to use it for development, add it to a `config/settings/development.local.yml`
+# file (this file is gitignored).
 paperclip:
   storage: filesystem
-  url: ":rails_relative_url_root/:attachment/:id_partition/:style/:safe_filename"
-  path: ":rails_root/public/:attachment/:id_partition/:style/:safe_filename"
+  url: ":rails_relative_url_root/system/:class/:attachment/:id_partition/:style/:safe_filename"
```
jhanggi, line 134: I could see an argument that we shouldn't change the default and just make the changes downstream, but I don't like where it's currently stored.

NU is already using the `class`-prefixed paths on S3.

giladshanan (approved):
One small non-blocking comment.  Looks good!  The branch just needs to be brought up to date.

PR #2369, opened 2020-08-03
Tech task: Add AASM dependency

jhanggi:
# Release Notes

Tech task add `after_commit_everywhere` which is apparently now a non-explicit AASM dependency.

# Additional Context

Fixes this message I'm seeing after https://github.com/tablexi/nucore-open/pull/2366

```
[DEPRECATION] :after_commit AASM callback is not safe in terms of race conditions and redundant calls.
              Please add `gem 'after_commit_everywhere', '~> 0.1', '>= 0.1.5'` to your Gemfile in order to fix that.
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem "devise-encryptable"
 
 ## models
 gem "aasm"
+gem "after_commit_everywhere", "~> 0.1", ">= 0.1.5" # Needed by AASM
 gem "paperclip"
 gem "paper_trail"
 gem "awesome_nested_set"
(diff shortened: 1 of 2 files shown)


jhanggi:
I looked at the AASM source code. The message is always when you `include AASM` if the gem isn't loaded, regardless of if you're using an `after_commit` callback or not. Their [readme](https://github.com/aasm/aasm#activerecord) even says to include it as a step when using ActiveRecord.

It seems to me it would be better to only throw the warning if you're using the `after_commit` callback, but this is what they're doing.

PR #2373, opened 2020-08-05
[WIP] Upgrade to Rails 6

jhanggi:
# Release Notes

Upgrade to Rails 6.

# Additional Context

Failing tests are addressed with the devise_saml_authenticatable update in #2372. 

- [ ] There's still a deprecation on `update_attributes!` (replace with `update!`).

- [ ] Does not introduce webpacker (on by default in new Rails 6 apps, but not default for upgraded apps)

I'm not quite sure we're ready to move forward with this, but I had some free time. It may be worth trying to backport some of the deprecations and other changes so this upgrade becomes cleaner.


config/boot.rb
```diff
@@ -1,6 +1,4 @@
-# frozen_string_literal: true
```
giladshanan, line 1: It seems like you remove these magic comments and switch from double to single quotes in a few places - what's the reason for these changes?

config/environments/development.rb
```diff
@@ -14,21 +12,27 @@
...
 
     config.cache_store = :null_store
   end
 
-  config.action_mailer.raise_delivery_errors = true
+  # Store uploaded files on the local file system (see config/storage.yml for options).
+  config.active_storage.service = :local
+
+  # Don't care if the mailer can't send.
+  config.action_mailer.raise_delivery_errors = false
```
giladshanan, line 35: Is this a Rails 6 related change, or unrelated?  I'm just curious.

config/environments/production.rb
```diff
@@ -22,28 +20,25 @@
 
   # Disable serving static files from the `/public` folder by default since
   # Apache or NGINX already handles this.
-  config.public_file_server.enabled = ENV["RAILS_SERVE_STATIC_FILES"].present?
+  config.public_file_server.enabled = ENV['RAILS_SERVE_STATIC_FILES'].present?
 
-  # Compress JavaScripts and CSS.
-  config.assets.js_compressor = Uglifier.new(harmony: true)
```
giladshanan, line 28: Why did this change?

config/boot.rb
jhanggi: What I do for a rails update is run `rake app:update` and allow it to overwrite all the files with the new versions. Then I go through and look at the changes and revert as necessary. I go back and forth on what to do with these kind of changes. The Rails templates it spits out use single quotes and don't include the `frozen_string_literal` lines. 

On one hand, changing it matches our style guide (well, we don't really enforce rubocop anymore). On the other hand, it's extra noise for future Rails upgrades. This file is identical between 5.2 and 6.0 so if we had not rubocopped it before, there would be no change to the file.

I could be convinced either way. This was just where I was leaning when I did the commit.

config/environments/development.rb
jhanggi: It's based on the `app:update` default `development.rb`. I'm not sure why we had it set to `true`.

config/environments/development.rb
```diff
@@ -71,7 +76,8 @@
 
   # Use an evented file watcher to asynchronously detect changes in source code,
   # routes, locales, etc. This feature depends on the listen gem.
-  # config.file_watcher = ActiveSupport::EventedFileUpdateChecker
+  config.file_watcher = ActiveSupport::EventedFileUpdateChecker
```
jhanggi, line 79: Note: this doesn't work in docker. For docker we need to use `ActiveSupport::FileUpdateChecker`. We often use an `ENV` check to see if we're in docker mode.

config/environments/production.rb
jhanggi: This is another `app:update` template change. I'm not sure if this needs to go away or not. I think it goes away for sure if we switch to webpacker.

PR #2377, opened 2020-08-06
Tech task: switch to newer prawn-rails library

jhanggi:
# Release Notes

Tech task: switch to newer prawn-rails library.

# Additional Context

The older [prawn_rails](https://github.com/Whoops/prawn-rails) hasn't been updated since 2012. I'm amazed it still continued to work. It will no longer work in Rails 6.

The newer [prawn-rails](https://github.com/cortiz/prawn-rails) library has the same API, but is actually maintained

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -60,7 +60,7 @@ gem "rubyzip"
 
 ## controllers
 gem "prawn"
-gem "prawn_rails"
+gem "prawn-rails"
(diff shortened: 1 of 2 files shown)


jhanggi:
Yeah, for testing I generated a couple statements locally. I had originally tested this on the Rails 6 branch, but I also tested it off `master` when I backported this.

PR #2382, opened 2020-08-06
Tech task: Add `case_sensitive` flag to uniqueness validations

jhanggi:
# Release Notes

Tech task: Tech task: Add `case_sensitive` flag to uniqueness validations.

# Additional Context

Rails 6 throws a deprecation warning if this is not there. 

I left `email` as case-sensitive since there are 4 users in NU where the email is not unique. They're all users who have both a netid and non-netid account (it appears both directions: at least one is currently external and at least one is currently internal).

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -28,7 +28,7 @@ class User < ApplicationRecord
 
   validates_presence_of :username, :first_name, :last_name
   validates :email, presence: true, email_format: true
-  validates_uniqueness_of :username, :email
+  validates_uniqueness_of :username, :email, case_sensitive: false
   validates :suspension_note, length: { maximum: 255 }
 
   accepts_nested_attributes_for :account_users, allow_destroy: true
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -27,8 +27,10 @@ class User < ApplicationRecord
   has_many :user_preferences, dependent: :destroy
 
   validates_presence_of :username, :first_name, :last_name
-  validates :email, presence: true, email_format: true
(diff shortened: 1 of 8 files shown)


app/models/user.rb
```diff
@@ -27,8 +27,10 @@ class User < ApplicationRecord
   has_many :user_preferences, dependent: :destroy
 
   validates_presence_of :username, :first_name, :last_name
-  validates :email, presence: true, email_format: true
-  validates_uniqueness_of :username, :email
+  # There are existing duplicate emails in NU -- they're all users that have both a netid
+  # account and a non-netid account.
+  validates :email, presence: true, email_format: true, uniqueness: { case_sensitive: true }
+  validates :username, uniqueness: true
```
giladshanan, line 33: Do we need to specify the case sensitivity here?

spec/models/price_group_spec.rb
```diff
@@ -4,25 +4,36 @@
...
   before :each do
-    @facility = FactoryBot.create(:facility)
-    @price_group = FactoryBot.create(:price_group, facility: facility)
+    @facility = facility
+    @price_group = price_group
   end
 
-  it "should create using factory" do
-    expect(@price_group).to be_valid
+  it "is valid using the factory using factory" do
```
giladshanan, line 15:
```suggestion
  it "is valid using the factory" do
```

app/models/user.rb
giladshanan:
```suggestion
  validates :username, uniqueness: { case_sensitive: false }
```

rdunlop (approved):

PR #2384, opened 2020-08-14
[#154227] Properly handle commas in facility names when setting the "from" name in bulk email

jhanggi:
# Release Notes

Fix issue in bulk email when the facility has a comma in the name.

# Additional Context

Dartmouth was having deliveries get blocked because they would see multiple "From" addresses in the email header. This was being caused because of how the comma was being interpreted. With this change, the from header will have quotes where necessary to probably escape everything.

This was from the facility "Electron Microscopy - Irradiation, Pre-clinical Imaging and Microscopy Resource".

```
From: <Electron.Microscopy.-.Irradiation@radar-app01.dartmouth.edu>,
Pre-clinical Imaging and Microscopy Resource
<Cancer.Center.Core.Operations@Dartmouth.edu>
```

### DIFF
--- a/vendor/engines/bulk_email/app/mailers/bulk_email/mailer.rb
+++ b/vendor/engines/bulk_email/app/mailers/bulk_email/mailer.rb
@@ -14,11 +14,9 @@ module BulkEmail
     end
 
     def sender
-      if @facility.try(:single_facility?)
-        "#{@facility.name} <#{default_params[:from]}>"
-      else
-        default_params[:from]
-      end
+      address = Mail::Address.new(default_params[:from])
+      address.display_name = @facility.name if @facility.try(:single_facility?)
+      address
     end
 
   end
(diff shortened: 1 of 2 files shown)


vendor/engines/bulk_email/app/mailers/bulk_email/mailer.rb
```diff
@@ -14,11 +14,9 @@ def send_mail(recipient:, subject:, body:, reply_to: nil, facility:)
     end
 
     def sender
-      if @facility.try(:single_facility?)
-        "#{@facility.name} <#{default_params[:from]}>"
-      else
-        default_params[:from]
-      end
+      address = Mail::Address.new(default_params[:from])
```
giladshanan, line 17: Are there any other places we might want to make use of this?

giladshanan:
Mystery solved 🕵️

vendor/engines/bulk_email/app/mailers/bulk_email/mailer.rb
jhanggi: I may want to look into one of the products DC has where they have two emails separated by a semicolon. I believe it gets used in the `reply_to` of bulk email, but I'm not sure.

I know that `Mail::Address.new` would throw away the second one. I'll put in another ticket to look into that. I'd like to get this fixed since it's preventing people from receiving email.

PR #2385, opened 2020-08-14
Update validation so it works properly with expense_accounts feature disabled

jhanggi:
# Release Notes

Tech task: Update `Product#account` validation so it works properly in tests when the feature is disabled.

# Additional Context

Originally, this validation would never be added at all if the `expense_accounts` feature were disabled (like it is a UConn, and I'm guessing _was_ at UIC). Then, when we turned on the feature in the specs (see `spec/models/product_spec.rb:172`), the validation would not be run because it's not there. Now it will conditionally run based on the setting at runtime rather than compile/load time.

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -40,15 +40,13 @@ class Product < ApplicationRecord
             email_format: true,
             allow_blank: true
 
-  if SettingsHelper.feature_on? :expense_accounts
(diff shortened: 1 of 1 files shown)


jhanggi:
Good call. I was a little too trigger happy on the merge, but I'll push a change up for that

PR #2395, opened 2020-08-25
Tech task: Update default PDF paper size to letter

jhanggi:
# Release Notes

Tech task: update default paper size in PDFs to US Letter.

# Additional Context

When we upgraded the Prawn gems in #2377, that changed the default paper size to A4. Fortunately, CI caught some errors in NU because one of the tables wasn't wide enough in the new paper size.

The documentation [suggests putting this configuration into an initializer](https://github.com/cortiz/prawn-rails#default-configuration), but this is the only place we make PDFs, this seems to be a more obvious spot a developer might see to change it if they're working on making a custom statement. (A4 is a very standard paper size outside the US).

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -63,10 +63,8 @@ gem "fine_uploader", path: "vendor/engines/fine_uploader"
 gem "fullcalendar", path: "vendor/engines/fullcalendar"
 gem "rubyzip"
 
-## controllers
-gem "prawn"
+## PDF generation
 gem "prawn-rails"
-gem "prawn-table"
 
 ## other
 gem "delayed_job_active_record"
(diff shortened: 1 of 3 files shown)


Gemfile
```diff
@@ -63,10 +63,8 @@ gem "fine_uploader", path: "vendor/engines/fine_uploader"
 gem "fullcalendar", path: "vendor/engines/fullcalendar"
 gem "rubyzip"
 
-## controllers
-gem "prawn"
```
jhanggi, line 67: `prawn-rails` includes these dependencies, so we don't need to explicitly include them ourselves anymore.

PR #2401, opened 2020-09-02
TECH TASK: Update merge_describer script

giladshanan:
PR numbers in the thousands are from nucore-open, in the hundreds and below are from the forks.  At the moment NU has the most PRs, with almost 600, so I think this will last us a while.


jhanggi (commented):
While we're in here, can we update the output format a bit? As opposed to two lines, just put it on one with the link maybe in parens? I think it'll make the output easier to read.

```
* Tech task: Update merge_describer script (#2401)
```

jhanggi (approved):

PR #2413, opened 2020-09-23
[#154551] Use correct app name in mailers

giladshanan:
# Release Notes

Set the app name dynamically so that each school's emails will include the appropriate app name (CORUM, Radar. or NUCore).

### DIFF
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -57,7 +57,7 @@ class Notifier < ActionMailer::Base
     @order_detail = OrderDetail.includes(:order, :order_status, :product).find(order_detail_id)
     mail(
       to: @order_detail.order.user.email,
-      subject: "[NUcore #{@order_detail.facility.abbreviation}] Order Status Changed To: #{@order_detail.order_status.name}"
+      subject: "[#{text(:app_name)} #{@order_detail.facility.abbreviation}] Order Status Changed To: #{@order_detail.order_status.name}"
     )
   end
 
--- a/app/mailers/notifier.rb
+++ b/app/mailers/notifier.rb
@@ -53,11 +53,12 @@ class Notifier < ActionMailer::Base
     send_nucore_mail args[:user].email, text("views.notifier.statement.subject", facility: @facility)
   end
 
-  def order_detail_status_changed(order_detail_id)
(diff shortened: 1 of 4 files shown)


app/mailers/notifier.rb
```diff
@@ -57,7 +57,7 @@ def order_detail_status_changed(order_detail_id)
     @order_detail = OrderDetail.includes(:order, :order_status, :product).find(order_detail_id)
     mail(
       to: @order_detail.order.user.email,
-      subject: "[NUcore #{@order_detail.facility.abbreviation}] Order Status Changed To: #{@order_detail.order_status.name}"
+      subject: "[#{text(:app_name)} #{@order_detail.facility.abbreviation}] Order Status Changed To: #{@order_detail.order_status.name}"
```
jhanggi, line 60: Would it be worth pulling the whole string out and `text(:subject, facility: @order_detail.facility.abbreviation, status: @order_detail.order_status.name)`?

```
subject: "!app_name! %{facility} Order Status Changed To: %{status}"
```

app/mailers/notifier.rb
```diff
@@ -57,7 +57,7 @@ def order_detail_status_changed(order_detail_id)
     @order_detail = OrderDetail.includes(:order, :order_status, :product).find(order_detail_id)
```
jhanggi, line 57: Would it be worth it while we're here to switch to using the globalid version where we pass in the order detail itself and let activejob handle the lookup?

app/mailers/notifier.rb
```diff
@@ -53,11 +53,12 @@ def statement(args)
     send_nucore_mail args[:user].email, text("views.notifier.statement.subject", facility: @facility)
   end
 
-  def order_detail_status_changed(order_detail_id)
-    @order_detail = OrderDetail.includes(:order, :order_status, :product).find(order_detail_id)
+  def order_detail_status_changed(order_detail)
+    facility = order_detail.facility.abbreviation
```
jhanggi, line 57: I I think we do need the `@order_detail = order_detail` so it gets passed to the view

jhanggi (changes requested):

jhanggi (approved):

PR #2414, opened 2020-09-25
TECH TASK: Use one common email address for staging emails

giladshanan:
This will make it easier to manage forwarded staging emails from the various schools.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -43,7 +43,8 @@ email:
   from: 'noreply@example.com'
   fake:
     enabled: false
-    to: nucore@tablexi.com
+    to:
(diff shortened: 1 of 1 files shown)


config/settings.yml
```diff
@@ -43,7 +43,7 @@ email:
   from: 'noreply@example.com'
   fake:
     enabled: false
-    to:
+    to: nucore@tablexi.com
```
jhanggi, line 46: I might recommend formatting this as a list. I think that'll make the downstream merge conflicts just _a little_ easier.

```yaml
to:
  - nucore@tablexi.com
```

jhanggi (approved):

PR #2416, opened 2020-09-29
Tech Task: Bump Rails version for security

giladshanan:
Security patch: https://github.com/advisories/GHSA-cfjv-5498-mph5


jhanggi (approved):
Maybe we should look into why we didn't get a dependabot update?

giladshanan:
We got an alert: https://github.com/tablexi/nucore-open/network/alert/Gemfile.lock/actionview/open
I tried to use the "Create Dependabot security update" button there, but it failed.  There's also a bit of warning text on that page that says "Dependabot cannot update to the required version".  I suspect that Dependabot tries to update `action-view` without the rest of rails, and that's why it's failing.

PR #2420, opened 2020-10-02
[#154684] Update bulk email queries

giladshanan:
# Release Notes

*Users on Training Request List*: exclude users who have requested training on inactive products or products that no longer have an access list
*Authorized Users*: exclude users who are authorized on inactive products or products that no longer have an access list
*Customers*: exclude customers who have only purchased products that are now inactive
*Account Owners*: exclude owners whose accounts have only been used to purchase that are now inactive

Also refactors the `RecipientSearcher` queries to make them more consistent and easier to understand.

### DIFF
--- a/app/views/product_users/index.html.haml
+++ b/app/views/product_users/index.html.haml
@@ -19,7 +19,7 @@
 - elsif @product_users.empty?
   %p.notice= text("none")
 - else
-  - if @product.respond_to?(:product_access_groups) && @product.product_access_groups.any?
+  - if @product.has_access_list?
     = form_for @product, url: [current_facility, @product, :update_restrictions], method: :put do |f|
       = render "table", f: f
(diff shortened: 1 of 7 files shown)


app/views/product_users/index.html.haml
```diff
@@ -19,7 +19,7 @@
 - elsif @product_users.empty?
   %p.notice= text("none")
 - else
-  - if @product.respond_to?(:product_access_groups) && @product.product_access_groups.any?
+  - if @product.has_access_list?
```
jhanggi, line 22: I feel like this name is misleading. This isn't checking for an access list (that's `requires_approval?`). it's checking for product access groups (which are related to schedule rules, so only instruments and rooms have them). I like the DRYing here, but maybe we should rename it? `has_product_access_groups?` perhaps?

jhanggi (approved):
I had one thought about an already-existing method name, but otherwise looks good.

PR #2426, opened 2020-10-12
[#154764] Fix issue where quantity was not being updated

jhanggi:
# Release Notes

Fix an issue when ordering on behalf if you update the quantity field, the value of quantity nor note and reference id are not persisted.

# Additional Context

To replicate:
* Add an item with a note to the cart
* Fill in the note and reference id fields
* Change the quantity field
* Click "Purchase"
* You will see the "Quantities have change. Please review" message, but the note, quantity, and reference fields all revert to their previous values.
* Clicking Update does the same thing

It always reverted to it's previous vlue:
Put in 3, click update -> reverts to 1
Put in 4, click update -> field shows 3

## Tech notes

While I had no problem replicating this in the browser, I struggled to get a failing spec. I'm not sure what the difference in behavior is. I left in the specs, but they were still passing before the other changes. I eventually gave up.

Ultimately, the issue is the order details getting updated are not the same objects as those that were getting displayed. So when you submitted the form, the it would read the database values into one set of objects, update another, and then update the first set of objected.

Both of the changes: the `inverse_of` and the `select` fixed the issue on its own, but it felt like both were reasonable things to put in.

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -239,11 +239,12 @@ class OrdersController < ApplicationController
       @order.assign_attributes(order_params)
 
       if OrderDetailUpdater.new(@order, order_update_params).update
-        # Must show instead of render to maintain "more options" state when
+        # Must render instead of redirect to maintain "more options" state when
         # ordering on behalf of
+        flash.now[:notice] = "Cart has been updated"
       else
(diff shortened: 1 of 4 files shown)


giladshanan (approved):
Strange that the spec isn't catching this.  LGTM.

PR #2431, opened 2020-10-16
[#155076] Use html helper in place of the view hook for welcome helper text

giladshanan:
# Release Notes

A view hook is more than we need for the wording change we need to make in the UMass fork.

See https://github.com/tablexi/nucore-umass/pull/82

### DIFF
--- a/app/views/facilities/index.html.haml
+++ b/app/views/facilities/index.html.haml
@@ -1,7 +1,7 @@
 - if SettingsHelper.feature_on?(:equipment_list)
   = render "shared/nav/equipment_list_nav"
 
-.alert.alert-info= html(".welcome")
+.alert.alert-info= html("welcome")
 
 - if azlist_on?
   .az_list_container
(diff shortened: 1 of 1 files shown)


app/views/facilities/index.html.haml
```diff
@@ -1,9 +1,7 @@
 - if SettingsHelper.feature_on?(:equipment_list)
   = render "shared/nav/equipment_list_nav"
 
-.alert.alert-info
-  = text(".welcome")
-  = render_view_hook("after_welcome")
+.alert.alert-info= html(".welcome")
```
jhanggi, line 4: I'm not sure if it's important or not, but I believe that `html` will be wrapped in a `div` or a `p`. Can you double check and confirm no extra padding or anything gets added here from the change. If it's a problem, you should be able to use `html(".welcome", inline: true)`.

Also, we tend to not use the leading dot when using the text helper (rather than the rails helper `t`). Not a big deal, though.

jhanggi (approved):

app/views/facilities/index.html.haml
giladshanan: I tested this actually, using either `text` or `html` there's no visual difference between them.  I can clean up that `.` though.

giladshanan:
@jhanggi it looks like NU uses an override for this as well, so each school has its own version of the welcome text.

PR #2450, opened 2020-11-16
[#153160] Add a CSV export for global user roles

giladshanan:
# Release Notes

In support of https://github.com/tablexi/nucore-nu/pull/602

Add view hooks for additional user attributes on the Global User Roles page
Add an Export to CSV function on the Global User Roles page


vandrijevik (approved):
:shipit:

PR #2497, opened 2021-01-26
[#156326] Handle reservation save issues

giladshanan:
# Release Notes

When adding actual times to an order detail, it's possible for the reservation update to fail silently.  This results in the order detail being moved out of the problem queue even though the reservation record is still missing actual times.

Also:
- Adds paper_trail tracking to the Reservation model.  
- Adds an `object_changes` column to the Versions table to make working with paper trail easier
- Remove temporary code.

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -321,7 +321,7 @@ class Reservation < ApplicationRecord
   end
 
   def has_actuals?
-    actual_start_at.present? && actual_end_at.present?
+    actual_start_at.present? && actual_end_at.present? && actual_end_at > actual_start_at
   end
 
   def started?
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -5,6 +5,7 @@ require "date"
 class Reservation < ApplicationRecord
 
   acts_as_paranoid # soft deletes
+  has_paper_trail
 
   include DateHelper
   include Reservations::DateSupport
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -379,9 +379,8 @@ class Reservation < ApplicationRecord
     self.billable_minutes = calculated_billable_minutes
   end
 
-  # FIXME: Temporary override to include reconciled orders, so we can backfill them
   def calculated_billable_minutes
-    if (order_detail&.complete? || order_detail&.reconciled?) && order_detail&.canceled_at.blank? && price_policy.present?
+    if order_detail&.complete? && order_detail&.canceled_at.blank? && price_policy.present?
       case price_policy.charge_for
       when InstrumentPricePolicy::CHARGE_FOR.fetch(:reservation)
         TimeRange.new(reserve_start_at, reserve_end_at).duration_mins
--- a/app/services/order_details/param_updater.rb
+++ b/app/services/order_details/param_updater.rb
@@ -70,7 +70,9 @@ class OrderDetails::ParamUpdater
 
     @order_detail.transaction do
       assign_price_changed_by_user
-      @order_detail.reservation.save_as_user(@editing_user) if @order_detail.reservation
+      if @order_detail.reservation
+        log_and_rollback unless @order_detail.reservation.save_as_user(@editing_user)
+      end
       if order_status_id && order_status_id.to_i != @order_detail.order_status_id
         change_order_status(order_status_id, @options[:cancel_fee]) || raise(ActiveRecord::Rollback)
       else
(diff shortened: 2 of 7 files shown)


app/services/order_details/param_updater.rb
```diff
@@ -84,12 +86,17 @@ def update_attributes(params)
...
   def trigger_notifications
-    OrderDetails::DisputeResolvedNotifier.new(@order_detail,current_user: @editing_user).notify
+    OrderDetails::DisputeResolvedNotifier.new(@order_detail, current_user: @editing_user).notify
     OrderDetails::AssignmentNotifier.new(@order_detail).notify
   end
 
   private
 
+  def log_and_rollback
+    logger.error("errors for reservation #{@order_detail.reservation.id}: #{@order_detail.reservation.errors.full_messages}")
```
jhanggi, line 96: Would it be worth using the ActiveSupport notifications? That should trigger a notification to rollbar, which would actively inform us rather than us thinking to check the logs.

```ruby
ActiveSupport::Notifications.instrument("background_error", exception: e, information: "Errors for reservation #{@order_detail.reservation.id}: #{@order_detail.reservation.errors.full_messages}")
```

app/services/order_details/param_updater.rb
```diff
@@ -70,7 +70,9 @@ def update_attributes(params)
 
     @order_detail.transaction do
       assign_price_changed_by_user
-      @order_detail.reservation.save_as_user(@editing_user) if @order_detail.reservation
+      if @order_detail.reservation
+        log_and_rollback unless @order_detail.reservation.save_as_user(@editing_user)
```
jhanggi, line 74: Would it be worth adding a comment here? Or do you think a `git blame` will be enough for future us.

jhanggi (commented):
Looks good, but I had one thought on the logging

app/services/order_details/param_updater.rb
giladshanan: Can you think of any scenario where a rollback here would cause problems?  This is going to run every time an order details is updated via `OrderManagement::OrderDetailsController#update` so it will no longer be possible to update an order detail if the reservation is invalid.

jhanggi: I'm not sure. There's probably a situation where it's a problem, but if the reservation is invalid then it seems we have another problem. I'm inclined towards leaving the rollback in--assuming it presents some error notification to the user--because then the user will know that the updates to the reservation times didn't work.

And hopefully, we'll see the notification and get enough info from it to finally figure out what's going wrong with these reservations. Having the user aware there was an issue could provide us additional insight as well.

app/services/order_details/param_updater.rb
```diff
@@ -84,12 +86,19 @@ def update_attributes(params)
...
-    OrderDetails::DisputeResolvedNotifier.new(@order_detail,current_user: @editing_user).notify
+    OrderDetails::DisputeResolvedNotifier.new(@order_detail, current_user: @editing_user).notify
     OrderDetails::AssignmentNotifier.new(@order_detail).notify
   end
 
   private
 
+  def log_and_rollback
+    # TODO: Determine why reservations are failing to save at this point, then remove logging.
+    ActiveSupport::Notifications.instrument("background_error", exception: e, information: "Failed save for reservation #{@order_detail.reservation.id}: #{@order_detail.reservation.errors.full_messages}")
```
jhanggi, line 97: Oh, we don't have an exception here. The undefined `e` is causing test failures.

Also, we should definitely do some sort of test to verify that `information` is ending up in rollbar.

giladshanan: I looked at this and decided to stick with just logging to rollbar directly.  The `information` key is used by `ExceptionNotifier`, not Rollbar.

app/models/reservation.rb
```diff
@@ -320,7 +321,7 @@ def valid_before_purchase?
   end
 
   def has_actuals?
-    actual_start_at.present? && actual_end_at.present?
+    actual_start_at.present? && actual_end_at.present? && actual_end_at > actual_start_at
```
jhanggi, line 324: Just curious, what is this fixing?

jhanggi (approved):

app/models/reservation.rb
giladshanan: In testing this I realized that if a user enters "0" for the actual duration, the reservation fails validations and moves the order detail out of the problem queue.  We've got 2 instances of ` "actual_duration_mins"=>"0"` updates in the logs at [2021-01-13T12:54:24.636039 #10142] and [2021-01-13T12:54:27.001807 #10154].  I think this is the "smoking gun".

I had to adjust the `has_actuals?` method to handle this specific case.  Otherwise the `@order_detail` won't have the proper `problem_description_key`.  Check out `Reservation#problem_description_key` and `#requires_but_missing_actuals?`

giladshanan:
I'll take a look at the failing spec tomorrow

PR #2507, opened 2021-02-03
[#152645] Handle LDAP auth when LdapAuthentication gem isn't installed

giladshanan:
# Release Notes

Only NU uses LDAP authentication.

### DIFF
--- a/app/services/users/auth_checker.rb
+++ b/app/services/users/auth_checker.rb
@@ -21,11 +21,15 @@ module Users
 
       if @user.authenticated_locally?
         @user.valid_password?(@password)
-      elsif defined?(LdapAuthentication) && LdapAuthentication.configured?
+      elsif ldap_enabled?
         @user.valid_ldap_authentication?(@password)
       elsif Settings.saml.present?
         # TODO
       end
+    end
+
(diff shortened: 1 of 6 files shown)


app/services/users/auth_checker.rb
```diff
@@ -21,7 +21,7 @@ def authenticated?
 
       if @user.authenticated_locally?
         @user.valid_password?(@password)
-      elsif LdapAuthentication.configured?
+      elsif defined?(LdapAuthentication) && LdapAuthentication.configured?
```
vandrijevik, line 24: Instead of resorting to `defined?`, should `uses_ldap_authentication` maybe become a configuration setting on `Rails.application.config` that defaults to `false`, and we set to `true` only for NU? It seems like it would be easier to manage in tests, too (we can change that boolean at will to exercise the code path we’re interested in, without any fake classes just to have a constant present).

vandrijevik (approved):

app/services/users/auth_checker.rb
giladshanan: I like the idea, just moved the configuration to `settings.yml`.

PR #2523, opened 2021-02-19
[#153934] Use a view hook instead of dynamic template lookup

giladshanan:
# Release Notes

Using a ViewHook here for consistency.


vendor/engines/c2po/app/views/c2po/facility_accounts/show/_additional_account_fields.html.haml
```diff
@@ -0,0 +1,5 @@
+- if @account.is_a?(CreditCardAccount)
```
rdunlop, line 1: I think that this should be `account` instead. (since we're passing it this way).

rdunlop (approved):

PR #2524, opened 2021-02-22
[#156362] Add hourly refersh for My Reservations Page

giladshanan:
# Release Notes

Adds additional logging for reservation switch failures, and an hourly refresh to the My Reservations page.

The refresh is intended to prevent a possible edge case:
* The user starts a reservation, and leaves the "My Reservations" tab open with an "End Reservation" link on it
*  The user doesn't end the reservation, and 12 hours later the reservation is moved to the problem queue
*  The user clicks the stale "End Reservation" link
*  The user will not see an error, and may think they have ended their reservation

### DIFF
--- a/app/assets/javascripts/page_refresh.js
+++ b/app/assets/javascripts/page_refresh.js
@@ -1,5 +1,5 @@
 /***
- * Triggers a full refreseh of the page at a specified internval.
+ * Triggers a full refresh of the page at a specified interval.
  *
  * Usage:
    = content_for :head_content do
--- a/app/assets/javascripts/page_refresh.js
+++ b/app/assets/javascripts/page_refresh.js
@@ -1,0 +1,20 @@
+/***
+ * Triggers a full refreseh of the page at a specified internval.
+ *
+ * Usage:
+   = content_for :head_content do
+     = javascript_include_tag "page_refresh"
+
+   .js--pageRefresh{ data: { refresh_interval: 1.hour } }
+***/
+
+document.addEventListener("DOMContentLoaded", function() {
+  const page = document.getElementsByClassName("js--pageRefresh")[0];
+  const refreshInterval = page.dataset["refreshInterval"] || 60; // in seconds
+
+  function refresh() {
+    window.location.reload()
+  }
+
+  window.setInterval(refresh, refreshInterval * 1000);
+});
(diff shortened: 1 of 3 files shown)


app/views/reservations/list.html.haml
```diff
@@ -1,3 +1,8 @@
+= content_for :head_content do
```
rdunlop, line 1: Technically, this doesn't need to be `= content_for`, but could be `- content_for`, but I see that this is a pattern which is pretty consistently applied to this codebase.

rdunlop (approved):

app/assets/javascripts/page_refresh.js
```diff
@@ -0,0 +1,20 @@
+/***
+ * Triggers a full refreseh of the page at a specified internval.
```
KaraAJC, line 2:
```suggestion
 * Triggers a full refresh of the page at a specified interval.
```

app/assets/javascripts/page_refresh.js
```diff
@@ -0,0 +1,20 @@
...
+ * Usage:
+   = content_for :head_content do
+     = javascript_include_tag "page_refresh"
+
+   .js--pageRefresh{ data: { refresh_interval: 1.hour } }
+***/
+
+document.addEventListener("DOMContentLoaded", function() {
+  const page = document.getElementsByClassName("js--pageRefresh")[0];
+  const refreshInterval = page.dataset["refreshInterval"] || 60; // in seconds
```
KaraAJC, line 13: for accessible purposes, it's useful to communicate to the user beforehand that there is a timeout for this activity. https://knowbility.org/blog/2018/WCAG21-226Timeouts/

KaraAJC (commented):
comments/typos to review, otherwise approve! 
@giladshanan

app/assets/javascripts/page_refresh.js
giladshanan: This is a good point, but the refresh is indirectly related.  The page contains links that allow users to take action on their reservations, and the opportunity to make changes expires 12 hours after the reservation end time.  This is true whether or not the page refreshes.  I could see adding a separate ticket to make this window more transparent, but there's enough complexity there is quite a but of complexity around when reservations can be updated so I'd like to address it on its own.

PR #2535, opened 2021-03-04
TECH TASK: docker-compose config update

KaraAJC:
# Release Notes

Currently, there is not a context set for the mailcatcher build, so when running docker-compose is defaults to the production docker-file, which is not what's needed for a local environment setup.


wangminghao (approved):
LGTM

giladshanan (approved):
👍🏼

PR #2542, opened 2021-03-08
[#156645] Upgrade ruby to 2.6.6

giladshanan:
# Release Notes

Ruby 2.5.5 is nearing the end of support.


Gemfile.lock
```diff
@@ -669,7 +669,7 @@ DEPENDENCIES
   will_paginate
 
 RUBY VERSION
-   ruby 2.5.5p157
+   ruby 2.6.3p62
```
rdunlop, line 672: This is unexpected. Why isn't it 2.6.6p146?

rdunlop (changes requested):

Gemfile.lock
giladshanan: Good catch, thanks

rdunlop (approved):
Don't forget to upgrade the ruby version in `infrastructure`

PR #2544, opened 2021-03-11
[#156687] Manage auto_cancel process via secrets.yml

giladshanan:
# Release Notes

The `ENV["HOSTNAME"]` is not set when the eye config gets loaded.  Instead, use the value of `run_auto_cancel` to control which instances run the auto_cancel process.

NOTE:  I've already set the value on the server.

### DIFF
--- a/config/eye.yml.erb
+++ b/config/eye.yml.erb
@@ -56,8 +56,8 @@ processes:
       user_commands:
         rotate: "kill -USR1 {PID}"
 
-# This should only run on 1 instance.
-# See run_auto_cancel in secrets.yml
+  # This should only run on 1 instance.
+  # See run_auto_cancel in secrets.yml
   - name: auto_cancel
     config:
       start_command: ruby lib/daemons/auto_cancel.rb start -- --no-monitor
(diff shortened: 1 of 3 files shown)


config/eye.yml.erb
```diff
@@ -58,7 +56,8 @@ processes:
       user_commands:
         rotate: "kill -USR1 {PID}"
 
-<% if is_first_host %>
+# This should only run on 1 instance.
```
rdunlop, line 59: I don't think that you can/should indent this at this level. Otherwise, it may not be correct yaml.

rdunlop (changes requested):

config/eye.yml.erb
```diff
@@ -58,7 +56,8 @@ processes:
...
+# See run_auto_cancel in secrets.yml
```
rdunlop, line 60:
```suggestion
  # This should only run on 1 instance.
  # See run_auto_cancel in secrets.yml
```

giladshanan: This is valid, but I think it should be indented

rdunlop (approved):

PR #2558, opened 2021-03-23
[#156428] CSV Export fixes

giladshanan:
# Release Notes

New/In Process orders haven't been fulfilled yet so:
- Use ordered_at instead of fulfilled_at dates.
- Costs here will mostly be estimated (or "Unassigned"), so cost headers should be clearly labeled with "Estimated".  In some cases the cost may be actual, but we are ok considering that a "very good estimate".

### DIFF
--- a/app/controllers/concerns/new_inprocess_controller.rb
+++ b/app/controllers/concerns/new_inprocess_controller.rb
@@ -14,7 +14,7 @@ module NewInprocessController
                                               TransactionSearch::DateRangeSearcher).search(order_details, @search_form)
     @order_details = @search.order_details.includes(:order_status).joins_assigned_users.reorder(sort_clause)
     @date_range_field = @search_form.date_params[:field]
-    @use_estimated_cost_header = true
+    @use_estimated_cost_label = true
 
     respond_to do |format|
       format.html { @order_details = @order_details.paginate(page: params[:page]) }
--- a/app/controllers/concerns/new_inprocess_controller.rb
+++ b/app/controllers/concerns/new_inprocess_controller.rb
@@ -13,6 +13,8 @@ module NewInprocessController
                                               TransactionSearch::OrderStatusSearcher,
                                               TransactionSearch::DateRangeSearcher).search(order_details, @search_form)
     @order_details = @search.order_details.includes(:order_status).joins_assigned_users.reorder(sort_clause)
+    @date_range_field = @search_form.date_params[:field]
+    @use_estimated_cost_header = true
 
     respond_to do |format|
       format.html { @order_details = @order_details.paginate(page: params[:page]) }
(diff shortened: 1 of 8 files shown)


app/controllers/concerns/new_inprocess_controller.rb
```diff
@@ -13,6 +13,8 @@ def index
                                               TransactionSearch::OrderStatusSearcher,
                                               TransactionSearch::DateRangeSearcher).search(order_details, @search_form)
     @order_details = @search.order_details.includes(:order_status).joins_assigned_users.reorder(sort_clause)
+    @date_range_field = @search_form.date_params[:field]
+    @use_estimated_cost_header = true
 
     respond_to do |format|
       format.html { @order_details = @order_details.paginate(page: params[:page]) }
```
rdunlop, line 20: I think that the html view's header label(s) (2 controllers) may need to be adjusted too.

rdunlop (changes requested):

app/models/reports/account_transactions_report.rb
```diff
@@ -60,9 +61,9 @@ def headers
       Reservation.human_attribute_name(:actual_end_at),
       OrderDetail.human_attribute_name(:quantity),
       OrderDetail.human_attribute_name(:user),
-      OrderDetail.human_attribute_name(:cost),
-      OrderDetail.human_attribute_name(:subsidy),
-      OrderDetail.human_attribute_name(:total),
+      actual_or_estimated_label_for(OrderDetail.human_attribute_name(:cost)),
```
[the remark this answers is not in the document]
giladshanan: Yes.  The amounts here should always be estimates, but even if they are the actual price, that would be a very accurate estimate.  I talked this through with the team at NU and they are comfortable with changing the headers here.

app/controllers/concerns/new_inprocess_controller.rb
giladshanan: The views use a generic label of "Price".  The costs listed are green for actual, red for estimated, or display "Unassigned" if unassigned.  I think it's worth adding a comment but we don't need to change the header text.

app/models/reports/account_transactions_report.rb
```diff
@@ -60,9 +61,9 @@ def headers
       Reservation.human_attribute_name(:actual_end_at),
       OrderDetail.human_attribute_name(:quantity),
       OrderDetail.human_attribute_name(:user),
-      OrderDetail.human_attribute_name(:cost),
-      OrderDetail.human_attribute_name(:subsidy),
-      OrderDetail.human_attribute_name(:total),
+      OrderDetail.human_attribute_name("#{@label_key_prefix}_cost"),
```
giladshanan, line 64: See https://github.com/tablexi/nucore-open/blob/master/config/locales/en.models.yml#L329

rdunlop (approved):

PR #2559, opened 2021-03-24
[#155426] Lockout password-based user after 5 failed login attempts

wangminghao:
https://pm.tablexi.com/issues/155426

- Devise set up for locking accounts with custom error message
- Set User as lockable
- Create LogEvent when User becomes locked

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -200,6 +200,5 @@ class User < ApplicationRecord
   def lock_access!(opts = { })
     super
     LogEvent.log(self, :access_locked, self, metadata: {cause: "Too many login attempts"})
-    puts "\n========= HELLO"
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -6,7 +6,7 @@ class User < ApplicationRecord
   include Nucore::Database::WhereIdsIn
 
   # ldap_authenticatable is included via a to_prepare hook if ldap is enabled
-  devise :database_authenticatable, :encryptable, :trackable, :recoverable
+  devise :database_authenticatable, :encryptable, :trackable, :recoverable, :lockable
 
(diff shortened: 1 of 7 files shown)


app/models/user.rb
```diff
@@ -197,4 +197,9 @@ def create_default_price_group!
     price_group_members.find_or_create_by!(price_group: default_price_group)
   end
 
+  def lock_access!(opts = { })
+    super
+    LogEvent.log(self, :access_locked, self, metadata: {cause: "Too many login attempts"})
+    puts "\n========= HELLO"
```
rdunlop, line 203: probably don't need this

rdunlop (commented):
I seem to recall discussion with the client that "doing a password reset should clear the locked-out state" so that if they did a reset-password, they could log in immediately.

Is this possible? desired?

wangminghao:
> I seem to recall discussion with the client that "doing a password reset should clear the locked-out state" so that if they did a reset-password, they could log in immediately.
> 
> Is this possible? desired?

My impression was that the desired behavior was the "normal" devise behavior, but I don't have the client context. @giladshanan do you have insight into this?

giladshanan:
The goal here is to improve security with minimal effort and impact on users.  We have a couple of options here:
- the user waits 5 min
- use the `:email` or `:both` unlock strategy, which are unlocked by password reset (https://github.com/heartcombo/devise/issues/2090) 

I will check in with the client this afternoon.

giladshanan:
The client is in favor of the `:email` unlock strategy instead of the 5 min cool off.

giladshanan (changes requested):
One small request, looks good!

rdunlop (approved):

rdunlop (approved):

giladshanan (approved):

PR #2566, opened 2021-04-02
[#156428] Use string instead of symbol

giladshanan:
Fixes: ActiveJob::SerializationError (Unsupported argument type: Symbol)


rdunlop (approved):
Is there any way to add a spec for this? would a OrderDetailCsvExport spec expose this issue?

PR #2572, opened 2021-04-14
[#155473] Increase password complexity requirements

KaraAJC:
# Release Notes

This PR implements `devise-security` gem which increases the password length to minimum 10 characters and complexity rules, which require one uppercase and lowercase letter, one digit, and one symbol.

This change will apply to new or updated email-based users. Existing users will not be affected until they update their password. 

This PR also refactors the `User` factory to reflect netid users by default.


app/models/user.rb
```diff
@@ -89,7 +97,7 @@ def update_password(params)
     end
 
     errors.add(:password, :empty) if params[:password].blank?
-    errors.add(:password, :password_too_short) if params[:password] && params[:password].strip.length < 6
+    errors.add(:password, :password_too_short) if params[:password] && params[:password].strip.length < 11
```
rdunlop, line 100: this is confusing. if the length is 10...isn't that a long-enough password? (based on `config.password_length = 10..255`, it implies that 10 is a valid length)

rdunlop (changes requested):
I see in the readme: "If you seeded the demo data, you can log in with admin@example.com/password".

Is this also going to be affected by these changes?

Gemfile
```diff
@@ -24,6 +24,7 @@ gem "mysql2"
 gem "cancancan"
 gem "devise"
 gem "devise-encryptable"
+gem 'devise-security'
```
giladshanan, line 27:
```suggestion
gem "devise-security"
```

app/models/user.rb
giladshanan: The minimum password length should be 10
```suggestion
    errors.add(:password, :password_too_short) if params[:password] && params[:password].strip.length < 10
```

giladshanan:
@KaraAJC just a few small requests.  There's also a merge conflict and the readme should get updated from "password" to whatever we're setting the demo_seed.

spec/models/user_spec.rb
```diff
@@ -77,15 +77,12 @@
     end
   end
 
-  it { is_expected.to be_authenticated_locally }
+  it { is_expected.not_to be_authenticated_locally }
 
-  it "can not be locally authenticated" do
-    user.encrypted_password = nil
+  it "can be locally authenticated as an email user" do
```
giladshanan, line 82:
```suggestion
  it "can be locally authenticated as a netid user (email_user? => false)" do
```

app/models/user.rb
```diff
@@ -53,6 +53,14 @@ class User < ApplicationRecord
   scope :with_recent_orders, ->(facility) { distinct.joins(:order_details).merge(OrderDetail.recent.for_facility(facility)) }
   scope :sort_last_first, -> { order(Arel.sql("LOWER(users.last_name), LOWER(users.first_name)")) }
 
+  # Password validations are required only for external users. This method overrides default logic in Devise-Security Gem.
```
giladshanan, line 56:
```suggestion
  # Password validations are required only for email users. This method overrides default logic in the devise-security gem.
  # See https://github.com/devise-security/devise-security/blob/master/lib/devise-security/models/secure_validatable.rb#L54
```

giladshanan (approved):

PR #2574, opened 2021-04-21
Prevent invalid query parameters from causing exceptions

rdunlop:
# Release Notes

Introduce rack-utf8_sanitizer in order to prevent malicious/invalid query parameters.

This should prevent our exception-monitoring from having many results if/when an outside system scans.


giladshanan (changes requested):
One small change

giladshanan (approved):
👍🏼

PR #2575, opened 2021-04-22
[#155203] Allow resending statement emails

giladshanan:
# Release Notes

Adds a "Resend" link to each statement in the Statement History tab.

# Screenshot

Screen Shot 2021-04-22 at 6 06 20 PM

### DIFF
--- a/app/controllers/facility_statements_controller.rb
+++ b/app/controllers/facility_statements_controller.rb
@@ -72,7 +72,7 @@ class FacilityStatementsController < ApplicationController
     redirect_to action: :new
   end
 
-  # GET /facilities/:facility_id/statements/:id/resend_emails
+  # POST /facilities/:facility_id/statements/:id/resend_emails
   def resend_emails
     if SettingsHelper.feature_on?(:send_statement_emails)
       statement = Statement.find(params[:id])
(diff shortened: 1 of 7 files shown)


app/controllers/facility_statements_controller.rb
```diff
@@ -72,6 +72,16 @@ def create
     redirect_to action: :new
   end
 
+  # GET /facilities/:facility_id/statements/:id/resend_emails
```
rdunlop, line 75: I don't think that this should be a "GET" action, because it has side-effects (sending an e-mail)

app/views/shared/_statements_table.html.haml
```diff
@@ -22,6 +22,9 @@
             %br
             - path = current_facility ? statement_path(s) : account_statement_path(s.account, s, format: :pdf)
             = link_to t("statements.pdf.download"), path
+            - if SettingsHelper.feature_on?(:send_statement_emails) && current_facility
+              %br
+              = link_to(t("statements.resend"), resend_emails_facility_statement_path(current_facility, s))
```
rdunlop, line 27: Can we make this a `put` or ` post` (because this link has a side-effect).

rdunlop (changes requested):

app/views/shared/_statements_table.html.haml
giladshanan: @rdunlop I couldn't find any useful guidance for an action like this that sends an email.  We're not altering records, which is why I went with `get`.  Do you prefer `put` or `post` here?

rdunlop (approved):

PR #2577, opened 2021-04-25
[#155473] Fix password complexity logic

giladshanan:
# Release Notes

When merging #2572 into NU I discovered a failing spec that revealed password complexity validations were being run when updating user info for external (email-based) users.  Failing NU PR: https://github.com/tablexi/nucore-nu/pull/632

I also added a hint clarifying the new complexity requirements.

# Screenshot

Screen Shot 2021-04-25 at 4 37 34 PM
Screen Shot 2021-04-25 at 4 38 01 PM


app/models/user.rb
```diff
@@ -53,13 +53,22 @@ class User < ApplicationRecord
   scope :with_recent_orders, ->(facility) { distinct.joins(:order_details).merge(OrderDetail.recent.for_facility(facility)) }
   scope :sort_last_first, -> { order(Arel.sql("LOWER(users.last_name), LOWER(users.first_name)")) }
 
-  # Password validations are required only for email users. This method overrides default logic in the devise-security gem.
+  # This method is only used by devise-security to determine
+  # whether or not password validations should run.
+  # A better name would be passowrd_validations_required?
```
rdunlop, line 58:
```suggestion
  # A better name would be password_validations_required?
```

app/models/user.rb
```diff
@@ -53,13 +53,22 @@ class User < ApplicationRecord
...
+  # but that's just my humble opinion.
   # See https://github.com/devise-security/devise-security/blob/master/lib/devise-security/models/secure_validatable.rb#L54
+  #
+  # By default passwords are validated for new records,
+  # or when the password or password_confirmation are being set.
+  # We over-ride here because we only want to run validations
```
rdunlop, line 64:
```suggestion
  # We override here because we only want to run validations
```

rdunlop (approved):

app/views/devise/registrations/edit.html.haml
```diff
@@ -1,3 +1,4 @@
+- # This file is currently unused, we aren't using Devise#registerable
```
wangminghao, line 1: This file is currently unused in nucore, but may be used downstream...

wangminghao (approved):
open question about devise Registerable views but those are not blockers.

KaraAJC (approved):

PR #2582, opened 2021-04-30
Tech task: Allow setting encryption options

giladshanan:
Allow setting optional encryption method in `ldap.yml`. Valid values: simple_tls or start_tls

### DIFF
--- a/config/ldap.yml.template
+++ b/config/ldap.yml.template
@@ -13,3 +13,4 @@ development:
   admin_user: "cn=admin,dc=example,dc=org"
   admin_password: admin
   additional_user_attributes: [] # attributes to update on log in
+  encryption_method: "" # either simple_tls or start_tls
--- a/config/ldap.yml.template
+++ b/config/ldap.yml.template
@@ -13,4 +13,4 @@ development:
   admin_user: "cn=admin,dc=example,dc=org"
   admin_password: admin
   additional_user_attributes: [] # attributes to update on log in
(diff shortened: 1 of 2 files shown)


vendor/engines/ldap_authentication/lib/ldap_authentication.rb
```diff
@@ -39,6 +41,18 @@ def self.additional_user_attributes
     config.fetch("additional_user_attributes", [])
   end
 
+  def self.additional_user_attributes
```
rdunlop, line 44: this looks like a bad merge..duplicating the lines above it.

rdunlop (changes requested):

config/ldap.yml.template
```diff
@@ -13,3 +13,4 @@ development:
   admin_user: "cn=admin,dc=example,dc=org"
   admin_password: admin
   additional_user_attributes: [] # attributes to update on log in
+  encryption_method: "" # either simple_tls or start_tls
```
giladshanan, line 16:
```suggestion
  encryption_method: "" # optional - valid options are either simple_tls, start_tls, or ''
```

rdunlop (approved):

joshea0:
I realize this is already merged, but I just took a look and this looks great! Thanks @giladshanan for figuring this out. This is a big improvement from my POV

giladshanan:
@joshea0 sure thing! I'm glad we found a way to get this working, and this will be more flexible for future use as well.

PR #2587, opened 2021-05-06
[#144829] Display Date Added on product access lists

wangminghao:
Added new "Date added" column to the User access list for Product and likewise on the Product access list for Users. This column reads from the existing `ProductUser.approved_at` field and indicates when a person was granted access to a restricted product.

User Access List:
user-access-list

Product Access List
product-access-list

### DIFF
--- a/app/helpers/date_helper.rb
+++ b/app/helpers/date_helper.rb
@@ -49,6 +49,10 @@ module DateHelper
     date.strftime("%B %e, %Y")
   end
 
+  def month_day_year(date)
+    date.strftime("%m/%d/%Y")
+  end
+
   def human_time(dt)
     return nil if dt.nil?
     begin
--- a/app/views/product_users/_table.html.haml
+++ b/app/views/product_users/_table.html.haml
@@ -21,7 +21,7 @@
         %td
           = link_to user.last_first_name, [current_facility, user]
 
-        %td= product_user.approved_at.strftime("%Y/%M/%d")
+        %td= month_day_year(product_user.approved_at)
         %td= user.username
         %td= user.email
         - if local_assigns[:f]
--- a/app/views/product_users/_table.html.haml
+++ b/app/views/product_users/_table.html.haml
@@ -3,7 +3,7 @@
     %tr
       %th
       %th= User.human_attribute_name(:name)
-      %th= User.human_attribute_name(:date_added)
+      %th= ProductUser.human_attribute_name(:date_added)
       %th= User.human_attribute_name(:username)
       %th= User.human_attribute_name(:email)
       - if local_assigns[:f]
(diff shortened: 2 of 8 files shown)


app/views/users/access_list.html.haml
```diff
@@ -24,21 +24,26 @@
           %tr
             %th.approval-column= Product.human_attribute_name(:can_be_used_by?)
             %th.product-column= Product.model_name.human
+            %th.product-column= ProductUser.human_attribute_name(:date_added)
             %th.scheduling-group-column= ProductAccessGroup.model_name.human
 
         %tbody
           - products.each do |product|
+            - access_granted_at = @access_granted_date_by_product_id[product.id]
```
giladshanan, line 32: I'd vote for using the `find_product_user` method instead:
```suggestion
            - access_granted_at = product.find_product_user(@user)
```
I think this is a little bit simpler, and wouldn't require any changes to the controller logic.  What do you think?

giladshanan (commented):
Looks good to me, just one suggestion.  Let me know what you think.

app/views/product_users/_table.html.haml
```diff
@@ -20,6 +21,7 @@
         %td
           = link_to user.last_first_name, [current_facility, user]
 
+        %td= product_user.approved_at.strftime("%Y/%M/%d")
```
giladshanan, line 24: `%M` is going to display minutes, `%m` is for month.  Also, most other dates in the app are displaying the year at the end, so let's go for consistency:
```suggestion
        %td= product_user.approved_at.strftime("%m/%d/%Y")
```

app/views/users/access_list.html.haml
```diff
@@ -24,21 +24,26 @@
...
+                  access_granted_at.present?,
                   aria: {label: product.name}
               %td.product-column
                 - if training_requested_for?(product)
                   %span.label.label-important
                     = text("users.access_list.training_requested")
                 = link_to product, [product.facility, product, :users]
+              %td.product-column
+                - if access_granted_at.present?
+                  = access_granted_at.strftime("%Y/%M/%d")
```
giladshanan, line 46:
```suggestion
                  = access_granted_at.strftime("%m/%d/%Y")
```

spec/controllers/users_controller_spec.rb
```diff
@@ -113,6 +113,7 @@
     end
 
     before(:each) do
+      FactoryBot.create(:product_user, user: user, product: instruments.last)
```
giladshanan, line 116: I'd rather not add to these older controller specs, I prefer system specs instead.  Could you add a spec to `user_access_list_spec.rb` to show the new column is present?

giladshanan (changes requested):

app/helpers/date_helper.rb
```diff
@@ -49,6 +49,10 @@ def human_date(date)
     date.strftime("%B %e, %Y")
   end
 
+  def month_day_year(date)
```
giladshanan, line 52: 💯

giladshanan (approved):
Looks great!

PR #2603, opened 2021-05-25
[#142305] New Product Access List CSV export

wangminghao:
Creates a new Report that is a CSV export for data about Product Access lists and send it to the email address supplied by the user.  Importing Product Access Lists will come in a follow up PR.

### DIFF
--- a/app/models/reports/instrument_access_list_export_raw.rb
+++ b/app/models/reports/instrument_access_list_export_raw.rb
@@ -1,0 +1,75 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module Reports
+
+  class InstrumentAccessListExportRaw
+    include DateHelper
+    include TextHelpers::Translation
+
+    def initialize(product_users)
+      @product_users = product_users
+    end
+
+    def to_csv
+      CSV.generate do |csv|
+        csv << headers
+        @product_users.each do |product_user|
+          csv << build_row(product_user)
+        end
+      end
+    end
+
+    def filename
+      "instrument_access_list.csv"
+    end
+
+    def description
+      text(".subject")
+    end
+
+    def text_content
+      text(".body")
+    end
+
+    def has_attachment?
+      true
+    end
+
+    def translation_scope
+      "reports.instruments.product_users"
+    end
+
+    private
+
+    def headers
+      [
+        User.human_attribute_name(:username),
+        User.human_attribute_name(:full_name),
+        text(".suspended_or_expired"),
+        User.human_attribute_name(:email),
+        ProductAccessGroup.model_name.human,
+        ProductUser.human_attribute_name(:date_added),
+      ]
+    end
+
+    def build_row(product_user)
+      user = product_user.user
+      [
+        user.username,
+        user.full_name,
+        suspended_or_expired(user),
+        user.email,
+        product_user.product_access_group.try(:name),
+        format_usa_datetime(product_user.approved_at),
+      ]
+    end
+
+    def suspended_or_expired(user)
+      return "" if user.active?
+      user.suspended? ? "Suspended" : "Expired"
+    end
+  end
+
+end
--- a/app/models/reports/instrument_access_list_export_raw.rb
+++ b/app/models/reports/instrument_access_list_export_raw.rb
@@ -38,7 +38,7 @@ module Reports
     end
 
     def translation_scope
-      "reports.instruments.product_users"
+      "reports.product_users"
     end
 
     private
@@ -50,7 +50,7 @@ module Reports
         text(".suspended_or_expired"),
         User.human_attribute_name(:email),
         ProductAccessGroup.model_name.human,
-        ProductUser.human_attribute_name(:date_added),
+        text(".date_added"),
       ]
     end
 
--- a/app/views/product_users/index.html.haml
+++ b/app/views/product_users/index.html.haml
@@ -20,6 +20,12 @@
   %p.notice= text("none")
 - else
   - if @product.has_product_access_groups?
+    = form_tag facility_instrument_users_path(format: :csv), method: :get, class: "search_form" do
+      = hidden_field_tag :email, current_user.email, disabled: true
+      = hidden_field_tag :format, params[:format], disabled: true
+
+    = link_to t("reports.instruments.product_users.export"), url_for(format: :csv), class: "js--exportSearchResults pull-right", data: { form: ".search_form" }
+
     = form_for @product, url: [current_facility, @product, :update_restrictions], method: :put do |f|
       = render "table", f: f
       = f.submit text("update", plural_label: ProductAccessGroup.model_name.human.pluralize),
--- a/app/views/product_users/index.html.haml
+++ b/app/views/product_users/index.html.haml
@@ -13,9 +13,7 @@
 
 %p= text("explanation", product_type: @product.model_name.human.downcase)
 
-%p
-  = link_to text("add"), [:new, current_facility, @product, :user], class: "btn-add"
-
+%p= link_to text("add"), [:new, current_facility, @product, :user], class: "btn-add"
 
 - if @product_users.nil?
 - elsif @product_users.empty?
@@ -36,4 +34,3 @@
     = render "table"
 
   = will_paginate(@product_users)
-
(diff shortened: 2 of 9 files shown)


app/models/reports/instrument_access_list_export_raw.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    def build_row(product_user)
+      user = product_user.user
+      [
+        user.username,
+        user.full_name,
+        suspended_or_expired(user),
+        user.email,
+        product_user.product_access_group.try(:name),
+        format_usa_datetime(product_user.approved_at),
```
rdunlop, line 65: I'm confused why the header uses `:date_added`, but the row uses `approved_at`.

app/views/product_users/index.html.haml
```diff
@@ -20,6 +20,12 @@
   %p.notice= text("none")
 - else
   - if @product.has_product_access_groups?
+    = form_tag facility_instrument_users_path(format: :csv), method: :get, class: "search_form" do
+      = hidden_field_tag :email, current_user.email, disabled: true
+      = hidden_field_tag :format, params[:format], disabled: true
```
rdunlop, line 25: this should be unnecessary, since you have `format: :csv`. is it necessary?

app/views/product_users/index.html.haml
```diff
@@ -20,6 +20,12 @@
   %p.notice= text("none")
 - else
   - if @product.has_product_access_groups?
+    = form_tag facility_instrument_users_path(format: :csv), method: :get, class: "search_form" do
+      = hidden_field_tag :email, current_user.email, disabled: true
```
rdunlop, line 24: Having a hidden field like this is a smell. Can we instead use `current_user.email` in the receiving controller method?

app/views/product_users/index.html.haml
```diff
@@ -20,6 +20,12 @@
   %p.notice= text("none")
 - else
   - if @product.has_product_access_groups?
+    = form_tag facility_instrument_users_path(format: :csv), method: :get, class: "search_form" do
```
rdunlop, line 23: I'm confused...where's the action button for this form? Appears to be related to the link_to...hmm..do we need the form? can it be a button only?

rdunlop (changes requested):

app/models/reports/instrument_access_list_export_raw.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+require "csv"
+
+module Reports
+
+  class InstrumentAccessListExportRaw
```
giladshanan, line 7: Instruments aren't the only product that can have an access list, and Export Raw is an existing feature.  How about:
```suggestion
  class ProductAccessListCsvReport
```

app/models/reports/instrument_access_list_export_raw.rb
```diff
@@ -0,0 +1,75 @@
...
+    def text_content
+      text(".body")
+    end
+
+    def has_attachment?
+      true
+    end
+
+    def translation_scope
+      "reports.instruments.product_users"
```
giladshanan, line 41:
```suggestion
      "reports.product_users"
```

app/models/reports/instrument_access_list_export_raw.rb
```diff
@@ -0,0 +1,75 @@
...
+      CSV.generate do |csv|
+        csv << headers
+        @product_users.each do |product_user|
+          csv << build_row(product_user)
+        end
+      end
+    end
+
+    def filename
+      "instrument_access_list.csv"
```
giladshanan, line 25: How about we include the product name here:
```suggestion
      "#{product.name}_access_list.csv"
```
You could pass in the product on initialization.

app/models/reports/instrument_access_list_export_raw.rb
```diff
@@ -0,0 +1,75 @@
...
+        user.email,
+        product_user.product_access_group.try(:name),
+        format_usa_datetime(product_user.approved_at),
+      ]
+    end
+
+    def suspended_or_expired(user)
+      return "" if user.active?
+      user.suspended? ? "Suspended" : "Expired"
+    end
```
giladshanan, line 72: I think I'd prefer calling this column "User Status":
```suggestion
    def user_status(user)
      if user.active?
        "Active"
      elsif user.suspended?
         "Suspended"
      else
         "Expired"
       end
    end
```

app/views/product_users/index.html.haml
giladshanan, line 23: This needs to be pulled up to line 22, out of the conditional, so the link appears regardless of the presence of product access groups

app/views/product_users/index.html.haml
giladshanan: It looks like `CsvEmailAction` defaults to `current_user.email`, and we don't need to prompt the user for an email address, so I think we can drop the hidden fields.

giladshanan (changes requested):
Can you add a spec or two?  Something simple like `global_user_roles_report_spec.rb` just to cover the happy path basics.

Let me know if you want to talk any of this through in real time.

app/models/reports/instrument_access_list_export_raw.rb
giladshanan: This is probably the result of an existing confusing practice of using `ProductUser.human_attribute_name(:date_added)` to get "Date added".  There is no `date_added` attribute.

This is only used in a couple other places, let's clean it up.  I think we should use i18n instead.

app/views/product_users/index.html.haml
```diff
@@ -20,6 +20,12 @@
...
+      = hidden_field_tag :email, current_user.email, disabled: true
+      = hidden_field_tag :format, params[:format], disabled: true
+
+    = link_to t("reports.instruments.product_users.export"), url_for(format: :csv), class: "js--exportSearchResults pull-right", data: { form: ".search_form" }
```
giladshanan, line 27: Let's keep this simple and just send the email to the current user instead of prompting.

app/views/product_users/index.html.haml
wangminghao: Oh I thought maybe we wouldn't want to display a link that would produce an empty file - but yeah I can see that it's good for consistency with how the rest of the site functions.

app/views/product_users/index.html.haml
wangminghao: I thought this was the consistent way of doing things for the other Export to CSV links that I saw? Also, as an admin might I want to send the file to someone else?

app/views/product_users/index.html.haml
wangminghao: This is the mechanism that other Export to CSV links in the app are using. They have a form with hidden fields that a javascript snippet populates through a pop-up input box and that's how we are able to pass the input email on to the backend through an xhr call. I was duplicating this functionality (though I agree it is not ideal)

app/views/product_users/index.html.haml
giladshanan: Good points, let's keep the prompt.

app/views/product_users/index.html.haml
giladshanan: If we have `@product_users` then we should have rows in the CSV export.  Putting the link on line 22 shouldn't create an empty file.  Is it possible you're mixing up `ProductAccessGroup` and `ProductUser`?

app/views/product_users/index.html.haml
```diff
@@ -13,12 +13,20 @@
 
 %p= text("explanation", product_type: @product.model_name.human.downcase)
 
-%p= link_to text("add"), [:new, current_facility, @product, :user], class: "btn-add"
+%p
+  = link_to text("add"), [:new, current_facility, @product, :user], class: "btn-add"
```
giladshanan, line 17: Is this change intended?  Seems like the `%p` doesn't need it's own line.

giladshanan (approved):
LGTM!  One small non-blocking question.

app/views/product_users/index.html.haml
giladshanan, line 17:
```suggestion
%p = link_to text("add"), [:new, current_facility, @product, :user], class: "btn-add"
```

app/views/product_users/index.html.haml
```diff
@@ -13,12 +13,20 @@
...
+
```
giladshanan, line 18:
```suggestion
%p= link_to text("add"), [:new, current_facility, @product, :user], class: "btn-add"
```

PR #2605, opened 2021-05-27
[#157607] Handle admin and offline reservations in kiosk view

giladshanan:
# Release Notes

The kiosk page doesn't behave when a switchable instrument is placed offline or on an admin hold.
Resolves `NoMethodError: undefined method 'full_name' for nil:NilClass`

### DIFF
--- a/app/presenters/reservation_user_action_presenter.rb
+++ b/app/presenters/reservation_user_action_presenter.rb
@@ -39,6 +39,9 @@ class ReservationUserActionPresenter
   end
 
   def kiosk_user_actions
+    # no actions allowed for Offline and Admin Reservations
+    return [] unless order_detail.present?
+
     actions = []
     actions << kiosk_accessories_link if accessories?
     actions << kiosk_switch_actions if can_switch_instrument?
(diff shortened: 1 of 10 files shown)


app/presenters/reservation_user_action_presenter.rb
```diff
@@ -39,6 +39,9 @@ def user_actions
   end
 
   def kiosk_user_actions
+    # no actions allowed for Offline and Admin Reservations
+    return [] unless order_detail.present?
```
rdunlop, line 43: Can we make OfflineReservation and AdminReservation return falsey to the below booleans, making this unnecessary?

app/views/kiosk_reservations/_reservations_table.html.haml
```diff
@@ -10,7 +10,8 @@
       - reservations.each do |res|
         %tr
           %td
-            = res.user.full_name
+            - # Admin and Offline Reservations have a created_by user and no order user
+            = res.user&.full_name || res.created_by&.full_name || text("no_user")
```
rdunlop, line 14: Can we put these into some methods on Reservation/AdminReservation/OfflineReservation?

rdunlop (changes requested):

rdunlop (approved):

PR #2616, opened 2021-06-30
[#144493] Display order detail note on hover in admin calendars

giladshanan:
# Release Notes

Allows toggling the visibility of order detail notes in admin calendars (per facility). 

When enabled for a facility, order notes will appear when hovering on reservations in admin calendars:
- Admin Daily View
- Create/edit Admin Holds
- Admin Instrument reservation calendar

All other places where the order detail note appears will be unaffected by the new setting.


app/views/facility_reservations/_reservation.html.haml
```diff
@@ -12,7 +12,7 @@
       %span{ class: order_detail.order_status.to_s.downcase }
         = "- #{order_detail.order_status}"
 
-      - if order_detail.note.present?
+      - if order_detail.note.present? && order_detail.facility.show_order_note?
```
wangminghao, line 15: seems like this and the logic in calendar_presenter, line 20 are similar. Might want to consider moving this "return order_note if present" logic into the order_detail object itself that can be called from both of these places.

wangminghao (commented):
One suggestion but overall looks good. Also curious about why you moved the FullCalendarConfig out of coffeescript? Seems like the other defined classes have been done in coffeescript. I don't have an opinion on this, just curiosity. :)

app/views/facility_reservations/_reservation.html.haml
giladshanan: I lean towards holding off abstracting this logic until we need it in a 3rd place ([rule of 3](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))).  These 2 spots are slightly different and the extra abstraction didn't seem worth the work right now.  I also try to avoid adding anything to the OrderDetail class, since it is already 1010 lines long :)

giladshanan:
@wangminghao  Most of the beneficial features of coffeescript have since been added to ES6, which is much more widely used these days.  My goal is to move away from coffeescript completely, so I have been in the habit of decaf'ing every file I need to change as I go along.  

Now I need a coffee ☕

wangminghao (approved):

PR #2620, opened 2021-07-09
[#142305] Add import for adding approved users

giladshanan:
# Release Notes

The import process should only add new approved users when no errors occur.
Any error should prevent the import from adding new approved users.

# Screenshots

Import includes one existing user and one new one:
Screen Shot 2021-07-11 at 4 40 12 PM

All users already have access:
Screen Shot 2021-07-11 at 4 39 40 PM

One username is not found:
Screen Shot 2021-07-11 at 4 39 20 PM

### DIFF
--- a/app/controllers/product_user_imports_controller.rb
+++ b/app/controllers/product_user_imports_controller.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+class ProductUserImportsController < ApplicationController
+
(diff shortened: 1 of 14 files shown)


giladshanan:
@rdunlop I need to add some specs, but wanted to get your feedback on the approach so far.

rdunlop (approved):

PR #2628, opened 2021-07-16
[#127146] Update file count when result files are uploaded or removed on the order show page

KaraAJC:
# Release Notes
This PR implements javascript that will refresh the results files count on the order management show page when a result file is uploaded via modal. This also improves accessibility of the action links and the table. 

# Screenshot
## Before
Screen Shot 2021-07-16 at 6 03 57 PM

## After
Screen Shot 2021-07-16 at 5 58 09 PM


app/views/facility_orders/show.html.haml
```diff
@@ -14,60 +17,26 @@
 = render "merge_order"
 
 %table.order-list.table.table-striped.table-hover#order-management
+  %caption Order detail listings
```
giladshanan, line 20: How about 'Order details' or 'Order detail information'?

app/assets/javascripts/order_table_refresh.js
```diff
@@ -0,0 +1,33 @@
+/***
+ * This is used for the order detail page for a facility
+ * that needs to have the result file count updated
+ * when a user uploads or removes files via modal
+ *
```
giladshanan, line 5:
```suggestion
 * Upates the file count when result files are uploaded or removed
 * via modal on the admin facility Order show page.
```

app/assets/javascripts/app/results-file-upload.js
```diff
@@ -0,0 +1,16 @@
+function makeAModal() {
```
giladshanan, line 1: How about - 
```suggestion
function setResultsFileUploadModals() {
```

app/assets/javascripts/app/results-file-upload.js
```diff
@@ -0,0 +1,16 @@
...
+  new AjaxModal('a.results-file-upload', '#results-file-upload-modal', {
```
giladshanan, line 2: This class and id are only used in this order show table.  I think we should move this code into `app/assets/javascripts/order_table_refresh.js`

giladshanan: Also let's use " instead of '

app/assets/javascripts/order_table_refresh.js
```diff
@@ -0,0 +1,33 @@
...
+***/
+
+document.addEventListener("DOMContentLoaded", function() {
+  const table = document.getElementsByClassName("js--orderTableRefresh")[0];
```
giladshanan, line 9: This is only use in `fetchAndRefresh`, can you move it into that function?

app/assets/javascripts/order_table_refresh.js
```diff
@@ -0,0 +1,33 @@
...
+          table.innerHTML = body;
+          makeAModal()
+         });
+      }
+    });
+  }
+
+  $("#results-file-upload-modal").on("hidden", function() {
+    fetchAndRefresh()
+  });
```
giladshanan, line 30: I think this is the same as:

```suggestion
  $("#results-file-upload-modal").on("hidden", fetchAndRefresh);
```

app/views/facility_orders/_result_file_icon.html.haml
```diff
@@ -2,4 +2,5 @@
   .inline-icons
     - result_file_count = od.stored_files.sample_result.count
     = link_to admin_order_detail_path(:files, od), class: "results-file-upload" do
-      %i{class: result_file_count > 0 ? "fa fa-file" : "fa fa-file-o", data: { toggle: "tooltip" }, title: t("file_uploads.sample_result.count", count: result_file_count) }
+      %i{ class: result_file_count > 0 ? "fa fa-file" : "fa fa-file-o", data: { toggle: "tooltip" }, title: "Upload" }
```
giladshanan, line 5: Let's use i18n here for `title: "Upload"`

app/views/facility_orders/show.html.haml
giladshanan: I'd vote for i18n here, even though it's not used consistently in this file.

app/views/facility_orders/show.html.haml
```diff
@@ -14,60 +17,26 @@
...
-              = tooltip_icon "fa fa-plus-square", t("product_accessories.pick_accessories.title")
-
-          - if od.reservation.present?
-            = link_to(fa_icon("calendar"),
-              facility_order_order_detail_url(current_facility, @order, od, format: :ics),
-              title: t("ical.calendar_tooltip"))
-          = render "order_file_icon", od: od
-          = render "result_file_icon", od: od
+      %th.order-id{scope: "col"}= OrderDetail.human_attribute_name(:id)
+      %th.badges{scope: "col"} Status
```
giladshanan, line 28: I don't feel strongly, but if you decide to use some i18n here, these headers should all go into a translation file as well

app/views/facility_orders/_order_table.html.haml
```diff
@@ -0,0 +1,39 @@
...
+      %td.product
+        = od.product
+        - if od.time_data.present?
+          %br
+          = od.time_data
+      %td.action
+        - if od.add_accessories?
+          = link_to new_facility_order_order_detail_accessory_path(current_facility, @order, od), class: ["has_accessories", "persistent", "undecorated"] do
+            = tooltip_icon "fa fa-plus-square", t("product_accessories.pick_accessories.title")
+            %span Add Accessories
```
giladshanan, line 19: i18n please

app/views/facility_orders/_order_table.html.haml
```diff
@@ -0,0 +1,39 @@
...
+
+        - if od.reservation.present?
+          = link_to facility_order_order_detail_url(current_facility, @order, od, format: :ics) do
+            = tooltip_icon "fa fa-calendar", t("ical.calendar_tooltip")
+            %span Download event
```
giladshanan, line 24: i18n

giladshanan (changes requested):
Looks good, just requesting some clean up and nitpicks.  Let me know if you want to talk through anything.

app/views/facility_orders/show.html.haml
```diff
@@ -1,6 +1,9 @@
 - content_for :h1 do
   = current_facility
 
+= content_for :head_content do
+  = javascript_include_tag "order_table_refresh"
```
rdunlop, line 5: I'd vote for including this javascript in the application.js package, and not having to download a special JS file for this view...instead, have that javascript only enable itself if a certain element-type is present on the page.

I think it already does this...since if `#results-file-upload-modal` isn't present, nothing will be bound, right?

giladshanan (approved):
Nice!  @KaraAJC This [function rename](https://github.com/tablexi/nucore-open/pull/2628/commits/c289bc13c27d7eb27ad4a29547a34baf1733d34e#diff-c3d6588846bc57a666505fd61680de0a56661a841e762a6caeed91e143a374d3R21) fixed the spec failures. When the function got renamed from `makeAModal` to `setResultsFileUploadModals`, there was still code calling `makeAModal`, which would have been undefined.    System specs that load JS (and one controller spec that renders views) were tripping over this.  

I tested locally with the undefined method call `$(makeAModal);` in place - the specs fail, but the behavior when testing manually is fine.  I can create a price policy with this JS error, no problem.  It's just the specs that are not giving helpful error messages.  I have an idea for improving that, I'll open a new PR for it.

rdunlop (approved):

PR #2629, opened 2021-07-19
[#154945] Display reminder modal during year end cutoff window

giladshanan:
# Release Notes

As far as I know, NU is the only school using journal cutoff dates.  I am not sure if HKP or OSU plan to use them, but UMass, Dartmouth and UConn do not use them.  

# Screenshots

Managing reminders:
Screen Shot 2021-07-31 at 1 09 18 AM

Modal on clicking "Create" during the reminder date range:
Screen Shot 2021-07-31 at 12 58 20 AM


app/models/journal_cutoff_date.rb
```diff
@@ -23,6 +23,33 @@ def past?
...
+  # This is known as the 'year end closing window' - the date range between the
+  # end of the fiscal year and the journal cutoff date for the last month of the fiscal year.
+  #
+  # For example:
+  #   With FY22 starting 09/01/21, the closing window would
+  #   start Sep 1, 2021 and end on the August 2021 cutoff date:
+  #   Sep 1, 2021 - Sep 8, 2021.
+  #
+  # Returns true or false.
+  def self.year_end_closing_window?
```
rdunlop, line 36: Is there a spec which shows whether the `now < year_end_cutoff.cutoff_date` is working as expected when the `Time.current` is on the same date as the cutoff_date.

Or when the Time.now is the same date as the `last_valid_date`?

rdunlop (changes requested):

rdunlop (approved):

config/locales/en.yml
```diff
@@ -266,13 +266,10 @@ en:
...
-      header: Year End Closing Reminder
-      message: |
-        We are in the year-end closing window.  Any journals for the previous fiscal year
-        must be submitted by %{cutoff_time}.  If you are creating a journal for %{year},
-        make sure to update the Journal Date field.
-      proceed: Proceed
+    journal_closing_modal:
+      header: Journal Closing Reminder
+      proceed: Create Journal
+      cancel: Edit Journal Date
```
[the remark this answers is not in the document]
giladshanan: Fair point - the idea is that in order to edit the journal date you need to cancel (close) the modal.

app/controllers/journal_closing_reminders_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+class JournalClosingRemindersController < ApplicationController
+
+  before_action :authenticate_user!
+  load_and_authorize_resource
+
+  layout "two_column"
+  before_action { @active_tab = "global_settings" }
+
+  def index
+    @journal_closing_reminders = JournalClosingReminder.order(ends_at: :desc)
```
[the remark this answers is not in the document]
giladshanan: They are planning to just have one record, which they will update as needed.

app/views/journal_closing_reminders/index.html.haml
```diff
@@ -0,0 +1,26 @@
...
+  = t("pages.global_settings")
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_global", sidenav_tab: "journal_closing_reminders"
+
+%h2= JournalClosingReminder.model_name.human(count: 2)
+
+= link_to t(".new"), new_journal_closing_reminder_path, class: "btn-add"
+
+%table.table
```
KaraAJC, line 11: should have caption element here

app/views/journal_closing_reminders/index.html.haml
```diff
@@ -0,0 +1,26 @@
...
+  %thead
+    %th
+    %th= t(".start_date")
+    %th= t(".end_date")
+    %th= t(".fiscal_year")
+    %th= t(".message")
```
KaraAJC, line 17: should have scope="col" on these headers

app/views/journal_closing_reminders/index.html.haml
```diff
@@ -0,0 +1,26 @@
...
+  %tbody
+    - @journal_closing_reminders.each do |journal_closing_reminder|
+      %tr{class: journal_closing_reminder.past? ? "row-warning" : "" }
+        %td= link_to t(".edit"), edit_journal_closing_reminder_path(journal_closing_reminder)
```
KaraAJC, line 21: might be where you'd have scope="row" but it would also need unique value, like "edit reminder #1" or something

config/locales/en.yml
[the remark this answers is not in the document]
KaraAJC: unless this action takes you back to editing the record where you can change the date, it might be better to say "Cancel & edit" or "Go Back and Edit Date"

KaraAJC (changes requested):

config/locales/en.yml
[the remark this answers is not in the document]
giladshanan: The button will hide the modal, revealing the form where the user can edit the Journal Date field.  I agree this is a little odd.  I would just go with "Continue"/"Cancel" myself, but the client specifically requested this text for the buttons.

app/controllers/journal_closing_reminders_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  end
+
+  private
+
+  def default_next_window
+    [JournalClosingReminder.maximum(:starts_at), Time.current].compact.max
+  end
+
+  def journal_closing_reminder_params
+    params.require(:journal_closing_reminder).permit(:message, starts_at: :date, ends_at: :date)
```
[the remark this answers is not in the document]
giladshanan: Yes, because of the `DateDropdownInput` is nesting the `starts_at` input within a `date` input.  I am going to try another, simpler approach.

app/inputs/date_dropdown_input.rb
```diff
@@ -0,0 +1,21 @@
...
+#   date_time_inputable :starts_at
+# end
+#
+# ... and then in the view:
+#   = f.input :starts_at, as: :date_dropdown
+class DateDropdownInput < SimpleForm::Inputs::Base
+  def input(wrapper_options)
+    value = object.public_send("#{attribute_name}_date_time_data")
+    @builder.simple_fields_for attribute_name, value do |f|
+      f.input :date, include_blank: value.present?, label: false, input_html: { class: "datepicker"  }
```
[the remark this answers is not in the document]
giladshanan: This is copy/pasted from TimeDropdownInput.  I am not sure the intent or if it is needed.

app/views/journal_closing_reminders/index.html.haml
giladshanan: I thought scope=row is for a `th` element, is that right?

spec/models/journal_creation_reminder_spec.rb
```diff
@@ -0,0 +1,95 @@
...
+          expect(reminder.errors.messages).to eq({ starts_at: ["must be after Ending Date"] })
+        end
+      end
+
+      context "when starts_at is the same as ends_at" do
+        let(:current_time) { Time.current }
+        let(:starts_at) { current_time }
+        let(:ends_at) { current_time }
+
+        it { is_expected.to be_valid }
```
rdunlop, line 39: I was surprised that this was valid, until I realized that the `ends_at` is being `.end_of_day`'d, which means that they aren't actually equal. Is this the intention of this spec? perhaps adding:
```
it "sets the end_date to the end of day" do
  expect(reminder.ends_at).to eq(current_time.end_of_day)
end
```
...so that it's clear/explicit...

but then I see that you do almost this exact thing on line 44...so maybe it's unnecessary.

rdunlop (approved):

spec/models/journal_creation_reminder_spec.rb
giladshanan: Good call

app/views/journal_closing_reminders/index.html.haml
KaraAJC: yes, that would be correct, and there's typically a header field on a row, which i was determining to be the edit field

KaraAJC (approved):

PR #2631, opened 2021-07-21
[#126556] Update conditions for create order button on Facility page

KaraAJC:
# Release Notes

Previously, if the facility chooses to "Show Order Form for Facility" (allows adding of multiple products simultaneously) but all the products are hidden, only the "Create Order" buttons are shown to the user, creating confusion.

This change adds a condition for the form to show if the above choice is made, AND there is at least one active product to show along with the buttons.

For admins, they will be able to see the "Create Order" button, even if all the products are hidden.


rdunlop (approved):

ahorner (approved):

KaraAJC (commented):
I like the refactor, @giladshanan !

PR #2692, opened 2021-08-03
[#142306] Holiday access rules

giladshanan:
# Release Notes

Allows setting holiday access restriction at the product level, and then allowing access at the level of scheduling group (ProductAcessGroup).

My current plan is to move a lot of the basic holiday logic out of the `jxml` engine when I merge this into the NU repo.


app/views/admin/products/_product_fields.html.haml
```diff
@@ -6,9 +6,10 @@
   = f.input :initial_order_status_id, collection: OrderStatus.initial_statuses(current_facility).collect {|cf| [cf.name_with_level, cf.id] }, hint: text("hints.initial_order_status"), include_blank: false
 
   %fieldset.well
-    = f.input :requires_approval, as: :boolean, label: false, inline_label: text("hints.requires_approval"), input_html: { data: { disables: ".#{f.object.class.model_name.to_s.underscore}_allows_training_requests"} }
+    = f.input :requires_approval, as: :boolean, label: false, inline_label: text("hints.requires_approval"), input_html: { data: { disables: ".#{f.object.class.model_name.to_s.underscore}_allows_training_requests, .instrument_restrict_holiday_access"} }
```
giladshanan, line 9: This probably should be `.#{f.object.class.model_name.to_s.underscore}_restrict_holiday_access` - or the `= f.input :restrict_holiday_access` element should only be shown for Instruments

app/views/holidays/index.html.haml
```diff
@@ -0,0 +1,24 @@
...
+  = t("pages.global_settings")
+
+= content_for :sidebar do
+  = render "admin/shared/sidenav_global", sidenav_tab: "holidays"
+
+%h2= t(".title")
+
+= link_to t(".new"), new_holiday_path, class: "btn-add"
+
+%table.table
```
giladshanan, line 11: I need to make some accessibility adjustments for this table

app/views/holidays/shared/_tab.html.haml
```diff
@@ -0,0 +1 @@
+= tab t("pages.holidays"), holidays_path, sidenav_tab == "holidays"
```
giladshanan, line 1: This file isnt  being used

app/views/product_access_groups/index.html.haml
```diff
@@ -13,6 +13,8 @@
       %th
       %th= ProductAccessGroup.human_attribute_name(:name)
       %th= User.model_name.human.pluralize
+      - if @product.restrict_holiday_access?
```
giladshanan, line 16: Not sure how I feel about hiding this column

db/migrate/20210803022837_create_holidays.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class CreateHolidays < ActiveRecord::Migration[5.2]
+  def change
+    create_table :holidays do |t|
+      t.datetime :date, null: false
```
rdunlop, line 6: I think that there should. be an index on this column in order to support the `.on` scope.

rdunlop (commented):
looking good

app/support/holiday_validations.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module HolidayValidations
+
+  extend ActiveSupport::Concern
+
+  included do
+    validate :holiday_accessible, if: :starts_on_holiday?
+  end
+
+  def holiday_accessible
```
rdunlop, line 11: The way that this concern is written, adds a lot of methods to the including-object.

It does not appear that you actually want/need to expose `holiday_accessible`, `user_restricted_on_holidays?`, etc to any users of the including object.

I propose extracting this logic into a real PORO.

doing so would have these benefits:
- Explicitly make clear which attributes of the including object are being used
- Allow for easier testing, by making the attributes explicit.

Example:

```ruby
module HolidayValidations
  extend ActiveSupport::Concern

  included do
    validate :holiday_accessible
  end

  def holiday_accessible
    # this makes explicit which attributes you are depending on
    holiday_validator = HolidayValidator.new(
                                        product: product,
                                        user: user
                                        event_start_date: event_start_date

    return unless holiday_validator.restricted?

    errors.add(:base, :holiday_access_restricted)
end

class HolidayValidator
  ....
  def restricted?
    return false unless starts_on_holiday?

    product.restrict_holiday_access? && user_restricted_on_holidays?
  end
  ...
end
```

vendor/engines/secure_rooms/app/services/secure_rooms/access_rules/holiday_access_rule.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module SecureRooms
+
+  module AccessRules
+
+    class HolidayAccessRule < BaseRule
```
rdunlop, line 7: I thought that we would need to reference this class from the `CheckAccess` list of rules, no?

rdunlop (changes requested):

vendor/engines/secure_rooms/app/models/secure_rooms/occupancy.rb
```diff
@@ -6,6 +6,7 @@ class Occupancy < ApplicationRecord
 
     include DateTimeInput::Model
 
+    belongs_to :product
```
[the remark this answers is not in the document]
giladshanan: Good catch - this is left over from an earlier point when the occupancy was getting validated (like a reservation).  We're not calling the `holiday_access` method to validate occupancies at all, so we don't need this association.

rdunlop (approved):

PR #2705, opened 2021-08-18
[#152051] Fail gracefully when replacing account owner with existing member

mtorres27:
# Release Notes

This PR adds error handling for attempts to replace an account owner with a user that is already an account member.

# Screenshot

Screen Shot 2021-08-18 at 12 20 26 PM

### DIFF
--- a/app/controllers/facility_account_users_controller.rb
+++ b/app/controllers/facility_account_users_controller.rb
@@ -7,7 +7,6 @@ class FacilityAccountUsersController < ApplicationController
   before_action :check_acting_as
   before_action :init_current_facility
   before_action :init_account
-  before_action :init_user, only: :create
 
   load_and_authorize_resource class: AccountUser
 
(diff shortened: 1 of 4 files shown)


rdunlop (approved):

app/controllers/facility_account_users_controller.rb
```diff
@@ -33,7 +34,6 @@ def new
 
   # POST /facilities/:facility_id/accounts/:account_id/account_users
   def create
-    @user = User.find(params[:user_id])
     @account_user = AccountUser.grant(@user, create_params[:user_role], @account, by: session_user)
 
     if @account_user.persisted?
```
giladshanan, line 39: What do you think about adding a branch to the existing conditional like:
```suggestion
    if @account_user.nil?
      flash[:error] = text("create.existing_member_error", user: @user.full_name)
      redirect_to facility_account_members_path(current_facility, @account)
    elsif @account_user.persisted?
```
This logic isn't as specific but I find the expected outcomes easier to parse.  

You could also do something like:
```'
def existing_member_becoming_owner?
  create_params[:user_role] == AccountUser::ACCOUNT_OWNER &&
    @account.account_users.map(&:user_id).include?(@user.id)
end
```
to cover this specific case, but I'm ok with using the more specific error message to cover any nil account_user.
We can adjust in the future if we need to.

spec/controllers/facility_account_users_controller_spec.rb
```diff
@@ -215,6 +215,36 @@
             do_request
           end
         end
+
+        context "from anything to owner" do
```
giladshanan, line 219: Can you add this as a simple system spec for this instead?  I should have mentioned this earlier but we are moving away from controller specs in the codebase.

giladshanan (changes requested):
Looks good!  I am blocking on account of the controller specs since we are trying to move away from those.  Let me know what you think about my other suggestion.

giladshanan (approved):

PR #2713, opened 2021-08-24
[#142306] Update future scope and helper text

giladshanan:
# Release Notes

Updates a scope to be Oracle-friendly.  
Also makes the caption a bit broader and puts more detail info in hint text at the bottom of the page.

### DIFF
--- a/app/models/holiday.rb
+++ b/app/models/holiday.rb
@@ -4,7 +4,7 @@ class Holiday < ApplicationRecord
 
   validates_presence_of :date
 
-  scope :future, -> { where("holidays.date >= ?", Time.current.to_date) }
+  scope :future, -> { where(date: Time.current.to_date..Float::INFINITY) }
   scope :on, -> (date) { where(date: date..date.end_of_day) }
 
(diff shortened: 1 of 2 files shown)


app/models/holiday.rb
```diff
@@ -4,7 +4,7 @@ class Holiday < ApplicationRecord
 
   validates_presence_of :date
 
-  scope :future, -> { where("holidays.date >= ?", Time.current.to_date) }
```
rdunlop, line 7: How hard would it be to add a spec to this which would show that it's failing on Oracle?

rdunlop (approved):

app/models/holiday.rb
giladshanan: I'm not sure how to do that (other than testing against Oracle).

rdunlop: Oh, I forgot, I thought our CI was testing against oracle

giladshanan: It does for nucore-nu, but not open

PR #2714, opened 2021-08-24
[#157244] Don't clean up reconciled order details

giladshanan:
# Release Notes

It is possible to add a service product to an existing order and mark it Complete at the same time.
If this service requires a survey or order form it will not be added to the existing order, but will appear as a transaction and can be reconciled.

This has only happened once, so for now we are just going to exclude reconciled order details from the cleanup task.
I manually merged the order detail that was in this odd state.

### DIFF
--- a/lib/tasks/order_details.rake
+++ b/lib/tasks/order_details.rake
@@ -14,7 +14,7 @@ namespace :order_details do
 
   desc "task to remove merge orders that have been abandoned. See Task #48377"
   task remove_merge_orders: :environment do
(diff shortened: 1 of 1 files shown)


rdunlop (commented):
Do ALL `Order` have a `OrderDetail` associated with them? by adding `join` you are going to be inner-joining the tables, is that ok?

giladshanan:
Good catch

rdunlop (approved):

PR #2760, opened 2021-09-29
[#158493][WIP] Add i_class_number to users

giladshanan:
# Release Notes

WIP

Adding inputs for iClass number to users

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/user_extension.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/user_extension.rb
@@ -8,6 +8,7 @@ module SecureRooms
 
     included do
       validates :card_number, uniqueness: { allow_blank: true }
+      validates :i_class_number, uniqueness: { allow_blank: true }
     end
 
   end
--- a/vendor/engines/secure_rooms/spec/models/user_spec.rb
+++ b/vendor/engines/secure_rooms/spec/models/user_spec.rb
@@ -5,4 +5,5 @@ require "rails_helper"
 RSpec.describe User do
   subject(:user) { build(:user) }
   it { is_expected.to validate_uniqueness_of :card_number }
+  it { is_expected.to validate_uniqueness_of :i_class_number }
 end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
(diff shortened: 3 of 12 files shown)


vendor/engines/secure_rooms/app/models/secure_rooms/user_extension.rb
```diff
@@ -8,6 +8,11 @@ module UserExtension
 
     included do
       validates :card_number, uniqueness: { allow_blank: true }
+      validates :i_class_number, uniqueness: { allow_blank: true }
+
+      def self.find_by_card_number(card_number)
```
giladshanan, line 13: This method name could be confusing because Rails [magically defines finder methods](https://guides.rubyonrails.org/active_record_querying.html#dynamic-finders) for each attribute.  This is not going to cause an issue with behavior, just future readability.  

If I saw this:
`User.find_by_card_number(params[:card_number])`

 I would assume it was doing
`User.find_by(card_number: params[:card_number])`

which is what Rails does under the hood.

```suggestion
      def self.for_card_number(card_number) 
```

vendor/engines/secure_rooms/spec/models/user_spec.rb
```diff
@@ -3,6 +3,36 @@
 require "rails_helper"
 
 RSpec.describe User do
-  subject(:user) { build(:user) }
+  subject(:user) { build(:user, :netid) }
```
giladshanan, line 6: You can DRY things up a bit here by setting card_number in the subject call:
```suggestion
  subject(:user) { build(:user, :netid, card_number: card_number) }
```

Then in each context block, you just need to define `card_number`:
`let(:card_number) { "12345-123" }`
... and you can drop the `before` blocks (`before { user.update(card_number: card_number) }`)

giladshanan (commented):
Just a couple small changes to make.

vendor/engines/secure_rooms/spec/models/user_spec.rb
Stamates: We tried this but it didn't seem to DRY up the code much since we're changing the card_number so much and the tests started failing.

giladshanan: I pushed up a commit that does this.  It doesn't DRY things up much, but the intent is a little clearer (for me at least).  Calling `user.update` was actually persisting the record to the db, which makes the specs pass.  I refactored a bit so the user under test is instantiated up front before we look it up by the card number.

PR #2762, opened 2021-10-04
[#152780] Handle orders with invalid Payment Source

mtorres27:
# Release Notes

This PR handles orders with invalid Payment Sources, making it explicit to the user how they should proceed in order to continue with their orders

# Screenshot

### DIFF
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -44,6 +44,11 @@ class OrdersController < ApplicationController
     facility_ability = Ability.new(session_user, @order.facility, self)
     @order.being_purchased_by_admin = facility_ability.can?(:act_as, @order.facility)
     @order.validate_order! if @order.new?
+
+    return if @order.has_valid_payment?
+    @order.errors.add(:base, "Account is not valid for the orderer")
+    @order.errors.add(:base, "Please Change Payment Source or remove any invalid products from the Cart before continuing")
+    flash[:error] = "There are errors in your order:<br>#{@order.errors.full_messages.join('<br>')}".html_safe
   end
 
   # PUT /orders/:id/clear
@@ -110,7 +115,12 @@ class OrdersController < ApplicationController
         rescue NUCore::MixedFacilityCart
           @order.errors.add(:base, "You can not add a product from another facility; please clear your cart or place a separate order.")
         rescue => e
-          @order.errors.add(:base, "An error was encountered while adding the product #{@product}.")
+          if e.message == "Validation failed: Account is not valid for the orderer"
+            @order.errors.add(:base, e.message)
+            @order.errors.add(:base, "Please Change Payment Source or remove any invalid products from the Cart before continuing")
+          else
+            @order.errors.add(:base, "An error was encountered while adding the product #{@product}.")
+          end
           Rails.logger.error(e.message)
           Rails.logger.error(e.backtrace.join("\n"))
         end
--- a/app/views/orders/_form.html.haml
+++ b/app/views/orders/_form.html.haml
@@ -41,11 +41,21 @@
       .backdate_fields= render "edit_date", f: f
 
     %ul.inline{ style: "float: left" }
-      %li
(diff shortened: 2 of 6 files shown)


app/views/orders/_product_description.html.haml
```diff
@@ -1,5 +1,6 @@
 - unless order_detail.order.validated?
   .label.label-warning= order_detail.validate_for_purchase
-
+- unless order_detail.order.has_valid_payment?
+  .label.label-warning= "Invalid"
```
giladshanan, line 4: I think this should say:
```suggestion
  .label.label-warning= "You are not approved to use this account."
```

app/controllers/orders_controller.rb
```diff
@@ -44,6 +44,11 @@ def show
     facility_ability = Ability.new(session_user, @order.facility, self)
     @order.being_purchased_by_admin = facility_ability.can?(:act_as, @order.facility)
     @order.validate_order! if @order.new?
+
+    return if @order.has_valid_payment?
+    @order.errors.add(:base, "Account is not valid for the orderer")
+    @order.errors.add(:base, "Please Change Payment Source or remove any invalid products from the Cart before continuing")
+    flash[:error] = "There are errors in your order:<br>#{@order.errors.full_messages.join('<br>')}".html_safe
```
giladshanan, line 51: I think we should say something like:
```
"The payment source is not valid for the orderer.<br>  Please (clear your cart) [link to clear cart] or (select a different payment source) [link to change payment source] to continue."
```
... and put this in a translation file (i18n).

Then we can reuse in the places we need it like:
```suggestion
    @order.errors.add(:base, t("account_invalid_for_orderer"))
    flash[:error] = "There are errors in your order:<br>#{@order.errors.full_messages.join('<br>')}".html_safe
```

app/controllers/orders_controller.rb
```diff
@@ -110,7 +115,12 @@ def add
         rescue NUCore::MixedFacilityCart
           @order.errors.add(:base, "You can not add a product from another facility; please clear your cart or place a separate order.")
         rescue => e
-          @order.errors.add(:base, "An error was encountered while adding the product #{@product}.")
+          if e.message == "Validation failed: Account is not valid for the orderer"
+            @order.errors.add(:base, e.message)
+            @order.errors.add(:base, "Please Change Payment Source or remove any invalid products from the Cart before continuing")
```
giladshanan, line 120: Then we can reuse in 3 places we need it like:
```suggestion
    @order.errors.add(:base, t("account_invalid_for_orderer"))
```

app/views/orders/_form.html.haml
```diff
@@ -41,11 +41,21 @@
...
+              class: %w(btn btn-primary),
+              data: { disable_with: t("shared.purchase") }
+      - else
+        - alert_error = "Please Change Payment Source or remove any invalid products from the Cart before continuing"
+        %li
+          %a{ class: "btn disabled", data: { toggle: "tooltip" }, title: "#{alert_error}" }
+            = t("shared.update")
+        - if order.validated?
+          %li
+            %a{ class: "btn btn-primary disabled", data: { toggle: "tooltip" }, title: "#{alert_error}" }
```
giladshanan, line 60:
```suggestion
        - alert_error = t("account_invalid_for_orderer")
        %li
          %a{ class: "btn disabled", data: { toggle: "tooltip" }, title: alert_error }
            = t("shared.update")
        - if order.validated?
          %li
            %a{ class: "btn btn-primary disabled", data: { toggle: "tooltip" }, title: alert_error }
```

giladshanan (changes requested):
Looks good!  I left a few small comments, let me know what you think about adding the links in the flash message.

app/views/orders/_form.html.haml
giladshanan: Actually, we can't put links in the tooltip.  I would still vote for putting the text in a translation file though.

PR #2790, opened 2021-10-28
[#158849] Fix card number and iclass indexes for mysql

giladshanan:
Oracle doesn't consider an empty string when enforcing uniqueness, but MySQL does
https://decipherinfosys.wordpress.com/2007/11/30/multiple-null-values-in-a-unique-index-in-sql-serverdb2-luw/

This prevents saving empty strings to the database, since MySQL enforces uniqueness to non-NULL values.

### DIFF
--- a/vendor/engines/secure_rooms/app/models/secure_rooms/user_extension.rb
+++ b/vendor/engines/secure_rooms/app/models/secure_rooms/user_extension.rb
@@ -7,11 +7,19 @@ module SecureRooms
     extend ActiveSupport::Concern
 
     included do
+      NULL_ATTRS = %w(card_number i_class_number)
+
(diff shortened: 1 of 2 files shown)


ahorner (commented):
What's the motivation for clearing these out before save, rather than on initial value assignment? This is the pattern used in the [`attribute_normalizer`](https://github.com/mdeering/attribute_normalizer) gem, and generally makes it a bit easier to reason about the contents of a field.

We probably don't want to pull in a full gem to implement this; we could accomplish this in a more lightweight manner with something like:

```
def card_number=(value)
  super value.presence
end

def i_class_number=(value)
  super value.presence
end
```

KaraAJC (approved):

PR #2794, opened 2021-11-02
[#158793] Check the connection every 6 hours

giladshanan:
Gives us early warning if something goes wrong with curl commands from the server to itself
For UMass this currently fails with:
`SSL_connect returned=1 errno=0 state=error: certificate verify failed (unable to get local issuer certificate)`

NOTE:
Instances hosted on k8s don't use the `whenever` gem.  
I think the `CronJob` objects will complain if they fail, but haven't looked into it or tested.


boatrite (approved):

giladshanan:
Closing in favor of a larger refactor -> https://github.com/tablexi/nucore-open/pull/2957

PR #2797, opened 2021-11-04
Moves Relay details for instruments into its own tab

ktweeden:
# Release Notes

This PR moves the relay information for instruments from the "details" tab to their own "relays" tab. When creating a new instrument, the relay type will be set to "reservation only".

# Screenshot

Screenshot 2021-11-05 at 12 20 46

Screenshot 2021-11-05 at 12 21 03

Screenshot 2021-11-08 at 18 14 10

Screenshot 2021-11-08 at 18 12 23


app/controllers/instrument_relays_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+  def update
+  end
+
+  def new
+  end
+
+  private
+
+  def init_product
+    id_param = params.except(:facility_id).keys.detect { |k| k.end_with?("_id") }
```
giladshanan, line 27: Since we are just using this for instruments, I think we  can do:
`@product = current_facility.instruments.find_by!(url_name: params[:instrument_id])`

app/controllers/products_common_controller.rb
```diff
@@ -110,7 +110,7 @@ def resource_params
                                                       :auto_cancel_mins, :lock_window, :cutoff_hours,
                                                       :problems_resolvable_by_user, :restrict_holiday_access,
                                                       relay_attributes: [:ip, :ip_port, :outlet, :username, :password, :type,
-                                                                         :auto_logout, :auto_logout_minutes, :id])
+                                                                         :auto_logout, :auto_logout_minutes, :id, :mac_address, :building_room_number, :circuit_number, :ethernet_port_number ])
```
giladshanan, line 113: I might break this into a new line after `:mac_address` just to keep the lines a bit shorter.  Not a big deal though.

app/views/instrument_relays/_form.html.haml
```diff
@@ -0,0 +1,41 @@
...
+    selected: @product.control_mechanism || :manual
+
+  .well#power-relay
+    %h3= text("instruments.instrument_fields.relay.label.relay")
+    %p= text("instruments.instrument_fields.relay.intro")
+    = f.simple_fields_for :relay, @product.relay || @product.build_relay do |r|
+      = r.input :type, collection: options_for_relay, include_blank: false
+      = r.input :ip
+      = r.input :ip_port,
+      hint: text("instruments.instrument_fields.relay.instruct.ip_port")
```
giladshanan, line 17: I find this easier to read with the additional args indented to align with `input`, like below on line 26-29

app/views/instrument_relays/index.html.haml
```diff
@@ -0,0 +1,41 @@
...
+  = current_facility
+= content_for :sidebar do
+  = render "admin/shared/sidenav_product", sidenav_tab: @product.class.name.pluralize.underscore
+= content_for :tabnav do
+  = render "admin/shared/tabnav_product", secondary_tab: "relays"
+
+%h2= @product
+
+- if @product.relay.nil?
+  %label= t 'instruments.manage.label.control_mech'
```
giladshanan, line 11: Can you update this file to use parens and double quotes?
```suggestion
  %label= t("instruments.manage.label.control_mech")
```
We don't have linters running in CI but I try to clean things up as we go.

app/controllers/instrument_relays_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  admin_tab :all
+  before_action :authenticate_user!
+  before_action :init_product
+  before_action :manage
+
+  layout "two_column"
+
+  # GET /facilities/:facility_id/instrument/:instrument_id/relays
+  def index
```
giladshanan, line 11: This feels more like a `show` action to me, the relationship to relays is a `has_one`.  What do you think?

app/views/instruments/_instrument_manage_fields.html.haml
```diff
@@ -4,8 +4,6 @@
 
 = f.input :problems_resolvable_by_user, as: :boolean, hint: false
 
-= render "relay_manage", f: f
```
giladshanan, line 7: Is this partial used anywhere else?  If not then let's remove it.

spec/controllers/instrument_relays_controller_spec.rb
```diff
@@ -0,0 +1,41 @@
+# frozen_string_literal: true
+
+require "rails_helper"
```
giladshanan, line 3: I should have mentioned this up front - we are moving away from controller specs where we can.  Do you mind re-writing this as a system spec instead?  For a simple example check out `access_list_tab_spec.rb`, or any other spec in the system directory that has a name ending like `...controller_spec.rb`.

spec/controllers/instruments_controller_spec.rb
```diff
@@ -276,6 +276,10 @@
                                       type: RelaySynaccessRevA.name,
                                       auto_logout: true,
                                       auto_logout_minutes: 15,
```
giladshanan, line 278: I think the relay related logic in this file should be moved into the new system spec, since it is describing a code path that won't be used any more

giladshanan (changes requested):
This is looking good!  I left a handful of suggestions, let me know if you want to talk through anything.  

Also, do you think it's worth including a notification after creation that would link to the relays tab?

app/controllers/instrument_relays_controller.rb
ktweeden: yeah, I went back and forth on this. I think I ended up settling on index because you aren't selecting a specific relay to view but rather the tab is for all the relays an instrument has (which happens to be one). Happy to switch to show though, I don't feel super strongly.

ktweeden: Thinking about this more, I feel like index also makes sense for the tab path because the instrument might not have a relay

ktweeden:
@giladshanan with the notification pointing the user to the relay tab - I landed on not including that mainly because we don't specifically highlight that other tabs may need editing/are set to defaults, other than the pricing (and here I think we only have a something that notifies the user these are missing if you're in the instrument index view). 

I'm not sure if it's important that users be aware that they need to take a look at the relay tab than any other information they haven't completed: what do you think? In effect I'm wondering: if the tabs had always been organised this way, would we have included some ui on save that points to the relay above other tabs from the outset? 

Happy to add it in though if you think it's the right way to go, especially if we think the user base will expect changes in UI to be called out explicitly, based on notifications etc that we've put in place for previous changes.

spec/controllers/instruments_controller_spec.rb
```diff
@@ -275,7 +275,7 @@
                                       password: "password",
                                       type: RelaySynaccessRevA.name,
                                       auto_logout: true,
-                                      auto_logout_minutes: 15,
```
giladshanan, line 278: There are [some specs ](https://github.com/tablexi/nucore-open/blob/f22ff0b945db10cbd1c866ead2446114e12043b5/spec/controllers/instruments_controller_spec.rb#L267)in this file that describe controller logic that I don't think we're using any more.
Can you please check if that's the case and if so, move the test logic from this file into the new system spec?

app/controllers/instrument_relays_controller.rb
```diff
@@ -0,0 +1,72 @@
...
+  end
+
+  # GET /facilities/:facility_id/instrument/:instrument_id/relays/:id/edit
+  def edit
+  end
+
+  def update
+    # TODO - handle change in relay type
```
giladshanan, line 21: Adding a spec or two for this would be good

giladshanan (commented):
- add spec: when changing relay types, the old relay should be deleted
- rspec spec/system/admin/instrument_relay_tab_spec.rb is failing, not sure why
- resourceful approach for form generation (simple_form_for [@relay])
- error validation should appear on the fields
- fields that are not required should not have *

giladshanan (approved):

PR #2710, opened 2021-08-20
[#142337] Implements bulk reconcile note

KaraAJC:
# Release Notes

This PR implements a bulk reconciliation note for the purchase order and credit card reconciliation step, and updates some accessibility functionality to the surrounding table

# Screenshots
## credit card reconciliation page with individual notes available
Screen Shot of credit card reconciliation page with individual notes available

## credit card reconciliation page with bulk note active
Screen Shot of credit card reconciliation page with bulk note active

PR #2810, opened 2021-11-18
Fix reconcile note bug, add specs

giladshanan:
# Release Notes

In [#142337] Implements bulk reconcile note (tablexi/nucore-open#2710) we introduced a regression that causes individually applied reconcile notes to not be saved to the db.


ahorner (approved):
A strong testament to the power of input normalization! Definitely a subtle issue. Good catch, and a straightforward fix.

giladshanan:
@ahorner thanks - while combing through logs to recreate some lost data for NU, I uncovered another input normalization issue.  Follow up PR for that one is coming soon.

PR #2824, opened 2021-12-08
[#155286] Order Detail Split Amounts

giladshanan:
# Release Notes
This PR Adds a list of subaccount split information on an order detail when payment sources include them. This also dynamically updates the split info when quantity or price changes.

# Screenshot

split accounts updating based on quantity or cost changing


giladshanan:
@KaraAJC you can go ahead and merge this in if you support the changes I added.  If not, add more changes or comments so we can pick this up when I'm back online in January 🎉

KaraAJC (approved):

PR #2834, opened 2021-12-15
[#159184] Don't pass nil as a possible mime type

giladshanan:
Addresses ArgumentError: Invalid formats: nil
Silently ignored before rails 6.0
https://github.com/rails/rails/commit/0eb9e1e51a9df554e1cf7a42ec339ca6f0120a04#

Tested locally by patching `formats_with_html_fallback`:
```
  def formats_with_html_fallback
    [nil] + [:html]
  end
```
... when clicking something I can't see while "acting as" another user:
```
  rescue_from NUCore::NotPermittedWhileActingAs, with: :render_acting_error
  def render_acting_error
    render "/acting_error", status: 403, layout: "application", formats: formats_with_html_fallback
  end
```
... the call to render raises `ArgumentError: Invalid formats: nil`.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -203,7 +203,8 @@ class ApplicationController < ActionController::Base
   end
 
   def formats_with_html_fallback
-    request.formats.map(&:symbol) + [:html]
+    # request.formats returns a collection of Mime::Type objects, which can have nil value for @symbol
+    request.formats.map(&:symbol).compact + [:html]
   end
 
 end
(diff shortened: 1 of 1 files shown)


KaraAJC (approved):

ahorner (approved):

giladshanan:
On further digging, I found that this problem is coming from places where the front end makes calls to the server with mime type `"*/*"`.  This gets instantiated as a `Mime::Type` with symbol value of `nil`.

See
https://github.com/rails/rails/blob/78e402077d4d69b08db7daf8944a32eef0cf6ac4/actionpack/lib/action_dispatch/http/mime_type.rb#L255
and
https://github.com/rails/rails/blob/6-0-stable/actionview/lib/action_view/lookup_context.rb#L284

ahorner (approved):

PR #2890, opened 2022-01-25
[#155215] Specify a stable version and use cdn.form.io

giladshanan:
As recommended by form.io support team

### DIFF
--- a/app/views/layouts/formio.html.haml
+++ b/app/views/layouts/formio.html.haml
@@ -3,9 +3,10 @@
   %head
     %meta{ "http-equiv" => "content-type", "content" => "text/html; charset=UTF-8" }
     = render_view_hook "head"
-    = stylesheet_link_tag "https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css", media: "all"
-    = stylesheet_link_tag "https://unpkg.com/formiojs@3.28.0/dist/formio.full.min.css", media: "all"
-    = javascript_include_tag "https://unpkg.com/formiojs@3.28.0/dist/formio.full.min.js"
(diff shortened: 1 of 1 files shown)


app/views/layouts/formio.html.haml
```diff
@@ -3,9 +3,10 @@
   %head
     %meta{ "http-equiv" => "content-type", "content" => "text/html; charset=UTF-8" }
     = render_view_hook "head"
-    = stylesheet_link_tag "https://maxcdn.bootstrapcdn.com/bootstrap/3.3.7/css/bootstrap.min.css", media: "all"
-    = stylesheet_link_tag "https://unpkg.com/formiojs@3.28.0/dist/formio.full.min.css", media: "all"
-    = javascript_include_tag "https://unpkg.com/formiojs@3.28.0/dist/formio.full.min.js"
+    - # using resource links based on https://github.com/formio/formio.js#form-building
```
boatrite, line 6: I think this technically should be `-#` for a haml comment

boatrite (approved):

app/views/layouts/formio.html.haml
giladshanan: TIL 🎉

giladshanan: I wonder if HAML actually parses them differently

boatrite: Comments are likely stripped earlier, whereas non-haml-comments ruby comments have to go through the ruby parser, and so is presumably slower to get the final result

giladshanan, line 6:
```suggestion
    -# using resource links based on https://github.com/formio/formio.js#form-building
```

PR #2913, opened 2022-02-09
[#159818][WIP] Update mailer syntax

madcowley:
# Release Notes

This updates the mailer syntax to use .with

Need to confirm potential issue here after upgrading to ruby 3.0.  If we have issues with that upgrade, this PR may help solve them.

### DIFF
--- a/app/controllers/concerns/csv_email_action.rb
+++ b/app/controllers/concerns/csv_email_action.rb
@@ -25,7 +25,7 @@ module CsvEmailAction
     yield_email_and_respond_for_report do |email|
       # TODO: in order to use ActiveJob's `deliver_later`, the report needs to be
       # specifically serializable, which many of our reports are not.
-      CsvReportMailer.delay.csv_report_email(email, report)
+      CsvReportMailer.with(to_address: email, report: report).delay.csv_report_email
     end
   end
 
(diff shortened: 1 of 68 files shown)


app/controllers/concerns/csv_email_action.rb
```diff
@@ -25,7 +25,7 @@ def queue_csv_report_email(report)
     yield_email_and_respond_for_report do |email|
       # TODO: in order to use ActiveJob's `deliver_later`, the report needs to be
       # specifically serializable, which many of our reports are not.
-      CsvReportMailer.delay.csv_report_email(email, report)
+      CsvReportMailer.with(to_address: email, report: report).delay.csv_report_email
```
rdunlop, line 28: Is `delay` the old way of doing `deliver_later`? Can we update this to use `deliver_later` (I don't know whether they are equivalent).

rdunlop (changes requested):

app/controllers/concerns/csv_email_action.rb
```diff
@@ -25,7 +25,7 @@ def queue_csv_report_email(report)
     yield_email_and_respond_for_report do |email|
       # TODO: in order to use ActiveJob's `deliver_later`, the report needs to be
       # specifically serializable, which many of our reports are not.
-      CsvReportMailer.delay.csv_report_email(email, report)
+      CsvReportMailer.with(to_address: email, report: report).csv_report_email.deliver_later
```
rdunlop, line 28: Perhaps this comment was here on purpose, and that's why we should stick with `.delay`? I think that I'd like @giladshanan to weigh in on this.

giladshanan:
We are running ruby 3 without these changes, closing

PR #2914, opened 2022-02-10
[#158143] Change hourly rate precision

petebrooks:
# Release Notes

Change the hourly rate precision to 8 decimal places (from 4). Ensure that displayed price is truncated to 4 digits (not rounded).


giladshanan (changes requested):
One small suggestion, and looks like some specs need updating.

giladshanan (approved):
👍🏼

PR #2925, opened 2022-02-16
[#159255] Give time data inputs more space

giladshanan:
# Release Notes

Reservation order details can have date and time inputs in some cases, and they need more space on the page.

# Screenshots
## With date/time inputs + split account
Screen Shot 2022-02-16 at 9 14 08 AM

## No date inputs, no split account
Screen Shot 2022-02-16 at 9 15 25 AM

### DIFF
--- a/app/helpers/accounts_helper.rb
+++ b/app/helpers/accounts_helper.rb
@@ -3,8 +3,7 @@
 module AccountsHelper
 
   def account_input(form)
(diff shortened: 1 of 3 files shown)


rdunlop (approved):

giladshanan:
@aaronrosen26 The time inputs for occupancies will appear in the same wider column as reservations, so we should be good on that front.

PR #2927, opened 2022-02-17
[#158143] Change usage subsidy precision

petebrooks:
# Release Notes

Change the usage subsidy precision to 8 decimal places.

PR #2932, opened 2022-02-25
[#159929] Add rake task for updating price policy usage_rate and usage_subsidy

giladshanan:
# Release Notes

A task to reset values so we can start using improved precision.

See https://github.com/tablexi/nucore-open/pull/2914 and https://github.com/tablexi/nucore-open/pull/2927

### DIFF
--- a/lib/tasks/price_policies.rake
+++ b/lib/tasks/price_policies.rake
@@ -24,7 +24,7 @@ namespace :price_policies do
   task :update_usage_rate, [:commit] => :environment do |_t, args|
     commit = args[:commit].to_s == "commit"
 
-    policies = PricePolicy.current.where.not(usage_rate: nil).count
+    policies = PricePolicy.current.where.not(usage_rate: nil)
 
     puts "Updating #{policies.count} price rules ..."
     policies.each do |policy|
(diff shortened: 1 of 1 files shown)


lib/tasks/price_policies.rake
```diff
@@ -18,4 +18,27 @@ namespace :price_policies do
...
+
+    puts "Updating #{policies.count} price rules ..."
+    policies.each do |policy|
+      if policy.usage_rate.nil? || policy.usage_subsidy.nil?
+        puts "Skipped #{policy.id}"
+      else
+        puts "#{policy.id} Before: #{policy.usage_rate.to_f}, #{policy.usage_subsidy.to_f}"
+        policy.usage_rate = policy.hourly_usage_rate.to_f.round(2)
+        policy.usage_subsidy = policy.hourly_usage_subsidy.to_f.round(2)
+        policy.save if commit
```
rdunlop, line 37: if this fails, it will return `false`, would you rather use `save!` so that it is obvious that it failed? or check the return code and `puts` if it fails? (so that you don't stop the update for others)

rdunlop (approved):

PR #2953, opened 2022-03-15
[#158590] Tech Task - Add rubocop-rails and -performance, update rule names

giladshanan:
# Release Notes

Updates rubocop for development use.

Follow up work will involve:
- [ ] upgrading the rubocop gem 
- [ ] team discussion around style rules
- [ ] address rubocop todo's


giladshanan (commented):
👍🏼

PR #2957, opened 2022-03-18
[#158793] Refactor recurring tasks

giladshanan:
# Release Notes

This consolidates 1 and 5 minutes recurring tasks into a single daemon process.

# Additional Context

- We have had some issues with the pattern of using curl to call ourselves.  SSL certs, which we may or may not have control over, could result in these jobs failing silently.  Using a daemon allows better monitoring (via `eye` or k8s).
- Maintains the ability for engines to hook in and add tasks.
- As in the previous implementation, this does not require loading the application every ~60 seconds.


app/models/recurring_task_config.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+class RecurringTaskConfig
+
+  # Returns an array of recurring task classes and their calling methods.
+  # Engines can append to this list.
+  def self.recurring_tasks
+    @@recurring_tasks ||= [
+      [AutoExpireReservation, :perform],
```
giladshanan, line 9: Some of these were running every 5 minutes, but in testing they are pretty fast (<1 second each) and I couldn't see any problem in running them more frequently, other than reducing the general load.  I could see adding an element here to configure cadence like:
`[AutoExpireReservation, :perform, 5]`
... and then we could maintain the same behavior as before.  Thoughts?

lib/daemons/recurring_tasks.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+require File.expand_path("base", File.dirname(__FILE__))
+
+Daemons::Base.new("recurring_tasks").start do
+  RecurringTaskConfig.recurring_tasks.each do |(worker, method)|
+    worker.new.public_send(method)
```
ahorner, line 7: It feels like there's room for an abstraction here. Maybe an internal `Task` class in our `RecurringTaskConfig` that handles the method chaining we're doing here? Our config could then implement `Enumerable`, and yield a series of tasks, which could let us simplify this daemon to something like `RecurringTaskConfig.each(&:invoke!)`.

lib/daemons/recurring_tasks.rb
```diff
@@ -0,0 +1,12 @@
...
+  end
+
+  sleep 1.minute.to_i
```
rdunlop, line 10: I thought that it was somewhat important to run as close to the top-of-the-minute as possible, for billing reasons....so, wouldn't we want this to sleep only as many seconds as needed to get to the next minute? (maybe I'm being overly complicated)

config/deploy.rb
```diff
@@ -14,3 +13,13 @@
 set :linked_dirs, fetch(:linked_dirs, []).concat(
   %w(bin log tmp/pids tmp/cache tmp/sockets vendor/bundle public/system public/files),
 )
+
+namespace :eye do
+  task :set_recurring_tasks do
+    on roles :all do |host|
+      set :eye_env, -> { { rails_env: fetch(:rails_env), recurring: host.roles_array.include?(:db) } }
```
rdunlop, line 20: why do this complicated thing, when line 19 seems to allow for `on roles :db do |host|`, no?

rdunlop (approved):

lib/daemons/recurring_tasks.rb
giladshanan: I think this is a good idea, to be safe

PR #2966, opened 2022-03-24
[#158844] Updates environment cache settings & c2po loading 

jossim:
# Release Notes

Because of the way classes and code were loaded, there was an issue with the c2po engine that caused the menu to load incorrectly, producing duplicate menu items. This was not noticed in development.

The `development` and `test` environments have been updated to better match `stage` and `production` and the issue with c2po has been corrected.

# Screenshot

On stage
Screen Shot 2022-03-24 at 3 11 31 PM

In development, now.
Screen Shot 2022-03-24 at 5 09 14 PM

### DIFF
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -6,10 +6,10 @@ Rails.application.configure do
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.cache_classes = true
 
   # Do not eager load code on boot.
-  config.eager_load = false
+  config.eager_load = true
 
   # Show full error reports.
   config.consider_all_requests_local = true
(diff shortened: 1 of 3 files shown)


config/environments/development.rb
```diff
@@ -6,10 +6,10 @@
   # In the development environment your application's code is reloaded on
   # every request. This slows down response time but is perfect for development
   # since you don't have to restart the web server when you make code changes.
-  config.cache_classes = false
+  config.cache_classes = true
```
giladshanan, line 9: I prefer setting this to false most of the time.  It slows down development to have to reset the server on each code change.  We can always set it to `true` for rails upgrades and such.  
```suggestion
  config.cache_classes = false
```

giladshanan (commented):
I don't want to change the `cache_classes` value for development.  Thoughts?

giladshanan (approved):

PR #2972, opened 2022-03-30
[#159295] Upgrade to Rails 6.1

jossim:
# Release Notes

Upgrades from 6.0 to 6.1. This largely involved removing some methods that no longer exist (`add_template_helper` and `fixture_file_upload`) and updating configuration files based on `rails app:update`


config/initializers/backtrace_silencers.rb
```diff
@@ -3,7 +3,8 @@
...
 
 # You can add backtrace silencers for libraries that you're using but don't wish to see in your backtraces.
-# Rails.backtrace_cleaner.add_silencer { |line| line =~ /my_noisy_library/ }
+# Rails.backtrace_cleaner.add_silencer { |line| /my_noisy_library/.match?(line) }
 
-# You can also remove all the silencers if you're trying to debug a problem that might stem from framework code.
-# Rails.backtrace_cleaner.remove_silencers!
+# You can also remove all the silencers if you're trying to debug a problem that might stem from framework code
+# by setting BACKTRACE=1 before calling your invocation, like "BACKTRACE=1 ./bin/rails runner 'MyClass.perform'".
+Rails.backtrace_cleaner.remove_silencers! if ENV["BACKTRACE"]
```
giladshanan, line 10: missing newline

config/initializers/filter_parameter_logging.rb
```diff
@@ -3,4 +3,6 @@
 # Be sure to restart your server when you modify this file.
 
 # Configure sensitive parameters which will be filtered from the log file.
-Rails.application.config.filter_parameters += [:password]
+Rails.application.config.filter_parameters += [
+  :password, :secret, :token, :_key, :crypt, :salt, :certificate, :otp, :ssn
+]
```
giladshanan, line 8: missing newline

db/schema.rb
```diff
@@ -12,7 +12,7 @@
 
 ActiveRecord::Schema.define(version: 2022_02_16_233321) do
 
-  create_table "account_facility_joins", id: :integer, options: "ENGINE=InnoDB DEFAULT CHARSET=utf8", force: :cascade do |t|
+  create_table "account_facility_joins", id: :integer, charset: "utf8mb3", force: :cascade do |t|
```
giladshanan, line 15: Why are these changing?

giladshanan (changes requested):
This is looking good, just a couple of small comments/questions.  There are some spec failures that need to be addressed.

config/application.rb
```diff
@@ -13,7 +13,7 @@ module Nucore
 
   class Application < Rails::Application
 
-    config.load_defaults 6.0
+    config.load_defaults 6.1
```
rdunlop, line 16: Have you gone through all of the rails 6.1 defaults to ensure that they work well, for example:

```
# Specify cookies SameSite protection level: either :none, :lax, or :strict.
#
# This change is not backwards compatible with earlier Rails versions.
# It's best enabled when your entire app is migrated and stable on 6.1.
# Rails.application.config.action_dispatch.cookies_same_site_protection = :lax

# Generate CSRF tokens that are encoded in URL-safe Base64.
#
# This change is not backwards compatible with earlier Rails versions.
# It's best enabled when your entire app is migrated and stable on 6.1.
# Rails.application.config.action_controller.urlsafe_csrf_tokens = true
```

db/schema.rb
jossim: @giladshanan this was an automatic change from `rails app:update`.

It looks like "utf8" is a lie in MySQL and is an alias for "utf8mb3", which is 3 bytes instead of 4, so I believe this is just making that explicit. True utf8 is actually called "utf8mb4."

https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html

https://adamhooper.medium.com/in-mysql-never-use-utf8-use-utf8mb4-11761243e434

giladshanan: Interesting.  Did you run `rails db:migrate` to see if these changes persist?

config/application.rb
giladshanan: @rdunlop I think these are ok to test in production - using [this article](https://lilyreile.medium.com/rails-6-1-new-framework-defaults-what-they-do-and-how-to-safely-uncomment-them-c546b70f0c5e) as a reference.

config/environments/production.rb
```diff
@@ -87,6 +87,12 @@
   # Send deprecation notices to registered listeners.
   config.active_support.deprecation = :notify
 
+  # Log disallowed deprecations.
+  config.active_support.disallowed_deprecation = :log
```
giladshanan, line 91: I don't think we want deprecation warnings in production logs - can this be set to `:none`?

config/initializers/zeitwerk.rb
```diff
@@ -1,3 +1,5 @@
-Rails.autoloaders.main.inflector.inflect(
-  "csv_helper" => "CSVHelper"
-)
\ No newline at end of file
+Rails.autoloaders.each do |autoloader|
+  autoloader.inflector.inflect(
+    "csv_helper" => "CSVHelper",
+  )
+end
```
giladshanan, line 5: newline

giladshanan (commented):
Couple small comments

config/environments/production.rb
jossim: `:none` doesn't appear to be an option, so I changed it to `:silence`

giladshanan (approved):
🎉

PR #2988, opened 2022-04-14
Updates mini_racer & removes libv8

jossim:
# Release Notes

The version of `mini_racer` depends on `libv8`, which is not under active development and does not work on ARM. This updates `mini_racer` to the latest version, which depends on `libv8-node`, which is under active development and works on ARM.


Dockerfile
```diff
@@ -16,6 +16,9 @@ COPY Gemfile Gemfile.lock .ruby-version /app/
 # We reference the engines in the Gemfile, so we need them to be there, too
 COPY vendor/engines /app/vendor/engines
 
+# Install Bundler 2
+RUN gem install bundler --version=2.3.11
```
rdunlop, line 20: Instead of doing this, can we read the version of bundler from the `Gemfile.lock`?
```suggestion
RUN grep "BUNDLED WITH" Gemfile.lock; if [ $? -eq 0 ]; then gem install bundler -v "$(grep -A 1 "BUNDLED WITH" Gemfile.lock | tail -n 1)"; fi
```

rdunlop (changes requested):

Gemfile
```diff
@@ -41,11 +41,7 @@ gem "paranoia"
 gem "sass-rails"
 gem "coffee-rails"
 gem "uglifier", "= 4.1.18" # 4.1.19 has an issue https://github.com/mishoo/UglifyJS2/issues/3245
-# TO DO: consider using nodejs instead of mini_racer
-# libv8 8+ does not like to compile on our boxes. It's easier to lock down
-# libv8 and mini_racer for now than to try to get them upgraded on the server.
-gem "mini_racer", "< 0.3"
-gem "libv8", "< 8"
+gem "mini_racer", "~> 0.6.2"
```
giladshanan, line 44: 🎉

giladshanan (commented):
I agree with @rdunlop's comment, otherwise this looks good.

Dockerfile
jossim: I've updated this and am able to get through most of the Dockerfile steps, though because I'm on M1, I'm unable to fully run docker-compose.

Can someone double check this and see if they have any issues?

giladshanan (approved):
I tested this locally and was able to start the application with docker compose.  Just need to address the merge conflict.

PR #2991, opened 2022-04-20
[#159884] Remove mini_racer

giladshanan:
# Release Notes

Using node instead of mini_racer, as mini_racer is failing to compile on the TXI stage server.


rdunlop:
You may want to add "NodeJS" to the README section about tools installed, next to "Ruby"

PR #2994, opened 2022-04-22
Tech task - Use .bashrc to determine which servers should run recurring_tasks

giladshanan:
# Release Notes

I've discovered 2 problems with the implementation for managing which servers run recurring tasks:
1) capistrano sets the ENV variable during deployment, but restarting the application manually with `eye-patch load config/eye.yml.erb` will not have the variable set, so recurring_tasks will not be reloaded
2) The recurring_tasks are running on both UMass production servers
Rather than debug the current approach, I'm using `.bashrc` to control the behavior.  
This is explicit, easy to reason about, and it works.

# Additional Context

I've updated `.bashrc` on UMass and TXI servers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ Because we use squash and merge, you should be able to see the changes by lookin
 at the [commit log](https://github.com/tablexi/nucore-open/commits/master). However, we have begun keeping track of breaking changes
 or optional rake tasks.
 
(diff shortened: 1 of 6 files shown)


rdunlop (commented):
I wonder what `eye` will think of the server which tries to run the daemon, but it exits immediately (because it's not configured to run). Will it consider that a "crash" and try to restart it?

ahorner (approved):

PR #3001, opened 2022-04-27
Update rails to 6.1.5.1 for security warning

rdunlop:


giladshanan:
See https://github.com/rails/rails/pull/45027

PR #3007, opened 2022-05-09
[#158778] improve payment source search

jossim:
# Release Notes

`@count` was erroneously removed, which caused an error when the
search_results template was rendered and the number of results was over
the limit.

This adds an `additional_results_notice` helper which checks its inputs
before they are used.

### DIFF
--- a/app/controllers/account_price_group_members_controller.rb
+++ b/app/controllers/account_price_group_members_controller.rb
@@ -13,7 +13,9 @@ class AccountPriceGroupMembersController < ApplicationController
     searcher = AccountSearcher.new(params[:search_term], scope: Account.for_facility(current_facility))
 
     if searcher.valid?
-      @accounts = searcher.results
+      @search_results = searcher.results
+      @count = @search_results.count
+      @accounts = @search_results.limit(@limit)
     else
       flash.now[:errors] = "Search terms must be 3 or more characters."
     end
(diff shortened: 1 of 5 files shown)


app/controllers/account_price_group_members_controller.rb
```diff
@@ -13,7 +13,9 @@ def search_results
     searcher = AccountSearcher.new(params[:search_term], scope: Account.for_facility(current_facility))
 
     if searcher.valid?
-      @accounts = searcher.results
+      @search_results = searcher.results
+      @count = @search_results.count
+      @accounts = @search_results.limit(@limit)
```
giladshanan, line 18: No need for an ivar here, we're not using `search_results` in the view:
```suggestion
      search_results = searcher.results
      @count = search_results.count
      @accounts = search_results.limit(@limit)
```

spec/helpers/account_price_group_members_helper_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      expect(additional_results_notice(count: nil, limit: 25)).to be_nil
+      expect(additional_results_notice(count: 14, limit: nil)).to be_nil
+      expect(additional_results_notice(count: 14, limit: 25)).to be_nil
+    end
+
+    it "returns nil if the accounts_count isn't over the limit" do
+      expect(additional_results_notice(count: 12, limit: 25)).to be_nil
+	end
+
+    it "returns a notice if accounts_count is over the limit" do
```
giladshanan, line 17: Looks like some indenting issues here.  Can you set 2 spaces as the default in VSCode?

giladshanan (commented):
Looks good, just a couple small nitpicks

giladshanan (approved):

PR #3015, opened 2022-05-16
[#158988] Adds setting for stage header color

jossim:
# Release Notes

Sometimes, it's not obvious that you're working in the stage environment or the production environment.

This makes the stage environment more clear by adding an optional setting which changes the color of the site's header when it's in the stage environment.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,8 +21,8 @@ module ApplicationHelper
   end
 
   def staging_header_styles
-    apply_stage_color = (Rails.env == "stage" && Settings.stage_color.present?)
-    "background-image: none; background-color: #{Settings.stage_color}" if apply_stage_color
+    stage_color = ENV.fetch("stage_color", nil)
+    "background-image: none; background-color: #{stage_color}" if stage_color
   end
 
   def sortable(column, title = nil)
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -21,7 +21,7 @@ module ApplicationHelper
   end
 
   def staging_header_styles
-    stage_color = ENV.fetch("stage_color", nil)
+    stage_color = Settings.stage_color
     "background-image: none; background-color: #{stage_color}" if stage_color
   end
 
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -20,6 +20,11 @@ module ApplicationHelper
     (full_title + app_name).html_safe
   end
 
+  def staging_header_styles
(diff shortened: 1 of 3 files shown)


app/views/shared/_header.html.haml
```diff
@@ -1,5 +1,6 @@
 .navbar.navbar-static-top
-  .navbar-inner
+  - styles = (Rails.env == "stage" && Settings.stage_color.present?) ? "background-image: none; background-color:  #{Settings.stage_color}" : ""
```
giladshanan, line 2:
```suggestion
  - apply_stage_color = Rails.env.staging? && Settings.stage_color.present?
  - stage_style = apply_stage_color ? "background-image: none; background-color:  #{Settings.stage_color}" : ""
```

giladshanan (approved):
I recommend breaking up the logic into 2 lines, what do you think?

app/views/shared/_header.html.haml
rdunlop: It appears that the env is `stage` not `staging`, so it would be `Rails.env.stage?`, not `Rails.env.staging?`

jossim: @giladshanan @rdunlop I added a helper method to make the view a bit cleaner https://github.com/tablexi/nucore-open/pull/3015/files#diff-5f3fc5e3977d242572aa1d08551f5eb557de0ccaff30370838ee9df5386ea0daR23

app/helpers/application_helper.rb
```diff
@@ -20,6 +20,11 @@ def html_title(title = nil)
     (full_title + app_name).html_safe
   end
 
+  def staging_header_styles
+    apply_stage_color = (Rails.env == "stage" && Settings.stage_color.present?)
+    "background-image: none; background-color: #{Settings.stage_color}" if apply_stage_color
```
giladshanan, line 25:
```suggestion
    stage_color = ENV.fetch("stage_color", nil)
    "background-image: none; background-color: #{stage_color}" if stage_color
```

giladshanan (changes requested):
Let's read from the ENV as Robin suggests.

rdunlop (approved):

config/settings.yml
```diff
@@ -1,5 +1,6 @@
 header_logo_path: ~
 display_app_name_in_header: true
+stage_color: ~
```
giladshanan, line 3: Let's read from ENV here, where devs might expect to find an option like this.
Can you test what happens when no value is set?
```suggestion
stage_color: <%= ENV.fetch("STAGE_COLOR", nil) %>
```

PR #3021, opened 2022-05-23
[#153891] TimedServices adjustment in product bundles

jossim:
# Release Notes

When TimedServices are part of a bundle, the time is preset, it's not possible for a user to adjust the amount of time they'd like.

This updates the behavior of bundled TimedServices so that the time can be adjusted by the user on checkout.

# Screenshot

Screen Shot 2022-05-31 at 12 48 56 PM

### DIFF
--- a/app/views/bundle_products/new.html.haml
+++ b/app/views/bundle_products/new.html.haml
@@ -18,7 +18,7 @@
       required: true,
       collection: @bundle.products_for_group_select,
       group_method: :last,
-      input_html: { class: "js--bundleProducts__productSelect", data: { timed_quantity_types: TimedService.model_name.human(count: 2) } }
+      input_html: { class: "js--bundleProducts__productSelect", data: { timed_quantity_types: '' } }
 
     .js--bundleProducts__quantityField
       = f.input :quantity, required: true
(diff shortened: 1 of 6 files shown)


spec/system/placing_an_order_spec.rb
```diff
@@ -114,8 +114,8 @@ def add_to_cart
 
       it "adds the item as a single line item" do
```
giladshanan, line 115: I think this spec needs a new name

spec/system/placing_an_order_spec.rb
```diff
@@ -114,8 +114,8 @@ def add_to_cart
...
         add_to_cart
-        expect(page).to have_content(timed_service.name).once
-        expect(page).to have_content("1:30")
+        expect(page).to have_content(timed_service.name)
```
giladshanan, line 117: How many times does the content appear?

giladshanan: Expect "00:01"
update one of the service quantity inputs and submit

app/views/bundle_products/new.html.haml
```diff
@@ -18,7 +18,7 @@
       required: true,
       collection: @bundle.products_for_group_select,
       group_method: :last,
-      input_html: { class: "js--bundleProducts__productSelect", data: { timed_quantity_types: TimedService.model_name.human(count: 2) } }
+      input_html: { class: "js--bundleProducts__productSelect", data: { timed_quantity_types: '' } }
```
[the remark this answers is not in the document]
giladshanan: I'm ok with cleaning this up in a follow up PR

giladshanan (approved):
LGTM, pending CI

PR #3030, opened 2022-05-27
[#160264] Add axe-core specs

giladshanan:
# Release Notes

Adds a handful of accessibility specs and a few small accessibility improvements to make those specs pass.

### DIFF
--- a/app/views/layouts/two_column.html.haml
+++ b/app/views/layouts/two_column.html.haml
@@ -11,8 +11,10 @@
       .container
         .row
           .span12
-            %h1= yield :h1
-            %h2= yield :h2
+            - if yield(:h1).present?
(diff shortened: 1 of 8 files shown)


jossim (approved):

app/views/layouts/two_column.html.haml
```diff
@@ -11,8 +11,10 @@
       .container
         .row
           .span12
-            %h1= yield :h1
-            %h2= yield :h2
+            - if yield(:h1).present?
```
rdunlop, line 14: there's a better helper method for this. I think this is clearer:
```suggestion
            - if content_for?(:h1)
```

rdunlop (changes requested):

PR #3048, opened 2022-06-21
[#160400] Simplify logic for user type (internal/external)

giladshanan:
# Release Notes

We currently have 2 ways to determine if a user is internal or external - checking for a stored password or comparing username and email.  We'd like to be able to support SSO that uses email as username, and this allows that while also moving to a single way of checking user type.  

I've confirmed this change with all the schools.  Some schools will benefit from a little user record cleanup after this is released.

### DIFF
--- a/app/forms/user_form.rb
+++ b/app/forms/user_form.rb
@@ -42,7 +42,7 @@ class UserForm < SimpleDelegator
   end
 
   def admin_editable?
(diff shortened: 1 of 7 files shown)


jossim (approved):

jossim:
Looks good, pending specs passing

PR #3066, opened 2022-06-29
[#153891] Adds dynamic hint logic to new bundle product view

jossim:
# Release Notes

For Timed Services, the quantity field may be confusing for some users, since its meaning has recently changed.

This adds some informational hint text that dynamically appears when a Timed Service is being added to the bundle. It also makes it trivial to add hints for other product types.

# Screenshot

Screen Shot 2022-07-06 at 4 23 03 PM

### DIFF
--- a/app/assets/javascripts/app/bundle_products.js
+++ b/app/assets/javascripts/app/bundle_products.js
@@ -1,34 +1,42 @@
-// On bundle_products#new, switch between a regular quantity field and a time
-// quantity field based on the type of product selected.
-$(function() {
-  $productSelect = $(".js--bundleProducts__productSelect");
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/app/bundle_products.js
```diff
@@ -2,29 +2,33 @@
 // quantity field based on the type of product selected.
 $(function() {
   $productSelect = $(".js--bundleProducts__productSelect");
+  const hints = document.querySelectorAll('.hint');
```
giladshanan, line 5: Let's use " instead of ' for consistency

giladshanan (approved):

PR #3078, opened 2022-07-11
Disables class caching in the development environment

jossim:
# Release Notes

Cached classes requires a lot of server restarts when developing. This disables class caching.

It also adjusts some loading behavior of the `C2po` engine so the application works correctly with respect to the caching. The way `C2po` was setup to load, `AccountConfig`'s reconcilable accounts would disappear. This make the reconcilable account code a bit more clear.

### DIFF
--- a/app/models/account_config.rb
+++ b/app/models/account_config.rb
@@ -30,7 +30,7 @@ class AccountConfig
   end
 
   def reconcilable_account_types
-    @reconcilable_account_types ||= statement_account_types.map(&:constantize).select(&:reconcilable?).map(&:to_s)
+    statement_account_types.uniq.map(&:constantize).select(&:reconcilable?).map(&:to_s)
   end
 
   # Returns an array of subclassed Account object names that support affiliates.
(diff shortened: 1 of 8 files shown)


app/support/reconcilable_account.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
 module ReconcilableAccount
-
+  def reconcilable?
```
giladshanan, line 4: I think it's worth adding a comment here to make it more obvious that this is a class method

app/models/account_config.rb
```diff
@@ -30,7 +30,7 @@ def statement_account_types
   end
 
   def reconcilable_account_types
-    statement_account_types.map(&:constantize).select { |t| t < ReconcilableAccount }.map(&:to_s)
+    statement_account_types.uniq.map(&:constantize).select(&:reconcilable?).map(&:to_s)
```
giladshanan, line 33: Did you decide not to memo'ize here?

giladshanan (approved):
Couple small comments for you

app/models/account_config.rb
giladshanan: I'm fine with `uniq` as long as it works, no need to monkey patch.

PR #3114, opened 2022-08-10
Adds spec changes from OSU

jossim:
# Release Notes

These changes were needed because OSU has `price_policy_note_options` and `order_detail_price_change_reason_options` drop down menus, so the selectors are slightly different.


vendor/engines/secure_rooms/spec/system/admin_edit_occupancy_spec.rb
```diff
@@ -34,7 +34,7 @@
     let(:entry_at) { Time.zone.parse("2017-09-12 08:30") }
     let(:exit_at) { Time.zone.parse("2017-09-12 09:30") }
 
-    it "can change the times" do
+    it "can change the times", :blank_price_change_reason_options do
```
jossim, line 37: This change is in OSU, but I'd guess we don't want it here, since if we ever have price change reason options, this test will ignore them.

giladshanan: Good point, let's take that out.

giladshanan (approved):

PR #3115, opened 2022-08-12
Adjusts _account_fields

jossim:
# Release Notes

It did not work correctly for existing accounts. This update correctly displays the account number for existing accounts.


giladshanan (approved):
Pending CI

PR #3119, opened 2022-08-17
Replace `account_field_value` with `account_number_part_value_or_default`

jossim:
# Release Notes

This replaces `account_field_value` in a helper with `account_number_part_value_or_default` in `Accounts::AccountNumberSectionable`

Due to special cases, `account_field_value` needed to be much more complicated to work in the needed scenarios -- on edit, and when there's only a single account number part. Pushing the logic to the `Account` model, via the `Accounts::AccountNumberSectionable` module is a simpler solution.


giladshanan:
@jossim looks like a spec failure here

giladshanan (approved):

PR #3122, opened 2022-08-19
[#160499] Change revenue_account to string on FacilityAccount

jossim:
# Release Notes

Some revenue accounts start with a leading 0. However, since `revenue_account` on `FacilityAccount` is an integer, it's not possible to save those. This changes `revenue_account` from an integer to a string, which will allow accounts with leading 0s.

### DIFF
--- a/db/migrate/20220819183819_change_revenue_account_to_string_in_facility_accounts.rb
+++ b/db/migrate/20220819183819_change_revenue_account_to_string_in_facility_accounts.rb
@@ -1,0 +1,9 @@
+class ChangeRevenueAccountToStringInFacilityAccounts < ActiveRecord::Migration[6.1]
+  def up
+    change_column :facility_accounts, :revenue_account, :string
+  end
+
+  def down
+    change_column :facility_accounts, :revenue_account, :integer
+  end
+end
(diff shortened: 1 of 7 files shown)


db/migrate/20220819183819_change_revenue_account_to_string_in_facility_accounts.rb
```diff
@@ -0,0 +1,9 @@
+class ChangeRevenueAccountToStringInFacilityAccounts < ActiveRecord::Migration[6.1]
+  def up
+    change_column :facility_accounts, :revenue_account, :string
```
jossim, line 3: Looks like this change will probably have the most implications in Northwestern.

giladshanan: What happens to existing data when this is run?  I think this change should happen in the OSU repo, no need to change this for other schools if they dont need it

jossim: I tested migrating this up and down on existing data. It's converted between strings and integers, and doesn't seem to have any issues.

jossim:
Closing this in favor of https://github.com/Oregon-State/nucore-osu/pull/160

PR #3135, opened 2022-08-26
Tech task - Fix recharge spec

giladshanan:
the have_field matcher doesn't handle nil/empty string value checking:
https://github.com/teamcapybara/capybara/pull/1169#issuecomment-44575123


giladshanan:
@jossim I tested this in every repo, and it passes

jossim (approved):

PR #3136, opened 2022-08-26
[#160776] Include username and email when NETID login fails

giladshanan:
# Release Notes

To help clarify/confirm for the user and assist in debugging, it would be helpful to include the username and email that is returned from the SAML response when SSO fails.
Also logs a warning to rollbar with the complete user attributes.

### DIFF
--- a/vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
+++ b/vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
@@ -22,7 +22,7 @@ module Devise
       # authentication to call fail!(:saml_invalid), which allows us to specify a separate
       # failure message for authentication via SAML.
       def failed_auth(msg)
-        Rollbar.warn(msg, attributes: failed_auth_user_attributes) if defined?(Rollbar)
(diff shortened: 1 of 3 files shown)


vendor/engines/saml_authentication/lib/saml_authentication/saml_authenticatable_with_custom_error.rb
```diff
@@ -22,11 +22,26 @@ class SamlAuthenticatableWithCustomError < SamlAuthenticatable
       # authentication to call fail!(:saml_invalid), which allows us to specify a separate
       # failure message for authentication via SAML.
       def failed_auth(msg)
+        Rollbar.warn(msg, attributes: failed_auth_user_attributes) if defined?(Rollbar)
```
giladshanan, line 25: This might get noisy - what do you think @jossim ?

jossim: @giladshanan do we know anything else about the error? If so, perhaps specific errors that are not of interest could be filtered out or something like that.

giladshanan: @jossim good call

jossim (approved):

PR #3145, opened 2022-09-09
[#160767] Adds NetID as a term in the translation file

jossim:
# Release Notes

Not every school calls their online SSO account "NetID". This reorganizes the translation files so that "NetID" is easy to change to something else.

The order import template contains "Netid" and is a static file. So this also adds `order_import_template_name` to the `settngs.yml` file, in case a school would like to change to its own template with school specific language.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -230,7 +230,7 @@ en:
   facility_users:
     title: "!Facility! Staff"
     search:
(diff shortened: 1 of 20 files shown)


config/locales/en.yml
```diff
@@ -29,6 +29,10 @@ en:
   facility_downcase: facility
   facilities_downcase: facilities
 
+  Online_id: NetID
```
giladshanan, line 32: What about `sso_id`?  Or `netid`?

giladshanan (approved):
I'm leaning towards `netd` as the translation key (rather than online_id) but that's my only nitpick.  Looks great!

PR #3162, opened 2022-09-20
Translation updates

jossim:
# Release Notes

Cherry picks spec changes from https://github.com/Oregon-State/nucore-osu/pull/185. Also removes `_upcase` from translation keys.

There's an adjustment in the override file here that needs to be reflected in other schools that are using it

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -29,20 +29,20 @@ en:
   facility_downcase: facility
   facilities_downcase: facilities
 
-  Sso_id: NetID
-  sso_id_downcase: NetID
-  name_or_sso_id: name, NetID, or username
+  Sso_id: OID
+  sso_id_downcase: oid
+  name_or_sso_id: name, OID, or username
 
-  Chart_string: Chart String
-  Chart_strings: Chart Strings
-  chart_string_downcase: chart string
-  chart_strings_downcase: chart strings
+  Chart_string: Index Account
+  Chart_strings: Index Accounts
+  chart_string_downcase: index account
+  chart_strings_downcase: index accounts
 
-  Statement: Statement
-  Statements: Statements
-  statement_downcase: statement
-  statements_downcase: statements
-  statemented_downcase: statemented
+  Statement: Invoice
+  Statements: Invoices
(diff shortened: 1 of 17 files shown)


giladshanan:
Looking good, just a few spec failures

config/locales/en.yml
```diff
@@ -29,17 +29,17 @@ en:
   facility_downcase: facility
   facilities_downcase: facilities
 
-  sso_id_upcase: NetID
-  sso_id_downcase: NetID
-  name_or_sso_id: name, NetID, or username
+  Sso_id: OID
+  sso_id_downcase: oid
+  name_or_sso_id: name, OID, or username
```
giladshanan, line 34:
```suggestion
  Sso_id: NetID
  sso_id_downcase: NetID
  name_or_sso_id: name, NetID, or username
```

config/locales/override/en.yml
```diff
@@ -12,17 +12,18 @@ en:
...
+  #       one: Statement
+  #       other: Statements
+  #     facility_account:
+  #       one: Recharge Chart String
+  #       other: Recharge Chart Strings
+  #     nufs_account:
+  #       one: Chart String
+  #       other: Chart Strings
 
   # attributes:
```
giladshanan, line 28: attributes belongs nested under activerecord I believe

spec/system/admin/moving_transactions_spec.rb
```diff
@@ -47,19 +50,19 @@
       find("input[value='#{od.id}']").click
     end
 
-    click_button "Reassign Chart Strings"
+    click_button "Reassign #{chart_string_name_upcase}"
```
giladshanan, line 53: Should this be:
```suggestion
    click_button "Reassign #{chart_strings_name_upcase}"
```

giladshanan (changes requested):
A few small changes, looks good over all

config/locales/override/en.yml
jossim: Indeed! Done

giladshanan (approved):

PR #3167, opened 2022-09-26
[#159098] Adds "Recalculate pricing" button to order detail form

jossim:
# Release Notes

To recalculate the price of an order detail, the payment source can be toggled, which is a fairly unnatural way to do that. This adds a "Recalculate pricing" button that recalculates the price.

# Screenshot

Screen Shot 2022-09-26 at 2 08 27 PM

### DIFF
--- a/app/assets/javascripts/app/manage_orders.js
+++ b/app/assets/javascripts/app/manage_orders.js
@@ -228,10 +228,22 @@ class OrderDetailManagement {
 }
 OrderDetailManagement.initClass();
 
+function repricingButtonListener(orderDetailManagement) {
+  const repricingButton = document.querySelector(".js--recalculate-pricing");
+
+  if (repricingButton) {
+    repricingButton.addEventListener('click', () => orderDetailManagement.updatePricing());
+  }
+}
+
 $(function() {
   const prepareForm = function() {
     const elem = $('form.manage_order_detail');
-    if (elem.length > 0) { return new OrderDetailManagement(elem); }
+    if (elem.length > 0) {
+      const odm = new OrderDetailManagement(elem);
+      repricingButtonListener(odm);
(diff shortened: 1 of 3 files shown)


app/views/order_management/order_details/_form.html.haml
```diff
@@ -64,6 +64,11 @@
           = f.input :quantity, as: :order_detail_quantity, hint_html: { class: "help-inline" }
         .span7= render "costs", f: f
 
+    .row
+      .span7
+      .span3
+        .btn.btn-primary.js--recalculate-pricing Recalculate pricing
```
giladshanan, line 70: Can you put this into `en.yml`?

app/assets/javascripts/app/manage_orders.js
```diff
@@ -228,10 +228,23 @@ class OrderDetailManagement {
...
 OrderDetailManagement.initClass();
 
+function repricingButtonListener(orderDetailManagement) {
+  const repricingButton = document.querySelector(".js--recalculate-pricing");
+
+  if (repricingButton) {
+    repricingButton.addEventListener('click', () => orderDetailManagement.updatePricing());
+  }
+}
+
```
giladshanan, line 238: I think think this should happen as part of `initPriceUpdating()`, since that is where the constructor class assigns other elements event listeners using `this.updatePricing()`.  You may need to adjust things a bit:
```
    const repricingButton = this.$element.find(".js--recalculate-pricing");

    if (repricingButton) {
      repricingButton.bind("click", () => this.updatePricing());
    }
```

giladshanan (commented):
I think the event listener should get assigned in different place in the logic - what do you think?

giladshanan (approved):

PR #3186, opened 2022-10-10
[#160949] change ethernet port to string

jossim:
# Release Notes

This changes the `ethernet_port_number` on Relays from an integer to a string

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -55,34 +55,6 @@ ActiveRecord::Schema.define(version: 2022_10_10_164640) do
     t.index ["affiliate_id"], name: "index_accounts_on_affiliate_id"
   end
 
-  create_table "active_storage_attachments", charset: "utf8mb3", force: :cascade do |t|
-    t.string "name", null: false
-    t.string "record_type", null: false
-    t.bigint "record_id", null: false
-    t.bigint "blob_id", null: false
-    t.datetime "created_at", null: false
-    t.index ["blob_id"], name: "index_active_storage_attachments_on_blob_id"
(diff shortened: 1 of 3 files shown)


db/schema.rb
```diff
@@ -55,6 +55,34 @@
     t.index ["affiliate_id"], name: "index_accounts_on_affiliate_id"
   end
 
+  create_table "active_storage_attachments", charset: "utf8mb3", force: :cascade do |t|
```
giladshanan, line 58: This seems unrelated

db/schema.rb
```diff
@@ -185,7 +213,7 @@
     t.boolean "is_active", null: false
     t.integer "created_by", null: false
     t.datetime "created_at", null: false
-    t.integer "revenue_account", null: false
+    t.string "revenue_account", null: false
```
giladshanan, line 216: this too...

giladshanan (changes requested):
Looks like your schema file has some unrelated changes in it.  What happens to existing data when this migration runs?

db/schema.rb
jossim, line 58: When I ran `rails db:migrate`, these active storage and affiliate changes happened, though it didn't seem there was a specific migration that was run to make these changes. So I couldn't figure out how to prevent them and also migrate the relays change

giladshanan: @jossim try running `db:rollback` then check the schema, repeat until the active storage and affiliate changes are removed.  at that point you may have to get a little creative - perhaps deleting or commenting out the active storage migration - then run `db:migrate` again.

giladshanan (approved):

PR #3188, opened 2022-10-12
Changing AccountConfig's data structures to sets

jossim:
# Release Notes

`AccountConfig` uses arrays to store data that is assumed to be unique. This replaces those arrays with sets, to insure uniqueness.

Any school that adds account types to `AccountConfig` will need to update that action in its Engine to add to sets rather than adding to arrays.


giladshanan (approved):

vendor/engines/c2po/lib/c2po.rb
```diff
@@ -2,13 +2,13 @@
...
 
-  C2PO_ACCOUNT_TYPES = %w(CreditCardAccount PurchaseOrderAccount).freeze
+  C2PO_ACCOUNT_TYPES = Set["CreditCardAccount", "PurchaseOrderAccount"]
 
   C2PO_ACCOUNT_TYPES_APPENDER = proc do
-    Account.config.account_types.concat C2po::C2PO_ACCOUNT_TYPES
-    Account.config.facility_account_types.concat C2po::C2PO_ACCOUNT_TYPES
-    Account.config.statement_account_types.concat C2po::C2PO_ACCOUNT_TYPES
-    Account.config.affiliate_account_types.concat C2po::C2PO_ACCOUNT_TYPES
+    Account.config.account_types.add(C2po::C2PO_ACCOUNT_TYPES).flatten!
```
giladshanan, line 8: Why is `flatten!` needed here?  Should we be using [`Set#merge`](https://ruby-doc.org/stdlib-2.7.1/libdoc/set/rdoc/Set.html#method-i-merge) instead of `add`?

giladshanan (commented):
Left a question for you

PR #1803, opened 2018-12-13
Use-Manage Facility List

ColinMcNeil:
# Release Notes

Add facility thumbnail upload capability, as well as a tiled facility list. This works with css grid and has a flexbox fallback.

# Screenshot
## Rendered tiles
screen shot 2018-12-13 at 12 33 54 pm
As visible, EF has a thumbnail while FAC2 does not.

## Upload field well
screen shot 2018-12-13 at 12 45 06 pm
Uploading a thumbnail is done in the admin form.

# Additional Context

This includes use of a feature flag and works with the az-list. I needed imagemagic installed on my machine for this to work.

PR #3193, opened 2022-10-14
[#160368] Facility card view

jossim:
# Release Notes

This adds the `facility_tile_list` feature flag. When set to `true`, it will turn the list of facilities on the home page into a tile grid, where each facility can have it's on image.

The images are handled by `ActiveStorage` or `Paperclip`, depending on the `active_storage` feature flag. If using `Paperclip`, a migration is required, which is outlined in the CHANGELOG.

# Screenshot
Screen Shot 2022-10-14 at 6 20 38 PM

# Additional context

This PR is based on this older PR from UConn https://github.com/tablexi/nucore-open/pull/1803


jossim:
To process images (e.g. resize them), looks like the `image_processing` gem is required for `ActiveStorage`. It [requires either libvips or ImageMagick](https://guides.rubyonrails.org/active_storage_overview.html#requirements) to be installed on the server

app/models/facility.rb
```diff
@@ -2,6 +2,10 @@
 
 class Facility < ApplicationRecord
 
+  # Migration is required if Paperclip is being used.
+  # The image is used in the UI if the facility_tile_list feature flag is enabled
+  include DownloadableFiles::Image
```
giladshanan, line 7: I think this needs a feature flag, right?  Maybe the code comment would be more at home in the Image concern?

app/views/facilities/_facility_fields.html.haml
```diff
@@ -2,6 +2,9 @@
   = f.input :name, hint: text(".hints.name")
   = f.input :abbreviation, hint: text(".hints.abbreviation")
   = f.input :url_name, hint: facility_url(f.object.url_name || "url-name"), as: :string
+  - if SettingsHelper.feature_on?(:facility_tile_list)
+    = f.input :file, label: "Image"
```
giladshanan, line 6: i would expect this to be:
```suggestion
    = f.input :file_field, :file, label: "Image"
```
... but if its working then maybe I'm wrong :)

config/settings.yml
```diff
@@ -126,7 +127,7 @@ feature:
   kiosk_view: true
   bypass_kiosk_auth: false
   revenue_account_editable: false
-  active_storage: false
+  active_storage: true
```
giladshanan, line 130: I would rather do this in a separate PR

giladshanan (commented):
Looks like a great start!  Left a few coments

jossim:
It'll probably be a good idea to consider image processing to ensure the images people upload aren't too large. This will make the infrastructure more complicated however.

spec/system/facility_list_tile_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+      expect(page).to have_selector(".tile-image")
+    end
+
+    context "image manpulation" do
+      before do
+        login_as admin_user
+        visit edit_facility_path(facility)
+      end
+
+      context "with image attached" do
```
giladshanan, line 28: I'd suggest making with/without an attached image the top level context.  Fine to do this in a follow up PR

giladshanan (approved):

PR #3221, opened 2022-10-28
[#160368] Add image processing to DownloadableFiles::Image

jossim:
# Release Notes

Users can attached image files of any size and dimension to a Facility, which can make the `facility_tile_list` grid layout in unexpected ways. This uses `ActiveStroage#variant` to make images the same size and dimension, it adds colored padding if the image is the wrong aspect ratio.

This depends on `libvips` to do the image processing. That or `ImageMagick` must be installed on the server for this to work.

# Screenshot

Screen Shot 2022-10-28 at 12 48 03 PM

### DIFF
--- a/app/models/concerns/downloadable_files/image.rb
+++ b/app/models/concerns/downloadable_files/image.rb
@@ -21,6 +21,14 @@ module DownloadableFiles
       @remove_file = !value.to_i.zero?
     end
 
+    def padded_image400x200
+      if SettingsHelper.feature_on?(:active_storage)
+        download_url.variant(resize_and_pad: [400, 200, { background: "rgb(231, 231, 231)" }])
+      else
+        download_url
+      end
+    end
+
   end
 
 end
--- a/app/models/concerns/downloadable_files/image.rb
+++ b/app/models/concerns/downloadable_files/image.rb
@@ -21,9 +21,9 @@ module DownloadableFiles
       @remove_file = !value.to_i.zero?
     end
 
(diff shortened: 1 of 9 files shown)


app/models/concerns/downloadable_files/image.rb
```diff
@@ -21,6 +21,14 @@ def remove_file=(value)
       @remove_file = !value.to_i.zero?
     end
 
+    def padded_image400x200
```
jossim, line 24: Since `Image` is a module that can be used wherever, I thought it might be a good idea if this method name was descriptive and didn't assume much with regards to `Facility`

giladshanan: I might vote for `padded_image(width: 400, height: 200, background: "rgb(231, 231, 231)")`

giladshanan (approved):
Are you still planning to address the limit to 300 characters and adding help text?  Fine to do in a follow up PR.

jossim:
> Are you still planning to address the limit to 300 characters and adding help text? Fine to do in a follow up PR.

@giladshanan ahh, forgot the help text. I can add that here. Do you know if there's an example of help text elsewhere in the code base?

The 300 character limit is a feature flag that OSU has turned on

giladshanan:
@jossim you can use the `hint` option in simpleform, something like:
`= f.input :file, hint: text("hints.image_dimensions")`

giladshanan (approved):

PR #3225, opened 2022-11-01
[#16088] SciShield adapter for safety certs

giladshanan:
# Release Notes

This adds an adapter and client class for integrating NUCore with SciShield for the purpose of checking the status of research safety certifications.  We now have 2 schools using SciShield (formerly Bioraft) so this makes it easy to share the logic.  Credentials and API endpoint are configurable via ENV variables.  Also, this makes it possible to specify an adapter class to use for research safety lookups via ENV variable.  This way we have more control over which adapter is used in production without needing to do a separate release.


app/services/research_safety_certification_lookup.rb
```diff
@@ -4,7 +4,7 @@ class ResearchSafetyCertificationLookup
 
   # Replace the value of this attribute in your engine to connect to the school's
   # custom API.
```
jossim, line 6: This comment should be updated or deleted

app/services/research_safety_certification_lookup.rb
```diff
@@ -4,7 +4,7 @@ class ResearchSafetyCertificationLookup
...
-  cattr_accessor(:adapter_class) { ResearchSafetyAlwaysCertifiedAdapter }
+  cattr_accessor(:adapter_class) { Settings.research_safety_adapter.class_name.constantize }
```
jossim, line 7: Should this being using `Settings` or an environment variable in `Settings`?

spec/services/scishield_api_adapter_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    end
+  end
+
+  describe "the user is found but the response includes an error" do
+    let(:response) { File.expand_path("../../fixtures/scishield/errors.json", __dir__) }
+    before do
+      stub_request(:get, api_endpoint)
+        .to_return(
+          body: File.new(response),
+          status: 200,
```
jossim, line 77: Should this be `200` or `403`? Looks like this is a 403 Forbidden response, rather than an 200 OK response

jossim (approved):

PR #3229, opened 2022-11-02
[#160958] Sets a minimum width of the status column on the order details table

jossim:
# Release Notes

At smaller viewport widths, the Status column of the order details table is too narrow. This sets a minimum width for that column

# Screenshot

Screen Shot 2022-11-02 at 5 45 24 PM

### DIFF
--- a/app/assets/stylesheets/app/tables.scss
+++ b/app/assets/stylesheets/app/tables.scss
@@ -25,16 +25,8 @@ table {
   }
 }
 
-/*
- * This makes sure the status column, in particular the "Status.." dropdown menu,
- * is clearly visible on the FacilityReservationsController#index page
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/app/tables.scss
```diff
@@ -25,6 +25,18 @@ table {
   }
 }
 
+/*
+ * This makes sure the status column, in particular the "Status.." dropdown menu,
+ * is clearly visible on the FacilityReservationsController#index page
+**/
+table.reservations {
```
giladshanan, line 32: I'd prefer adding a class to the dropdown input directly.  What do you think?

giladshanan (approved):

PR #3241, opened 2022-11-09
[#160883] Configurable iat offset for SciShield

giladshanan:
# Release Notes

Running into an issue with ~1 min lag between server time and API time.

### DIFF
--- a/app/services/research_safety_adapters/scishield_api_client.rb
+++ b/app/services/research_safety_adapters/scishield_api_client.rb
@@ -15,14 +15,19 @@ module ResearchSafetyAdapters
       return @token if @token.present?
       return nil unless keys_present?
 
-      now = Time.current.to_i
       payload = {
-        iat: now,
-        exp: now + 3600,
+        iat: iat,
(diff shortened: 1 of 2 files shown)


jossim (approved):

jossim:
On infrastructure, should we add sync'ing with a NTP server?

Seems like servers should have pretty accurate time

giladshanan:
@jossim Good question - I jumped on the server and it turns out the RTC was 1 minute behind UTC for some reason.
I ran `hwclock --systohc` to "set the hardware clock from the current system time" and things are synced up now.

jossim:
@giladshanan interesting. It may still drift over time. Not sure if this is Ubuntu or not, but it's pretty easy to setup Ubuntu to sync with atomic clocks

https://www.digitalocean.com/community/tutorials/how-to-set-up-time-synchronization-on-ubuntu-20-04

giladshanan:
@jossim it looks like the sync is enabled for the `corum` user but not for `root`.  I'm going to close this for now since we have a way forward.

PR #3246, opened 2022-11-14
[#161124] Add rake task to add template_file to order detail

jossim:
# Release Notes

In the case that a service is modified to require a form after it has been purchased, the order detail will state that it's missing a form, but there's no way to add the needed template_result file via the UI after purchase. This task adds a template_result file to an order detail by copying the template file from the product that was purchased.

### DIFF
--- a/lib/tasks/order_details.rake
+++ b/lib/tasks/order_details.rake
@@ -48,4 +48,18 @@ namespace :order_details do
       end
     end
   end
+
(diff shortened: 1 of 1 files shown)


giladshanan (approved):

giladshanan:
@jossim looks like a failing spec - probably just need to re-run it

PR #3266, opened 2022-12-07
[#161053] PurchaseOrder & Account updates from OSU

jossim:
# Release Notes

Some changes from https://github.com/Oregon-State/nucore-osu/pull/225

This PR:
* Creates a feature flag to include `AffiliateAccount` in purchase order or not. This allows an affiliate account to be optional if false.
* Updates views to depend on `PurchaseOrderAccount`'s attribute names rather than field names in the translation file.
* Adjusts account_spec to properly work with name-spaced facility accounts

### DIFF
--- a/vendor/engines/c2po/config/locales/en.yml
+++ b/vendor/engines/c2po/config/locales/en.yml
@@ -23,7 +23,6 @@ en:
       account_fields:
         purchase_order_account:
           account_number: Account Number
-          ar_number_label: Account Receivable Number
           ar_number_hint: if applicable
           outside_contact_label: Outside Contact Info
           outside_contact_hint: Contact information (phone number, email, fax, etc.)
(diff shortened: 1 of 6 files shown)


vendor/engines/c2po/config/locales/en.yml
```diff
@@ -46,6 +46,7 @@ en:
     attributes:
       purchase_order_account:
         account_number: Payment Source
+        ar_number: Account Receivable Number
```
giladshanan, line 49: I think we should edit the key on line [26](https://github.com/tablexi/nucore-open/pull/3266/files#diff-a6b0d91195ef79c1415d24e9d1d0fa1d6015b113b07d88c6072cb40b392415cbR26) instead.  What do you think?

giladshanan (changes requested):
Left a couple small comments.  I think the change to `ar_number` is important to reduce potential confusion.

vendor/engines/c2po/config/locales/en.yml
giladshanan: @jossim thoughts?

PR #3137, opened 2022-08-26
[#160616] Adds ActiveStorage feature flag

jossim:
# Release Notes

- [x] Add specs for `read_attached_file`
- [x] Refactor the skipped `journal_spec`
- [x] Add to the readme
- [x] Consider refactoring `DownloadableFile` into 2 modules, `DownloadableFile` and `DownloadablePaperclipFile`

PR #3325, opened 2023-01-13
[#161127] Docs tab for users

jossim:
# Release Notes

This adds a document tab for users that admins can use to upload files associated with a user.

This PR is a cherry-picked of https://github.com/Oregon-State/nucore-osu/pull/238

# Screenshot

Screen Shot 2023-01-09 at 4 53 31 PM

### DIFF
--- a/app/views/admin/shared/_tabnav_users.html.haml
+++ b/app/views/admin/shared/_tabnav_users.html.haml
@@ -26,4 +26,6 @@
   - if ResearchSafetyCertificate.any?
     = tab text("views.user_research_safety_certifications.tab"), facility_user_user_research_safety_certifications_path(current_facility, @user), secondary_tab == "user_certificates"
 
+  = tab "Docs", facility_user_user_file_uploads_path(current_facility, @user), secondary_tab == "docs"
+
   = render_view_hook "after", secondary_tab: secondary_tab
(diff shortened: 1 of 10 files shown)


giladshanan (commented):
@jossim did you run the migration locally?  I think the specs are failing because the timestamp version in `schema.rb` should match the filename of the most recent migration file:
```
ActiveRecord::Schema.define(version: 2022_12_12_231558) do
```

db/schema.rb
```diff
@@ -213,7 +213,7 @@
     t.boolean "is_active", null: false
     t.integer "created_by", null: false
     t.datetime "created_at", null: false
-    t.integer "revenue_account", null: false
+    t.string "revenue_account", null: false
```
giladshanan, line 216: @jossim this doesn't look right.  I think your local db might be in a funky state. See https://github.com/tablexi/nucore-open/pull/3137

giladshanan (changes requested):

app/views/admin/shared/_tabnav_users.html.haml
```diff
@@ -26,4 +26,6 @@
   - if ResearchSafetyCertificate.any?
     = tab text("views.user_research_safety_certifications.tab"), facility_user_user_research_safety_certifications_path(current_facility, @user), secondary_tab == "user_certificates"
 
+  = tab "Docs", facility_user_user_file_uploads_path(current_facility, @user), secondary_tab == "docs"
```
giladshanan, line 29: Can you use the translation file here?
```suggestion
  = tab t(".user_docs"), facility_user_user_file_uploads_path(current_facility, @user), secondary_tab == "docs"
```

giladshanan (approved):
One small nitpick

giladshanan (approved):

stevens101:
It looks like the documents table column headings don't align with the data in the rows. For example the "File" column is actually the Delete column. Might need another th tag.

PR #3332, opened 2023-01-17
[#161102] Create `facility_tile_list_admin` feature flag

jossim:
# Release Notes

This splits the tile list feature into 2 feature flags

The `facility_tile_list_admin` feature flag enables the feature for the facility admin pages. And the `facility_tile_list` feature flag enables it for the public pages, but only if the facility_tile_list_admin` feature flag is also enabled.

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -36,9 +36,15 @@ class FacilitiesController < ApplicationController
     @facilities = Facility.active.alphabetized
     @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = SettingsHelper.feature_on?(:use_manage) ? "use" : "home"
-    @list_layout = SettingsHelper.feature_on?(:facility_tile_list) ? "tile" : "list"
     @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.includes(:facility).alphabetized
     @azlist = build_az_list(@facilities)
+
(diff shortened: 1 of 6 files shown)


app/views/facilities/_facilities_list.html.haml
```diff
@@ -4,7 +4,7 @@
     - @azlist.map.with_index do |(letter, matched_facilities), index|
       - matched_facilities.each do |facility|
         %li.facility_listing.js--facility_listing{ class: [ az_classname_for_facility(index, letter), @list_layout ] }
-          - if SettingsHelper.feature_on?(:facility_tile_list)
+          - if SettingsHelper.feature_on?(:facility_tile_list) && SettingsHelper.feature_on?(:facility_tile_list_admin)
```
giladshanan, line 7: I think I would lean towards leaving this as is - if someone wanted to use the tile view on the front end with the default image, that's fine with me

app/controllers/facilities_controller.rb
```diff
@@ -36,9 +36,15 @@ def index
     @facilities = Facility.active.alphabetized
     @recently_used_facilities = MostRecentlyUsedSearcher.new(acting_user).recently_used_facilities.alphabetized
     @active_tab = SettingsHelper.feature_on?(:use_manage) ? "use" : "home"
-    @list_layout = SettingsHelper.feature_on?(:facility_tile_list) ? "tile" : "list"
     @recent_products = MostRecentlyUsedSearcher.new(acting_user).recently_used_products.includes(:facility).alphabetized
     @azlist = build_az_list(@facilities)
+
```
giladshanan, line 41: Not sure we need this change

giladshanan (approved):
I'd lean towards keeping these two feature flags more isolated, not a blocking comment though

giladshanan (approved):

PR #3339, opened 2023-01-20
[#161308] Updates rake task to add missing template to order detail

jossim:
# Release Notes

This updates the rake task that adds a missing template to order details.

Now it will only add a file if the form is missing, a file is created instead of using the product's file. And it's now possible to pass multiple ids to the rake task.

### DIFF
--- a/lib/tasks/order_details.rake
+++ b/lib/tasks/order_details.rake
@@ -66,7 +66,7 @@ namespace :order_details do
         file: StringIO.new("Placeholder text for missing template."),
         file_type: "template_result",
         name: "placeholder.csv",
-        created_by: "order_detail.user.id"
+        created_by: order_detail.user.id
       )
 
       order_detail.save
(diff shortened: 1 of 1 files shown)


lib/tasks/order_details.rake
```diff
@@ -54,12 +53,23 @@ namespace :order_details do
...
+    order_details = OrderDetail.where id: args[:ids].split
+
+    order_details.each do |order_detail|
+      next unless order_detail.missing_form?
+
+      order_detail.stored_files << StoredFile.new(
+        file: StringIO.new("Placeholder text for missing template."),
+        file_type: "template_result",
+        name: "placeholder.csv",
+        created_by: "order_detail.user.id"
```
giladshanan, line 69:
```suggestion
        created_by: order_detail.user.id
```

giladshanan (changes requested):
Can you add a couple of `puts` statements to give us a sense of how the run is going?

giladshanan (approved):

PR #3348, opened 2023-01-24
[#161102] Adds `active_storage_for_images_only` feature flag

jossim:
# Release Notes

This adds the `active_storage_for_images_only` feature flag to allow for the use of `ActiveStorage` just for `DownloadableFiles::Image`.

Since UMass would like to use images but is using Paperclip otherwise, this will allow them to add images via `ActiveStorage` so there's less to migrate over to `ActiveStorage` when the time comes.

### DIFF
--- a/app/lib/settings_helper.rb
+++ b/app/lib/settings_helper.rb
@@ -29,6 +29,10 @@ module SettingsHelper
     setting "relays.#{Rails.env}.reservation_enabled"
   end
 
+  def self.active_storage_enabled?
+    feature_on?(:active_storage) || feature_on?(:active_storage_for_images_only)
+  end
+
   #
   # Used to query the +Settings+ under feature:
   # [_feature_]
--- a/app/lib/settings_helper.rb
+++ b/app/lib/settings_helper.rb
@@ -29,10 +29,6 @@ module SettingsHelper
     setting "relays.#{Rails.env}.reservation_enabled"
   end
 
-  def self.active_storage_enabled?
-    feature_on?(:active_storage) || feature_on?(:active_storage_for_images_only)
-  end
-
   #
   # Used to query the +Settings+ under feature:
   # [_feature_]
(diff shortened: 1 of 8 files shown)


app/models/concerns/active_storage_file.rb
```diff
@@ -9,7 +9,7 @@ module ActiveStorageFile
     has_one_attached :file
 
     def assign_attributes(new_attributes)
-      if SettingsHelper.feature_on?(:active_storage)
+      if SettingsHelper.active_storage_enabled?
```
giladshanan, line 12: Do we need this feature flag here?  I can't recall if we need it for some reason or perhaps it was left over from a previous iteration of the code, before we separated things into modules

app/lib/settings_helper.rb
```diff
@@ -29,6 +29,10 @@ def self.relays_enabled_for_reservation?
     setting "relays.#{Rails.env}.reservation_enabled"
   end
 
+  def self.active_storage_enabled?
+    feature_on?(:active_storage) || feature_on?(:active_storage_for_images_only)
```
giladshanan, line 33: I think I'm leaning towards letting each feature flag do its own thing - what's your thinking about intertwining here?  

If we keep this logic I'd vote for renaming to `active_storage_for_images_enabled?`

jossim: Because of how the flags are named, it seemed like `active_storage` could/would be a superset of `active_storage_for_images_only`. Also doing it this way would mean no change in OSU's settings.yml. Neither of these may be interesting enough reasons, don't have a strong opinion.

app/models/concerns/active_storage_file.rb
jossim: I also can't recall. It passes CI with it removed though, so I suppose it's fine to leave it out. I recall this method causing a lot of problems, so likely extra stuff was left in it while working up to a solution.

giladshanan (approved):
🎉

PR #3397, opened 2023-02-27
[#161362] Price group based discounts

jossim:
# Release Notes

This updates how discounts work, relating them to price groups. Now each global `PriceGroup` has its own discount for each `ScheduleRule`. For non-global `PriceGroup`s, the discount they have will fallback to the global price group that corresponds to its `is_internal` status.

Because `ScheduleRule#discount_percent` is no longer used to calculate the discount, every schedule rule needs to be updated to have `PriceGroupDiscount`s for the global price groups. There is a rake task, `schedule_rule:add_price_group_discounts`, to automate this process -- it adds price group discounts for every global price group to every schedule rule.

# Screenshot

Screen Shot 2023-03-13 at 5 40 10 PM

Screen Shot 2023-03-13 at 5 40 32 PM


app/models/schedule_rule.rb
```diff
@@ -130,8 +133,13 @@ def as_calendar_objects(options = {})
     ScheduleRuleCalendarPresenter.new(self, options).to_json
   end
 
-  def discount_for(start_at, end_at)
-    percent_overlap(start_at, end_at) * discount_percent.to_f
+  def discount_for(start_at, end_at, price_group)
+    price_group_discount = PriceGroupDiscount.where(
+      price_group: price_group,
+      schedule_rule: self
+    ).first
```
giladshanan, line 140: I think this logic could be more succint like:
```suggestion 
    price_group_discount = price_group_discounts.find_by(price_group: price_group)
```

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -15,10 +15,13 @@
       = f.label :end_time
       = time_select24(f, :end, hours: (0..24), minute_step: @product.reserve_interval)
 
-= f.input :discount_percent, label: false, hint: false do
-  = f.label :discount_percent
-  = f.hint :discount_percent
-  = f.number_field :discount_percent, size: 4
+%h3 Global Price Group Discounts
```
giladshanan, line 18: i18n please

db/migrate/20230223221431_create_price_group_discounts.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePriceGroupDiscounts < ActiveRecord::Migration[6.1]
```
giladshanan, line 1: Can you add `# frozen_string_literal: true` here?

app/models/price_group_discount.rb
```diff
@@ -0,0 +1,4 @@
+class PriceGroupDiscount < ApplicationRecord
```
giladshanan, line 1: `# frozen_string_literal: true`

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -15,10 +15,13 @@
...
+
+= f.association :price_group_discounts do |pgd|
+  = pgd.label pgd.object.price_group.name
+  = pgd.hint :discount_percent
```
giladshanan, line 22: Is this showing up like `"During the period covered by this schedule rule, overall cost will be discounted by"`?  I don't see it in the screenshot

spec/factories.rb
```diff
@@ -3,6 +3,11 @@
 require File.expand_path("factories_env", File.dirname(__FILE__))
 
 FactoryBot.define do
+  factory :price_group_discount do
```
giladshanan, line 6: This belongs in its own factory file `price_group_discounts.rb`

spec/models/price_group_discount_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe PriceGroupDiscount, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
```
giladshanan, line 4: I'd say either add some specs here or remove the file

giladshanan (changes requested):
- [x] The main missing piece left is dealing with the specs.  Some specs are failing because they are using the schedule rule factory which is trying to set a value for `discount_percent`.  I think you can change the factory definition so that it creates $0 `price_group_discounts` for the globals by default.  You can also create a `transient` attribute to make it easier to pass in a single value and update all the discounts (mimic old behavior in specs).  Let me know if you have questions or want to pair on that.

- [x] Once you fix that, there may be other adjustments that need to be made.  I'd expect to see changes and/or additions to instrument_price_policy_calculations_spec.rb, schedule_rule_spec.rb, and price_policies/time_based_price_calculator_spec.rb.  I'd like to have at least one spec that covers setting discounts and at least one that shows they are behaving as expected in calculations.  

- [ ] Then we need to figure out how to handle existing discount values.  Either a one-off rake task or migration would work.

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -15,10 +15,13 @@
       = f.label :end_time
       = time_select24(f, :end, hours: (0..24), minute_step: @product.reserve_interval)
 
-= f.input :discount_percent, label: false, hint: false do
-  = f.label :discount_percent
-  = f.hint :discount_percent
-  = f.number_field :discount_percent, size: 4
+%h3 Global Price Group Discounts
+
+= f.association :price_group_discounts do |pgd|
```
jossim, line 20: can put a `label` and `hint` here

app/views/schedule_rules/_schedule_rule_fields.html.haml
jossim: This should be either "Discounts" or "Price Group Discounts"

app/views/schedule_rules/_schedule_rule_fields.html.haml
giladshanan: We should add a hint here (or maybe just at the top level) indicating that non-global price groups will get the discount based on internal/external discount.

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -15,15 +15,19 @@
...
       = time_select24(f, :end, hours: (0..24), minute_step: @product.reserve_interval)
 
-%h3
-  = t("views.schedule_rules.discount")
+%h3= t("views.schedule_rules.discount")
 
-= f.association :price_group_discounts do |pgd|
+%p= t("views.schedule_rules.discount_hint")
+
+= f.association :price_group_discounts, label: "hello", hint: "hi" do |pgd|
```
giladshanan, line 22: I think we can drop the label and hint here

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -15,15 +15,19 @@
...
   = pgd.label pgd.object.price_group.name
-  = pgd.hint :discount_percent
-  = pgd.number_field :discount_percent, size: 4, label: false
+  = pgd.number_field :discount_percent, size: 4
   = pgd.hidden_field :price_group_id, value: pgd.object.price_group.id
   = "%"
+  - if pgd.object.price_group == PriceGroup.base 
+    %p= t("views.schedule_rules.base_discount_hint")
+  - elsif pgd.object.price_group == PriceGroup.external
+    %p= t("views.schedule_rules.external_discount_hint")
```
giladshanan, line 30: Can you see if `pgd.hint :base_discount_hint` would work here?  You may need to move the copy back to `simple_form_en.yml` under hints/schedule_rule/base_discount_hint.

giladshanan (approved):
Pending CI

PR #3419, opened 2023-03-15
[#161362] Remove `ignored_columns` from `ScheduleRule`

jossim:
# Release Notes

Looks like this is needed for the `schedule_rule:add_price_group_discounts` rake task to work


giladshanan (approved):
@jossim after the rake task is run we should deprecate this column and then remove it.  Can you add a ticket to nucore-open to remind us?

PR #3421, opened 2023-03-17
[#160937] Display reconciliation/close date and times on journal and order details

jossim:
# Release Notes

This logs and displays when a journal or order detail is closed. These logged events will show up in the event log as well as on the journal and statements pages

# Screenshot

Screen Shot 2023-03-29 at 12 10 54 PM

Screen Shot 2023-03-29 at 12 11 28 PM

### DIFF
--- a/spec/presenters/statement_presenter_spec.rb
+++ b/spec/presenters/statement_presenter_spec.rb
@@ -64,24 +64,25 @@ RSpec.describe StatementPresenter do
     end
   end
 
-  describe "#closed_by_user_full_names" do
-    before do
-      LogEvent.log(statement, :closed, user1)
-      LogEvent.log(statement, :closed, user2)
+  context "with closed events" do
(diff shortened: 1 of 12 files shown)


spec/presenters/statement_presenter_spec.rb
```diff
@@ -60,4 +64,22 @@
       end
     end
   end
+
+  context "with closed events" do
+    let!(:log_events) { [LogEvent.log(statement, :closed, user1), LogEvent.log(statement, :closed, user2)] }
```
giladshanan, line 69: Any reason not to give each of these LogEvents its own `let!` block?

spec/system/admin/journal_close_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+
```
giladshanan, line 5: extra newline

giladshanan (approved):
🎉 
just a few small nitpicks

PR #3427, opened 2023-03-27
[#159818] Ruby 2.7.5 

giladshanan:
# Release Notes

Also moves the spec suite over to Github Actions.
Auto deployments are not running at the moment, we will fix that after the ruby upgrade process is complete.


giladshanan:
@jossim we can ignore the CircleCI Pipeline check failure.

jossim (approved):

PR #3534, opened 2023-06-01
TECH TASK: Updates creating a journal spec to get account from Settings

jossim:
# Release Notes

Different schools have different account types. This allows the account type used in `spec/system/creating_a_journal_spec.rb` to be specified in `config/settings.yml`


giladshanan (approved):
Can you add something about this in `doc/coding_standards.md` - maybe under the `Fork-specific changes` section?

jossim:
@giladshanan I put a mention of it in the list

PR #3588, opened 2023-07-14
[#160013] Allows global admins to update fulfillment date on completed orders

jossim:
# Release Notes

This allows global admins to update the `fulfilled_at` date on order details that are in the complete state.

It also logs an event when this happens.

# Screenshot

Screen Shot 2023-07-19 at 11 25 46 AM

Logging

Screen Shot 2023-07-19 at 12 06 06 PM

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -66,7 +66,6 @@ class OrderDetail < ApplicationRecord
   has_many   :notifications, as: :subject, dependent: :destroy
   has_many   :stored_files, dependent: :destroy
   has_many   :sample_results_files, -> { sample_result }, class_name: "StoredFile"
-  has_many   :log_events, as: :loggable
 
   # This is a _temporary_ associaton to make up for the fact that the
   # vestal versions gem is no longer in the project. It's here to
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -66,6 +66,7 @@ class OrderDetail < ApplicationRecord
   has_many   :notifications, as: :subject, dependent: :destroy
   has_many   :stored_files, dependent: :destroy
   has_many   :sample_results_files, -> { sample_result }, class_name: "StoredFile"
+  has_many   :log_events, as: :loggable
 
   # This is a _temporary_ associaton to make up for the fact that the
   # vestal versions gem is no longer in the project. It's here to
@@ -931,7 +932,14 @@ class OrderDetail < ApplicationRecord
     return unless complete? && manual_fulfilled_at_time
     return if manual_fulfilled_at_time.beginning_of_day == fulfilled_at.beginning_of_day
 
+    @updated_fulfilled_at = true
     make_complete
+  end
+
+  # OrderManagement::OrderDetailsController#update uses this to determine if
+  # a LogEvent should be logged  
+  def updated_completed_fulfilled_at?
+    @updated_fulfilled_at
   end
 
   private
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -927,6 +927,13 @@ class OrderDetail < ApplicationRecord
     ActiveModel::Type::Boolean.new.cast(resolve_dispute)
   end
 
+  def update_completed_fulfilled_at
+    return unless complete?
+    return if manual_fulfilled_at_time.beginning_of_day == fulfilled_at.beginning_of_day
+
+    make_complete
+  end
+
   private
 
   # Is there enough information to move an associated order to complete/problem?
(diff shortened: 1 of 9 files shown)


app/models/order_detail.rb
```diff
@@ -66,6 +66,7 @@ def set_problem_order
   has_many   :notifications, as: :subject, dependent: :destroy
   has_many   :stored_files, dependent: :destroy
   has_many   :sample_results_files, -> { sample_result }, class_name: "StoredFile"
+  has_many   :log_events, as: :loggable
```
giladshanan, line 69: Do we need this?

app/models/order_detail.rb
```diff
@@ -927,6 +928,20 @@ def resolve_dispute?
     ActiveModel::Type::Boolean.new.cast(resolve_dispute)
   end
 
+  def update_completed_fulfilled_at
```
giladshanan, line 931: I think you've got the `@editing_user` in the param updater.  If you pass that in, could you do the logging hare as part of this method?  That might simplify things a bit.
```suggestion
  def update_completed_fulfilled_at(@editing_user)
```

app/views/order_management/order_details/_form.html.haml
```diff
@@ -19,7 +19,16 @@
       .span2
         - if @order_statuses
           = f.association :order_status, collection: @order_statuses, label_method: :name_with_level, include_blank: false
-          - if f.object.fulfilled_at.blank? || f.object.fulfilled_at_changed?
+          - od = f.object
+          - if od.state == "complete" && current_user.administrator?
```
giladshanan, line 23: I think you can do `od.complete?` here
```suggestion
          - if od.complete? && current_user.administrator?
```

giladshanan, line 23: What do you think about creating a view helper module to hold a new method that takes a user and an order detail record and returns a boolean:
```
#helpers/order_detail_helper.rb

module OrderDetailHelper

  def can_edit_fulfilled_at?(od, user)
    return true if od.complete? && user.administrator?

    od.fulfilled_at.blank? || od.fulfilled_at_changed?
  end

end
```
... then we can call that method to determine whether or not to show the form input.  Less duplicated code, less logic in the view, and we could even write some simple tests around the new helper method.

config/locales/en.yml
```diff
@@ -492,6 +492,8 @@ en:
     update:
       success: Cutoff date successfully updated
 
+  order_detail:
```
giladshanan, line 495: This would be more at home in `config/locales/views/admin/en.log_events.yml`

spec/system/admin/manage_order_detail_spec.rb
```diff
@@ -10,14 +10,17 @@
   let(:reservation) { create(:purchased_reservation, product: instrument) }
   let(:order_detail) { reservation.order_detail }
   let(:order) { reservation.order }
-
+  let(:administrator) { create(:user, :administrator) }
   let(:director) { create(:user, :facility_director, facility: facility) }
+  let(:logged_in_user) {}
```
giladshanan, line 15: How about 
```suggestion
  let(:logged_in_user) { director }
```
... since that's what most of these specs are based on

giladshanan (changes requested):
I made a few suggestions for organizing things, let me know what you think

spec/system/admin/manage_order_detail_spec.rb
```diff
@@ -245,4 +246,36 @@
...
+
+      it "logs fulfilled at date updates" do
+        expect(LogEvent).to be_exists(loggable: order_detail, event_type: :updated_fulfilled_at, user: logged_in_user)
+      end
+    end
+
+    context "as non-global admin" do
+      let(:logged_in_user) { director }
+
+      it "cannont update the fulfilled at date"
```
giladshanan, line 278: I think it would be good to write a spec for this case

giladshanan (commented):
This is looking good!  I think there's just one more spec to write

spec/system/admin/manage_order_detail_spec.rb
```diff
@@ -245,4 +246,38 @@
...
+      it "logs fulfilled at date updates" do
+        expect(LogEvent).to be_exists(loggable: order_detail, event_type: :updated_fulfilled_at, user: logged_in_user)
+      end
+    end
+
+    context "as non-global admin" do
+      let(:logged_in_user) { director }
+
+      it "cannont update the fulfilled at date" do
+        expect(page).to have_content("Fulfilled Date").once
```
giladshanan, line 279: I was thinking something a little more specific, like:
```suggestion
        expect(page).to have_no_selector("input", "order_detail[fulfilled_at]")
```

giladshanan (commented):
One more comment for you

giladshanan (approved):

PR #3594, opened 2023-07-21
[#161847] Rake tasks for adding new PriceGroups to ScheduleRules

jossim:
# Release Notes

When a global price group is created, it needs to be added to all the schedule rules for it to appear on the UI for each product.

This created a rake tasks which does that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,22 +288,6 @@ It is possible to track deprecation warnings locally with [deprecation_toolkit](
 
 `deprecation_toolkit` is configured in [`spec/deprecation_toolkit_env.rb`](spec/deprecation_toolkit_env.rb).
 
-## Creating Global Price Groups
-
-A global price group is a price group that's accessible across all facilities, and it must be created via the Rails console.
-
-```
-pg = PriceGroup.new(name: "New Price Group Name", display_order: PriceGroup.globals.count + 1, is_internal: false)
-pg.save(validate: false)
-```
-
-Once the price group is created, the `schedule_rule:add_new_price_groups` rake task should be run to add the price group to all existing schedule rules. If you would like the discount to be something other than 0%, the task accepts the discount as a parameter. This example will set a 20% discount for all the price group it will add to all schedule rules
-
-```
-rake schedule_rule:add_new_price_groups[20]
-```
-
-
 ## Optional Modules
 
 The following modules are provided as optional features via
--- a/README.md
+++ b/README.md
@@ -288,6 +288,22 @@ It is possible to track deprecation warnings locally with [deprecation_toolkit](
 
 `deprecation_toolkit` is configured in [`spec/deprecation_toolkit_env.rb`](spec/deprecation_toolkit_env.rb).
 
+## Creating Global Price Groups
+
+A global price group is a price group that's accessible across all facilities, and it must be created via the Rails console.
+
+```
+pg = PriceGroup.new(name: "New Price Group Name", display_order: PriceGroup.globals.count + 1, is_internal: false)
+pg.save(validate: false)
+```
+
+Once the price group is created, the `schedule_rule:add_new_price_groups` rake task should be run to add the price group to all existing schedule rules. If you would like the discount to be something other than 0%, the task accepts the discount as a parameter. This example will set a 20% discount for all the price group it will add to all schedule rules
+
+```
+rake schedule_rule:add_new_price_groups[20]
+```
+
+
 ## Optional Modules
 
 The following modules are provided as optional features via
--- a/app/models/price_group.rb
+++ b/app/models/price_group.rb
@@ -41,6 +41,53 @@ class PriceGroup < ApplicationRecord
     price_group
   end
 
+  def is_not_global?
+    !global?
+  end
+
+  def global?
+    facility.nil?
+  end
+
+  def can_manage_price_group_members?
+    is_not_global? || SettingsHelper.feature_on?(:can_manage_global_price_groups)
+  end
+
+  def can_purchase?(product)
+    !PriceGroupProduct.find_by(price_group_id: id, product_id: product.id).nil?
+  end
+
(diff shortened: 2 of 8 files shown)


lib/tasks/schedule_rules.rake
```diff
@@ -20,4 +23,34 @@ namespace :schedule_rule do
...
   end
+
+  # Add missing global price groups to schedule rules.
+  #
+  # Run this whenever a new global price group is created. You can set a discount
+  # percent for all schedule rules. The below example will set a 20% discount
+  #
+  # `rake schedule_rule:add_new_price_groups[20]`
+  desc "Adds PriceGroupDiscounts for global price groups that have yet to be associated with schedule rules"
+  task :add_new_price_groups, [:discount] => :environment do |_t, args|
```
giladshanan, line 34: - Move to new file, price_groups.rake
- Add relevant documentation in that file
- Task accepts a pg_name, is_internal, admin_editable, and discount
- Task uses a new class method PriceGroup.new_global

README.md
```diff
@@ -288,6 +288,22 @@ It is possible to track deprecation warnings locally with [deprecation_toolkit](
...
+```
+pg = PriceGroup.new(name: "New Price Group Name", display_order: PriceGroup.globals.count + 1, is_internal: false)
+pg.save(validate: false)
+```
+
+Once the price group is created, the `schedule_rule:add_new_price_groups` rake task should be run to add the price group to all existing schedule rules. If you would like the discount to be something other than 0%, the task accepts the discount as a parameter. This example will set a 20% discount for all the price group it will add to all schedule rules
+
+```
+rake schedule_rule:add_new_price_groups[20]
+```
```
giladshanan, line 304: I think this would be more at home in the CHANGELOG file

app/models/price_group.rb
```diff
@@ -78,4 +86,63 @@ def external?
...
+  # Find a global price group by name, or create it if it does not exist
+  def self.find_or_create_global(name:, is_internal: false, admin_editable: true)
+    global_price_groups = PriceGroup.globals
+    found_price_group = global_price_groups.find_by(name:)
+
+    if found_price_group
+      log_string = "Global price group '#{name}' already exists."
+
+      puts log_string
+      logger.info log_string
```
giladshanan, line 100: Do we need the explicit logging here and in other methods related to setup_global?  I think that method will be called from the console or rake task to `puts` should do the trick.

app/models/price_group.rb
```diff
@@ -78,4 +86,63 @@ def external?
...
+
+      pg
+    end
+  end
+
+  # Creates price group discounts for every schedule rule that does not have a
+  # price group discount for this price group
```
giladshanan, line 123:
```suggestion
  # Creates price group discounts for this price group, if they do not exist
```

lib/tasks/pirce_groups.rake
```diff
@@ -0,0 +1,19 @@
...
+  # This creates a new global price group, if it does not exist, and adds
+  # a PriceGroupDiscount for it to every ScheduleRule. If the price group
+  # already exists, this will just add PriceGroupDiscounts for it to ScheduleRules
+  # that are missing then.
+  #
+  # rake price_group:add_global_price_group["new group",true,false,15]
+  desc "Create a new global price group and/or setup schedule rules for it"
+  task :add_global_price_group, [:name, :internal, :admin_editable, :discount] => :environment do |_t, args|
+    name = args[:name]
+    is_internal = args[:internal].casecmp("true").zero?
```
giladshanan, line 13: Not a blocking comment, but curious why you opted for `casecmp` instead of `== "true"` here

app/models/price_group.rb
```diff
@@ -33,6 +33,14 @@ def self.external
     globals.find_by(name: Settings.price_group.name.external)
   end
 
+  # Create a global price group, if it does not exist, and setup all the
+  # schedule rules with price group discounts for the price group.
+  def self.setup_global(name:, is_internal: false, admin_editable: true, discount_percent: 0)
```
giladshanan, line 38: Would this work in the seeds file?

giladshanan (commented):
nice!

lib/tasks/pirce_groups.rake
jossim: @giladshanan this was a RuboCop suggestion https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancecasecmp

app/models/price_group.rb
jossim: @giladshanan indeed it does

giladshanan (commented):
There are a handful of other places where PriceGroups are created, can you take a look and see if we can use the new method instead:
- `demo_seed.rake`
- `rails_helper.rb`
- school-specific seed files like `vendor/engines/umass_corum/db/seeds.rb` (we can update those as we pull this down into the school repos)

app/models/price_group.rb
giladshanan, line 38: maybe this should have an input for `display_order`?

spec/rails_helper.rb
```diff
@@ -132,9 +132,8 @@
     Affiliate.OTHER
 
     # initialize price groups
-    @nupg = PriceGroup.find_or_create_by(name: Settings.price_group.name.base, is_internal: true, display_order: 1)
-    @nupg.save(validate: false)
-    PriceGroup.find_or_create_by(name: Settings.price_group.name.external, is_internal: false, display_order: 3).save(validate: false)
+    PriceGroup.setup_global(name: Settings.price_group.name.base, is_internal: true)
```
giladshanan, line 135: I think this will fix the specs:
```suggestion
    @nupg = PriceGroup.setup_global(name: Settings.price_group.name.base, is_internal: true)
```

giladshanan (approved):

PR #3597, opened 2023-07-25
[#161857] Add new dropdown for Billing Mode when creating new products

jossim:
# Release Notes

This adds a billing mode attribute to products that global admins can set only on creation. Users who are not global admins cannot set this attribute (it defaults to "Default").

# Screenshot

Screen Shot 2023-07-26 at 1 28 03 PM

Screen Shot 2023-07-26 at 1 27 21 PM

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -55,6 +55,11 @@ class Product < ApplicationRecord
     errors.add(:contact_email, text("errors.models.product.attributes.contact_email.required")) unless email.present?
   end
 
+  def self.billing_modes
+    ["Default", "Skip Review", "Nonbillable"]
+  end
+  validates :billing_mode, inclusion: Product.billing_modes
+
   scope :active, -> { where(is_archived: false, is_hidden: false) }
   scope :alphabetized, -> { order(Arel.sql("LOWER(products.name)")) }
   scope :archived, -> { where(is_archived: true) }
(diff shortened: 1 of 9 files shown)


spec/system/managing_product_billing_mode_spec.rb
```diff
@@ -0,0 +1,46 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe "Managing products" do
+  let(:instrument) { FactoryBot.create(:setup_instrument) }
+  let(:facility) { instrument.facility }
+  let(:administrator) { create(:user, :administrator) }
+  let(:facility_administrator) { create(:user, :facility_administrator, facility: facility) }
+  let(:user) { nil }
```
giladshanan, line 10: Do you need this?

giladshanan (changes requested):
Looks good just a couple requests

app/models/product.rb
```diff
@@ -67,6 +67,11 @@ class Product < ApplicationRecord
     left_outer_joins(:product_display_group_product).where(product_display_group_products: { id: nil })
   }
 
+  def self.billing_modes
+    ["Default", "Skip Review", "Nonbillable"]
+  end
+  validates :billing_mode, inclusion: Product.billing_modes
```
giladshanan, line 73: I'd move this up with the other validations, like line 52 or so

giladshanan (approved):
One small nitpick

PR #3615, opened 2023-08-03
[#161858] Skip Review billing mode

jossim:
# Release Notes

This implements the "Skip Review" billing mode, which will reconcile an order detail whenever it goes into the "complete" state, rather than going through the billing workflow. This is done for order details of products that are in the "Skip Review" billing mode.

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -377,7 +377,7 @@ class OrderDetail < ApplicationRecord
     end
 
     event :to_reconciled do
-      transitions to: :reconciled, from: :complete
+      transitions to: :reconciled, from: :complete, guard: :actual_total
     end
 
     event :to_canceled do
@@ -386,7 +386,8 @@ class OrderDetail < ApplicationRecord
   end
 
   def reconcile_if_skip_review
-    change_status!(OrderStatus.reconciled) if complete? && product.billing_mode == "Skip Review"
+    should_change_status = complete? && product.billing_mode == "Skip Review"
+    change_status!(OrderStatus.reconciled) if should_change_status
   end
 
   # block will be called after the transition, but before the save
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -352,6 +352,8 @@ class OrderDetail < ApplicationRecord
     end
   end
 
+  after_save :reconcile_if_skip_review, if: :complete?
+
   include AASM
 
   CANCELABLE_STATES = [:new, :inprocess, :complete].freeze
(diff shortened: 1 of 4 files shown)


spec/models/order_detail_spec.rb
```diff
@@ -2017,4 +2017,27 @@ def set_cancellation_cost_for_all_policies(cost)
     end
 
   end
+
+  describe "#reconcile_if_skip_review" do
+    before do
+      order_detail.actual_cost = 10
+      order_detail.actual_subsidy = 0
```
jossim, line 2024: Look into a spec when these aren't set. Look at how this can happen.

Make a model and system spec.

spec/system/admin/billing_mode_workflow_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        click_button "Save"
+        expect(order_detail.reload.reconciled?).to be true
+      end
+    end
+
+    context "invalid account" do
+      let(:accepts_po) { false }
+      let(:account) { po_account }
+
+      it "should not have a valid cart" do
```
jossim, line 49: Try to place an order, instead of looking at a cart

spec/system/admin/billing_mode_workflow_spec.rb
```diff
@@ -0,0 +1,94 @@
...
+  let(:order_detail) { order.order_details.first }
+  let(:director) { create(:user, :facility_director, facility:) }
+  let(:user) { order.account.owner.user }
+  let(:logged_in_user) { nil }
+
+  before do
+    login_as logged_in_user
+  end
+
+  describe "'Default' billing mode"
```
giladshanan, line 20: You can remove this

app/models/order_detail.rb
```diff
@@ -383,6 +385,14 @@ def self.ordered_or_reserved_in_range(start_date, end_date)
     end
   end
 
+  def reconcile_if_skip_review
+    should_change_status = actual_total &&
+                           complete? &&
+                           product.billing_mode == "Skip Review"
+
+    change_status!(OrderStatus.reconciled) if should_change_status
+  end
```
giladshanan, line 394: I'd pull this out into a little method like:
```
def auto_reconcile?
  product.billing_mode == "Skip Review" &&
    complete? &&
    actual_total
end
```
...then 
```suggestion
  def reconcile_if_skip_review
    change_status!(OrderStatus.reconciled) if auto_reconcile?
  end
```

giladshanan (approved):
Couple small suggestions, nothing blocking

PR #3631, opened 2023-08-11
[#161905] Invalid image format uploads

jossim:
# Release Notes

It was possible to upload an unprocessable image format (e.g. JFIJ) without getting any sort of error in the UI.

This adds `processable_image` to the validation, which will provide an error in the UI, and it lists the valid image formats

# Screenshot

Screen Shot 2023-08-11 at 5 39 51 PM


giladshanan (approved):

giladshanan:
@jossim In your testing you said that jfij images are actually uploaded, even if it can't be processed. Can you clarify the steps and outcomes?  Can you successfully save and attach a jfij file?

When I test this locally, I am not able to attach a jfij file.  There is some processing happening in the background, but I don’t think the file is actually uploaded.

PR #3641, opened 2023-08-22
[#161924] Journal 90 day popup bug

jossim:
# Release Notes

The "Select All" link does not have an event listener on it. So when transactions are selected using that link, the  Journal 90 day popup is not setup correctly.

This adds a listener to the "Select All" link


jossim:
Probably should up `spec/system/creating_a_journal_spec.rb` to use the "Select All" link

giladshanan:
@jossim did you confirm with Lin that this as a likely culprit? we should fix and add a spec either way.  might also be worth taking some time to refactor `spec/system/creating_a_journal_spec.rb` so it doesn't fail randomly.

giladshanan (approved):

PR #3645, opened 2023-08-28
[#161859] Moves nonbillable user creation to the NonbillableAccount class

jossim:
# Release Notes

Creating the user who owns the Nonbillable account is more at home in the NonbillableAccount class than it is in the User class.


giladshanan (approved):
Pending CI

PR #3676, opened 2023-09-15
Fix external user and nonbillable instrument bug

jossim:
# Release Notes

- Addresses this Rollbar error: https://app.rollbar.com/a/oregonstate/fix/item/RELMS/38/occurrence/365621698022#detail
- Allows external users to reserve or purchase any Nonbillable product
- Requires a payment source selection for Skip Review products
- Updates the setup_global method
- Big improvement in improved spec coverage

## Manual Tasks

This adds a new `after_create` callback on `Product`. Instead of creating a price policy for just the base price group, it creates one for all global price groups. For existing `skip_review_mode?` products to work correctly, they will need to have these price policies added manually:
`bundle exec rake 'price_policies:update_skip_review[commit]'`

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -24,7 +24,7 @@ class Product < ApplicationRecord
   has_one :product_display_group, through: :product_display_group_product
 
   after_create :create_default_price_group_products
-  after_create :create_nonbillable_price_policy, if: :skip_order_review?
+  after_create :create_skip_review_price_policies, if: :skip_order_review?
 
   email_list_attribute :training_request_contacts
 
@@ -209,8 +209,8 @@ class Product < ApplicationRecord
     end
   end
 
-  def create_nonbillable_price_policy
-    PricePolicyBuilder.create_nonbillable_for(self)
+  def create_skip_review_price_policies
+    PricePolicyBuilder.create_skip_review_for(self)
   end
 
   def available_for_purchase?
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -24,7 +24,8 @@ class Product < ApplicationRecord
   has_one :product_display_group, through: :product_display_group_product
 
   after_create :create_default_price_group_products
-  after_create :create_skip_review_price_policies, if: :skip_order_review?
+  after_create :create_skip_review_price_policies, if: :skip_review_mode?
+  after_create :create_nonbillable_price_policy, if: :nonbillable_mode?
 
   email_list_attribute :training_request_contacts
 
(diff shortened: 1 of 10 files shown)


giladshanan:
@jossim It looks like there are only a few places where we ask an order detail for its price groups. One of them is in this PR.  Another one already has similar logic:
```
  def can_purchase_order_detail?(order_detail)
    price_group_ids = if skip_order_review?
                        [PriceGroup.nonbillable.id]
                      else
                        order_detail.price_groups.map(&:id)
                      end

    can_purchase? price_group_ids
  end
```

There are only 2 other places I can see it called, in `cheapest_price_policy` and `in_window?` method.  
I think it's worth exploring a refactor of `OrderDetail#price_groups`:
```
  def price_groups
    return PriceGroup.nonbillable if product.nonbillable_mode?

    groups = user.price_groups
    groups += account.price_groups if account
    groups.uniq
  end
```

You could start by making the change and seeing if there are any specs that complain.
I'd recommend also doing some browser testing locally to understand any side effects.
Finally, I'd ask you to confirm if we should be doing this for both Skip Review and Nonbillable, or only Nonbillable.  I think that we just want Nonbillable here, since Skip Review products should require the user to have access to a Billable payment source.

app/services/price_policy_builder.rb
```diff
@@ -14,23 +14,25 @@ def self.get_new_policies_based_on_most_recent(product, start_date)
...
-      cancellation_cost: 0,
-      usage_subsidy: 0,
-      unit_cost: 0,
-      unit_subsidy: 0,
-      can_purchase: true,
-      charge_for: "reservation",
-      note: "Price rule automatically created because of billing mode"
-    )
+  def self.create_skip_review_for(product)
+    PriceGroup.globals.each do |price_group|
```
giladshanan, line 18: What about non-global price groups?  I think it's possible to add users and payment sources to non-global price groups, but not sure if this breaks things or not.  Please add specs to confirm.

app/models/product.rb
```diff
@@ -209,8 +209,8 @@ def create_default_price_group_products
     end
   end
 
-  def create_nonbillable_price_policy
-    PricePolicyBuilder.create_nonbillable_for(self)
+  def create_skip_review_price_policies
```
giladshanan, line 212: I'm not sure about this - maybe it makes sense to have different logic for Skip Review vs Nonbillable.

app/services/price_policy_builder.rb
giladshanan: Also, we're adding a new dependency between Price Groups and price rules here.  I think `setup_global` should create price rules for all skip review products.

giladshanan (commented):
Some questions for you, let me know what you think

app/services/price_policy_builder.rb
giladshanan: Also - can you add a rake task for updating any existing skip review products?

giladshanan: Users or accounts should always belong to at least on global price group - so I think we are good here

PR #3695, opened 2023-09-28
Security update for jquery-ui-rails gem

giladshanan:
https://github.com/jquery-ui-rails/jquery-ui-rails/issues/146#issuecomment-1655225232

This gem does not appear to be actively maintained...

We should probably use something else

https://github.com/tablexi/nucore-open/security/dependabot/43
https://github.com/tablexi/nucore-open/security/dependabot/44
https://github.com/tablexi/nucore-open/security/dependabot/45


thukim (changes requested):
👍

jossim (approved):

PR #3704, opened 2023-10-02
[#161863] Instrument quantity on reconciled statements

jossim:
# Release Notes

Reservations for reconciled order details have their `billable_minutes` set to `nil`. OSU and UMass use this attribute in their statements, which means the purchased quantity is blank. Dartmouth and NU use `PriceDisplayment#wrapped_quanity` to display the quantity on the statement. This method relies on `actual_duration_mins` and `duration_mins` instead of `billable_minutes`

This PR makes `PriceDisplayment#wrapped_statement_quanity`, which uses a calculated value of, instead of the saved value for `billable_minutes`.

This method can be used by OSU and UMass to address the blank quantity on statements.

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -358,7 +358,7 @@ class Reservation < ApplicationRecord
   end
 
   def update_billable_minutes
-    update_column(:billable_minutes, calculated_billable_minutes)
+    update_column(:billable_minutes, calculated_billable_minutes) if set_billable_minutes?
   end
 
   private
@@ -384,7 +384,9 @@ class Reservation < ApplicationRecord
   end
 
   def set_billable_minutes
-    self.billable_minutes = calculated_billable_minutes if set_billable_minutes?
+    if set_billable_minutes?
+      self.billable_minutes = calculated_billable_minutes
+    end
   end
 
   def set_billable_minutes?
@@ -392,6 +394,7 @@ class Reservation < ApplicationRecord
   end
 
   def calculated_billable_minutes
+    # binding.pry
     case price_policy.charge_for
     when InstrumentPricePolicy::CHARGE_FOR.fetch(:reservation)
       TimeRange.new(reserve_start_at, reserve_end_at).duration_mins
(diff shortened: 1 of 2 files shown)


app/models/reservation.rb
```diff
@@ -388,7 +388,7 @@ def set_billable_minutes
   end
 
   def calculated_billable_minutes
```
giladshanan, line 390: @jossim looks like this is only called in hook:
```
  before_save :set_billable_minutes
```
... I think I'd rather skip `set_billable_minutes` for reconciled order details - what do you think?
I'd want to test that out with Skip Review/Nonbillable but I think it makes more sense to keep the previous billable min than set a new one that we expect to be the same anyway.

app/support/price_displayment.rb
```diff
@@ -82,6 +97,18 @@ def build_quantity_presenter
...
+    case reservation.price_policy.charge_for
+    when InstrumentPricePolicy::CHARGE_FOR.fetch(:reservation)
+      TimeRange.new(reservation.reserve_start_at, reservation.reserve_end_at).duration_mins
+    when InstrumentPricePolicy::CHARGE_FOR.fetch(:usage)
+      TimeRange.new(reservation.actual_start_at, reservation.actual_end_at).duration_mins
+    when InstrumentPricePolicy::CHARGE_FOR.fetch(:overage)
+      end_time = [reservation.reserve_end_at, reservation.actual_end_at].max
+      TimeRange.new(reservation.reserve_start_at, reservation.end_time).duration_mins
+    end
+  end
```
giladshanan, line 110: Could you put this logic in `Reservation` instead and call it `billable_duration_mins`?

Then you'd have:
```
def calculated_billable_minutes
  set_billable_minutes? ? billable_duration_mins : nil
end
```

This allows us to reuse the logic and avoid having two methods named `calculated_billable_minutes` that do subtly different things

giladshanan (changes requested):

giladshanan (approved):
👍🏼

PR #3720, opened 2023-10-05
[#161863] Tech tasks: make Reservation#billable_duration_mins public 

jossim:
# Release Notes

`billable_duration_mins` needs to be called on `Reservation`, not just within the class

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -361,17 +361,15 @@ class Reservation < ApplicationRecord
     update_column(:billable_minutes, calculated_billable_minutes)
   end
 
-  def calculated_billable_minutes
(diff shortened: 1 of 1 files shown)


giladshanan (approved):
CI is complaining

app/models/reservation.rb
```diff
@@ -361,6 +361,18 @@ def update_billable_minutes
     update_column(:billable_minutes, calculated_billable_minutes)
   end
 
+  def billable_duration_mins
```
giladshanan, line 364: @jossim I realized you had made the wrong method public here - it's `billable_duration_mins` we need to reference in statement creation

PR #3728, opened 2023-10-09
[#161895] Price group reodering

jossim:
# Release Notes

# Screenshot

Optional.

# Additional Context

### DIFF
--- a/app/controllers/price_groups_controller.rb
+++ b/app/controllers/price_groups_controller.rb
@@ -21,7 +21,7 @@ class PriceGroupsController < ApplicationController
 
   # GET /facilities/:facility_id/price_groups
   def index
-    @price_groups = current_facility.price_groups
+    @price_groups = current_facility.price_groups.reorder(display_order: :asc, name: :asc)
   end
 
   # GET /facilities/:facility_id/price_groups/:id
--- a/app/controllers/schedule_rules_controller.rb
+++ b/app/controllers/schedule_rules_controller.rb
@@ -39,7 +39,7 @@ class ScheduleRulesController < ApplicationController
       end_min: 0,
     )
 
-    PriceGroup.globals.each do |price_group|
+    PriceGroup.reorder(display_order: :asc, name: :asc).globals.each do |price_group|
       @schedule_rule.price_group_discounts.build(price_group: price_group, discount_percent: 0)
     end
   end


giladshanan (changes requested):
A few small nitpicks - and the schema file needs to be cleaned up

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -19,15 +19,17 @@
...
-    = pgd.hint :external_discount_hint
+- unless @highlighted_price_group_discounts&.empty?
+  .well
+    %h4= t("views.schedule_rules.highlighted_heading")
+    = f.association :price_group_discounts, collection: @highlighted_price_group_discounts do |pgd|
+      = render partial: "price_group_discount", locals: { pgd: pgd }
+
+- unless @non_highlighted_price_group_discounts&.empty?
+  .well
+    %h4= t("views.schedule_rules.non_highlighted_heading")
```
giladshanan, line 30: Maybe we should only show the label that says “Other price groups” if highlighted price groups exist.  This makes it easy to keep the existing look if that’s what folks want.  Thoughts?

giladshanan (approved):
Nice!  Left you one non-blocking suggestion, plus I think it would be good to update the seeds file

app/views/schedule_rules/_schedule_rule_fields.html.haml
```diff
@@ -27,7 +27,10 @@
 
 - unless @non_highlighted_price_group_discounts&.empty?
   .well
-    %h4= t("views.schedule_rules.non_highlighted_heading")
+    - if !@highlighted_price_group_discounts.present? || @highlighted_price_group_discounts&.empty?
```
giladshanan, line 30: @jossim would this work instead?
`- if @highlighted_price_group_discounts&.blank?`

jossim: @giladshanan ahh yeah. I added this because of a spec failure and didn't clean it up

CorumUMass:
I noticed this weird duplication in the list of discounts, I think just added one to the highlighted section and two to the non-highlighted section, then Saved.
I don't have access to the original ticket, so sorry if this is still in progress...
161895_weird_discount_list

LeticiaErrandonea:
Yes, this has been reported and we have a fix here https://github.com/tablexi/nucore-open/pull/3757

PR #3735, opened 2023-10-11
[#161991] [Shared dev] Duration based pricing - Select pricing mode

LeticiaErrandonea:
# Release Notes
Add pricing mode to instrument. It can either be Schedule Rule or Duration (defaults to the first one). For new/edit, user sees a radio button. Also, it's in the Details tab.

# Screenshots
## Form
image

## View
image

# Additional Context

### DIFF
--- a/app/models/concerns/products/schedule_rule_support.rb
+++ b/app/models/concerns/products/schedule_rule_support.rb
@@ -9,8 +9,6 @@ module Products::ScheduleRuleSupport
     has_many :product_access_groups, foreign_key: :product_id, inverse_of: :product
   end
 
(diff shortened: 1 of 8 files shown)


app/models/concerns/products/schedule_rule_support.rb
```diff
@@ -9,6 +9,8 @@ module Products::ScheduleRuleSupport
     has_many :product_access_groups, foreign_key: :product_id, inverse_of: :product
   end
 
+  PRICING_MODES = ["Schedule Rule", "Duration"].freeze
```
giladshanan, line 12: Since we only want this on instruments, can you move this back to the `Instrument` class?

PR #3756, opened 2023-10-19
[#161993] Duration based pricing - (3) Create duration and schedule tabs

LeticiaErrandonea:
# Release Notes
Both Schedule Rule discounts and duration-based discounts cannot be set for the same instrument.

# Screenshot
#### View
image

#### New 
image

#### Edit
image

### Duration pricing tab
image

### Scheduling tab
image

### Duration pricing tab
image

# Additional Context

### DIFF
--- a/app/controllers/instrument_duration_rates_controller.rb
+++ b/app/controllers/instrument_duration_rates_controller.rb
@@ -47,6 +47,8 @@ class InstrumentDurationRatesController < ApplicationController
   def init_instrument_duration_rates
     @product = Product.find_by!(url_name: params[:id])
 
+    return unless @product.is_a?(Instrument)
+
     set_product_duration_rates
   end
 
--- a/app/views/admin/shared/_tabnav_product.html.haml
+++ b/app/views/admin/shared/_tabnav_product.html.haml
@@ -16,8 +16,7 @@
       = tab t("views.admin.products.tabnav.pricing_rules"), [current_facility, @product, PricePolicy], (secondary_tab == "pricing_rules")
   - if @product.is_a?(Instrument)
     = tab t("views.admin.products.tabnav.restrictions"), edit_facility_price_group_product_path(current_facility, @product), (secondary_tab == "restrictions")
-    - if @product.duration_pricing_mode?
-      = tab t("views.admin.products.tabnav.duration_pricing"), edit_facility_instrument_duration_rate_path(current_facility, @product), (secondary_tab == "duration_pricing")
+    = tab t("views.admin.products.tabnav.duration_pricing"), edit_facility_instrument_duration_rate_path(current_facility, @product), (secondary_tab == "duration_pricing")
     = tab t("views.admin.products.tabnav.reservations"), facility_instrument_schedule_path(current_facility, @product), (secondary_tab == "reservations")
   = tab t("views.admin.products.tabnav.documentation"),
     [current_facility, @product, :file_uploads, file_type: "info"],
--- a/spec/system/admin/instrument_scheduling_tab_spec.rb
+++ b/spec/system/admin/instrument_scheduling_tab_spec.rb
@@ -97,6 +97,7 @@ RSpec.describe "Instrument Scheduling Tab" do
       it "shows all Schedule rule fields" do
         expect(page).to have_content("Editing Schedule Rule")
         expect(page).to have_content("Price Group Discounts")
+        # As the instrument has no highlighted price groups, header should not be displayed
         expect(page).not_to have_content("Highlighted Price Groups")
         expect(page).to have_content("Price Groups")
       end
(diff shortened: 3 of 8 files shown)


app/controllers/instrument_duration_rates_controller.rb
```diff
@@ -47,6 +47,8 @@ def manage
   def init_instrument_duration_rates
     @product = Product.find_by!(url_name: params[:id])
 
+    return unless @product.is_a?(Instrument)
```
giladshanan, line 50: are we showing this tab for other product types?

app/views/admin/shared/_tabnav_product.html.haml
```diff
@@ -16,8 +16,7 @@
       = tab t("views.admin.products.tabnav.pricing_rules"), [current_facility, @product, PricePolicy], (secondary_tab == "pricing_rules")
   - if @product.is_a?(Instrument)
     = tab t("views.admin.products.tabnav.restrictions"), edit_facility_price_group_product_path(current_facility, @product), (secondary_tab == "restrictions")
-    - if @product.duration_pricing_mode?
-      = tab t("views.admin.products.tabnav.duration_pricing"), edit_facility_instrument_duration_rate_path(current_facility, @product), (secondary_tab == "duration_pricing")
+    = tab t("views.admin.products.tabnav.duration_pricing"), edit_facility_instrument_duration_rate_path(current_facility, @product), (secondary_tab == "duration_pricing")
```
giladshanan, line 19: I think we should add a conditional like line 17 `- if @product.is_a?(Instrument)`

spec/system/admin/instrument_scheduling_tab_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    let!(:instrument) do
+      FactoryBot.create(:setup_instrument, pricing_mode: "Duration", facility: facility)
+    end
+
+    context "new schedule rule" do
+      before do
+        visit new_facility_instrument_schedule_rule_path(facility, instrument)
+      end
+
+      it "shows common Schedule rule fields" do
```
giladshanan, line 23:
```suggestion
      it "shows common Schedule rule fields, but NOT Price Group Discounts" do
```

spec/system/admin/instrument_scheduling_tab_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+    let!(:instrument) do
+      FactoryBot.create(:setup_instrument, pricing_mode: "Schedule Rule", facility: facility)
+    end
+
+    context "new schedule rule" do
+      before do
+        visit new_facility_instrument_schedule_rule_path(facility, instrument)
+      end
+
+      it "shows common Schedule rule fields" do
```
giladshanan, line 73:
```suggestion
      it "shows all Schedule rule fields" do
```

spec/system/admin/instrument_scheduling_tab_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+
+    context "edit schedule rules" do
+      before do
+        visit edit_facility_instrument_schedule_rule_path(facility, instrument, instrument.schedule_rules.first)
+      end
+
+      it "shows all Schedule rule fields" do
+        expect(page).to have_content("Editing Schedule Rule")
+        expect(page).to have_content("Price Group Discounts")
+        expect(page).not_to have_content("Highlighted Price Groups")
```
giladshanan, line 100: Can you add a code comment here about the intent of asserting that "Highlighted Price Groups" is absent?

spec/system/admin/instrument_scheduling_tab_spec.rb
```diff
@@ -0,0 +1,106 @@
...
+        expect(page).not_to have_content("Discount (%)")
+      end
+    end
+
+    context "edit schedule rules" do
+      before do
+        visit edit_facility_instrument_schedule_rule_path(facility, instrument, instrument.schedule_rules.first)
+      end
+
+      it "shows common Schedule rule fields" do
```
giladshanan, line 47:
```suggestion
      it "shows common Schedule rule fields, but NOT Price Group Discounts" do
```

giladshanan (approved):
Some non-blocking nitpicks - looks good!

app/controllers/instrument_duration_rates_controller.rb
LeticiaErrandonea: No, only Instruments

app/views/admin/shared/_tabnav_product.html.haml
LeticiaErrandonea: It's indented so the one from line 17 is also applied, right?

giladshanan: Right!

app/controllers/instrument_duration_rates_controller.rb
giladshanan: So why do we need this check?

LeticiaErrandonea: First, sorry I merged before resolving this comment!

I usually prefer to add those checks in case someone gets to the action by using the URL directly (not going through the buttons in the app).

PR #3764, opened 2023-10-24
[#161994] [Shared dev] Duration based pricing - (4) Track changes in event log

LeticiaErrandonea:
# Release Notes
When a duration pricing rule is either deleted, created or updated, it is logged. 

# Screenshot

Optional.

# Additional Context
Given the current strategy for Duration Rates is deleting all of them and creating new ones, I decided against just logging in the model. After all, it's not true that the user actively deleted the entry.

My criteria was:
- Updated: there was another duration rate with the same `min_duration` but different `rate`
- Created: there wasn't any other duration rate with that `min_duration`
- Deleted: there was a duration rate with a `min_duration` that is no longer set
- In any other case, nothing is logged


app/controllers/instrument_duration_rates_controller.rb
```diff
@@ -59,4 +62,29 @@ def set_product_duration_rates
...
+    end
+
+    changes = false
+
+    previous_duration_rates.zip(@product.duration_rates).each do |previous_duration_rate, new_duration_rate|
+      unless previous_duration_rate.min_duration == new_duration_rate.min_duration && previous_duration_rate.rate == new_duration_rate.rate
+        changes = true
+        break
+      end
+    end
```
giladshanan, line 82: What if you use `map` here instead:
```suggestion
    changes = previous_duration_rates.zip(@product.duration_rates).map do |previous_duration_rate, new_duration_rate|
     previous_duration_rate.min_duration != new_duration_rate.min_duration || previous_duration_rate.rate != new_duration_rate.rate
    end
```
then...
```
 LogEvent.log(...) if changes.any?
```
https://apidock.com/ruby/Enumerable/any%3F
Reads a little clearer to me, what do you think?

giladshanan (commented):
Left a couple of suggestions

giladshanan (commented):
Left a few small comments

PR #3783, opened 2023-11-01
[#162031] [Shared dev] Duration based pricing - Move form inputs to Pricing tab

LeticiaErrandonea:
# Release Notes
Instruments with Duration pricing mode have new fields in the Pricing rules table. There Stepped billing rates can be set.

Adjustment input now has a `-` before it, so we're explicit about it being a discount.

Pending:
- Validate that Stepped billing rate <= base rate
- Validate that Stepped billing adjustment <= base rate
- Improve error messages

# Screenshot
image

# Additional Context

### DIFF
--- a/app/models/duration_rate.rb
+++ b/app/models/duration_rate.rb
@@ -2,10 +2,20 @@
 
 class DurationRate < ApplicationRecord
 
-  belongs_to :instrument, foreign_key: :product_id
+  belongs_to :price_group
+  belongs_to :rate_start
 
-  validates :min_duration, presence: true
-  validates :min_duration, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
-  validates :rate, presence: true
+  validate :rate_or_subsidy
   validates :rate, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
+  validates :subsidy, numericality: { greater_than: 0, allow_blank: true }
+
+  # TODO: Validate both rate and subsidy are lesser than or equal to base rate
+  # Question: Are all base rate price groups global? Is that the way of knowing whether a price group is the Base one?
+
+  private
+  def rate_or_subsidy
+    if rate.blank? && subsidy.blank?
+      errors.add(:base, "Either Rate or Adjustment must be provided")
+    end
+  end
 end
--- a/app/models/duration_rate.rb
+++ b/app/models/duration_rate.rb
@@ -3,7 +3,7 @@
 class DurationRate < ApplicationRecord
 
   belongs_to :price_group
-  belongs_to :rate_start
+  belongs_to :rate_start, required: true
 
   validate :rate_or_subsidy
   validates :rate, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
--- a/app/models/duration_rate.rb
+++ b/app/models/duration_rate.rb
@@ -9,8 +9,14 @@ class DurationRate < ApplicationRecord
   validates :rate, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
   validates :subsidy, numericality: { greater_than: 0, allow_blank: true }
 
-  # TODO: Validate both rate and subsidy are lesser than or equal to base rate
-  # Question: Are all base rate price groups global? Is that the way of knowing whether a price group is the Base one?
+  # TODO: Validate:
+  # - Rate is <= base rate (if Price group is the “base” group or external).
+  # - Subsidy is <= base rate (if Price group is internal but not the “base”).
+  # For this it looks like we'll need to link Rate Start to Price Policy, instead of Instrument.
+  # In this context we only have a Price Group (which can have N Price policies) and an Instrument (which can have N Price policies).
+  # We need a Price Policy as that's where we store the usage_rate.
+  # Also, makes sense that Steps for an Instrument can change in time but we may want to keep the historic data.
+
 
   private
   def rate_or_subsidy
--- a/app/services/price_policy_updater.rb
+++ b/app/services/price_policy_updater.rb
@@ -62,10 +62,16 @@ class PricePolicyUpdater
         end
       end
     end
+
+    @price_policies.each do |price_policy|
+      price_policy.assign_attributes(price_group_attributes(price_policy.price_group))
+    end
   end
 
   def save_for_stepped_billing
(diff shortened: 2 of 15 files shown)


app/services/price_policy_updater.rb
```diff
@@ -39,10 +40,27 @@ def assign_attributes
...
   end
 
+  def save_product
+    keys_from_built_rate_starts = []
+
+    rate_starts_params = permitted_product_params[:rate_starts_attributes].clone
+    rate_starts_params.values.each_with_index do |dr, index|
+      if dr[:min_duration].blank?
+        if dr[:id].present?
+          rate_starts_params["#{index}"].merge! ({ _destroy: '1' })
```
giladshanan, line 54: Curious how the various validations work together - for example:
- can you have a rate start with no duration rates?
- if you have a rate start with duration rates, and you remove the rate start, what happens?

app/services/price_policy_updater.rb
```diff
@@ -39,10 +40,27 @@ def assign_attributes
...
+        else
+          keys_from_built_rate_starts << "#{index}"
+        end
+      end
+    end
+
+    @product.update({ rate_starts_attributes: rate_starts_params.except(*keys_from_built_rate_starts) })
```
giladshanan, line 61: A note for later, if we have time: I think we should have a separate method that assembles the appropriate data from params.  Then we should be able to pass it over to the price_policy:
```
def price_policy_attributes
  price_group_attributes(price_policy.price_group).merge(product_attributes)
end

def product_attributes
 keys_from_built_rate_starts = []
  ...
  { rate_starts_attributes: rate_starts_params.except(*keys_from_built_rate_starts) }
end

price_policy.assign_attributes(price_policy_attributes)

```

app/services/price_policy_updater.rb
LeticiaErrandonea: > can you have a rate start with no duration rates?
Pushed this to when I add Duration rates. I was planning on validating that and returning an error. Not sure where I'd place it, though. Maybe in the Updater itself?

> if you have a rate start with duration rates, and you remove the rate start, what happens?
Should be destroyed

giladshanan: @LeticiaErrandonea for the second case

> if you have a rate start with duration rates, and you remove the rate start, what happens?

I think we should either show an error or use JS to display a confirmation modal

LeticiaErrandonea: So if a user wants to delete the Step, they need to delete all duration rates? I think that may be a lot of work for the user, but I don't know how many price groups is common to have

app/models/duration_rate.rb
```diff
@@ -2,10 +2,26 @@
...
-  validates :min_duration, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
-  validates :rate, presence: true
+  validate :rate_or_subsidy
   validates :rate, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
+  validates :subsidy, numericality: { greater_than: 0, allow_blank: true }
+
+  # TODO: Validate:
+  # - Rate is <= base rate (if Price group is the “base” group or external).
+  # - Subsidy is <= base rate (if Price group is internal but not the “base”).
+  # For this it looks like we'll need to link Rate Start to Price Policy, instead of Instrument.
```
giladshanan, line 15: Totally in support of implementing this later, but wanted to share my thoughts.  Each product should only have one _current_ price policy per price group - and in the context of the controller I think we should have all the relevant data by way of the price policy object. The price policy knows its price group and the price groups knows the duration rate.  I don't think we need rate start here, do we?

app/models/duration_rate.rb
```diff
@@ -2,10 +2,26 @@
...
+  # In this context we only have a Price Group (which can have N Price policies) and an Instrument (which can have N Price policies).
+  # We need a Price Policy as that's where we store the usage_rate.
+  # Also, makes sense that Steps for an Instrument can change in time but we may want to keep the historic data.
```
giladshanan, line 18: We're going to log changes to the Event Log, any other historic tracking is out of scope for now.

app/services/price_policy_updater.rb
```diff
@@ -39,10 +44,47 @@ def assign_attributes
 
   def save
     ActiveRecord::Base.transaction do
-      @price_policies.all?(&:save) || raise(ActiveRecord::Rollback)
+      (assign_attributes && @price_policies.all?(&:save)) || raise(ActiveRecord::Rollback)
+    end
+  end
+
+  def assign_attributes_for_stepped_billing
```
giladshanan, line 51: We should consider how to refactor this in the next sprint

giladshanan (commented):
This is looking pretty good to me - are you planning to add specs here?  I would vote for doing logging in the next sprint unless you already have work in progress for that

app/services/price_policy_updater.rb
```diff
@@ -39,10 +44,53 @@ def assign_attributes
 
   def save
     ActiveRecord::Base.transaction do
-      @price_policies.all?(&:save) || raise(ActiveRecord::Rollback)
+      (assign_attributes && @price_policies.all?(&:save)) || raise(ActiveRecord::Rollback)
```
giladshanan, line 47: @LeticiaErrandonea should this change be dependent on the product pricing mode?

giladshanan: Actually, I think this method is only called on line 33, so we should revert this change

app/models/duration_rate.rb
LeticiaErrandonea: I understand what you mean.

I tried to rise 2 different questions here:
1. Business: does it make sense to have only one set of duration rates per instrument? Should I stick to a set of intervals for all my price policies? I definitely lack knowledge here, but I'm wondering whether it makes sense to say that for one price policy my intervals were 1-3, 3-5, 5+ hours, and then for my new Price policy I want them to be 2-4, 4-5, 5+
2. How to get the base rate. Now I understand I could just look for the current Price policy for the instrument and that's it 🙂

app/services/price_policy_updater.rb
LeticiaErrandonea: Yes, 100%!

giladshanan (approved):
Pending CI

PR #3795, opened 2023-11-06
[#161345] Order import account validation

jossim:
# Release Notes

Currently when orders are imported, the account validation is checked against the current time, rather than the time the order was actually fulfilled.

This PR changes the validation to check against the time the order was fulfilled, rather than the current time.

These changes are ported in from https://github.com/UMass-CORUM/nucore-umass/pull/574

### DIFF
--- a/spec/factories/accounts.rb
+++ b/spec/factories/accounts.rb
@@ -17,14 +17,6 @@ FactoryBot.define do
     end
   end
 
-  trait :expired do
-    transient do
-      expired_date { 1.month.ago }
-    end
-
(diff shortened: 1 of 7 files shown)


spec/factories/accounts.rb
```diff
@@ -17,6 +17,14 @@
     end
   end
 
+  trait :expired do
```
giladshanan, line 20: I think we can drop this

giladshanan (commented):
LGTM - just one comment

giladshanan (approved):

PR #3804, opened 2023-11-08
[#162084] Change model association for DurationRate

LeticiaErrandonea:
# Release Notes
[Tech task] Update DB so Stepped billing rates can be validated agains Price policy usage rate

### DIFF
--- a/app/models/rate_start.rb
+++ b/app/models/rate_start.rb
@@ -2,7 +2,7 @@
 
 class RateStart < ApplicationRecord
 
-  has_many :duration_rates
+  has_many :duration_rates, dependent: :destroy
   belongs_to :instrument, foreign_key: :product_id
 
   validates :min_duration, presence: true
--- a/app/models/rate_start.rb
+++ b/app/models/rate_start.rb
@@ -2,7 +2,7 @@
 
 class RateStart < ApplicationRecord
 
-  has_many :duration_rates, dependent: :destroy
+  has_many :duration_rates
   belongs_to :instrument, foreign_key: :product_id
 
   validates :min_duration, presence: true
--- a/app/services/price_policy_updater.rb
+++ b/app/services/price_policy_updater.rb
@@ -51,7 +51,7 @@ class PricePolicyUpdater
   def assign_attributes_for_stepped_billing
     @price_policies.each do |price_policy|
       price_group_id = price_policy.price_group.id
-      duration_rates = @params["price_policy_#{price_group_id}"]["duration_rates_attributes"]
+      duration_rates = @params.dig("price_policy_#{price_group_id}", "duration_rates_attributes")
 
       if duration_rates.present?
         duration_rates.values.each_with_index do |dr, index|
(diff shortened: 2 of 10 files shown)


app/models/rate_start.rb
```diff
@@ -2,7 +2,7 @@
 
 class RateStart < ApplicationRecord
 
-  has_many :duration_rates
+  has_many :duration_rates, dependent: :destroy
```
LeticiaErrandonea, line 5: This change doesn't actually belong here, but it's needed. I can undo it and add this in the next PR

db/migrate/20231108210613_change_associations_duration_rates.rb
```diff
@@ -0,0 +1,17 @@
+class ChangeAssociationsDurationRates < ActiveRecord::Migration[7.0]
```
LeticiaErrandonea, line 1: Do we want to do something about existing DurationRates? At least destroy them all?

app/models/rate_start.rb
giladshanan: Smaller PRs are good - if this isn't related to model changes, save it for later

app/services/price_policy_updater.rb
```diff
@@ -126,15 +126,12 @@ def permitted_params
...
-          :id,
-          :subsidy,
-          :rate,
-          :price_group_id,
-          :rate_start_id
-        ]
+        :subsidy,
+        :rate,
+        :price_policy_id,
+        :rate_start_id
```
giladshanan, line 134: This feels right

db/migrate/20231108210613_change_associations_duration_rates.rb
giladshanan: 💣  Yes let's start fresh

giladshanan (commented):
Looks good so far - this feels like the right direction

db/migrate/20231108210613_change_associations_duration_rates.rb
```diff
@@ -0,0 +1,19 @@
...
```
giladshanan, line 1: Please add frozen string literal magic comment

db/migrate/20231108210613_change_associations_duration_rates.rb
```diff
@@ -0,0 +1,19 @@
...
+  def up
+    DurationRate.destroy_all
```
giladshanan, line 3: Can you use either SQL or define `DurationRate` in the migration file?  Just to be [safe](https://jakeyesbeck.com/2021/04/10/avoid-models-in-migrations/)

db/migrate/20231108210613_change_associations_duration_rates.rb
LeticiaErrandonea: I'll have to add a sticky note somewhere so I don't keep forgetting to do it 🤦‍♀️

giladshanan (approved):

db/migrate/20231108210613_change_associations_duration_rates.rb
giladshanan: we should set up rubocop to help us remember things like this

PR #3807, opened 2023-11-09
[#162019] Update billable duration minutes calculation

jossim:
# Release Notes

# Screenshot

Optional.

# Additional Context

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -368,8 +368,9 @@ class Reservation < ApplicationRecord
     when InstrumentPricePolicy::CHARGE_FOR.fetch(:usage)
       TimeRange.new(actual_start_at, actual_end_at).duration_mins
     when InstrumentPricePolicy::CHARGE_FOR.fetch(:overage)
+      start_time = [reserve_start_at, actual_start_at].min
       end_time = [reserve_end_at, actual_end_at].max
-      TimeRange.new(reserve_start_at, end_time).duration_mins
+      TimeRange.new(start_time, end_time).duration_mins
     end
   end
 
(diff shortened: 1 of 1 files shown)


app/models/reservation.rb
```diff
@@ -368,8 +368,9 @@ def billable_duration_mins
     when InstrumentPricePolicy::CHARGE_FOR.fetch(:usage)
       TimeRange.new(actual_start_at, actual_end_at).duration_mins
     when InstrumentPricePolicy::CHARGE_FOR.fetch(:overage)
+      start_time = [reserve_start_at, actual_start_at].min
       end_time = [reserve_end_at, actual_end_at].max
```
giladshanan, line 372: Have you confirmed the cause of this?
I think the problem is raised on this line - according to [rollbar](https://app.rollbar.com/a/corum/fix/item/nucore-umass/66).  In my local testing, `TimeRange` seems to handle `nil`s just fine:
```
[3]nucore-open(main)> TimeRange.new(Time.current, nil).duration_mins
nil
[4]nucore-open(main)> TimeRange.new(nil, Time.current).duration_mins
nil
```
... while `max` raises the error seen in rollbar:
```
[2]nucore-open(main)> [Time.current, nil].max
ArgumentError: comparison of NilClass with ActiveSupport::TimeWithZone failed
```

giladshanan (changes requested):
I don't think we're solving the right problem here - and this code is shared between statements and other pricing logic.  I'd like to see some specs that demonstrate the problem.  I suspect that `actual_end_at` is not always set, and perhaps it should be for overage instruments.  Maybe this is a problem order?  Maybe looking at the statements in the ticket could shine some light on this

jossim:
The changes need to be made in the UMass and OSU engines instead of in Open

PR #3811, opened 2023-11-13
[#161345] Fix: Nonbillable account open parameter

jossim:
# Release Notes

In https://github.com/tablexi/nucore-open/pull/3795, `Account#validate_against_product` adds a `fulfillment_time` parameter to what is passes into `account_open?`, making it necessary for each account subclass to add that parameter to its method signature.

This PR adds `fulfillment_time`  to `NonbillableAccount#account_open?`

### DIFF
--- a/app/models/nonbillable_account.rb
+++ b/app/models/nonbillable_account.rb
@@ -10,7 +10,7 @@ class NonbillableAccount < Account
     first || create
   end
 
-  def account_open?(_account_number)
+  def account_open?(_account_number, fulfillment_time: nil)
     true
   end
 
--- a/app/models/nonbillable_account.rb
+++ b/app/models/nonbillable_account.rb
@@ -10,7 +10,10 @@ class NonbillableAccount < Account
     first || create
   end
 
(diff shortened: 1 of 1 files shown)


LeticiaErrandonea (approved):

app/models/nonbillable_account.rb
```diff
@@ -10,7 +10,7 @@ def self.singleton_instance
     first || create
   end
 
-  def account_open?(_account_number)
+  def account_open?(_account_number, fulfillment_time: nil)
```
LeticiaErrandonea, line 13: Sorry, forgot to add a nit comment: can we prefix with `_` as this param isn't used?

giladshanan: I agree with Leti here

jossim: @giladshanan @LeticiaErrandonea since `fulfillment_time` is a named parameter, putting `_` in front gives an error (`ArgumentError: unknown keyword: :fulfillment_time`), so I updated the method to accept anything.

giladshanan (approved):

PR #3819, opened 2023-11-16
[#161995] Stepped billing calculations

LeticiaErrandonea:
# Release Notes

Implements the calculation logic to apply the stepped billing rates set by admin users.

Includes a couple small changes as well:
- Store rate and subsidy as per-minute values for consistency with price policies
- A small visual fix for displaying the Initial Rate start as 0hrs in a read only input

### DIFF
--- a/app/models/duration_rate.rb
+++ b/app/models/duration_rate.rb
@@ -2,8 +2,6 @@
 
 class DurationRate < ApplicationRecord
 
-  SIXTY_MIN = 60.0
-
   belongs_to :price_policy, required: true
 
   validate :rate_or_subsidy
@@ -14,6 +12,24 @@ class DurationRate < ApplicationRecord
   validate :subsidy_lesser_than_or_equal_to_base_rate
 
   scope :sorted, -> { order(min_duration_hours: :asc) }
+
+  def rate=(hourly_rate)
+    super
+    self[:rate] /= 60.0 if self[:rate].respond_to? :/
+  end
+
+  def subsidy=(hourly_subsidy)
+    super
+    self[:subsidy] /= 60.0 if self[:subsidy].respond_to? :/
+  end
+
+  def hourly_rate
+    rate.try :*, 60
+  end
+
+  def hourly_subsidy
+    subsidy.try :*, 60
+  end
 
   private
 
@@ -27,7 +43,7 @@ class DurationRate < ApplicationRecord
     return unless price_group.external? || price_group.master_internal?
     return unless price_policy.usage_rate && rate
 
-    if rate / SIXTY_MIN > price_policy.usage_rate
+    if rate > price_policy.usage_rate
       errors.add(:base, "Rate must be lesser than or equal to Base rate")
     end
   end
@@ -36,7 +52,7 @@ class DurationRate < ApplicationRecord
     return if price_group.external? || price_group.master_internal?
     return unless price_policy.usage_rate && subsidy
 
-    if subsidy / SIXTY_MIN > price_policy.usage_rate
+    if subsidy > price_policy.usage_rate
       errors.add(:base, "Subsidy must be lesser than or equal to Base rate")
     end
   end
--- a/app/models/duration_rate.rb
+++ b/app/models/duration_rate.rb
@@ -2,6 +2,8 @@
 
 class DurationRate < ApplicationRecord
 
+  SIXTY_MIN = 60.0
+
   belongs_to :price_policy, required: true
 
   validate :rate_or_subsidy
@@ -15,20 +17,20 @@ class DurationRate < ApplicationRecord
 
   def rate=(hourly_rate)
     super
-    self[:rate] /= 60.0 if self[:rate].respond_to? :/
+    self[:rate] /= SIXTY_MIN if self[:rate].respond_to? :/
   end
 
   def subsidy=(hourly_subsidy)
     super
-    self[:subsidy] /= 60.0 if self[:subsidy].respond_to? :/
+    self[:subsidy] /= SIXTY_MIN if self[:subsidy].respond_to? :/
   end
 
   def hourly_rate
-    rate.try :*, 60
+    rate.try :*, SIXTY_MIN
   end
 
   def hourly_subsidy
-    subsidy.try :*, 60
+    subsidy.try :*, SIXTY_MIN
   end
 
   private
--- a/app/services/price_policies/time_based_price_calculator.rb
+++ b/app/services/price_policies/time_based_price_calculator.rb
@@ -66,7 +66,7 @@ module PricePolicies
     end
 
     def build_intervals
-      sorted_duration_rates = price_policy.duration_rates.sort_by { |dr| dr.min_duration_hours || 1_000 }
+      sorted_duration_rates = price_policy.duration_rates.sorted
 
       intervals = [{ interval_start: 0, interval_end: sorted_duration_rates[0]&.min_duration_hours || Float::INFINITY, rate: usage_rate * 60, subsidy: usage_subsidy * 60 }]
 
--- a/app/services/price_policies/time_based_price_calculator.rb
+++ b/app/services/price_policies/time_based_price_calculator.rb
@@ -67,15 +67,17 @@ module PricePolicies
 
     def build_intervals
       sorted_duration_rates = price_policy.duration_rates.sorted
+      default_rate = usage_rate * 60
+      default_subsidy = usage_subsidy * 60
 
-      intervals = [{ interval_start: 0, interval_end: sorted_duration_rates[0]&.min_duration_hours || Float::INFINITY, step_rate: usage_rate * 60, step_subsidy: usage_subsidy * 60 }]
+      intervals = [{ interval_start: 0, interval_end: sorted_duration_rates[0]&.min_duration_hours || Float::INFINITY, step_rate: default_rate, step_subsidy: default_subsidy }]
 
       sorted_duration_rates.each_with_index do |duration_rate, index|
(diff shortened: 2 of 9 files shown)


app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,47 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+
+      { cost: result[:cost], subsidy: result[:subsidy] }
+    end
+
+    def build_intervals
+      sorted_duration_rates = price_policy.duration_rates.sort_by { |dr| dr.min_duration_hours || 1_000 }
+      lower_intervals = sorted_duration_rates.dup
+      higher_intervals = sorted_duration_rates.dup
+
+      if sorted_duration_rates.first.min_duration_hours > 0
```
giladshanan, line 73: I think this will always be the case - in the model we are doing validating `greater_than: 0` now

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,47 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+      else
+        higher_intervals.shift
+      end
+
+      lower_intervals.zip(higher_intervals).map do |lower, higher|
+        hourly_rate = lower ? lower.rate : usage_rate * 60
+        hourly_subsidy = lower ? lower.subsidy : usage_subsidy * 60
+        interval_start = lower&.min_duration_hours || 0
+        interval_end = higher&.min_duration_hours || Float::INFINITY
+
```
giladshanan, line 84: I'm wondering if the approach of building lower and higher intervals and zipping them is still worth the complexity.  Instead, I think we could do `sorted_duration_rates.each_with_index` and use the index to handle logic - like when index is 0, use default values for hourly rate, hourly subsidy, and interval start.  For interval end, we need to get min_duration_hours from the record at `index +1`.  This seems easier to reason about to me, what do you think?

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,47 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+        acc[:subsidy] += interval_data[:subsidy] ? interval_data[:subsidy] * time_to_charge : 0
+
+        acc
+      end
+
+      { cost: result[:cost], subsidy: result[:subsidy] }
+    end
+
+    def build_intervals
+      sorted_duration_rates = price_policy.duration_rates.sort_by { |dr| dr.min_duration_hours || 1_000 }
```
giladshanan, line 69:
```suggestion
      sorted_duration_rates = price_policy.duration_rates.sorted
```

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,47 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+      result = build_intervals.reduce({ time_left: duration_in_hours, cost: 0, subsidy: 0 }) do |acc, interval_data|
+        time_left = acc[:time_left]
+
+        interval_length = interval_data[:interval_end] - interval_data[:interval_start]
+
+        time_to_charge = [time_left, interval_length].min
+
+        acc[:time_left] -= time_to_charge
+        acc[:cost] += interval_data[:rate] ? interval_data[:rate] * time_to_charge : 0
+        acc[:subsidy] += interval_data[:subsidy] ? interval_data[:subsidy] * time_to_charge : 0
```
giladshanan, line 60: We should only ever have one or the other:
```suggestion
        if interval_data[:rate].present?
          acc[:cost] += interval_data[:rate] * time_to_charge : 0
        elsif interval_data[:subsidy].present?
          acc[:subsidy] +=  interval_data[:subsidy] * time_to_charge : 0
        end
```

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,47 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
       end
     end
 
+    def duration_pricing_cost_and_subsidy(duration_mins)
+      duration_in_hours = duration_mins / 60.0
```
giladshanan, line 49: I think we should probably be storing duration rate `rate` and `subsidy` per minute, for consistency with `usage_rate` and `usage_subsidy`

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,47 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+        higher_intervals.shift
+      end
+
+      lower_intervals.zip(higher_intervals).map do |lower, higher|
+        hourly_rate = lower ? lower.rate : usage_rate * 60
+        hourly_subsidy = lower ? lower.subsidy : usage_subsidy * 60
+        interval_start = lower&.min_duration_hours || 0
+        interval_end = higher&.min_duration_hours || Float::INFINITY
+
+        { interval_start: , interval_end:, rate: hourly_rate, subsidy: hourly_subsidy }
```
giladshanan, line 85: how about:
```suggestion
        { interval_start: , interval_end:, step_rate: hourly_rate, step_subsidy: hourly_subsidy }
```

giladshanan (changes requested):

app/services/price_policies/time_based_price_calculator.rb
LeticiaErrandonea: Based on what I see in `[cost_and_subsidy](https://github.com/tablexi/nucore-open/blob/master/app/services/price_policies/time_based_price_calculator.rb#L33-L41)`, we should set both

app/services/price_policies/time_based_price_calculator.rb
LeticiaErrandonea: So we should have the user set the hourly rate or subsidy and then modify it before storing? I think that's what we do for `usage_rate`

giladshanan: Yeah, I think there are benefits to being consistent even though it adds some complexity up front

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,51 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+
+      { cost: result[:cost], subsidy: result[:subsidy] }
+    end
+
+    def build_intervals
+      sorted_duration_rates = price_policy.duration_rates.sorted
+      default_rate = usage_rate * 60
+      default_subsidy = usage_subsidy * 60
+
+      intervals = [{ interval_start: 0, interval_end: sorted_duration_rates[0]&.min_duration_hours || Float::INFINITY, step_rate: default_rate, step_subsidy: default_subsidy }]
```
giladshanan, line 73: Can you break this into multiple lines for readability?

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,51 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+
+      sorted_duration_rates.each_with_index do |duration_rate, index|
+        if duration_rate.rate.present?
+          hourly_rate = duration_rate.rate
+          hourly_subsidy = default_subsidy
```
giladshanan, line 78: Is this correct?  I would think this should be `nil`

app/services/price_policies/time_based_price_calculator.rb
```diff
@@ -40,6 +45,58 @@ def cost_and_subsidy(duration_mins, discount_multiplier)
...
+      result = build_intervals.reduce({ time_left: duration_in_hours, cost: 0, subsidy: 0 }) do |acc, interval_data|
+        time_left = acc[:time_left]
+
+        interval_length = interval_data[:interval_end] - interval_data[:interval_start]
+
+        time_to_charge = [time_left, interval_length].min
+
+        acc[:time_left] -= time_to_charge
+        acc[:cost] += interval_data[:step_rate] ? interval_data[:step_rate] * time_to_charge : 0
+        acc[:subsidy] += interval_data[:step_subsidy] ? interval_data[:step_subsidy] * time_to_charge : 0
```
LeticiaErrandonea, line 60:
```suggestion
        acc[:cost] += (interval_data[:step_rate] || 0) * time_to_charge
        acc[:subsidy] += (interval_data[:step_subsidy] || 0) * time_to_charge
```

@giladshanan what do you think is better? I don't have a strong opinion here

giladshanan (approved):

app/models/duration_rate.rb
```diff
@@ -15,6 +15,24 @@ class DurationRate < ApplicationRecord
 
   scope :sorted, -> { order(min_duration_hours: :asc) }
 
+  def rate=(hourly_rate)
+    super
+    self[:rate] /= SIXTY_MIN if self[:rate].respond_to? :/
+  end
+
+  def subsidy=(hourly_subsidy)
+    super
```
jossim, line 24: Is `super` needed?

db/migrate/20231116194445_update_duration_rate_rate_and_subsidy_to_per_minute_values.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class UpdateDurationRateRateAndSubsidyToPerMinuteValues < ActiveRecord::Migration[7.0]
+  class DurationRate < ApplicationRecord; end
+
+  def change
```
jossim, line 6: Should this be `up` instead of `change`?

jossim (approved):

PR #3821, opened 2023-11-16
[#162000] SciShield local storage

jossim:
# Release Notes

Some instruments require training before users are allowed to use them. The SciShield API can be used to verify a user has the proper training to use the instrument. But, if the API is down, the user will get a 500 error instead of being able to reserve the instrument.

This PR creates a local copy of a user's trainings that are checked before the API, allowing the system to be used even if the API is down.


app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -0,0 +1,40 @@
...
+      @users = User.select(:id, :email)
+    end
+
+    def synchronize
+      return if api_unavailable?
+
+      ScishieldTraining.transaction do
+        ScishieldTraining.delete_all
+
+        @users.each do |user|
```
jossim, line 16:
```suggestion
        @users.find_each do |user|
```

spec/services/scishield_api_adapter_spec.rb
```diff
@@ -83,4 +83,17 @@
     end
   end
 
+  describe "Scishield training exists in database" do
```
jossim, line 86: Change this to `context`, and make a context for empty db

app/services/research_safety_adapters/scishield_api_client.rb
```diff
@@ -35,12 +35,15 @@ def keys_present?
     end
 
     def certifications_for(email)
+      training_api_request(email).body
```
jossim, line 38:
```suggestion
      response = training_api_request(email)
      response.body
```

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -0,0 +1,40 @@
...
+          adapter = ScishieldApiAdapter.new(user)
+          cert_names = adapter.certified_course_names_from_api
+
+          cert_names.each { |cert_name| ScishieldTraining.create(user_id: user.id, course_name: cert_name) }
+        end
+      rescue StandardError
```
jossim, line 22: Put in logging and Rollbar here

db/migrate/20231116155708_create_scishield_trainings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateScishieldTrainings < ActiveRecord::Migration[7.0]
```
jossim, line 1:
```suggestion
# frozen_string_literal: true

class CreateScishieldTrainings < ActiveRecord::Migration[7.0]
```

db/migrate/20231116155708_create_scishield_trainings.rb
```diff
@@ -0,0 +1,10 @@
...
+  def change
+    create_table :scishield_trainings do |t|
+      t.integer :user_id
```
jossim, line 4: add null false

spec/factories/scishield_trainings.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :scishield_training do
+    user_id { 1 }
```
jossim, line 5:
```suggestion
    user_id { User.first.id }
```

spec/services/research_safety_adapters/scishield_training_synchronizer_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    it "adds courses to database" do
+      expect(ScishieldTraining.count).to eq 0
+      synchronizer.synchronize
+      expect(ScishieldTraining.count).to eq 3
+      expect(ScishieldTraining.all.map(&:course_name)).to include(*course_names)
+    end
+  end
+
+  context "when the API reponses with error" do
```
jossim, line 35: Add some courses here, and test that they aren't removed

jossim:
This still needs a cron job setup

app/models/scishield_training.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class ScishieldTraining < ApplicationRecord
```
giladshanan, line 3: Please add a code comment explaining the intent/responsibility of this new class

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module ResearchSafetyAdapters
+
+  class ScishieldTrainingSynchronizer
```
giladshanan, line 5: Please add a code comment explaining the intent of this class

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -0,0 +1,45 @@
...
+    def initialize
+      @users = User.active.select(:id, :email)
```
giladshanan, line 7: Why are you just grabbing id and email here?  Is there a performance concern?  

We might want to know the username at some point.  Also, `ScishieldApiAdapter` depends on this data including everything it needs to know about users.  Right now that's not encoded anywhere.  I'd vote for either just grabbing all the user data like `User.active` or defining the list of attributes in `ScishieldApiAdapter` and the using it here like:
`@users = User.active.select(ScishieldApiAdapter.user_attributes)`.

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    def synchronize
+      return if api_unavailable?
```
giladshanan, line 11: I think we will want to know if the API is down - can you add a `Rollbar.error` here?

spec/services/research_safety_adapters/scishield_training_synchronizer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+  before do
+    stub_request(:get, api_endpoint)
+      .to_return(
+        status: status_code,
+        body: File.new(response)
+      )
+  end
+
+  context "when the API responses without error" do
```
giladshanan, line 20:
```suggestion
  context "when the API responds without error" do
```

spec/services/research_safety_adapters/scishield_training_synchronizer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    let(:status_code) { "200" }
+
+    it "adds courses to database" do
+      expect(ScishieldTraining.count).to eq 0
+      synchronizer.synchronize
+      expect(ScishieldTraining.count).to eq 3
+      expect(ScishieldTraining.all.map(&:course_name)).to include(*course_names)
```
giladshanan, line 27: Have you tried the [`contain_exactly`](https://medium.com/beambenefits/array-matchers-in-rspec-42b370e3983e) matcher here?
```suggestion
      expect(ScishieldTraining.all.map(&:course_name)).to contain_exactly(*course_names)
```

spec/services/research_safety_adapters/scishield_training_synchronizer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    end
+  end
+
+  context "when the API reponses with error" do
```
giladshanan, line 31:
```suggestion
  context "when the API responds with error" do
```

spec/services/research_safety_adapters/scishield_training_synchronizer_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    let(:status_code) { 500 }
```
giladshanan, line 32: Can you add specs for 404 and 403?

spec/services/scishield_api_adapter_spec.rb
```diff
@@ -4,83 +4,98 @@
...
-          status: 403,
-        )
+      3.times { create(:scishield_training, user_id: user.id) }
     end
 
-    it "raises an error" do
-      expect { adapter.certified?(anything) }.to raise_error(/Forbidden: This route can only be accessed by authenticated users/)
+    it "gets course names from database" do
+      expect(adapter.certified_course_names).to contain_exactly(
+        "Scishield Course Name 1",
```
giladshanan, line 95: These course names depend on the factory - which can be a little brittle over time.   I'd vote for either setting the names explicitly in the spec, or using `let!` to create the trainings and then referencing that variable to get the expected names.

giladshanan (changes requested):
Left you some suggestions, looks good!

app/services/research_safety_adapters/scishield_training_synchronizer.rb
jossim: Yeah, I was thinking about performance

giladshanan (approved):
Looks good, just a couple non-blocking comments

PR #3825, opened 2023-11-17
[#161996] Stepped billing Documentation

LeticiaErrandonea:
# Release Notes
Documentation for Stepped billing. Also includes already existing logic for pricing.

### DIFF
--- a/doc/product_pricing.md
+++ b/doc/product_pricing.md
@@ -1,5 +1,18 @@
 # Product Pricing
 ## All products
+### Price rules
+Products have different price groups, that can either be internal or external. Based on the user's price group and the rate set for it in the product price rules, the final charge will be different.
+
+#### Internal
+For internal price groups there is a base rate. There can be adjustments set for other internal price groups.
+
+#### External
+External price groups have a different rate and adjustments cannot be set.
+
+### Manual adjustment by admin
+In case an order needs to be charged a different sum, an admin can navigate to the Billing section and update the Price manually. It's required to also enter a pricing note.
+
+In the Order Detail view there's a button to Recalculate pricing, which returns the expected price based on the pricing rules.
 
 ## Instrument Pricing
 
--- a/doc/product_pricing.md
+++ b/doc/product_pricing.md
@@ -115,4 +115,4 @@ Reservation for 10 hours:
 = 90 + 74 + 60 + 81 = 305
 
 ### Cancel/reservation charges
-Reservations have a Reservation Cost, that is only invoked if user fail to cancel it within a set time window. For some instruments, there may not be a window, so there are charges when cancelling, regardless of the closeness to the reservation start time.
+Admins can set a Reservation Cost in the price rule, which is only invoked if the user fails to cancel their reservation within a set time window. If no window is set, cancelation charges are applied regardless of the closeness to the reservation start time.
--- a/doc/product_pricing.md
+++ b/doc/product_pricing.md
@@ -4,7 +4,7 @@
 Products are assigned to price rules which allow setting rates associated with different price groups, which can either be internal or external. Based on the user's price group and the rate set for it in the product price rules, the final charge will be different.
 
 #### Internal
-For internal price groups there is a base rate. There can be adjustments set for other internal price groups.
+For internal price groups there is a base rate. An adjustment amount can be configured for other internal price groups. This amount will be subtracted from the base rate.
 
 #### External
 External price groups have their own rates and adjustments cannot be set.
--- a/doc/product_pricing.md
+++ b/doc/product_pricing.md
@@ -70,3 +70,44 @@ Result: User is charged for 75 minutes (reservation plus 15 minutes of overage)
 
 Usage: 1:15-2:15pm
 Result: User is charged for 75 minutes (reservation plus 15 minutes of overage)
+
+### Pricing mode
+#### Schedule rule/price group based
+Schedule rules state what days and times an instrument is available for reservation.
+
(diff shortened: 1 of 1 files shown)


doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
+# Product Pricing
+## All products
+### Price rules
+Products have different price groups, that can either be internal or external. Based on the user's price group and the rate set for it in the product price rules, the final charge will be different.
```
giladshanan, line 4:
```suggestion
Products are assigned to price rules which allow setting rates associated with different price groups, which can either be internal or external. Based on the user's price group and the rate set for it in the product price rules, the final charge will be different.
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Internal
+For internal price groups there is a base rate. There can be adjustments set for other internal price groups.
```
giladshanan, line 7:
```suggestion
For internal price groups there is a base rate. An adjustment amount can be configured for other internal price groups. This amount will be subtracted from the base rate."
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### External
+External price groups have a different rate and adjustments cannot be set.
```
giladshanan, line 10:
```suggestion
External price groups have their own rates and adjustments cannot be set.
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+Reservation 1-2pm
+
+Usage: 1:15-1:45pm
+Result: User is charged for 60 minutes (the reservation length)
+
+Usage: 1:00-2:15pm
+Result: User is charged for 75 minutes (reservation plus 15 minutes of overage)
+
+Usage: 1:15-2:15pm
+Result: User is charged for 75 minutes (reservation plus 15 minutes of overage)
```
giladshanan, line 72: Is this correct?  I would think this one would be 60 min

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+### Pricing mode
+#### Schedule rule/price group based
+Schedule rules state what days and times an instrument is available for reservation.
+
+When using this Priicng mode, the user will pay an hourly rate defined by the price group they belong to.
```
giladshanan, line 78:
```suggestion
When using this Pricng mode, the user will pay an hourly rate defined by the Price Group they belong to.  Admins can set a percentage discount for each global Price Group during the times defined by each schedule rule.
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Duration based
+In this mode, there are up to 4 rates for each price policy. These rates apply to a given step, which is defined by setting a Rate start (hr). Price policies can either have no steps defined (just the usage rate) or all of them. It's invalid to set two or three steps.
```
giladshanan, line 81:
```suggestion
In this mode, sometimes referred to as "Stepped Billing" mode, there are up to 4 rates for each price policy. These rates apply to a given step, which is defined by setting a Rate start (hr). Price policies can either have no steps defined (just the usage rate) or all of them. It's invalid to set rates for one or two steps.
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+The first step will always start at 0 hours and its rate will be price policy's `usage_rate`. Note: This step is not created in the DB as a Rate Start, is inferred.
```
giladshanan, line 83:
```suggestion
The first step, also called the "Initial Rate", will always start at 0 hours and its rate will be price policy's `usage_rate`. Note: This step is not created in the DB as a Rate Start, it is inferred from the price policy record.
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+Step final price for internal groups is calculated as `Base step rate - Group step adjustment`.
```
giladshanan, line 85:
```suggestion
Step hourly rate for non-base internal groups is calculated as `Base step rate - step adjustment`.  See examples below for more info.
```

doc/product_pricing.md
```diff
@@ -0,0 +1,118 @@
...
+
+Reservation for 10 hours:
+  2 * (50 - 5)
++ 3 * (45 - 8)
++ 2 * (40 - 10)
++ 3 * (39 - 12)
+= 90 + 74 + 60 + 81 = 305
+
+### Cancel/reservation charges
+Reservations have a Reservation Cost, that is only invoked if user fail to cancel it within a set time window. For some instruments, there may not be a window, so there are charges when cancelling, regardless of the closeness to the reservation start time.
```
giladshanan, line 118: Is this accurate?
```suggestion
Admins can set a Reservation Cost in the price rule, which is only invoked if user fail to cancel their reservation within a set time window. If not window is set, cancelation charges are applied regardless of the closeness to the reservation start time.
```

giladshanan (changes requested):
Nice work, left some suggestions and questions

doc/product_pricing.md
LeticiaErrandonea: This is part of what I just copied from the previous file, but I think you're right.

doc/product_pricing.md
LeticiaErrandonea: I found this tricky. We say there are 4 steps (usage rate/subsidy + 3 duration rates) but then we want to say you either set all or none of them. I wonder what's the best way to be clear about this.

Maybe something like _It's invalid to set rates for one or two *additional* steps._?

giladshanan (approved):

PR #3826, opened 2023-11-17
[#162094] FIX: rake task to add missing global price groups to schedule rules

jossim:
# Release Notes

At least some existing schedule rules are missing price group discounts for some global price groups.

This rake task will check all schedule rules and add missing global price group discounts to them.

### DIFF
--- a/lib/tasks/schedule_rules.rake
+++ b/lib/tasks/schedule_rules.rake
@@ -24,8 +24,38 @@ namespace :schedule_rule do
     end
   end
 
+  # Usage:
+  # Dry run: rake "schedule_rule:add_missing_price_group_discounts[0]"
+  # Commit: rake "schedule_rule:add_missing_price_group_discounts[0,true]"
+  #
+  # You can skip the discount_percent argument to use existing legacy discounts if they exist
+  # Dry run: rake "schedule_rule:add_missing_price_group_discounts[,]"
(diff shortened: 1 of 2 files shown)


lib/tasks/schedule_rules.rake
```diff
@@ -23,4 +23,20 @@ namespace :schedule_rule do
       end
     end
   end
+
+  desc "Adds missing global price groups to existing schedule rules"
+  task add_missing_price_group_discounts: :environment do |_t, _args|
+    ScheduleRule.all.find_each do |schedule_rule|
```
jossim, line 29: This is similar to `PriceGroup#setup_schedule_rules`. That could be used, but we need to get the current `percent_discount` rather than setting it to 0

lib/tasks/schedule_rules.rake
```diff
@@ -23,4 +23,9 @@ namespace :schedule_rule do
       end
     end
   end
+
+  desc "Adds missing global price groups to existing schedule rules"
+  task add_missing_price_group_discounts: :environment do |_t, _args|
```
jossim, line 28: Make a discount_precent argument here

jossim, line 28: Don't use setup_schedule_rules, instead put logic in here, with dry run option

PR #3865, opened 2023-11-29
[#162000] Use batched requests to sync with the Scishield API

jossim:
# Release Notes

The Scishield API may have a rate limit in some cases, which causes it to stop responding.

This PR updates `ResearchSafetyAdapters::ScishieldTrainingSynchronizer#synchronize` to make batched requests to the API. The size of the batches, and the sleep time between them can be set via environment variables `SCISHIELD_BATCH_SIZE` and `SCISHIELD_BATCH_SLEEP_TIME`

### DIFF
--- a/app/services/research_safety_adapters/scishield_training_synchronizer.rb
+++ b/app/services/research_safety_adapters/scishield_training_synchronizer.rb
@@ -12,20 +12,40 @@ module ResearchSafetyAdapters
         Rails.logger.error(msg)
         Rollbar.error(msg) if defined?(Rollbar)
       else
-        ScishieldTraining.transaction do
-          ScishieldTraining.delete_all
+        user_certs = {}
 
-          users.find_each do |user|
-            adapter = ScishieldApiAdapter.new(user)
-            cert_names = adapter.certified_course_names_from_api
+        # In some cases, making many requests in a row results in the API not
+        # responding. Making requests in batches with sleep in between seems
+        # to make the API work in cases where this is a problem. `sleep` is put
+        # first here because `api_unavailable?` will have already made 10 API
+        # requests.
+        users.in_batches(of: batch_size) do |user_batch|
+          sleep(batch_sleep_time)
+
+          user_batch.each do |user|
+            adapter = ScishieldApiAdapter.new(user, api_client)
+            begin
+              cert_names = adapter.certified_course_names_from_api
+              user_certs[user.id.to_s] = cert_names if cert_names.presence
+            rescue => e
+              Rails.logger.error(e.message)
+              Rollbar.error(e.message) if defined?(Rollbar)
+            end
+          end
+        end
+
+        ScishieldTraining.transaction do
+          ScishieldTraining.delete(user_certs.keys.map(&:to_s))
+
+          user_certs.each do |user_id, cert_names|
             trainings_added = 0
 
-            cert_names.each do |cert_name|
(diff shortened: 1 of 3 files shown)


app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -12,20 +12,40 @@ def synchronize
...
+            rescue => e
+              Rails.logger.error(e.message)
+              Rollbar.error(e.message) if defined?(Rollbar)
+            end
+          end
+        end
+
         ScishieldTraining.transaction do
-          ScishieldTraining.delete_all
+          ScishieldTraining.delete(user_certs.keys.map(&:to_s))
```
giladshanan, line 38: Following up on https://github.com/Oregon-State/nucore-osu/pull/325#discussion_r1409517777...

I think we should start with the simplest approach - delete everything.  We want the db to reflect the API.  If the records are not there locally we check the API again anyway.  What do you think?

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -12,20 +12,40 @@ def synchronize
...
+            cert_names.each do |name|
+              st = ScishieldTraining.create(user_id:, course_name: name)
               trainings_added += 1 if st.errors.blank?
             end
 
-            puts "#{trainings_added} trainings added for user id #{user.id}"
+            puts "#{trainings_added} trainings added for user id #{user_id}"
           end
         rescue StandardError
           msg = "ScishieldTrainingSynchronizer error, rolling back transaction"
```
giladshanan, line 51:
```suggestion
        rescue StandardError => e
          msg = "Rolling back transaction, ScishieldTrainingSynchronizer error: #{e.message}"
```

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -12,20 +12,40 @@ def synchronize
...
-            adapter = ScishieldApiAdapter.new(user)
-            cert_names = adapter.certified_course_names_from_api
+          user_certs.each do |user_id, cert_names|
             trainings_added = 0
 
-            cert_names.each do |cert_name|
-              st = ScishieldTraining.create(user_id: user.id, course_name: cert_name)
+            cert_names.each do |name|
+              st = ScishieldTraining.create(user_id:, course_name: name)
               trainings_added += 1 if st.errors.blank?
```
giladshanan, line 45:
```suggestion
              trainings_added += 1 if st.persisted?
```

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -12,20 +12,40 @@ def synchronize
...
-          users.find_each do |user|
-            adapter = ScishieldApiAdapter.new(user)
-            cert_names = adapter.certified_course_names_from_api
+          user_certs.each do |user_id, cert_names|
             trainings_added = 0
 
-            cert_names.each do |cert_name|
-              st = ScishieldTraining.create(user_id: user.id, course_name: cert_name)
+            cert_names.each do |name|
+              st = ScishieldTraining.create(user_id:, course_name: name)
```
giladshanan, line 44: I'm not sure what would cause this to fail, but I think we should raise in that case:
```suggestion
              st = ScishieldTraining.create!(user_id:, course_name: name)
```

PR #3869, opened 2023-11-30
[#162000] Tech task: Update Scishield sync batch sleep time default

jossim:
# Release Notes

This updates the default sleep time to 20 seconds.

Updating environment variables for OSU is a bit cumbersome, so this just sets the default to what OSU would use


giladshanan (approved):
Are you planning to update the ENV for UMass then?  We should add a manual task to release notes

giladshanan:
@jossim let's save this for next release

PR #3871, opened 2023-12-01
[#162000] Scishield sync retry

jossim:
# Release Notes

Sometimes a request for a single user's trainings fails when sync'ing, and this user is skipped and their trainings are not sync'd.

This PR adds retries in the case of a failed request. `ScishieldTrainingSynchronizer#synchronize` will now retry up to 5 times, spacing out tries by 25% more sleep each time.


giladshanan (approved):
Nice!

PR #3872, opened 2023-12-01
[#161995] [Shared dev] Duration based pricing - (5) Duration discounts applied to user reservation

LeticiaErrandonea:
# Release Notes
* Update stepped billing calculation so it uses the step rate for internal price groups.
* Require that either none or all steps are filled.
* Show calculation for each step in the Price policy index (only internal groups).

# Screenshot
image
image


app/models/duration_rate.rb
```diff
@@ -6,12 +6,13 @@ class DurationRate < ApplicationRecord
 
   belongs_to :price_policy, required: true
 
-  validate :rate_or_subsidy
-  validates :rate, numericality: { greater_than_or_equal_to: 0, allow_blank: true }
+  validates :rate, presence: true
+  validates :rate, numericality: { greater_than_or_equal_to: 0 }
+  validates :subsidy, presence: true, if: -> { price_policy.price_group.is_internal? && !price_policy.price_group.master_internal? }
```
giladshanan, line 11: I think you can use the `price_group` method (line 61):
```suggestion
  validates :subsidy, presence: true, if: -> { price_group.is_internal? && !price_group.master_internal? }
```
... I think this is the 3rd validation where we need to check for this so maybe it's time for a method like `requires_subsidy?`.

app/assets/javascripts/price_policy.js
```diff
@@ -72,6 +72,22 @@ $(document).ready(function() {
...
+
+  function setRate(params) {
+    const columnIndex = params.dataset.index;
+    const stepBaseRate = params.value;
+    $(`input[name*='duration_rates_attributes][${columnIndex}][rate]']`).val(
+      stepBaseRate
+    );
+  }
+
+  $(".js--baseRate").each(function (_index, element) {
```
giladshanan, line 87: Small nitpick, but I'd move this up to line 75.  First we set up the page on load, then register the "change" event listener, then define the function.

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -241,30 +249,56 @@
...
+        uncheck "price_policy_#{cancer_center.id}[can_purchase]"
         uncheck "price_policy_#{external_price_group.id}[can_purchase]"
         uncheck "price_policy_#{cannot_purchase_group.id}[can_purchase]"
 
         fill_in "note", with: "This is my note"
 
         click_button "Add Pricing Rules"
 
-        expect(page).to have_content("Duration rates base Subsidy must be lesser than or equal to Base rate")
+        expect(page).to have_content("Duration rates is the wrong length (should be 3 characters)")
```
giladshanan, line 292: Let's give this a more user-friendly message like "Missing rate or subsidy for #{price_group}"

giladshanan (changes requested):
This is really close - just a few small asjustments

spec/system/admin/instrument_price_policies_controller_spec.rb
LeticiaErrandonea: Had in my TODO to update this message, will change to your suggestion

app/models/order_detail.rb
```diff
@@ -679,7 +679,13 @@ def assign_price_policy
   end
 
   def assign_actual_price
-    pp = product.cheapest_price_policy(self, time_for_policy_lookup)
+    # If time_for_policy_lookup is nil, then the default date is not set
```
giladshanan, line 682: Why is this change needed?  if I remove it, `spec/system/admin/manage_order_detail_spec.rb` still passes

db/migrate/20231206192754_delete_all_duration_rates.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class DeleteAllDurationRates < ActiveRecord::Migration[7.0]
+  class DurationRate < ApplicationRecord; end
+
+  def change
+    DurationRate.delete_all
```
giladshanan, line 7: We should make a note to do some manual testing around this before releasing

app/models/order_detail.rb
LeticiaErrandonea: That's interesting. I did a ton of debugging and updating that was the only way of having `product.cheapest_price_policy` return any policies at all

db/migrate/20231206192754_delete_all_duration_rates.rb
LeticiaErrandonea: You mean like checking that price policies aren't broken? Any other cases you can think of?

giladshanan: If we have any duration rate records on production, we should check if any orders have been placed against them.  I don't think anyone is using this yet but I just want to be safe

PR #3905, opened 2023-12-19
[#162105][WIP] Tech Task: Handle mime type errors

giladshanan:
# Release Notes

This type of error pops up from time to time, the result of a security scan.  This refactors error handling to a more modern pattern, while at the same time handling mime type errors properly so they don't create noise in Rollbar.

See:
- https://guides.rubyonrails.org/configuring.html#config-exceptions-app
- https://web-crunch.com/posts/custom-error-page-ruby-on-rails
- https://mattbrictson.com/blog/handling-rails-exceptions

## To Do
- [x] Remove unused boiler plate from generator
- [ ] Add and/or update specs
- [ ] Local testing of mime type handling


LeticiaErrandonea:
Couldn't fix the specs, but it looks like a config issue (works for local but consistently fails for test env).

I tried recommendations [here](https://stackoverflow.com/a/14141836) and didn't work. Issue is that the `ErrorsController` doesn't get hit when the exception is raised.

PR #3913, opened 2023-12-22
TECH TASK: [#162110] Add wait_for_ajax in order to wait for redirect

LeticiaErrandonea:
# Release Notes

# Screenshot

Optional.

# Additional Context
Source: https://aranair.github.io/posts/2016/07/27/capybara-and-waiting/


spec/support/matchers/wait_for_ajax.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+puts "Loading spec/support/matchers/wait_for_ajax.rb"
+
+module WaitForAjax
```
giladshanan, line 5: I think you need to call `include WaitForAjax` in the specs where you want to use this.
Also, I'd put this under `spec/support` directory, since this is not a matcher

giladshanan (changes requested):
I like the approach, curious to see how it works!

spec/support/matchers/wait_for_ajax.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def finished_all_ajax_requests?
+    page.evaluate_script("jQuery.active").zero?
+  end
+
+end
+
+RSpec.configure do |config|
+  config.include WaitForAjax, type: :js
```
giladshanan, line 20: Oh I see the idea now - it might be sufficient to move this file up a level (from /matchers into /support).  Right now the spec runs are finding the `wait_for_ajax` method is undefined.

LeticiaErrandonea:
@giladshanan specs passed now. Do we want to rebuild to check whether it's reliable?

giladshanan (approved):
@LeticiaErrandonea let's merge this in, we can always make more changes later if we find it is not reliable

PR #3934, opened 2024-01-03
[#162125] Improve error handling for case where data key is missing

giladshanan:
# Release Notes

We expect the SciShield API response to include a `data` key.
I'm not sure why it would be missing, so I want to know when this happens, but I don't want to create a lot of rollbar noise, retries from the `SciShieldTrainingSyncronizer`, or stop the sync process.

This logs a Rollbar error when the `data` key is missing. It also allows the number of retries to be set via the `SCISHIELD_SYNC_RETRY_MAX` environment variable.


LeticiaErrandonea (approved):

giladshanan (commented):
@jossim left some comments for you

spec/fixtures/scishield/empty_success.json
```diff
@@ -0,0 +1,359 @@
+{
+  "jsonapi": {
```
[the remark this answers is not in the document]
jossim: @giladshanan I removed all the data

app/services/research_safety_adapters/scishield_api_client.rb
```diff
@@ -11,6 +11,15 @@ class ScishieldApiClient
     PRIVATE_KEY = Rails.application.secrets.dig(:scishield, :rsa_private_key)
     API_ENDPOINT = Rails.application.secrets.dig(:scishield, :scishield_endpoint)
 
+    def invalid_response?(email)
+      response = training_api_request(email)
+      http_status_error = response.code.match?(/5|403|404/)
+      certification_data = JSON.parse(response.body)
+      certification_data_error = certification_data["data"].nil?
```
giladshanan, line 18: Why not use dig here?
```suggestion
      certification_data_error = certification_data.dig("data").nil?
```

jossim: Rubocop suggested this change. I can revert it

PR #3956, opened 2024-01-09
TECH TASK: [#162110] New attempt at fixing Journal specs

LeticiaErrandonea:
# Release Notes
* TECH TASK: In Journal specs, click button again if it's still visible.

# Screenshot

Optional.

# Additional Context

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -177,8 +177,13 @@ RSpec.configure do |config|
       WebMock.disable_net_connect!
     else
       WebMock.disable_net_connect!(allow_localhost: true)
+      old_default_max_wait_time = Capybara.default_max_wait_time
+      Capybara.default_max_wait_time = 5
+
       example.call
+
       WebMock.disable_net_connect!(allow_localhost: false)
+      Capybara.default_max_wait_time = old_default_max_wait_time
     end
   end
 
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -181,7 +181,7 @@ RSpec.configure do |config|
       # This should help with intermittent failures due to AJAX requests taking longer to finish.
       # Capybara.default_max_wait_time is explicitly used in wait_for_ajax to wait for AJAX requests to finish.
       old_default_max_wait_time = Capybara.default_max_wait_time
-      Capybara.default_max_wait_time = 5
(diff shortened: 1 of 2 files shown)


LeticiaErrandonea:
@giladshanan built a couple of times and worked. I think this is ready to be reviewed. I'd like to keep the wait time as low as possible, so if this makes specs pass I wouldn't increase it

spec/rails_helper.rb
```diff
@@ -177,8 +177,13 @@
       WebMock.disable_net_connect!
     else
       WebMock.disable_net_connect!(allow_localhost: true)
+      old_default_max_wait_time = Capybara.default_max_wait_time
```
giladshanan, line 180: Please add a code comment explaining the reason for this change

giladshanan (approved):
🤞🏼 looks good, just left a small request

LeticiaErrandonea:
@giladshanan given I've increased the wait time, but at least one spec still fail, I'll skip it (line 123). Let me know if you think I should skip more of them

spec/system/creating_a_journal_spec.rb
```diff
@@ -127,10 +127,12 @@
         expect(page).to have_content("We are in the year-end closing window.")
         click_button "Create Journal"
         wait_for_ajax
+        # Sometimes the first click doesn't work, so try again
+        click_button "Create Journal" unless page.has_content?("Pending Journal")
```
giladshanan, line 131: @LeticiaErrandonea what do you think of this idea?  It works for me locally, and I think I like it better than skipping the specs or sinking more time into debugging :) 

If you are not opposed to the approach, then:
1. Apply to the other flaky specs in this file
2. Test if we can remove the changes to rails_helper in this PR

spec/rails_helper.rb
```diff
@@ -177,8 +177,17 @@
       WebMock.disable_net_connect!
     else
       WebMock.disable_net_connect!(allow_localhost: true)
+      # Increase the default wait time for Capybara from 2 seconds to 5 seconds.
```
giladshanan, line 180: I think we can remove these changes and keep wait_for_ajax, since other specs are using that as well

LeticiaErrandonea: You're right

giladshanan (approved):

PR #3962, opened 2024-01-11
[#162138] Improvements stepped billing

LeticiaErrandonea:
# Release Notes
* Allow users to fill up to 4 steps for Stepped billing.
* Display per minute rate/subsidy for steps.

# Screenshot
image

# Additional Context

### DIFF
--- a/app/assets/javascripts/price_policy.js
+++ b/app/assets/javascripts/price_policy.js
@@ -21,6 +21,7 @@ $(document).ready(function() {
     const isDisabled = !$checkbox.prop("checked");
     $cells.toggleClass("disabled", isDisabled);
     $cells.find("input[type=text], input[type=hidden], input[type=checkbox]").not($checkbox).each((_i, elem) => hardToggleField($(elem), isDisabled));
+    $cells.find(".per-minute").toggleClass("hidden", isDisabled);
     // If we are enabling the row, make sure the cancellation cost field gets the correct state
     if (!isDisabled) { return $cells.find(".js--fullCancellationCost").trigger("change"); }
   };
--- a/spec/system/admin/instrument_price_policies_controller_spec.rb
+++ b/spec/system/admin/instrument_price_policies_controller_spec.rb
@@ -307,7 +307,7 @@ RSpec.describe InstrumentPricePoliciesController do
 
         click_button "Add Pricing Rules"
 
-        expect(page).to have_content("Duration rates min duration hours may not be blank")
+        expect(page).to have_content("Missing value for 'Rate start (hr)'")
       end
     end
 
@@ -329,7 +329,7 @@ RSpec.describe InstrumentPricePoliciesController do
       fill_in "price_policy_#{cancer_center.id}[usage_subsidy]", with: "25"
       fill_in "price_policy_#{cancer_center.id}[duration_rates_attributes][0][subsidy]", with: "15"
       fill_in "price_policy_#{cancer_center.id}[duration_rates_attributes][1][subsidy]", with: "15"
-      fill_in "price_policy_#{cancer_center.id}[duration_rates_attributes][2][subsidy]", with: "5"
+      fill_in "price_policy_#{cancer_center.id}[duration_rates_attributes][2][subsidy]", with: "4"
 
       fill_in "price_policy_#{external_price_group.id}[usage_rate]", with: "120.11"
       fill_in "price_policy_#{external_price_group.id}[minimum_cost]", with: "122"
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/price_policy.js
```diff
@@ -21,6 +21,7 @@ $(document).ready(function() {
     const isDisabled = !$checkbox.prop("checked");
     $cells.toggleClass("disabled", isDisabled);
     $cells.find("input[type=text], input[type=hidden], input[type=checkbox]").not($checkbox).each((_i, elem) => hardToggleField($(elem), isDisabled));
+    $cells.find(".per-minute").toggleClass("hidden", isDisabled);
```
giladshanan, line 24: why is this needed?  will any other pages be affected?

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -358,23 +357,32 @@
...
+      fill_in "price_policy_#{cancer_center.id}[duration_rates_attributes][2][subsidy]", with: ""
+      fill_in "price_policy_#{external_price_group.id}[duration_rates_attributes][2][rate]", with: ""
+
+      find_field("price_policy_#{base_price_group.id}[duration_rates_attributes][1][rate]").native.send_keys :tab # change focus to trigger the onchange event
 
-      find_field("price_policy_#{base_price_group.id}[duration_rates_attributes][2][rate]").native.send_keys :tab # change focus to trigger the onchange event
-      expect(page).to have_field("price_policy_#{cancer_center.id}[duration_rates_attributes][2][rate]", with: "20", type: :hidden)
-      expect(page).to have_field("price_policy_#{external_price_group.id}[duration_rates_attributes][2][rate]", with: "90.00")
+      expect(page).to have_field("price_policy_#{cancer_center.id}[duration_rates_attributes][1][rate]", with: "20", type: :hidden)
+      expect(page).to have_field("price_policy_#{external_price_group.id}[duration_rates_attributes][1][rate]", with: "100.00")
```
giladshanan, line 371: I'm not sure I know the intent here - can you add code comments to make it clear what this is testing?

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -358,23 +357,32 @@
...
 
       click_button "Save Rules"
 
       expect(page).to have_content("Price Rules were successfully updated.")
 
       expect(page).to have_content("Over 5 hrs")
+      expect(page).not_to have_content("Over 3 hrs")
+
       expect(page).not_to have_content("Over 4 hrs")
+      expect(page).not_to have_content("$40.00")
```
giladshanan, line 381: I'd group the `not_to` expectations together with a code comment explaining the intent

giladshanan (changes requested):
Looks good, just have a few questions and small suggestions

app/assets/javascripts/price_policy.js
LeticiaErrandonea: When this wasn't here, rate per minute was visible even when inputs were disabled (and showed no values). 

Staging:
image

Locally:
image


I can try a way of only changing this for Stepped billing, but I think it's pointless to show a "X rate per minute" if that group can't purchase and we don't show the rate value anywhere

spec/system/admin/instrument_price_policies_controller_spec.rb
LeticiaErrandonea: I grouped them by what they're testing and added comments

app/assets/javascripts/price_policy.js
giladshanan: If the behavior is the same for stepped billing and non-stepped billing then let's fix it in both places.  sound good?

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -358,28 +357,46 @@
       expect(page).to have_field("price_policy_#{base_price_group.id}[duration_rates_attributes][1][rate]", with: "40.00")
       expect(page).to have_field("price_policy_#{base_price_group.id}[duration_rates_attributes][2][rate]", with: "30.00")
 
-      fill_in "min_duration_2", with: "5"
-      fill_in "price_policy_#{base_price_group.id}[duration_rates_attributes][2][rate]", with: "20"
+      # Update values for Step 2
+      fill_in "min_duration_1", with: "5"
+      fill_in "price_policy_#{base_price_group.id}[duration_rates_attributes][1][rate]", with: "20"
+
+      # Remove all values for Step 3
```
giladshanan, line 364: What would you write in the `it` description if you had one here?  Is this accurate?
```
# When values for Step 3 are removed, you can still save Step 1 and 2
```

giladshanan (approved):
Looks good!

PR #3983, opened 2024-01-18
[#161789] Access List Enhancements

LeticiaErrandonea:
# Release Notes
* Add search feature to product Access List tab.
* Change page size to 50 for results in Product Access List tab.

# Screenshot
image

image

image

# Additional Context

### DIFF
--- a/app/assets/javascripts/_common.js
+++ b/app/assets/javascripts/_common.js
@@ -1,3 +1,30 @@
+function submitAjaxForm(e) {
+  e.preventDefault(); //Prevent the normal submission action
+  var form = $(this);
+  var submit = $("input[type='submit']", form);
+  var submit_val = submit.val();
+  submit.val("Please Wait...");
+  submit.attr("disabled", true);
+  jQuery.ajax({
+    type: "get",
+    data: form.serialize(),
+    url: form.attr("action"),
+    timeout: 25000,
+    success: function (r) {
+      $("#result").html(r);
+      submit.val(submit_val);
+      submit.attr("disabled", false);
+    },
+    error: function () {
+      $("#result").html(
+        "<p>There was an error retrieving results.  Please try again.</p>"
+      );
+      submit.val(submit_val);
+      submit.attr("disabled", false);
+    },
+  });
+};
+
 $(document).ready(function() {
   if ($("#login") && $("#username")) {
     $("#username").focus();
@@ -7,37 +34,7 @@ $(document).ready(function() {
     window.location = this.value;
   });
 
-  $("form#ajax_form").submit(function(e){
-    e.preventDefault(); //Prevent the normal submission action
-    var form = $(this);
-    var hideElementAttr = form.attr("data-hide-target");
-    var hideElement = $(`[${hideElementAttr}]`);
-    var submit = $("input[type='submit']",form);
-    var submit_val = submit.val();
-    submit.val("Please Wait...");
-    submit.attr("disabled", true);
-    jQuery.ajax({
-      type: "get",
-      data: form.serialize(),
-      url:  form.attr("action"),
-      timeout: 25000,
-      success: function(r) {
-        $("#result").html(r);
-        submit.val(submit_val);
-        submit.attr("disabled", false);
-
-        if (hideElement) {
-          var input_value = form.find("input[type=text]").val();
-          hideElement.attr("hidden", input_value.length > 0);
-        }
-      },
-      error: function() {
-        $("#result").html("<p>There was an error retrieving results.  Please try again.</p>");
-        submit.val(submit_val);
-        submit.attr("disabled", false);
-      }
-    });
-  });
+  $("form#ajax_form").submit(submitAjaxForm);
 
   $("#content").click(function(e){
     var element = $(e.target);
--- a/app/assets/javascripts/_common.js
+++ b/app/assets/javascripts/_common.js
@@ -1,16 +1,16 @@
 $(document).ready(function() {
-  if ($('#login') && $('#username')) {
-    $('#username').focus();
+  if ($("#login") && $("#username")) {
+    $("#username").focus();
   }
 
-  $('.link_select').change(function(e) {
+  $(".link_select").change(function(e) {
     window.location = this.value;
   });
 
-  $('form#ajax_form').submit(function(e){
+  $("form#ajax_form").submit(function(e){
     e.preventDefault(); //Prevent the normal submission action
     var form = $(this);
-    var hideElementAttr = form.attr('data-hide-target');
+    var hideElementAttr = form.attr("data-hide-target");
     var hideElement = $(`[${hideElementAttr}]`);
     var submit = $("input[type='submit']",form);
     var submit_val = submit.val();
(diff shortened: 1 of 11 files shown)


giladshanan (changes requested):
Nice work - just some minor changes and questions

app/views/product_users/_users_results.html.haml
```diff
@@ -0,0 +1,14 @@
+- if @product_users.nil?
```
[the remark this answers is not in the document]
LeticiaErrandonea: Yes, when product doesn't require approval

app/views/product_users/index.html.haml
```diff
@@ -25,10 +25,10 @@
     = hidden_field_tag :email, current_user.email, disabled: true
     = hidden_field_tag :format, params[:format], disabled: true
 
-  = link_to t("reports.product_users.export"), url_for(format: :csv), class: "js--exportSearchResults pull-right", data: { form: ".search_form" }
+  = link_to t("reports.product_users.export"), url_for(format: :csv), class: "js--exportSearchResults pull-right", data: { form: ".search_form", action: "export_csv" }
```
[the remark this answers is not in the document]
LeticiaErrandonea: My goal was to have something to identify the link that isn't used anywhere else. Also, both the class and the data-form are not accurate for this element

app/views/product_users/index.html.haml
```diff
@@ -20,22 +20,17 @@
         = text("import_button")
         %span{ style:"display:none" }= f.file_field :file, onchange: "form.submit()"
 
-- if @product_users.nil?
-- elsif @product_users.empty?
-  %p.notice= text("none")
```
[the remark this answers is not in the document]
LeticiaErrandonea: I moved it to users_results partial. I had to add `if @search` because depending on that we should say that there aren't any approved users or that the search results are empty. I didn't find a message I liked and that fitted well both cases

app/assets/javascripts/_common.js
```diff
@@ -23,6 +25,11 @@ $(document).ready(function() {
         $('#result').html(r);
         submit.val(submit_val);
         submit.attr("disabled", false);
+
+        if (hideElement) {
+          var input_value = form.find('input[type=text]').val();
```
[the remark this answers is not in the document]
LeticiaErrandonea: Any name suggestions? I can't think of anything that sounds good (`input-hide-element` is my best idea 😅 )

app/assets/javascripts/_common.js
```diff
@@ -1,65 +1,72 @@
...
+  $(".link_select").change(function(e) {
     window.location = this.value;
   });
 
-  $('form#ajax_form').submit(function(e){
+  $("form#ajax_form").submit(function(e){
     e.preventDefault(); //Prevent the normal submission action
     var form = $(this);
+    var hideElementAttr = form.attr("data-hide-target");
+    var hideElement = $(`[${hideElementAttr}]`);
```
giladshanan, line 14: would this work instead?  not a big deal, I just try to reach for "vanilla" JS when I can
`document.querySelectorAll("[data-hid-target]");`

app/views/product_users/_users_results.html.haml
```diff
@@ -0,0 +1,14 @@
...
+- elsif @product_users.empty?
+  -if @search_action_performed
+    %p.notice= text("no_results")
+  - else
+    %p.notice= text("empty_access_list")
```
giladshanan, line 6: Then I think we need this logic to cover all cases. maybe with some code comments:
```suggestion
- if @product_users.nil?
  - # Access list not required for this product
  %p.notice= text("none")
- elsif @product_users.empty? && @search_action_performed
  - # search results returned empty
  %p.notice= text("no_results")
-  elsif @product_users.empty?
  - # no users have access to this product, no search was performed
  %p.notice= text("empty_access_list")
```

app/views/product_users/index.html.haml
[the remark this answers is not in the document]
giladshanan: got it - I left a comment on the users_search partial

app/views/product_users/index.html.haml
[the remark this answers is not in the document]
giladshanan: interesting, should the class and data attr be changed then?

app/views/product_users/_users_results.html.haml
LeticiaErrandonea: When the product doesn't require approval, it just never loads this partial (tab isn't available). Maybe `@product_users.nil?` is unnecessary after all?

app/assets/javascripts/_common.js
LeticiaErrandonea: That doesn't work, as the `data-*` are not valid selectors to `querySelectorAll` (such as `.class` or `#id`).

app/views/product_users/index.html.haml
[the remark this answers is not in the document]
LeticiaErrandonea: Looks like the export JS code depends heavily on that class being used 😕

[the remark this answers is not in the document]
giladshanan: can you add a code comment then to explain that the class and data attribute reference searching but they are unrelated to search here

app/views/product_users/index.html.haml
```diff
@@ -20,22 +20,16 @@
...
-- else
+- if @product_users.present?
```
giladshanan, line 23: how about:
```suggestion
- if @product_users.present? && !@search_action_performed
```

app/views/product_users/_users_results.html.haml
giladshanan: Hm, I don't think we will hit `@product_users.empty?` in this context only when searching, right?  My instinct is to pull more of the `index` template (starting at line 23) into this partial so we can define the logic in one place.  I think it will be easier to reason about that way.  What do you think?

giladshanan (changes requested):
@LeticiaErrandonea I left some suggestions for reorganizing the code here - please let me know if they are clear enough or not.  It may be easier to talk through in real time

app/controllers/product_users_controller.rb
```diff
@@ -49,6 +45,18 @@ def index
...
   end
 
+  # GET /facilities/:facility_id/bundles/bundle_id/users/search
+  # GET /facilities/:facility_id/instruments/instrument_id/users/search
+  # GET /facilities/:facility_id/items/item_id/users/search
+  # GET /facilities/:facility_id/services/service_id/users/search
+  def search
+    @product_users = all_product_users(params[:search]).paginate(page: params[:page], per_page: USERS_PER_PAGE)
+
+    @search_action_performed = params[:search]
```
giladshanan, line 55: Let's called this `@search_term` since we're not using the boolean any more

app/views/product_users/_users_results.html.haml
```diff
@@ -0,0 +1,33 @@
...
+  - else
+    - if @search_action_performed.blank?
+      -# Search action wasn't performed or it was performed with the empty string
+      = render "export_csv"
+
+    - if @product.has_product_access_groups?
+      = form_for @product, url: [current_facility, @product, :update_restrictions], method: :put do |f|
+        = render "table", f: f
+        = f.submit text("update", plural_label: ProductAccessGroup.model_name.human.pluralize),
```
giladshanan, line 30: Does this translation still work in the new partial?

app/views/product_users/index.html.haml
```diff
@@ -20,22 +20,4 @@
         = text("import_button")
         %span{ style:"display:none" }= f.file_field :file, onchange: "form.submit()"
 
-- if @product_users.nil?
-- elsif @product_users.empty?
-  %p.notice= text("none")
```
giladshanan, line 25: I think there's an entry for this in `config/locales/views/admin/en.product_users.yml:7` that can be removed now

giladshanan (commented):
@LeticiaErrandonea a couple more small questions for you, almost there!

app/views/product_users/_users_results.html.haml
LeticiaErrandonea: Good catch!

giladshanan (approved):
🎉

PR #4011, opened 2024-01-30
TECH TASK: Add specs to catch ScishieldTrainingSynchronizer settings typo

jossim:
# Release Notes

This adds specs for `ScishieldTrainingSynchronizer` that test the type returned by its methods that wrap `Settings`


spec/services/research_safety_adapters/scishield_training_synchronizer_spec.rb
```diff
@@ -9,66 +9,88 @@
...
       let(:status_code) { "200" }
 
-      it "does not add courses to database" do
-        expect(ScishieldTraining.count).to eq 2
-        synchronizer.synchronize
-        expect(ScishieldTraining.count).to eq 2
+      it "adds courses to database" do
+        expect(ScishieldTraining.count).to eq 0
+        # Note, this will take take 20 seconds to run, to change this, update
+        # the value of `synchronizer.batch_sleep_time`
```
giladshanan, line 45: Can we change this value for this spec?  I think you can use metadata for `feature_setting` here, right?

giladshanan (approved):
@jossim looks good but I left one non-blocking suggestion for you

PR #4017, opened 2024-02-02
[#162140] Make price rule inputs consistent between different product types

jossim:
# Release Notes

This DRYs out some view code that's now more shared between duration billing and non-duration billing products.

### Testing notes

* Ensure all of the edit/show pages load
* Ensure the table headers and other copy make sense for each page
* Ensure the math is correct on the adjusted rates


LeticiaErrandonea:
Not sure this is the right place to have the "Billing Rates" header. In fact, I don't know if we need it for products that don't use stepped billing
image

app/models/product.rb
```diff
@@ -334,6 +334,10 @@ def alert
     nil
   end
 
+  def duration_pricing_mode?
+    false
+  end
```
LeticiaErrandonea, line 339: Given you added this, can you please update all occurrences of `@product.is_a?(Instrument) && @product.duration_pricing_mode?` to be `@product.duration_pricing_mode?`?

LeticiaErrandonea:
Hey @jossim, I think we're missing updating the pricing views for the rest of the products. 

This is what I get for an Item:
image

jossim:
@LeticiaErrandonea good catch! I've updated the other pages

e.g.
Screen Shot 2024-02-05 at 5 16 12 PM

LeticiaErrandonea:
Thanks @jossim 

A couple more places to update:
- Item price rules
image

- Service price rules
image

After these changes I think we're good

spec/system/admin/item_price_policies_controller_spec.rb
```diff
@@ -30,9 +30,9 @@
 
     click_button "Add Pricing Rules"
 
-    expect(page).to have_content(table_row("#{base_price_group.name} (Internal)", "$100.00", "$0.00", "$100.00"))
-    expect(page).to have_content(table_row("#{cancer_center.name} (Internal)", "$100.00", "$25.25", "$74.75"))
-    expect(page).to have_content(table_row("#{external_price_group.name} (External)", "$125.15", "$0.00", "$125.15"))
+    expect(page).to have_content(table_row("#{base_price_group.name} (Internal)", "$100.00", "$0.00\n$100.00"))
+    expect(page).to have_content(table_row("#{cancer_center.name} (Internal)", "$100.00", "$25.25\n$100.00"))
```
LeticiaErrandonea, line 34: Is this right? I think the math is off now (rate is $100 and subsidy $25.25, so the final rate should be $74.75)

LeticiaErrandonea:
@jossim related to my comment about specs:

I tried adding a price rule for an Internal price group and it worked fine
image

Although, it got me thinking whether it makes sense to have the first two columns, if then we display all that info in the last one

giladshanan (commented):
@jossim this is looking good, just some small updates needed for specs to pass, then we can merge in and start working on release prep

giladshanan (approved):
@jossim looks good!  can you add those testing notes to the ticket?  @LeticiaErrandonea  it might be worth adding to release notes as well

PR #4045, opened 2024-02-13
[#161990] Add project to bulk order import

jossim:
# Release Notes

This adds a project id to the bulk order importer, using the projects engine

### DIFF
--- a/app/services/order_row_importer.rb
+++ b/app/services/order_row_importer.rb
@@ -82,7 +82,7 @@ class OrderRowImporter
   end
 
   def import
-    add_product_to_order if has_valid_headers? && has_valid_fields?
+    add_to_order if has_valid_headers? && has_valid_fields?
   end
 
   def order_date
@@ -129,7 +129,7 @@ class OrderRowImporter
 
   private
 
-  def add_product_to_order
+  def add_to_order
     ActiveRecord::Base.transaction do
       begin
         @order = field(:order_number).present? ? existing_order : @order_import.fetch_or_create_order!(self)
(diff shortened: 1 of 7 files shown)


app/services/order_row_importer.rb
```diff
@@ -152,6 +154,14 @@ def add_product_to_order
     end
   end
 
+  def to_add_attributes
+    { note: field(:notes), account:, reference_id: field(:reference_id), project_id: field(:project_id) }.merge(custom_attributes)
```
LeticiaErrandonea, line 158: Should this include `project_id: field(:project_id)`? From what I understand, we use `merge(custom_attributes)` and have the `custom_attributes` from the extension add that field. Am I missing something?

LeticiaErrandonea (commented):
Left a question

app/services/order_row_importer.rb
giladshanan: @LeticiaErrandonea good catch - project id should be removed here

spec/services/order_row_importer_spec.rb
```diff
@@ -641,6 +641,7 @@
           "Order",
           "Reference ID",
           "Errors",
+          "Project ID",
```
LeticiaErrandonea, line 644: Not sure how this works, but is this okay even if a core isn't using the Project extension?

jossim: @LeticiaErrandonea this spec is in the engine too, so I just removed it here.

I explored updating it here to test the other columns, but since the engine is loaded, Project ID is always present

LeticiaErrandonea (approved):
Thanks for the changes!

PR #4052, opened 2024-02-16
[#160449] Walk up reservation flow

jossim:
# Release Notes

This adds quick reservation pages for instruments

# Screenshot

## New page
1
Screen Shot 2024-03-14 at 3 44 10 PM

2
Screen Shot 2024-03-11 at 11 34 11 AM

## Show page

End reservation
Screen Shot 2024-03-08 at 4 26 18 PM

When reservation can't be started right now
Screenshot 2024-03-15 at 11 10 45 AM

# Manual testing

If another reservation exists but has not been started, nothing should happen (the relay stays on) when the walkup reservation goes over into that reservation's time


app/controllers/quick_actions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  def create
+    index = reservation_params[:reservation_index].to_i
+    account = Account.find(reservation_params[:account])
+    reservation = @possible_reservations[index]
+
+    build_order
+
+    # Borrowed from ReservationCreator
```
giladshanan, line 26: Can we reuse instead of borrowing?

app/controllers/quick_actions_controller.rb
```diff
@@ -0,0 +1,76 @@
...
+    raise OrderPurchaseValidatorError, @order_detail if validator.invalid?
+    reservation.save_as_user!(current_user)
+    @order_detail.assign_estimated_price(reservation.reserve_end_at)
+    @order_detail.save_as_user!(current_user)
+
+    reservation.start_reservation!
+    redirect_to facility_instrument_quick_actions_path(@facility, @instrument)
+  end
+
+  def update
```
giladshanan, line 38: Is this a WIP?  If not, `start` might be a more intention-revealing name for the action. I'd expect `update` to be handling params

app/models/concerns/products/scheduling_support.rb
```diff
@@ -41,6 +41,31 @@ def next_available_reservation(after: Time.zone.now, duration: 1.minute, options
     reservation_in_week(after, duration, rules, options)
   end
 
+  def quick_action_reservations
+    quick_action_intervals.map { |i| next_available_reservation(duration: i.minutes) }
+                          .select { |r| r.valid?(:walkup_available) && r.send(:in_grace_period?) }
```
giladshanan, line 46:
```suggestion
    intervals = quick_action_intervals.map { |i| next_available_reservation(duration: i.minutes) }
    intervals.select { |r| r.valid?(:walkup_available) && r.send(:in_grace_period?) }
```

giladshanan: Why are you checking `in_grace_period?` here instead of `startable_now?` ?

app/models/reservation.rb
```diff
@@ -80,6 +80,8 @@ def self.joins_order
   scope :not_canceled, -> { where(order_details: { canceled_at: nil }) } # dubious
   scope :not_started, -> { where(actual_start_at: nil) }
   scope :not_ended, -> { where(actual_end_at: nil) }
+  scope :started, -> { where.not(actual_start_at: nil) }
+  scope :ongoing, -> { not_ended.started }
```
giladshanan, line 84: I'm not sure if we need these scopes - we already have a scope named `ongoing` on line [94](https://github.com/tablexi/nucore-open/pull/4052/files#diff-cdd375fa0e65d17f6ccd38013bbaf6a2751431bd677fefc0f6c78a10ae1b9c46R94), and we also have [current_in_use](https://github.com/tablexi/nucore-open/pull/4052/files#diff-cdd375fa0e65d17f6ccd38013bbaf6a2751431bd677fefc0f6c78a10ae1b9c46R96), which looks like what we want to use instead of `started`.

app/views/quick_actions/_reservation_information.html.haml
```diff
@@ -0,0 +1,12 @@
...
+
+%dl
+	%dt Reservation Date
+	%dd= human_date(reservation.reserve_start_at)
+
+	%dt Reservation Time
+	%dd= "#{human_time(reservation.reserve_start_at)} - #{human_time(reservation.reserve_end_at)}"
+
+	%dt Status
+	%dd= reservation.order_detail.state
```
giladshanan, line 12: This indentation is interesting - I'm surprised haml isn't yelling at you

app/views/quick_actions/new.html.haml
```diff
@@ -0,0 +1,21 @@
...
+		%dd= human_date(@reservation.reserve_start_at)
+
+		%dt Reservation Time
+		%dd= human_time(@reservation.reserve_start_at)
+
+	= form_with url: facility_instrument_quick_actions_path(@facility, @instrument) do |f|
+		%h4 Reservation Length
+		- @reservation_intervals.each_with_index do |interval, index|
+			= f.radio_button "reservation_index", index
+			= f.label "reservation_index_#{index}", "#{interval} mins"
```
giladshanan, line 15: This should work, but I was thinking the input would submit an interval as a value to the controller rather than an index.  What do you think about that?

app/models/reservation.rb
```diff
@@ -388,7 +390,7 @@ def in_grace_period?(at = Time.current)
 
     # Compare int values, not timestamps. If you do the
     # latter fractions of a second can cause false positives.
-    at >= grace_period_begin && at <= grace_period_end
+    grace_period_begin <= at && at <= grace_period_end
```
giladshanan, line 393: 💯

giladshanan (commented):
This is shaping up nicely! I left some comments

app/controllers/quick_reservations_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    @order.account = account
+    validator = OrderPurchaseValidator.new(@order_detail)
+    raise OrderPurchaseValidatorError, @order_detail if validator.invalid?
+    reservation.order_detail = @order_detail
+    reservation.save_as_user!(current_user)
+
+    @order_detail.assign_estimated_price(reservation.reserve_end_at)
+    @order_detail.save_as_user!(current_user)
+
+    @order.validate_order!
```
jossim, line 37: There's an `OrderPurchaser` class that `OrdersController` uses to purchase an order. It asks for a bit more than we have here to give. I think these 3 lines are enough to purchase an order, but I could've missed something

giladshanan:
@jossim after you finish specs, I would vote for refactoring so the form returns a duration value. Then in create/start action use that duration value to create and start a new reservation.  What do you think?

app/models/concerns/products/scheduling_support.rb
```diff
@@ -41,6 +41,31 @@ def next_available_reservation(after: Time.zone.now, duration: 1.minute, options
     reservation_in_week(after, duration, rules, options)
   end
 
+  def quick_reservation_reservations(after: Time.zone.now + 1.second)
```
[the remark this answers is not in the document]
jossim: Also not a fan, was planning on addressing it. There's no need for it outside of specs. I haven't checked, but with the refactor, maybe it's not needed for specs either

app/assets/stylesheets/app/quick_reservation.scss
```diff
@@ -0,0 +1,84 @@
+body.quick_reservation {
```
giladshanan, line 1: Can you use spaces instead of tabs for indenting here?

app/controllers/quick_reservations_controller.rb
```diff
@@ -0,0 +1,109 @@
+# frozen_string_literal: true
+
+class QuickReservationsController < ApplicationController
+  layout "quick_reservation"
```
giladshanan, line 4: What's the goal of the new layout here?

app/models/concerns/products/scheduling_support.rb
```diff
@@ -41,6 +41,35 @@ def next_available_reservation(after: Time.zone.now, duration: 1.minute, options
     reservation_in_week(after, duration, rules, options)
   end
 
+  def quick_reservation_data(after: Time.zone.now + 1.minute)
+    quick_reservation_reservations(after:).map { |r| { reserve_start_at: r.reserve_start_at, duration_mins: r.duration_mins } }
```
giladshanan, line 45: Would this work instead of relying on `quick_reservation_reservations`?
```suggestion
    quick_reservation_intervals.select { |i| walkup_available?(after, interval: i) }.compact
```

app/views/quick_reservations/_reservation_information.html.haml
```diff
@@ -0,0 +1,12 @@
+%h1.gray= "Order #: #{reservation.order_detail}"
+%h2= "#{@facility.abbreviation} / #{@instrument.name}"
+
+%dl
+	%dt Reservation Date
```
giladshanan, line 5: Change tabs to spaces

giladshanan, line 5: i18n please

app/views/quick_reservations/index.html.haml
```diff
@@ -0,0 +1,10 @@
+- if @ongoing
```
giladshanan, line 1: Tabs + i18n

app/views/quick_reservations/new.html.haml
```diff
@@ -0,0 +1,26 @@
+%h1= "#{@facility.abbreviation} / #{@instrument.name}"
```
giladshanan, line 1: Tabs + i18n

app/views/quick_reservations/index.html.haml
```diff
@@ -0,0 +1,10 @@
...
+	= render partial: "reservation_information", locals: { reservation: @ongoing }
+
+	= link_to "End Reservation", order_order_detail_reservation_switch_instrument_path(@ongoing.order_detail.order, @ongoing.order_detail, @ongoing, switch: "off", reservation_ended: "on"), class: "#{end_reservation_class(@ongoing)} end_link", data: { refresh_on_cancel: true }
```
giladshanan, line 4: This would be more readable broken into multiple lines

app/controllers/quick_reservations_controller.rb
jossim: The mockup lacks the normal header, footer, and side panel. So it seemed like another layout would be needed

giladshanan:
@jossim reminder for us to feature flag this

spec/system/quick_reservations_spec.rb
```diff
@@ -0,0 +1,150 @@
...
+      visit root_path
+      click_link "Logout"
+      login_as admin_user
+      visit show_problems_facility_reservations_path(facility)
+      expect(page).to have_content(reservation.order.id)
+      expect(page).to have_content("Missing Actuals")
+    end
+  end
+
+  # locally test: if another reservation exists but has not been started, nothing will happen (the relay stays on).
```
jossim, line 149: Make sure to test this on staging

app/models/concerns/products/scheduling_support.rb
```diff
@@ -41,6 +41,33 @@ def next_available_reservation(after: Time.zone.now, duration: 1.minute, options
     reservation_in_week(after, duration, rules, options)
   end
 
+  def quick_reservation_data(after: 1.minute.from_now)
```
jossim, line 44: `1.minute.from_now` is used here only because specs are frozen in time, and is not otherwise useful.

config/routes.rb
```diff
@@ -84,6 +84,11 @@
     resources :training_requests, only: [:index, :destroy] if SettingsHelper.feature_on?(:training_requests)
 
     resources :instruments do
+      if SettingsHelper.feature_on?(:walkup_reservations)
```
giladshanan, line 87: I think we can just flag the link - there's not much risk of folks using this without the QR codes

app/models/concerns/products/scheduling_support.rb
giladshanan: how does the time freeze cause issues?

giladshanan (changes requested):
I left some comments.  
Can you add accessibility specs please?
Also, I think this still is missing some error handling.

app/models/concerns/products/scheduling_support.rb
giladshanan: I want to make sure the spec failure isn't telling us something important

jossim: @giladshanan the issue was with `Reservations::Validations#in_the_future?`

`ReservationCreator#save` runs `ReservationCreator#save_reservation_and_order_detail`, which produces the error "Validation failed: Reserve Start must be in the future"

app/assets/stylesheets/app/quick_reservation.scss
```diff
@@ -0,0 +1,108 @@
...
+  div#content {
+    h1 {
+      font-size: 2em;
+      border: none;
+    }
+    h1.gray {
+      color: #595959;
```
jossim, line 8: This was `$navbarActiveBorderColor` but I changed it to get the accessibility spec to pass. Not sure if we want to update that variable value, or create an accessibility compliant alternative variable

app/assets/stylesheets/app/quick_reservation.scss
```diff
@@ -0,0 +1,108 @@
...
+        margin-bottom: 1em;
+        clear: both;
+      }
+      dd.underline {
+        border-bottom: 3px solid #005B8A;
+        float: left;
+      }
+    }
+    a.end_link {
+      background-color: #005B8A;
```
jossim, line 53: This was `$linkColor` but I changed it to get the accessibility spec to pass. Not sure if we want to update that variable value, or create an accessibility compliant alternative variable

giladshanan (approved):
I pushed a commit with small clean up changes, and left a few non-blocking comments.

Nice work!

app/assets/stylesheets/app/quick_reservation.scss
jossim: @giladshanan the change doesn't look very significant

Here's before

before

And here's after

after

app/views/quick_reservations/show.html.haml
```diff
@@ -0,0 +1,8 @@
+- if @reservation.startable_now?
+  = render partial: "reservation_information", locals: { reservation: @reservation }
+  
+  = form_with url: facility_instrument_quick_reservations_start_path(@facility, @instrument) do |f|
+    = f.submit t(".start")
+- else
```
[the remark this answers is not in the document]
jossim: @giladshanan I've updated this and added a screenshot

CorumUMass:
One small thing I noticed on corum-test is the header bar css is not quite right, the text is gray rather than white.
Maybe it's missing the .navbar wrapper and a.brand? See below...
```
<header><div class="navbar">
<div class="navbar-inner">
  <a class="brand" href="/"><span id="app_name">CORUM</span></a>
  </div></div>
</header>
```

PR #4066, opened 2024-02-22
FIX: [#161990] OrderRowImporter module loading issue

jossim:
# Release Notes

This removes the loading of `Projects::OrderRowImporterExtension` and instead puts that code in `OrderRowImporter::Overridable` to address an issue with `Projects::OrderRowImporterExtension` causing `OsuRelms::JournalRowExtension` not to load.

This also puts translations in `[order_row_importer_extension_spec.rb](https://github.com/tablexi/nucore-open/compare/161990-project-column-fix?expand=1#diff-ee151e6b0eb11f34bef758bcef1314bc065bab9303e53590e9d7bb6106b4a0f1)`

### DIFF
--- a/app/services/order_row_importer.rb
+++ b/app/services/order_row_importer.rb
@@ -4,6 +4,27 @@ require "date_helper" # parse_usa_import_date
 
 class OrderRowImporter
 
+  module Overridable
+
+    def validate_custom_attributes
+      if field(:project_name).present? && project.nil?
+        add_error(:project_not_found)
+      end
+    end
+
+    private
+
+    def project
+      facility.projects.active.find_by name: field(:project_name)
+    end
+
+    def custom_attributes
+      { project_id: project&.id }
+    end
+
+  end
+
+  prepend Overridable
   include DateHelper
 
   cattr_accessor(:importable_product_types) { [Item, Service, TimedService] }
@@ -18,6 +39,7 @@ class OrderRowImporter
       notes
       order_number
       reference_id
+      project_name
       errors
     ]
   end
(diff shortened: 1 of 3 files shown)


vendor/engines/projects/lib/projects/engine.rb
```diff
@@ -15,8 +15,6 @@ class Engine < Rails::Engine
       OrderDetailBatchUpdater.send :include, Projects::OrderDetailBatchUpdaterExtension
       OrdersController.send :include, Projects::OrdersControllerExtension
       Reservation.send :include, Projects::ReservationExtension
-      OrderRowImporter.send :prepend, Projects::OrderRowImporterExtension
```
LeticiaErrandonea, line 18: Should the `extension` file be removed as well?

app/services/order_row_importer.rb
```diff
@@ -4,6 +4,27 @@
...
+    def validate_custom_attributes
+      if field(:project_name).present? && project.nil?
+        add_error(:project_not_found)
+      end
+    end
+
+    private
+
+    def project
+      facility.projects.active.find_by name: field(:project_name)
```
LeticiaErrandonea, line 18: Don't know how this works, so maybe a dumb question: is this safe for schools that don't have `projects`?

LeticiaErrandonea (commented):
Left a couple of questions

giladshanan (approved):
Pending CI

app/services/order_row_importer.rb
giladshanan: A very good point - thank you @LeticiaErrandonea !

PR #4068, opened 2024-02-26
[#162170] Add Available for Cross-Facility Ordering? to products

LeticiaErrandonea:
# Release Notes
* Add option to toggle "Available for Cross-Facility Ordering" field for products.

# Screenshot
## Show
image

## Edit
image

# Additional Context


db/migrate/20240226142637_add_cross_facility_ordering_available_to_products.rb
```diff
@@ -0,0 +1,5 @@
+class AddCrossFacilityOrderingAvailableToProducts < ActiveRecord::Migration[7.0]
```
giladshanan, line 1: Can you add a frozen string comment please?

giladshanan (commented):
This looks great, my only question is about naming. The ticket refers to this feature as "Multi-Facility Projects" but that is a bit tricky.  First of all, "facility" is an overloaded term in the code (`FacilityFacilityAccountsController` 😄 ).  On top of that, we have somewhat complex translation logic that supports customization - Dartmouth uses "Resource" instead of "Facility":
```
  Facility: "!activerecord.models.facility.one!"
  Facilities: "!activerecord.models.facility.other!"
```
I've been calling this feature "Cross Core Projects" but I don't know if we have agreement on that yet. I might vote to leave this in PR until Thursday when we meetwith the Consortium

config/locales/en.models.yml
```diff
@@ -274,6 +274,7 @@ en:
         is_active: "Is Active?"
       product:
         can_be_used_by?: Approved?
+        cross_facility_ordering_available: "Available for Cross-Facility Ordering?"
```
giladshanan, line 277:
```suggestion
        cross_facility_ordering_available: "Available for Cross-!Facility! Ordering?"
```

config/locales/views/admin/en.products.yml
```diff
@@ -31,6 +31,7 @@ en:
             is_archived: "Inactivate the product, disallowing purchase and viewing"
             is_hidden: "Hide %{field} from end users; visible to staff when \"ordering on behalf\" of another user"
             billing_mode: "<b>Default</b>: Standard billing workflow, payment source required.<br><b>Skip Review</b>: Reconciled on completion, payment source required.<br><b>Nonbillable</b>: Reconciled on completion, NO payment source required."
+            cross_facility_ordering_available: "Available for Cross-Facility Ordering?"
```
giladshanan, line 34:
```suggestion
            cross_facility_ordering_available: "Available for Cross-!Facility! Ordering?"
```

giladshanan (commented):
Is there a system spec we can add to for this?  Just a happy-path spec that shows the value is saved to the db

config/locales/en.models.yml
giladshanan: If we need to call this "Cross-Resource Ordering" for Dartmouth, this is how we would do it

giladshanan (approved):

PR #4101, opened 2024-03-13
[#162204] Add sleep time to SciShield availability check

jossim:
# Release Notes

This adds a sleep of `batch_sleep_time` to each API request in `api_unavailable?`.

The sync for OSU is failing due to `api_unavailable?` returning `false` when API requests are generally possible. Adding the sleep time in `api_unavailable?` makes it more reliable.

### DIFF
--- a/app/services/research_safety_adapters/scishield_api_client.rb
+++ b/app/services/research_safety_adapters/scishield_api_client.rb
@@ -19,14 +19,14 @@ module ResearchSafetyAdapters
 
       http_status_error || certification_data_error
     end
-    
+
     def token
       return @token if @token.present?
       return nil unless keys_present?
 
       now = Time.current.to_i
       payload = {
-        iat: now,
+        iat: now - 2.minutes.to_i, # SciShield Support suggested setting this a couple of minutes into that past to make requests more reliable. There may be an issue with the API related to the time change
         exp: now + 3600,
         drupal: { uid: KEY }
       }
--- a/app/services/research_safety_adapters/scishield_api_client.rb
+++ b/app/services/research_safety_adapters/scishield_api_client.rb
@@ -26,7 +26,9 @@ module ResearchSafetyAdapters
 
       now = Time.current.to_i
       payload = {
-        iat: now - 2.minutes.to_i, # SciShield Support suggested setting this a couple of minutes into that past to make requests more reliable. There may be an issue with the API related to the time change
+        # SciShield Support suggested setting this a couple of minutes in the past to make requests more reliable.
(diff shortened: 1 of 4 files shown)


app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -80,6 +80,7 @@ def synchronize
     def api_unavailable?
       # Test API responses for 10 random users
       users.sample(10).map do |user|
+        sleep(batch_sleep_time)
```
giladshanan, line 83: I'm fine with this, but waiting 20 sec between requests seems significantly more cautious than what we do in the `synchronize` method, right?  Do we need to use a smaller batch size?  or add a 1-2 sec wait between requests?

giladshanan (commented):
I feel like `api_unavailable?` should be similar to `synchronize.  I'd like to know if there is a smaller batch size that consistently avoids (or at least reduces) retries and time outs when running `synchronize`.  If we learn something from testing, then both methods can be more alike each other - maybe both should have a few seconds of `sleep` time between each request, and use the same batch size.

giladshanan:
@jossim what do you think about trying to set `iat` a min or two in the past, as suggested by SciShield support?  we can test that on staging first but if it helps, I don't see an issue with it

app/services/research_safety_adapters/scishield_api_client.rb
```diff
@@ -19,14 +19,14 @@ def invalid_response?(email)
...
-    
+
     def token
       return @token if @token.present?
       return nil unless keys_present?
 
       now = Time.current.to_i
       payload = {
-        iat: now,
+        iat: now - 2.minutes.to_i, # SciShield Support suggested setting this a couple of minutes into that past to make requests more reliable. There may be an issue with the API related to the time change
```
giladshanan, line 29: How about making this a setting that can be set via env variable and defaults 0 minutes?
```suggestion
        # SciShield Support suggested setting this a couple of minutes in the past to make requests more reliable. 
        # There may be an issue with the API related to the time change for DST
        iat: now - Settings.scishield.iat_offset.minutes.to_i, 
```

app/services/research_safety_adapters/scishield_training_synchronizer.rb
```diff
@@ -80,6 +80,7 @@ def synchronize
     def api_unavailable?
       # Test API responses for 10 random users
       users.sample(10).map do |user|
```
giladshanan, line 82:
```suggestion
      users.sample(batch_size).map do |user|
```

giladshanan (approved):

PR #4102, opened 2024-03-14
[#162194] Create a new order when adding cross-core products to an existing order

LeticiaErrandonea:
⚠️ **This PR relies on Projects module being enabled. We should check this won't fail in Production and cause errors for any school that doesn't have it enabled.**

# Release Notes
* Create a new order when a product from a new facility is added to an existing order. If the order already exists, add to it.

# Screenshot
localhost_3000_facilities_example_orders_31

# Additional Context
## TODO

- [x] Confirm UI
- [x] Add system specs
  - Fix failing spec - it's reasonable that it's failing but it's worth it to improve the spec instead of just making it pass
- [x] Use I18n
- [x] Create partial for order table
- [x] Use `attributes` and remove `ordered_at` from `ItemAdder`

### Other PRs
- [ ] Accessibility
- [ ] Check time based input and quantity
- [ ] Consider using `merge_order` method and set `original_order` instead of setting `@merge_order`

#### #4112
- [x] Fix condition to show "CC Project ID", as it currently will cause issues if the first order detail has a project that isn't CC
- [x] Add separate table for each facility
- [x] Add total footer for new table
- [x] Add total for all the orders (top, next to CC Project ID)
- [x] Add headers

- [x] Prevent errors for schools that don't have Projects engine enabled
- [x] Change Order By to be the user that added the order detail

# Accessibility

### DIFF
--- a/app/forms/add_to_order_form.rb
+++ b/app/forms/add_to_order_form.rb
@@ -32,9 +32,9 @@ class AddToOrderForm
       facility_id: current_facility.id
     )
 
-    return add_to_order!(order_project) if @original_order.facility.id == @facility_id
-
-    if order_project.nil?
+    if @original_order.facility.id == @facility_id
+      add_to_order!(order_project)
+    elsif order_project.nil?
       create_cross_core_project_and_add_order!
     else
       facility_order_in_project = order_project.order_details.find { |od| od.order.facility_id == @facility_id }&.order
--- a/app/forms/add_to_order_form.rb
+++ b/app/forms/add_to_order_form.rb
@@ -123,7 +123,9 @@ class AddToOrderForm
 
   def add_to_order!(project = nil)
     OrderDetail.transaction do
-      merge_order.add(product, quantity, params).each do |order_detail|
+      item_adder_params = @original_order.facility_id == @facility_id ? params : params.merge(ordered_at: Time.zone.now)
+
+      merge_order.add(product, quantity, item_adder_params).each do |order_detail|
         backdate(order_detail)
 
         order_detail.set_default_status!
@@ -161,7 +163,7 @@ class AddToOrderForm
         created_by: created_by.id,
       )
 
-      product_order.add(product, quantity, params).each do |order_detail|
+      product_order.add(product, quantity, params.merge(ordered_at: Time.zone.now)).each do |order_detail|
         order_detail.set_default_status!
         order_detail.change_status!(order_status) if order_status.present?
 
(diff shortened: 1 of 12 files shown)


app/forms/add_to_order_form.rb
```diff
@@ -25,7 +25,28 @@ def initialize(original_order)
...
 
-    add_to_order!
+    @facility_id = @facility_id&.to_i
+
+    order_project = Projects::Project.find_by(
+      name: "#{current_facility.abbreviation}-#{original_order.id}",
+      facility_id: current_facility.id
+    )
+
+    return add_to_order!(order_project) if @original_order.facility.id == @facility_id
```
giladshanan, line 35: any reason to keep this separate from the if/else branch below?  I think it could be easier to reason about that way.  for example, what happens when `@original_order.facility.id == @facility_id` is truthy but `order_project.nil?` is also truthy?

app/forms/add_to_order_form.rb
```diff
@@ -25,7 +25,28 @@ def initialize(original_order)
...
+
+    if order_project.nil?
+      create_cross_core_project_and_add_order!
+    else
+      facility_order_in_project = order_project.order_details.find { |od| od.order.facility_id == @facility_id }&.order
```
giladshanan, line 40: this will need to change in the next PR, right?

giladshanan (approved):
Looks good!  Just one small non-blocking suggestion

app/forms/add_to_order_form.rb
LeticiaErrandonea: If we're adding to the original order, we don't care whether there's a project or not. If there is, we just add the new order detail to it. If there isn't, we just create the order detail.

app/forms/add_to_order_form.rb
LeticiaErrandonea: I don't think so. What's not right with this?

app/forms/add_to_order_form.rb
```diff
@@ -25,7 +25,28 @@ def initialize(original_order)
   def save
     raise(ActiveRecord::RecordInvalid, self) unless valid?
 
-    add_to_order!
+    @facility_id = @facility_id&.to_i
+
+    order_project = Projects::Project.find_by(
+      name: "#{current_facility.abbreviation}-#{original_order.id}",
+      facility_id: current_facility.id
+    )
```
giladshanan, line 33: I think we will need an association between the original_order and the cross-core project:
`belongs_to :cross_core_project`

giladshanan (approved):

PR #4157, opened 2024-04-04
[#162190] QR walk up fixes

jossim:
# Release Notes

This fixes a some  bugs in the QR walk up controller.

Handles that case
* when the minimum reserve time is larger than any of the reserve intervals
* when a non-actionable reservation is redirected to
* when the instrument has no schedule rules

### DIFF
--- a/app/controllers/quick_reservations_controller.rb
+++ b/app/controllers/quick_reservations_controller.rb
@@ -6,7 +6,7 @@ class QuickReservationsController < ApplicationController
   before_action :authenticate_user!
   load_resource :facility, find_by: :url_name
   load_resource :instrument, through: :facility, find_by: :url_name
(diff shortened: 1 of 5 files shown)


app/controllers/quick_reservations_controller.rb
```diff
@@ -133,6 +133,8 @@ def get_startable_reservation
     ongoing = reservations.ongoing.first
 
     @reservation = ongoing || startable
+
+    @reservation&.order_detail&.problem? ? nil : @reservation
```
giladshanan, line 137: I think we want:
```suggestion
    @reservation = nil if @reservation&.order_detail&.problem?
```
... right?
Can you call this `set_startable_reservation` instead of `get_startable_reservation`?

giladshanan (approved):

PR #4158, opened 2024-04-05
[Tech task] Spec clean up

giladshanan:
# Release Notes

A couple small changes to address unpredictable failures in `rspec ./spec/system/admin/adding_to_an_order_spec.rb:206`

This seems to still be failing in OSU and other repos, so we will skip for now and continue with our release.
We can keep debugging next sprint.


LeticiaErrandonea (approved):
Pending CI

PR #4189, opened 2024-04-17
[#162225] Multi-Facility - Duplicates being created

LeticiaErrandonea:
# Release Notes
* Fix bug for duplicate order details.

# Additional Context
Add a spec to add to facility order using Feature Flag.
New spec failed with old code as it found two "Make a reservation" buttons.

# Accessibility

### DIFF
--- a/app/forms/add_to_order_form.rb
+++ b/app/forms/add_to_order_form.rb
@@ -31,6 +31,7 @@ class AddToOrderForm
 
     if @original_order.facility.id == @facility_id
       add_to_order!
+      return true
     end
 
     if SettingsHelper.feature_on?(:cross_core_projects)
--- a/app/forms/add_to_order_form.rb
+++ b/app/forms/add_to_order_form.rb
@@ -31,6 +31,7 @@ class AddToOrderForm
 
     if @original_order.facility.id == @facility_id
       add_to_order!
+      return true
     end
 
     return true unless SettingsHelper.feature_on?(:cross_core_projects)
--- a/app/forms/add_to_order_form.rb
+++ b/app/forms/add_to_order_form.rb
@@ -31,10 +31,7 @@ class AddToOrderForm
 
     if @original_order.facility.id == @facility_id
       add_to_order!
-      return true
-    end
-
-    if SettingsHelper.feature_on?(:cross_core_projects)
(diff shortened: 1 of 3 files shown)


app/views/facility_orders/show.html.haml
```diff
@@ -13,8 +13,9 @@
       - if project_id
         %dl.span2
           %dt= t("views.facility_orders.show.cross_core_project_id")
-          = link_to facility_project_path(current_facility, project_id) do
-            %dd= project_id
+          %dd
+            = link_to facility_project_path(current_facility, project_id) do
+              = project_id
```
giladshanan, line 18: I think we can simplify this a bit:
```suggestion
          %dd  = link_to project_id, facility_project_path(current_facility, project_id)
```

app/forms/add_to_order_form.rb
```diff
@@ -31,6 +31,7 @@ def save
 
     if @original_order.facility.id == @facility_id
       add_to_order!
+      return true
     end
 
     return true unless SettingsHelper.feature_on?(:cross_core_projects)
```
giladshanan, line 37: I think I'd prefer to wrap the feature flagged changes in a conditional, instead of using a guard clause:
```
  if SettingsHelper.feature_on?(:cross_core_projects)
    if order_for_selected_facility.nil?
      create_cross_core_project_and_add_order!
    else
      add_to_order!
    end
  end
```
... then we can still git `true` on line 45.
This would be easier to read and clearer what needs to happen when we are ready to remove the flag.

giladshanan (changes requested):
Nice catch!  Left a couple small comments

app/forms/add_to_order_form.rb
```diff
@@ -31,14 +31,15 @@ def save
...
+      return true
     end
 
-    return true unless SettingsHelper.feature_on?(:cross_core_projects)
-
-    if order_for_selected_facility.nil?
-      create_cross_core_project_and_add_order!
-    else
-      add_to_order!
+    if SettingsHelper.feature_on?(:cross_core_projects)
```
giladshanan, line 37: maybe this should be `elsif`?

giladshanan (approved):

PR #4195, opened 2024-04-23
[#162263] Multi-Facility - 5c When an admin places a cross-core order, update valid payment source list

LeticiaErrandonea:
# Release Notes
* When an admin places a cross-core order, update valid payment source list.

# Screenshot
image

image

# Additional Context

# Accessibility

### DIFF
--- a/app/controllers/facility_accounts_controller.rb
+++ b/app/controllers/facility_accounts_controller.rb
@@ -12,6 +12,7 @@ class FacilityAccountsController < ApplicationController
   before_action :init_current_facility
   before_action :init_account, except: :search_results
   before_action :build_account, only: [:new, :create]
+  before_action :set_facility_accounts_for_user, only: [:accounts_available_for_order]
 
   authorize_resource :account
 
@@ -119,6 +120,14 @@ class FacilityAccountsController < ApplicationController
     end
   end
 
+  def accounts_available_for_order
+    respond_to do |format|
+      format.js do
+        render json: @facility_accounts_for_user
+      end
+    end
+  end
+
   private
 
   def available_account_types
@@ -156,6 +165,19 @@ class FacilityAccountsController < ApplicationController
       owner_user: @owner_user,
       params: params,
     ).build
+  end
+
+  def set_facility_accounts_for_user
+    order_id = params[:order_id]&.to_i
+
+    return unless order_id
+
+    order = Order.find(order_id)
+
+    return unless order
+
+    order_user = order.user
+    @facility_accounts_for_user = AvailableAccountsFinder.new(order_user, current_facility).accounts
   end
 
 end
--- a/app/controllers/facility_accounts_controller.rb
+++ b/app/controllers/facility_accounts_controller.rb
@@ -168,13 +168,13 @@ class FacilityAccountsController < ApplicationController
   end
 
   def set_facility_accounts_for_user
-    order_id = params[:order_id]&.to_i
+    order_id = params[:order_id]
     order = Order.find(order_id) if order_id
 
     return unless order
 
     order_user = order.user
-    @facility_accounts_for_user = AvailableAccountsFinder.new(order_user, current_facility).accounts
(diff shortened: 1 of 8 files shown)


app/lib/ability.rb
```diff
@@ -168,6 +168,10 @@ def facility_director_abilities(user, resource, controller)
   def facility_administrator_abilities(user, resource, controller)
     can :available_for_cross_core_ordering, Product if controller.is_a?(ProductsController)
 
+    if controller.is_a?(FacilityAccountsController) && user.facility_administrator_of_any_facility?
+      can [:accounts_available_for_order, :show], Account
```
LeticiaErrandonea, line 172: Had to add `show` as it's required by the `authorize_resource`

spec/system/admin/adding_to_an_order_spec.rb
```diff
@@ -287,9 +288,12 @@
         expect(page).not_to have_content("Cross Core Project ID")
         expect(page.has_selector?("option", text: cross_core_product_facility.name, visible: false)).to be(true)
         expect(page.has_selector?("option", text: product.name, visible: false)).to be(true)
+        expect(page.has_selector?("option", text: other_account.to_s, visible: false)).to be(false)
 
         select_from_chosen facility2.name, from: "add_to_order_form[facility_id]"
         select_from_chosen cross_core_product_facility2.name, from: "add_to_order_form[product_id]"
+        # This fails because the account is not valid for the facility
+        # select_from_chosen other_account.to_s, from: "Payment Source"
```
LeticiaErrandonea, line 296: For some reason `other_account` is not available for the facility, looks like a setup issue

LeticiaErrandonea: Fixed!

app/controllers/facility_accounts_controller.rb
```diff
@@ -158,4 +167,14 @@ def build_account
     ).build
   end
 
+  def set_facility_accounts_for_user
+    order_id = params[:order_id]
+    order = Order.find(order_id) if order_id
+
+    return unless order
```
giladshanan, line 174: What is the UX in this case?  should we return an empty collection instead?

app/views/facility_orders/_merge_order_form.html.haml
```diff
@@ -31,11 +32,19 @@
...
-          collection: @add_to_order_form.available_accounts,
-          include_blank: !@order.account.active?,
-          hint: @order.account.active? ? "" : t(".original_account_inactive", account: @order.account),
-          input_html: { class: "js--chosen" }
+        - if enable_cross_core_changes
+          = f.input :account_id,
+            collection: @add_to_order_form.available_accounts,
+            hint: @order.account.active? ? "" : t(".original_account_inactive", account: @order.account),
+            include_blank: !@order.account.active?,
+            input_html: { class: "js--edit-order__account js--chosen",
```
giladshanan, line 40: If this line is the only difference between the two cases, maybe we can set that value separately like:
```
input_html = enable_cross_core_changes ? { class: "js--edit-order__account js--chosen", data: { "include-blank" => !@order.account.active? } } : { class: "js--chosen" }
```

spec/system/admin/adding_to_an_order_spec.rb
```diff
@@ -275,6 +275,7 @@
 
     context "with admin role" do
       let(:user) { create(:user, :facility_administrator, facility:) }
+      let!(:other_account) { create(:account, :with_account_owner, type: "CreditCardAccount", owner: order.user, description: "Other Account", facility: facility2) }
```
giladshanan, line 278: how about:
```suggestion
      let!(:facility2_account) { create(:account, :with_account_owner, type: "CreditCardAccount", owner: order.user, description: "Other Account", facility: facility2) }
```

giladshanan (approved):
A few small suggestions, nothing blocking

app/controllers/facility_accounts_controller.rb
LeticiaErrandonea: UX was an ugly error in the console 😅 

Fixed now, thanks!

PR #4245, opened 2024-05-13
[Shared dev] [#162173] Add indicator for Cross-Core orders

LeticiaErrandonea:
# Release Notes
* See cross-core orders with an indicator in the Orders tab

# Screenshot
image

# Additional Context
## Pending work
- [ ] Confirm new column. Where should icon be placed?
- [ ] Confirm we want to have Ordered For and Cross Core select in same row
- [ ] Label for new select?
- [ ] Add specs
- [ ] Use I18n
- [ ] Check Accessibility
- [ ] Consider using a better option for `select`

# Accessibility


app/views/shared/transactions/_search.html.haml
```diff
@@ -2,7 +2,10 @@
   .row
     %fieldset.span6#search
       - @search.options.reject(&:multipart?).each do |searcher|
-        = f.input searcher.key, as: :transaction_chosen, collection: searcher.options, label: searcher.label, label_method: searcher.label_method, data_attrs: searcher.method(:data_attrs), input_html: { id: searcher.key }
+        - if searcher.key == "cross_cores"
+          = f.input searcher.key, collection: searcher.options, label: searcher.label, label_method: searcher.label_method, value_method: -> (option) { searcher.value_method(option) }, include_blank: false
```
LeticiaErrandonea, line 6: I went for an option that is the closest to what was already here, although I don't like this solution, so I'm up for a refactor (happy to hear opinions)

app/views/facility_orders/index.html.haml
```diff
@@ -34,6 +37,10 @@
               %td.centered= check_box_tag("order_detail_ids[]", od.id, false, {class: "toggle", id: nil })
               %td.centered= link_to od.order_id, facility_order_path(current_facility, od.order)
               %td.centered= link_to od.id, manage_order_detail_path(od), class: "manage-order-detail"
+              - if cross_core_order_view_enabled
+                %td
+                  - if od.order.cross_core_project.present? && od.order.cross_core_project.facility != current_facility
+                    = od.order.cross_core_project.facility
```
LeticiaErrandonea, line 43: Given clients said they don't want the Ordered By person I figured we may not need the icon and hover, so I started with this simpler version

app/controllers/concerns/new_inprocess_controller.rb
```diff
@@ -8,10 +8,18 @@ def index
...
-                                              TransactionSearch::DateRangeSearcher).search(order_details, @search_form)
+    searchers = [
+      TransactionSearch::ProductSearcher,
+      TransactionSearch::OrderedForSearcher,
+      TransactionSearch::OrderStatusSearcher,
+      TransactionSearch::DateRangeSearcher,
+    ]
+
+    if SettingsHelper.feature_on?(:cross_core_order_view)
+      searchers.insert(2, TransactionSearch::CrossCoreSearcher)
```
LeticiaErrandonea, line 19: View heavily relies on the searchers order, that's why I'm adding it in the third position

LeticiaErrandonea: Leave it at the end and make column smaller for this input

LeticiaErrandonea:
@giladshanan, it's missing specs but should be somewhat ready for your review

db/migrate/20240514183639_add_original_cross_core_order_to_orders.rb
```diff
@@ -0,0 +1,17 @@
+class AddOriginalCrossCoreOrderToOrders < ActiveRecord::Migration[7.0]
```
giladshanan, line 1: `# frozen_string_literal: true`

db/migrate/20240514183639_add_original_cross_core_order_to_orders.rb
```diff
@@ -0,0 +1,17 @@
...
+  def up
+    add_column :orders, :original_cross_core_order, :boolean, default: false, null: false
+
+    Projects::Project.find_each do |project|
+      next unless project.orders.any?
+
+      original_order = project.orders.where(facility_id: project.facility_id).first
+
+      original_order&.update!(original_cross_core_order: true)
```
giladshanan, line 10: I'm not sure we want to raise an exception here, maybe just log the project ID and error instead?

app/services/transaction_search/cross_core_searcher.rb
```diff
@@ -0,0 +1,39 @@
...
+    end
+
+    def search(params)
+      return order_details if params.blank?
+
+      params = params.first
+
+      # Based on params, filter by cross_core_project_id present in order_detail.order
+      if params == "no"
+        order_details.joins(:order).where(orders: { cross_core_project_id: nil })
```
giladshanan, line 17: I'm not 100% sure the desired logic for the queries here - can you ask the consortium about it?

app/views/facility_orders/index.html.haml
```diff
@@ -34,6 +37,14 @@
...
               %td.centered= link_to od.order_id, facility_order_path(current_facility, od.order)
               %td.centered= link_to od.id, manage_order_detail_path(od), class: "manage-order-detail"
+              - if cross_core_order_view_enabled
+                %td
+                  = od.order.created_by_user.full_name
+                  - if od.order.cross_core_project.present?
+                    - if od.order.cross_core_project.facility == current_facility
+                      %i.fa.fa-building{ data: { toggle: "tooltip" }, title: t("views.facility_orders.index.cross_core_icon_tooltip", facility: od.order.cross_core_project.facility.abbreviation)}
+                    - else
+                      %i.fa.fa-users{ data: { toggle: "tooltip" }, title: t("views.facility_orders.index.cross_core_icon_tooltip", facility: od.order.cross_core_project.facility.abbreviation)}
```
giladshanan, line 47: I might pull this into a partial so we can re-use it.  should we have a new column for the icons/facility abbreviations?

spec/system/admin/facility_orders_search_spec.rb
```diff
@@ -3,53 +3,191 @@
...
-      visit show_problems_facility_orders_path(facility)
+  context "cross-core orders", :js, feature_setting: { cross_core_order_view: true } do
+    # cross_core_order_originating_facility is a purchased order that originates in Facility (item)
+    # facility2_item is ordered and that new order is added to cross_core_project
+
+    # cross_core_order_originating_facility2 is a purchased order that originates in Facility2 (facility2_item)
+    # item is ordered and that new order is added to cross_core_project2
+
+    # cross_core_project has cross_core_order_originating_facility order and another one for Facility2
+    # cross_core_project2 has cross_core_order_originating_facility2 order and another one for Facility
```
giladshanan, line 67: This section was a little confusing to me - if there's useful context here for the future, I'd move it closer to the code it's describing.  The relations are clear from looking at the `let` statements, so code comments should tell us "What are we testing here?" or "why did we set up the test this way?"

giladshanan (approved):
Looks good to me, I left two small comments for you

app/views/facility_orders/index.html.haml
LeticiaErrandonea: Ticket said to use icons, but today Jeff said to use a new column. Maybe we can ask at the Consortium meeting?

spec/system/admin/facility_orders_search_spec.rb
LeticiaErrandonea: This started as notes for myself, as I was a bit confused while writing the spec. I updated the comments to only have the useful parts. Let me know what you think

app/views/facility_orders/index.html.haml
LeticiaErrandonea: Move icon to new column called Facility. Next to the icon, show the facility abbreviation (check whether it needs to be truncated to 6 chars). Keep icon on hover behavior

app/views/shared/_cross_core_facility_abbreviation_order_details.html.haml
```diff
@@ -0,0 +1,8 @@
+- if order_detail.order.cross_core_project.present?
+  - facility_abbreviation = order_detail.order.cross_core_project.facility.abbreviation
+  - facility_abbreviation = facility_abbreviation.length > 6 ? facility_abbreviation.truncate(7) : facility_abbreviation
```
giladshanan, line 3: Any reason not to just truncate everything?
```suggestion
  - facility_abbreviation = order_detail.order.cross_core_project.facility.abbreviation.truncate(7)
```

giladshanan (approved):
One small question, not blocking

app/views/shared/_cross_core_facility_abbreviation_order_details.html.haml
LeticiaErrandonea: `...` counts as part of the 6 chars. That means that if I have ABCDE, it displays "ABC..." instead of "ABCDE"

LeticiaErrandonea: You're right. I had tried a couple of options and saw something weird but now I couldn't reproduce it now, so I think I must have messed something up.

PR #4286, opened 2024-05-28
[Shared dev] [#162298] Allow Staff users to perform cross core actions

LeticiaErrandonea:
# Release Notes
* Allow Staff users to perform cross core actions

# Accessibility

### DIFF
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -166,12 +166,6 @@ class Ability
 
 
   def facility_administrator_abilities(user, resource, controller)
-    can :available_for_cross_core_ordering, Product if controller.is_a?(ProductsController)
-
-    if controller.is_a?(FacilityAccountsController) && user.facility_administrator_of_any_facility?
-      can [:accounts_available_for_order, :show], Account
-    end
-
     if resource.is_a?(PriceGroup) && user.facility_administrator_of?(resource.facility)
       if !resource.global?
         can :manage, [AccountPriceGroupMember, UserPriceGroupMember]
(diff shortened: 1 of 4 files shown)


app/lib/ability.rb
```diff
@@ -290,6 +257,42 @@ def facility_senior_staff_abilities(user, resource, controller)
 
 
   def facility_staff_abilities(user, resource, controller)
+    can :available_for_cross_core_ordering, Product if controller.is_a?(FacilityOrdersController) || controller.is_a?(ProductsController)
```
giladshanan, line 260: curious if this is necessary since we are skip the authorize resource call?

app/lib/ability.rb
```diff
@@ -290,6 +257,42 @@ def facility_senior_staff_abilities(user, resource, controller)
...
+
+    if controller.is_a?(FacilityAccountsController) && user.facility_staff_of_any_facility?
```
giladshanan, line 262: Use if/elsif for controller logic, consider moving cross core abilities into its own method

app/models/concerns/users/roles.rb
```diff
@@ -33,8 +33,12 @@ def operable_facilities
       end
     end
 
-    def facility_administrator_of_any_facility?
-      user_roles.find { |r| r.role == UserRole::FACILITY_ADMINISTRATOR }.present?
+    def facility_staff_of_any_facility?
+      user_roles.find { |r| [UserRole::FACILITY_ADMINISTRATOR, UserRole::FACILITY_SENIOR_STAFF, UserRole::FACILITY_STAFF].include?(r.role) }.present?
+    end
+
+    def facility_staff_or_above_of?(facility)
```
giladshanan, line 40:
```suggestion
    def facility_staff_or_manager_of?(facility)
```

app/models/concerns/users/roles.rb
```diff
@@ -33,8 +33,12 @@ def operable_facilities
       end
     end
 
-    def facility_administrator_of_any_facility?
-      user_roles.find { |r| r.role == UserRole::FACILITY_ADMINISTRATOR }.present?
+    def facility_staff_of_any_facility?
```
giladshanan, line 36:
```suggestion
    def facility_staff_or_manager_of_any_facility?
```

app/models/concerns/users/roles.rb
```diff
@@ -56,6 +60,11 @@ def manager_of?(facility)
       user_roles.manager?(facility) || administrator?
     end
 
+    def facility_staff_of?(facility)
+      return false if facility.blank?
```
giladshanan, line 64: can you add a newline after the guard clause?

giladshanan (commented):
nice work, left some minor comments and suggestions

app/lib/ability.rb
LeticiaErrandonea: Yes, it's still necessary. If I remove this, specs fail as no user is allowed to hit that endpoint

app/lib/ability.rb
```diff
@@ -290,6 +257,12 @@ def facility_senior_staff_abilities(user, resource, controller)
 
 
   def facility_staff_abilities(user, resource, controller)
+    cross_core_abilities(user, resource, controller)
+
+    if resource.is_a?(OrderDetail) && user.facility_staff_or_manager_of?(resource.facility)
+      can :manage, OrderDetail, order: { facility_id: resource.order.facility_id }
```
giladshanan, line 263: I'm not sure but this might be giving staff more abilities than they want/realize. Let's push this to stage and call it out in the testing notes

giladshanan (approved):
nice!

PR #4289, opened 2024-05-29
[Shared dev] [#162175] Show price group in orders table

LeticiaErrandonea:
# Release Notes
* Show price group (estimated or final) in orders table.

# Screenshot
## Orders tab
image

## Cross core orders tab
image

# Accessibility

### DIFF
--- a/app/views/facility_orders/index.html.haml
+++ b/app/views/facility_orders/index.html.haml
@@ -38,6 +38,7 @@
         %tbody
           - @order_details.each do |od|
             %tr
+              - order_detail = OrderDetailPresenter.new(od)
               %td.centered= check_box_tag("order_detail_ids[]", od.id, false, {class: "toggle", id: nil })
               %td.centered= link_to od.order_id, facility_order_path(current_facility, od.order)
               %td.centered= link_to od.id, manage_order_detail_path(od), class: "manage-order-detail"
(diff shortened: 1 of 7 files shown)


vendor/engines/projects/app/services/projects/cross_core_facility_searcher.rb
```diff
@@ -31,11 +32,16 @@ def label_method
...
     def label
-      I18n.t("projects.projects.cross_core_orders.filter_label")
+      text("projects.projects.cross_core_orders.filter_label")
     end
 
     def input_type
       :select
     end
+
+    def translation_scope
```
giladshanan, line 42: Can you add a comment explaining why this needs to be an empty string?

vendor/engines/projects/spec/system/cross_core_orders_spec.rb
```diff
@@ -66,7 +66,7 @@
 
   context "when selecting All" do
     before do
-      select "All", from: "Participating Facilities"
+      select "All", from: "Participating facilities"
```
giladshanan, line 69: I think this and line 94 are going to cause issues when merged into repos where another word is used.  I think you can use this helper method instead:
```suggestion
      select "All", from: "Participating #{facilities_route}"
```
... or just explicitly call `I18n.t("facilities_downcase")`

app/views/facility_orders/index.html.haml
```diff
@@ -51,6 +53,8 @@
               %td= od.order_status
               %td.currency
                 =  OrderDetailPresenter.new(od).wrapped_total
+              - if cross_core_order_view_enabled
+                %td= od.price_group&.name || od.estimated_price_group_name
```
giladshanan, line 57: how about putting this into a method int the `OrderDetailPresenter`?

giladshanan (changes requested):
Looks good, just a couple small suggestions.  Also I think the specs should reference `I18n.t("facilities_downcase")`

PR #4302, opened 2024-06-04
[#162306] Make Cross core orders purchased

LeticiaErrandonea:
# Release Notes
* Make Cross core orders purchased.

# Additional context
A "new" order in our codebase is a "cart". Once the user purchases it, the order changes to "purchased". 
Cross core orders were created in "new" state but never updated to purchased. For instruments (the majority of our tests) it was right, as it's only updated when the reservation is confirmed.
This was causing different problems, the main ones being: cross core order wasn't visible in the Orders tab for the receiving facility, and order's user cart had the last product that was intended to be added to the project.


app/forms/add_to_order_form.rb
```diff
@@ -163,20 +161,23 @@ def create_cross_core_project_and_add_order!
...
-        facility: product_facility,
-        account_id:,
-        user_id: original_order.user_id,
-        created_by: created_by.id,
-        cross_core_project: @order_project
-      )
+      product_order = find_or_create_product_order
+
+      # If the product that is going to be added requires a merge order, then it is created as part of the merge_order method.
+      # Otherwise, just use product_order.
```
giladshanan, line 167:
```suggestion
      # If the product that is going to be added requires a merge. order, 
      # then we create the merge order as part of the merge_order method.
      # We don't want to purchase the merge_order yet because we don't have 
      # all the detail needed from the user (reservation time, order form, etc)
```

app/forms/add_to_order_form.rb
```diff
@@ -163,20 +161,23 @@ def create_cross_core_project_and_add_order!
...
       original_order.update!(cross_core_project: @order_project)
 
-      product_order = Order.find_or_create_by!(
-        facility: product_facility,
-        account_id:,
-        user_id: original_order.user_id,
-        created_by: created_by.id,
-        cross_core_project: @order_project
-      )
+      product_order = find_or_create_product_order
```
giladshanan, line 164:
```suggestion
      # This is the order that will be purchased if we have the details to do so
      product_order = find_or_create_product_order
```

app/forms/add_to_order_form.rb
```diff
@@ -163,20 +161,23 @@ def create_cross_core_project_and_add_order!
...
 
-      product_order.add(product, quantity, params.merge(ordered_at: Time.zone.now)).each do |order_detail|
         order_detail.set_default_status!
         order_detail.change_status!(order_status) if order_status.present?
 
         order_detail.update!(project_id: @order_project.id)
       end
+
+      order_purchaser = OrderPurchaser.new(order: product_order, acting_as: original_order.user, order_in_past: nil, params: ActionController::Parameters.new(params), user: created_by)
+      order_purchaser.purchase_cross_core! if order_purchaser.purchase_for_cross_core?
```
giladshanan, line 180:
```suggestion
      order_purchaser.purchase_cross_core! unless requires_merge?
```

app/forms/add_to_order_form.rb
```diff
@@ -163,20 +161,23 @@ def create_cross_core_project_and_add_order!
...
-        account_id:,
-        user_id: original_order.user_id,
-        created_by: created_by.id,
-        cross_core_project: @order_project
-      )
+      product_order = find_or_create_product_order
+
+      # If the product that is going to be added requires a merge order, then it is created as part of the merge_order method.
+      # Otherwise, just use product_order.
+      target_order = requires_merge? ? merge_order : product_order
```
giladshanan, line 168: I think this should be if/else

app/forms/add_to_order_form.rb
```diff
@@ -230,15 +230,33 @@ def merge_order
                        created_by: created_by.id,
                        cross_core_project: @order_project,
                      )
-                   else
+                    else
```
giladshanan, line 233: nit pick - I think the indentation is one space off?

app/forms/add_to_order_form.rb
```diff
@@ -219,9 +220,8 @@ def order_status
   def merge_order
```
giladshanan, line 220: I think we can do our future selves a favor by adding some code comments here about what this method actually does/returns.  I would also support renaming to `target_order` or something else that is more revealing of intention.

app/forms/add_to_order_form.rb
```diff
@@ -230,15 +230,33 @@ def merge_order
...
                        cross_core_project: @order_project,
                      )
-                   else
+                    else
                      target_order
-                   end
+                    end
+  end
+
+  def find_or_create_product_order
```
giladshanan, line 238: Add a code comment to explain why are we finding first?
"Create the new order if the project doesn't have a cross core order for the selected facility yet"

app/forms/add_to_order_form.rb
```diff
@@ -230,15 +230,33 @@ def merge_order
...
+    Order.create!(
+      facility: product_facility,
+      account_id:,
+      user_id: original_order.user_id,
+      created_by: created_by.id,
+      cross_core_project: @order_project
+    )
+  end
+
+  def requires_merge?
```
giladshanan, line 250: Comment - intention is to determine if we should create a merge order or not based on the products

app/forms/add_to_order_form.rb
```diff
@@ -230,15 +230,33 @@ def merge_order
...
+    products = product.is_a?(Bundle) ? product.products : [product]
+
+    products.any?(&:requires_merge?)
   end
 
   def order_for_selected_facility
     return unless @order_project
 
-    @order_for_selected_facility ||= @order_project.orders.find { |o| o.facility_id == @facility_id }
+    @order_for_selected_facility ||= @order_project.orders.reload.find { |o| o.facility_id == @facility_id }
```
giladshanan, line 259: comment - why do we need to reload here?

app/services/order_purchaser.rb
```diff
@@ -42,6 +42,24 @@ def purchase!
     @errors = order.errors.full_messages
   end
 
+  def purchase_cross_core!
+    # Check if we need this
```
giladshanan, line 46: remove this and line 53 if we don't need it

app/services/order_purchaser.rb
```diff
@@ -50,6 +68,10 @@ def errors
     @errors.compact.uniq
   end
 
+  def purchase_for_cross_core?
```
giladshanan, line 71: don't think we need this

giladshanan (commented):
Getting close!

spec/system/admin/adding_to_a_cross_core_order_spec.rb
```diff
@@ -0,0 +1,271 @@
...
+          click_button "Create"
+
+          # After the reservation is created, the merge_order is cleared
+          expect(project.reload.orders.last.merge_with_order_id).to eq(nil)
+        end
+      end
+    end
+  end
+
+  describe "adding a product from another facility" do
```
giladshanan, line 166:
```suggestion
  describe "adding an item from another facility" do
```

giladshanan (approved):
🎉  one small suggestion

PR #4330, opened 2024-06-11
[Shared dev] [#162254] Redirect to correct order after "Cancel reservation" 

LeticiaErrandonea:
# Release Notes
* Redirect to correct order after "Cancel reservation" 

# Accessibility

### DIFF
--- a/app/controllers/order_details_controller.rb
+++ b/app/controllers/order_details_controller.rb
@@ -46,14 +46,7 @@ class OrderDetailsController < ApplicationController
         end
       end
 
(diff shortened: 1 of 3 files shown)


app/controllers/order_details_controller.rb
```diff
@@ -120,4 +121,15 @@ def order_detail_params
     params.require(:order_detail).permit(:account_id)
   end
 
+  def redirect_to_path
```
giladshanan, line 124: If we use this same logic in 3 or more places, I'd vote for putting this into a concern

giladshanan (approved):

PR #4368, opened 2024-06-18
[SHARED DEV] [#162198] Add filters to Projects tab

LeticiaErrandonea:
# Release Notes
* Add filters to Projects tab

# Screenshot
## When there are results
image

## When there are no results
image

# Additional Context
* In the next PR I'll add Project show with order details

# Accessibility

### DIFF
--- a/app/services/projects_search/cross_core_searcher.rb
+++ b/app/services/projects_search/cross_core_searcher.rb
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-module CrossCoreProjectsSearch
+module ProjectsSearch
 
   class CrossCoreSearcher < BaseSearcher
     def options
--- a/app/services/projects_search/cross_core_searcher.rb
+++ b/app/services/projects_search/cross_core_searcher.rb
@@ -30,16 +30,11 @@ module ProjectsSearch
     private
 
     def cross_core_projects
-      Projects::Project
-        .joins(:orders)
-        .where(orders: { facility_id: @current_facility_id })
+      Projects::Project.cross_core.for_facility(@current_facility_id)
     end
 
     def single_facility_projects
-      Projects::Project
-        .left_outer_joins(orders: [:facility, :cross_core_project])
-        .where(orders: { cross_core_project_id: nil })
-        .where(facility_id: @current_facility_id)
+      Projects::Project.for_single_facility(@current_facility_id)
     end
 
   end
--- a/app/services/projects_search/searcher.rb
+++ b/app/services/projects_search/searcher.rb
@@ -32,7 +32,7 @@ module ProjectsSearch
 
         search_params = params[searcher_class.key.to_sym]
 
-        # Options should not be restricted, they should search over the full order details
+        # Options should not be restricted, they should search over the full list of projects
         option_searcher = searcher_class.new(projects)
 
         Results.new(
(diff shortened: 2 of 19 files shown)


app/services/projects_search/cross_core_searcher.rb
```diff
@@ -0,0 +1,47 @@
...
+    def input_type
+      :select
+    end
+
+    private
+
+    def cross_core_projects
+      Projects::Project
+        .joins(:orders)
+        .where(orders: { facility_id: @current_facility_id })
```
giladshanan, line 35: I think we have a scope that does this same thing:
```suggestion
      Projects::Project.for_facility(@current_facility_id )
```
... but isn't this logic going to include single facility projects as well?
for OrderDetail we have this scope:
```
  scope :cross_core, lambda {
    joins(order: [:facility, :cross_core_project])
    .where.not(orders: { cross_core_project_id: nil })
  }
```
... can you add something like that to the Project model?
then we can do:
`Projects::Project.cross_core.for_facility(@current_facility_id )`

app/services/projects_search/cross_core_searcher.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    def single_facility_projects
+      Projects::Project
```
giladshanan, line 39:
```suggestion
      Projects::Project.single_facility.for_facility(@current_facility_id )
```
...
```
  scope :single_facility, lambda {
    joins(order: [:facility, :cross_core_project])
    .where(orders: { cross_core_project_id: nil })
  }
```

app/services/projects_search/searcher.rb
```diff
@@ -0,0 +1,73 @@
...
+      @searchers = Array(searchers)
+    end
+
+    def search(projects, params)
+      @searchers.reduce(Results.new(projects)) do |results, searcher_class|
+        searcher = searcher_class.new(results.projects, params[:current_facility_id])
+
+        search_params = params[searcher_class.key.to_sym]
+
+        # Options should not be restricted, they should search over the full order details
```
giladshanan, line 35:
```suggestion
        # Options should not be restricted, they should search over the full list of projects
```

vendor/engines/projects/app/controllers/projects/projects_controller.rb
```diff
@@ -83,6 +88,10 @@ def showing_inactive?
 
     private
 
+    def initial_projects
+      nil
```
giladshanan, line 92: Any reason to use `nil` here instead of an empty array?  curious why this needs a method of its own?  maybe the searcher should have this as an optional input?

giladshanan (changes requested):
Looks good, left a few comments/requests

app/services/projects_search/cross_core_searcher.rb
LeticiaErrandonea: Oh, you're right! I added a case for specs but those single facility projects don't have any orders attached.

vendor/engines/projects/app/controllers/projects/projects_controller.rb
LeticiaErrandonea: The searcher uses recursion to filter the previous list of projects, making the input optional would also mean switching the params. I can do that but I wonder if it's the best as it'd make this searcher different from the rest and could be a bit confusing.

It definitely doesn't need to be its own method, I'll fix that.

giladshanan (approved):
🎉

PR #4391, opened 2024-06-25
[Shared dev] [#162202] Edit project

LeticiaErrandonea:
# Release Notes
* Add ability to edit cross core projects

# Screenshot
## Order show
image

#### From another facility
image

#### From current facility
image

### Single facility
image

### DIFF
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -147,7 +147,7 @@ class FacilityOrdersController < ApplicationController
         @cross_core_orders_by_facility[order_facility] = order
       end
 
-      @cross_core_project_name = project.name
+      @cross_core_project = project
     end
   end
 
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -146,9 +146,9 @@ class FacilityOrdersController < ApplicationController
         @cross_core_order_details_by_facility[order_facility] = order.order_details.ordered_by_parents
         @cross_core_orders_by_facility[order_facility] = order
       end
-    end
 
-    @project_name = project&.name
+      @cross_core_project_name = project.name
+    end
   end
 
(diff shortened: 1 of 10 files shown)


app/controllers/facility_orders_controller.rb
```diff
@@ -147,6 +147,8 @@ def load_cross_core_order_details
         @cross_core_orders_by_facility[order_facility] = order
       end
     end
+
+    @project_name = project&.name
```
giladshanan, line 151: Is the `&` here to handle single-facility projects, or is there another case?
If it's just that, I would vote for something a little more verbose but intention revealing:
```
  if project.present?
    project_orders = project.orders.where.not(facility_id: @order.facility_id)
    project_orders.each do |order|
      order_facility = order.facility.to_s
      @cross_core_order_details_by_facility[order_facility] = order.order_details.ordered_by_parents
      @cross_core_orders_by_facility[order_facility] = order
    end
    @project_name = project.name
  else
    @project_name = nil
  end
```
also, maybe this should be `@cross_core_project`?

giladshanan (approved):
One non-blocking comment

app/views/facility_orders/show.html.haml
```diff
@@ -10,11 +10,11 @@
     = banner_label @order, :created_by_user
     - if SettingsHelper.feature_on?(:cross_core_projects)
       - project_id = @cross_core_order_details_by_facility.present? ? @order_details.first&.project_id : nil
-      - if project_id
+      - if @cross_core_project_name
```
giladshanan, line 13: based on spec failures I think maybe we need:
```suggestion
      - if @cross_core_project_name && project_id
```
... but not sure if there is a valid case where we have a cross core project with no order details

giladshanan (changes requested):
Looks like specs are unhappy

giladshanan (approved):
Small suggestion

PR #4405, opened 2024-07-02
Fix: [#162313] Purchase cross core order when it only has one instrument

LeticiaErrandonea:
# Release Notes
* Purchase cross core order when it only has one instrument

### DIFF
--- a/app/services/reservation_creator.rb
+++ b/app/services/reservation_creator.rb
@@ -36,6 +36,9 @@ class ReservationCreator
           # The purchase_order_path or cart_path will handle the backdating, but we need
           # to do this here for merged reservations.
           backdate_reservation_if_necessary(session_user)
+
+          purchase_original_cross_core_order
+
           @success = :merged_order
         elsif @order.order_details.one?
           @success = :instrument_only_order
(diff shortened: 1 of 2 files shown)


app/services/reservation_creator.rb
```diff
@@ -104,4 +107,19 @@ def status_q
     ActiveSupport::StringInquirer.new(@success.to_s)
   end
 
+  def purchase_original_cross_core_order
+    original_order = Order.find(@order.merge_with_order_id)
```
giladshanan, line 111: Do we need to handle the case where a record is not found?

app/services/reservation_creator.rb
```diff
@@ -104,4 +107,19 @@ def status_q
...
+
+    if original_order.order_details.one? && original_order.cross_core_project.present?
+      order_purchaser = OrderPurchaser.new(
+        order: original_order,
+        acting_as: original_order.user,
+        order_in_past: nil,
+        params: ActionController::Parameters.new({}),
+        user: original_order.created_by
+      )
+      order_purchaser.purchase_cross_core!
```
giladshanan, line 121: What happens if this line fails?

giladshanan (approved):
Left a couple non-blocking questions for you

app/services/reservation_creator.rb
LeticiaErrandonea: No. We only get to that if-block when the order has `merge_with_order_id`

app/services/reservation_creator.rb
LeticiaErrandonea: image

PR #4431, opened 2024-07-12
Tech task: [#162316] Cross Core refactor

LeticiaErrandonea:
# Release Notes
* Tech task: Cross Core refactor

# Additional Info
* This PR only moves almost all the extensions to the app.
* Views are still in the engine.
* Ability extension in still in the engine.


LeticiaErrandonea (commented):
Left a couple of comments, but this looks great!

Thanks for finishing it @giladshanan!

app/models/order_detail.rb
```diff
@@ -66,13 +67,13 @@ def set_problem_order
   has_many   :notifications, as: :subject, dependent: :destroy
   has_many   :stored_files, dependent: :destroy
   has_many   :sample_results_files, -> { sample_result }, class_name: "StoredFile"
+  has_many   :vestal_versions, as: :versioned
```
giladshanan, line 70: @LeticiaErrandonea why did this line move?

app/services/order_row_importer.rb
```diff
@@ -295,7 +296,17 @@ def validate_existing_order
     end
   end
 
-  # Projects adds a custom attribute and overrides this method to validate it
   def validate_custom_attributes
+    if field(:project_name).present? && project.nil?
```
giladshanan, line 300: I think this should be in its own method like the others, maybe `validate_project_name`

app/services/order_row_importer.rb
```diff
@@ -295,7 +296,17 @@ def validate_existing_order
...
+      add_error(:project_not_found)
+    end
+  end
+
+  def project
+    facility.projects.active.find_by name: field(:project_name)
+  end
+
+  def custom_attributes
+    { project_id: project&.id }
```
giladshanan, line 310: these aren't really custom attributes any more, right?  and we already define `custom_attributes` on line 162.  How about adding `project_id` to `to_add_attributes` instead?

Also, I think `custom_attributes` and `validate_custom_attributes` are probably un-used now, so we should add code comments to the methods

app/views/shared/_order_detail_action_form.html.haml
```diff
@@ -1,6 +1,8 @@
 - @facility_users ||= User.find_users_by_facility(current_facility) || []
 
 %td.action-form
+  = render "projects/shared/select_facility_project", facility: current_facility
+  - # TODO: Check if un-used and remove.  This was used by the old projects engine before it was brought into the application.
```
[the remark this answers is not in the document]
giladshanan: `"projects/shared/select_facility_project"` is definitely still used, but since we are rendering it explicitly now I'm not sure if anyone else is using the hook for `"batch_update_above_product_column"` below on line 6

spec/services/order_row_importer_spec.rb
```diff
@@ -491,6 +498,10 @@
             end
 
             context "and the order is invalid" do
+              before(:each) do
```
giladshanan, line 501: why did this change?

giladshanan (commented):
Geting close!

giladshanan (approved):
🎉

PR #4432, opened 2024-07-12
[#162199] Fix bug when navigating to My Transactions

LeticiaErrandonea:
# Release Notes
* Fix bug when navigating to My Transactions

### DIFF
--- a/app/views/shared/transactions/_cross_core_icon.html.haml
+++ b/app/views/shared/transactions/_cross_core_icon.html.haml
@@ -1,3 +1,3 @@
-- if order_detail.cross_core_activity_for_facility?(current_facility)
+- if order_detail.cross_core_activity_for_facility?(current_facility || order_detail.facility)
   %br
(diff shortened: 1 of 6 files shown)


app/views/shared/transactions/_cross_core_icon.html.haml
```diff
@@ -1,3 +1,3 @@
-- if order_detail.cross_core_activity_for_facility?(current_facility)
+- if order_detail.cross_core_activity_for_facility?(current_facility || order_detail.facility)
```
giladshanan, line 1: I'm wondering if this method should be changed a bit.
If we don't need to check `current_facility`, then we can just reference the order_detail's facility internally:
```
# app/models/order_detail.rb
def cross_core_activity_for_facility?
  order.cross_core_project.present? && order.cross_core_project.facility_id != facility.id
end
```
What do you think?

giladshanan (approved):
I left you a small refactoring suggestion -otherwise looks good

giladshanan (approved):

PR #4439, opened 2024-07-17
Fix: [#162350] Clear cross core project when clearing cart

LeticiaErrandonea:
# Release Notes
* Clear cross core project when clearing cart

### DIFF
--- a/app/models/order.rb
+++ b/app/models/order.rb
@@ -130,6 +130,7 @@ class Order < ApplicationRecord
     order_details.destroy_all
     self.facility = nil
     self.account = nil
+    self.cross_core_project_id = nil
     save
   end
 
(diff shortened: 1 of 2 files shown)


app/models/order.rb
```diff
@@ -130,6 +130,7 @@ def clear_cart?
     order_details.destroy_all
```
giladshanan, line 130: This method should probably not end with `?` - maybe `clear_cart` instead?

giladshanan (approved):

app/models/order.rb
LeticiaErrandonea: I think it may be related to it being used as a guard. Not sure, but I see the rest also follow that pattern

giladshanan: 👍🏼 good to merge then

PR #4452, opened 2024-07-24
[#162028] Separate Facilities List in Active/Inactive

LeticiaErrandonea:
# Release Notes
* Separate Facilities List in Active/Inactive

# Screenshot
image


app/controllers/facilities_controller.rb
```diff
@@ -68,9 +68,13 @@ def list
     @active_tab = "manage_facilites"
     if session_user.administrator?
       @facilities = Facility.alphabetized
```
giladshanan, line 70: It looks like we're not using `@facilities` in the view any more, is that right?
If so let's change it to a local variable `facilities`

giladshanan (approved):
One small comment, looks good!

PR #4465, opened 2024-07-31
[#162286] Instrument reservations updates

LeticiaErrandonea:
# Release Notes
* Move calendar above admin holds.
* Make admin holds end date configurable.

# Screenshot
image

image

# Additional Context
- [ ] Set to 16 or 17 weeks (~120 days) for OSU.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -16,6 +16,9 @@ accounts:
   product_default: 75340
   revenue_account_default: 50617
 
+admin_holds:
+  repeat_end_date_in_weeks: <%= ENV.fetch("ADMIN_HOLD_REPEAT_END_DATE_IN_WEEKS", 12) %>
+
 financial:
   # in the format MM-DD
   fiscal_year_begins: 09-01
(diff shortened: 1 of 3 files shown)


config/settings.yml
```diff
@@ -16,6 +16,9 @@ accounts:
   product_default: 75340
   revenue_account_default: 50617
 
+admin_holds:
+  repeat_end_date_in_weeks: <%= ENV.fetch("ADMIN_HOLD_REPEAT_END_DATE_IN_WEEKS", 12) %>
```
giladshanan, line 20: What do you think about documenting this and the other options managed via settings.yml?
We have `HOWTO_feature_flags.md` that describes some of it, but I think we should aim to document more of the settings either in that file, a new file, or perhaps in notion.  What do you think?

giladshanan (approved):
Looks good, just a question for you about improving documentation

config/settings.yml
LeticiaErrandonea: After talking about this, decided to consider documenting in the future

PR #4482, opened 2024-08-07
[SHARED DEV] [#162345] Allow facility staff to upload file to order for cross core

LeticiaErrandonea:
# Release Notes
* Fix file upload for cross core orders

# Additional Context
Tried to add specs [here](https://github.com/tablexi/nucore-open/blob/master/spec/system/admin/adding_to_a_cross_core_order_spec.rb#L20) but I run into the same problem as with [this one](https://github.com/tablexi/nucore-open/blob/master/spec/system/admin/adding_to_a_cross_core_order_spec.rb#L57) we ended up skipping.

### DIFF
--- a/app/assets/javascripts/app/order_table_refresh.js
+++ b/app/assets/javascripts/app/order_table_refresh.js
@@ -6,15 +6,17 @@
 document.addEventListener("DOMContentLoaded", function() {
 
   function fetchAndRefresh() {
-    const facility = this.getElementsByClassName('js--orderTableRefreshData')[0].dataset.updatedFacility
+    const updatedFacilityId = this.getElementsByClassName(
+      "js--orderTableRefreshData"
+    )[0].dataset.updatedFacility;
 
     const table = document.querySelector(
-      `.js--orderTableRefresh#facility_id_${facility}`
+      `.js--orderTableRefresh#facility_id_${updatedFacilityId}`
     );
     const url = new URL(document.location);
 
     url.searchParams.set("refresh", "true");
-    url.searchParams.set("updated_facility_id", facility);
+    url.searchParams.set("updated_facility_id", updatedFacilityId);
     const headers = { Accept: "text/html" };
 
     fetch(url, { headers: headers }).then(function (response) {
--- a/app/assets/javascripts/app/order_table_refresh.js
+++ b/app/assets/javascripts/app/order_table_refresh.js
@@ -6,20 +6,23 @@
 document.addEventListener("DOMContentLoaded", function() {
 
   function fetchAndRefresh() {
-    const table = document.getElementsByClassName("js--orderTableRefresh")[0];
+    const facility = this.getElementsByClassName('js--orderTableRefreshData')[0].dataset.updatedFacility
+
+    const table = document.querySelector(
+      `.js--orderTableRefresh#facility_id_${facility}`
+    );
     const url = new URL(document.location);
 
     url.searchParams.set("refresh", "true");
+    url.searchParams.set("updated_facility_id", facility);
     const headers = { Accept: "text/html" };
 
-    fetch(url, { headers: headers })
-    .then(function (response) {
+    fetch(url, { headers: headers }).then(function (response) {
       if (response.ok) {
-        response.text()
-        .then(function(body) {
+        response.text().then(function (body) {
           table.innerHTML = body;
-          setResultsFileUploadModals()
-         });
+          setResultsFileUploadModals();
+        });
       }
     });
   }
--- a/app/controllers/facility_orders_controller.rb
+++ b/app/controllers/facility_orders_controller.rb
@@ -40,6 +40,8 @@ class FacilityOrdersController < ApplicationController
         cross_core_order_details_for_facility = @cross_core_data_by_facility_id[updated_facility_id][:order_details]
         render partial: "order_table", locals: { order_details: cross_core_order_details_for_facility, cross_core: true, facility_id: updated_facility_id }
       end
+    else
+      @gross_order_value = @order.total + @cross_core_data_by_facility_id.values.map { |hash| hash[:order] }.sum(&:total)
     end
   end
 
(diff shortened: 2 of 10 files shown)


annaswims:
I don't understand the  data model enough to know what the fix is but It looks to me like the [skipped test you linked to]([this one](https://github.com/tablexi/nucore-open/blob/master/spec/system/admin/adding_to_a_cross_core_order_spec.rb#L57)) might be part of a bug. 

It redirect to a URL like `http://172.18.0.5:4000/facilities/facility5/orders/2612` ... but order is not associated with the `facility5`.  

As I looked more, I noticed that the fixture named `facility2` has  url_name: "facility5" and the fixture named `facility` has  url_name: "facility2".

app/controllers/order_management/order_details_controller.rb
```diff
@@ -111,4 +113,9 @@ def edit_disabled?
     @order_detail.in_open_journal? || @order_detail.reconciled?
   end
 
+  def init_order_detail
```
giladshanan, line 116: @LeticiaErrandonea why is this change needed?

giladshanan (commented):
@LeticiaErrandonea is the issue with the specs, or behavior of the application?
I think there is value in specs that document the behavior, even if the behavior is not ideal - like redirecting to a 404 page.

app/controllers/order_management/order_details_controller.rb
LeticiaErrandonea: @giladshanan because the order and order detail may not be from the `current_facility` but from some other facility (for cross core orders)

app/controllers/order_management/order_details_controller.rb
```diff
@@ -7,8 +7,10 @@ class OrderManagement::OrderDetailsController < ApplicationController
   before_action :authenticate_user!
 
   load_resource :facility, find_by: :url_name
-  load_resource :order, through: :facility
-  load_resource :order_detail, through: :order
+
+  load_resource :order, through: :facility, except: [:files]
+  load_resource :order_detail, through: :order, except: [:files]
+  before_action :init_order_detail, only: [:files]
```
giladshanan, line 13:
```suggestion
  # We can't load through the facility because of cross-core orders
  before_action :init_order_detail, only: [:files]
```

app/assets/javascripts/app/order_table_refresh.js
```diff
@@ -6,20 +6,23 @@
 document.addEventListener("DOMContentLoaded", function() {
 
   function fetchAndRefresh() {
-    const table = document.getElementsByClassName("js--orderTableRefresh")[0];
+    const facility = this.getElementsByClassName('js--orderTableRefreshData')[0].dataset.updatedFacility
```
giladshanan, line 9: How about `const updatedFacilityId`?
Let's use " for consistency here.

app/controllers/facility_orders_controller.rb
```diff
@@ -32,8 +32,13 @@ def initialize
   def show
     @order_details = @order.order_details.ordered_by_parents
     @order_details = @order_details.includes(:reservation, :order_status, :product, :order)
+    updated_facility_id = params[:updated_facility_id]
     if params[:refresh]
-      render partial: "order_table", locals: { order_details: @order_details, cross_core: false }
+      if updated_facility_id == current_facility.id.to_s
+        render partial: "order_table", locals: { order_details: @order_details, cross_core: false, facility: current_facility }
+      else
+        render partial: "order_table", locals: { order_details: @cross_core_data_by_facility_id[updated_facility_id][:order_details], cross_core: true, facility: updated_facility_id }
```
giladshanan, line 40: I think this would be more readable if you put the order detail into a variable:
```
cross_core_order_details = @cross_core_data_by_facility_id[updated_facility_id][:order_details]
```

app/lib/ability.rb
```diff
@@ -446,6 +446,16 @@ def cross_core_abilities(user, resource, controller)
         can [:add_accessories, :new, :show, :update, :cancel], OrderDetail if user.facility_staff_or_manager_of?(project.facility)
       end
     end
+
+    if controller.is_a?(FileUploadsController) && resource.is_a?(Facility) && SettingsHelper.feature_on?(:cross_core_projects)
+      can [:upload_sample_results, :destroy], StoredFile do |fileupload|
+        project = fileupload.order_detail.project
+
+        project&.cross_core? &&
+          (user.facility_staff_or_manager_of?(project.facility) || user.facility_staff_or_manager_of?(order_detail.order.facility)) &&
```
giladshanan, line 455: Can you break this into 2 lines?

app/views/facility_orders/show.html.haml
```diff
@@ -16,7 +16,7 @@
             = link_to @cross_core_project.name, facility_project_path(current_facility, @cross_core_project.id)
         %dl.span2
           %dt= t("views.facility_orders.show.cross_core_project_total")
-          - gross_order_value = @order.total + @cross_core_orders_by_facility.values.sum(&:total)
+          - gross_order_value = @order.total + @cross_core_data_by_facility_id.values.map { |hash| hash[:order] }.sum(&:total)
```
giladshanan, line 19: Could this happen in the controller instead?

app/views/facility_orders/show.html.haml
```diff
@@ -25,20 +25,20 @@
 
 %table.order-list.table.table-striped.table-hover#order-management
   = render "order_table_headers", cross_core: false
-  = render "order_table", order_details: @order_details, cross_core: false
+  = render "order_table", order_details: @order_details, cross_core: false, facility: current_facility.id
```
giladshanan, line 28: small nitpick, but I vote for `facility_id: current_facility.id`

giladshanan (approved):
Looks good, just some small formatting changes.
Are you going to change the default wait time in a separate PR?

giladshanan (approved):

PR #4488, opened 2024-08-09
[#162375] Allow adding multiple services to order

LeticiaErrandonea:
# Release Notes
* Allow adding multiple services to order without requiring multiple forms

# Screenshot
image

### DIFF
--- a/app/support/orders/item_adder.rb
+++ b/app/support/orders/item_adder.rb
@@ -21,14 +21,12 @@ class Orders::ItemAdder
     ods = case product
           when Bundle
             add_bundles(product, quantity, attributes)
-          when Service
-            add_services(product, quantity, attributes)
           when TimedService
             add_timed_services(product, quantity, duration, attributes)
           when Instrument
             add_instruments(product, quantity, attributes)
           else
-            [create_order_detail({ product: product, quantity: quantity }.merge(attributes))]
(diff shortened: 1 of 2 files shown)


giladshanan (commented):
I think we also need to change [this line.](https://github.com/tablexi/nucore-open/pull/4488/files#diff-a234c8d3e9e6fe620cf119b065dd65852174d88e4bbf72cd81907d274dfe748fR38)

It looks like this will also change the behavior of multi-adding services with a new order - is that right?  I haven't tested this, but I thought the goal was to align the behavior of creating a new order with adding to an existing order.

giladshanan (changes requested):
@LeticiaErrandonea I looked into this a bit more - the current behavior when ordering multiple services from the "multi-add" [view](https://nucore-open.herokuapp.com/facilities/demo) is to add X service order details with quantity of 1, where X is the value entered by the user.  Since this PR changes that behavior, we will also need to change [this line.](https://github.com/tablexi/nucore-open/pull/4488/files#diff-a234c8d3e9e6fe620cf119b065dd65852174d88e4bbf72cd81907d274dfe748fR38):

Screenshot 2024-08-12 at 5 32 21 PM

We should also make sure to call out clear testing notes for this in the release notes.

giladshanan (approved):

PR #4506, opened 2024-08-16
[#162345] Fix bug when form is required for cross core order

LeticiaErrandonea:
# Release Notes
* Fix bug when form is required for cross core order


app/controllers/order_detail_stored_files_controller.rb
```diff
@@ -81,4 +83,18 @@ def ability_resource
     @order_detail
   end
 
+  def redirect_to_facility_order(consider_merge_order: false)
```
giladshanan, line 86: If you can add some code comments here about why this logic needs to be this way that would be helpful for future developers.

giladshanan (approved):
LGTM!

PR #4528, opened 2024-08-28
[Tech task] [#162384] Make cart link be loaded with AJAX

LeticiaErrandonea:
# Release Notes
* Tech task: Improve page load times

### DIFF
--- a/app/assets/javascripts/app/load_cart_count.js
+++ b/app/assets/javascripts/app/load_cart_count.js
@@ -9,6 +9,9 @@ window.Cart = class Cart {
     if (!element) { return; }
 
     let url = element.data("url");
+
+    if (!url) { return; }
+
     url = new URL(url);
 
     fetch(url, { mode: "no-cors" }).then(function (response) {
--- a/app/assets/javascripts/app/load_cart_count.js
+++ b/app/assets/javascripts/app/load_cart_count.js
@@ -1,0 +1,36 @@
+window.Cart = class Cart {
+  constructor($cart) {
+    this.$cart = $cart;
+  }
+
+  loadCartCount() {
+    const element = $(this.$cart);
+
+    if (!element) { return; }
+
+    const url = element.data("url");
+    const link = element.data("link");
+
+    return $.ajax({
+      type: "get",
+      url,
+      dataType: "text",
+      success(data) {
+        data = JSON.parse(data);
+        const count = data.data.count;
+        const text = `Cart (${count})`;
+
+        const anchorElement = document.createElement("a");
+        anchorElement.setAttribute("href", link);
+        anchorElement.innerHTML = text;
+
+        element.append(anchorElement);
+      }
+    })
+  }
+}
+
+$(function () {
+  const cart = new Cart(".js--cart_count");
+  return cart.loadCartCount();
+});
--- a/app/assets/javascripts/app/load_cart_count.js
+++ b/app/assets/javascripts/app/load_cart_count.js
@@ -8,25 +8,21 @@ window.Cart = class Cart {
 
     if (!element) { return; }
 
-    const url = element.data("url");
-    const link = element.data("link");
+    let url = element.data("url");
+    url = new URL(url);
 
-    return $.ajax({
-      type: "get",
-      url,
-      dataType: "text",
-      success(data) {
-        data = JSON.parse(data);
-        const count = data.data.count;
-        const text = `Cart (${count})`;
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/app/load_cart_count.js
```diff
@@ -0,0 +1,36 @@
...
+
+  loadCartCount() {
+    const element = $(this.$cart);
+
+    if (!element) { return; }
+
+    const url = element.data("url");
+    const link = element.data("link");
+
+    return $.ajax({
```
giladshanan, line 14: let's use `fetch` here, we're trying to move away from jQuery

app/assets/javascripts/app/load_cart_count.js
```diff
@@ -0,0 +1,36 @@
...
+      type: "get",
+      url,
+      dataType: "text",
+      success(data) {
+        data = JSON.parse(data);
+        const count = data.data.count;
+        const text = `Cart (${count})`;
+
+        const anchorElement = document.createElement("a");
```
giladshanan, line 23: instead, let's update the existing `anchorElement.innerHTML`

app/views/shared/_header.html.haml
```diff
@@ -25,7 +25,7 @@
               %li.navbar-text= "#{acting_user.full_name} (#{acting_user.username})"
               %li.divider-vertical
               = render "/shared/support"
-              %li= link_to "#{t('pages.cart')} (#{current_cart.order_details.count})", :cart
+              %li.js--cart_count{ data: { url: orders_cart_count_path, link: cart_url } }
```
giladshanan, line 28: I'd vote for rendering the page with link that just says "Cart" without the count, then add the count later when we get the response via JS

giladshanan (changes requested):

app/assets/javascripts/app/load_cart_count.js
```diff
@@ -0,0 +1,35 @@
...
+
+    if (!element) { return; }
+
+    let url = element.data("url");
+
+    if (!url) { return; }
+
+    url = new URL(url);
+
+    fetch(url, { mode: "no-cors" }).then(function (response) {
```
giladshanan, line 17: Do we need `{ mode: "no-cors" }` here?

app/assets/javascripts/app/load_cart_count.js
```diff
@@ -0,0 +1,35 @@
...
+      if (response.ok) {
```
giladshanan, line 18: should we console.log something on failure?

app/assets/javascripts/app/load_cart_count.js
```diff
@@ -0,0 +1,35 @@
...
+
+          const anchorElement = element.find("a");
+          anchorElement.text(text);
+        });
+      }
+    });
+  }
+};
+
+$(function () {
```
giladshanan, line 32: This is fine with me, but I might consider something simpler to start.
I don't feel strongly about this but I would just define a function like `updateCartCount` that takes a string for the class selector - or just hard code `".js--cart_count")` to start.
Then when the page is done loading, run `updateCartCount`.  

What you have here works and there's nothing wrong with using a constructor, totally up to you.

app/controllers/orders_controller.rb
```diff
@@ -5,11 +5,13 @@ class OrdersController < ApplicationController
...
   before_action :authenticate_user!
-  before_action :check_acting_as,          except: [:cart, :add, :choose_account, :show, :remove, :purchase, :update_or_purchase, :receipt, :update]
-  before_action :init_order,               except: [:cart, :index, :receipt]
-  before_action :protect_purchased_orders, except: [:cart, :receipt, :confirmed, :index]
+  before_action :check_acting_as,          except: [:cart, :cart_order_details_count, :add, :choose_account, :show, :remove, :purchase, :update_or_purchase, :receipt, :update]
+  before_action :init_order,               except: [:cart, :cart_order_details_count, :index, :receipt]
+  before_action :protect_purchased_orders, except: [:cart, :cart_order_details_count, :receipt, :confirmed, :index]
   before_action :load_statuses, only: [:show, :update, :purchase, :update_or_purchase]
 
+  skip_after_action :verify_same_origin_request, only: [:cart_order_details_count]
```
giladshanan, line 13: This might be worth a code comment like "Using fetch instead of jQuery/XHR, and the only thing this action does it update the cart count"

giladshanan (approved):
Left you some comments - nothing blocking though.
Nice work!

PR #4585, opened 2024-09-19
[#162311] New reconciliation status for "abandoned" orders

LeticiaErrandonea:
# Release Notes
NOTE: run the rake task on heroku after deploying

* Admins can now mark as "Unrecoverable" statements that are unreconciled.

# Screenshot
## Reconcile
image

## Mark as "Unrecoverable"
image

## Statement history
image

# Additional Context
## Pending work
- [ ] Specs

# Accessibility

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -497,8 +497,8 @@ class OrderDetail < ApplicationRecord
   end
 
   def can_be_marked_as_unrecoverable?
-    # Can only be marked as unrecoverable if it has already been invoiced
-    state.to_sym == :complete && statement.present?
+    # Can only be marked as unrecoverable if it has already been completed
+    state.to_sym == :complete
   end
 
   def pending?
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -368,13 +368,14 @@ class OrderDetail < ApplicationRecord
 
   include AASM
 
-  CANCELABLE_STATES = [:new, :inprocess, :complete].freeze
+  CANCELABLE_STATES = [:new, :inprocess, :complete, :unrecoverable].freeze
   aasm column: :state do
     state :new, initial: true
     state :inprocess
     state :complete, enter: :make_complete
     state :reconciled, enter: :set_reconciled_at
     state :canceled, enter: :clear_costs
+    state :unrecoverable
 
     event :to_new do
       transitions to: :new, from: :inprocess
@@ -390,6 +391,10 @@ class OrderDetail < ApplicationRecord
 
     event :to_reconciled do
       transitions to: :reconciled, from: :complete, guard: :actual_total
+    end
+
+    event :to_unrecoverable do
+      transitions to: :unrecoverable, from: :complete, guard: :can_be_marked_as_unrecoverable?
     end
 
     event :to_canceled do
@@ -489,6 +494,11 @@ class OrderDetail < ApplicationRecord
   def cancelable?
     # can't cancel if the reservation isn't already canceled or if this OD has been added to a journal
     state_is_cancelable? && journal.nil? && !has_uncanceled_reservation?
+  end
+
+  def can_be_marked_as_unrecoverable?
+    # Can only be marked as unrecoverable if it has already been invoiced
+    state.to_sym == :complete && statement.present?
   end
 
   def pending?
@@ -752,7 +762,9 @@ class OrderDetail < ApplicationRecord
         change_status! order_status
       end
     else
+      return false if state.to_sym == :unreconciled
       return false unless reservation&.can_cancel?
+
       # must set canceled_at after calling #can_cancel?
       return false unless update(canceled_at: Time.current)
       clear_statement if cancellation_fee == 0
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -762,6 +762,7 @@ class OrderDetail < ApplicationRecord
         change_status! order_status
       end
     else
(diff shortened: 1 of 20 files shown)


app/services/order_details/reconciler.rb
```diff
@@ -12,8 +12,17 @@ class Reconciler
...
-    def initialize(order_detail_scope, params, reconciled_at, bulk_reconcile_note = nil, bulk_deposit_number = nil, bulk_reconcile_checkbox = nil)
+    def initialize(
+      order_detail_scope,
+      params,
+      reconciled_at,
+      bulk_reconcile_note = nil,
+      bulk_deposit_number = nil,
+      bulk_reconcile_checkbox = nil,
+      order_status = "reconciled"
+    )
```
giladshanan, line 23: I like this - it's much easier to read this way.  I think it should be indented to the right a bit though?

app/services/order_details/reconciler.rb
```diff
@@ -44,12 +53,20 @@ def to_be_reconciled
...
-    def reconcile(order_detail, params)
-      order_detail.reconciled_at = @reconciled_at
-      order_detail.assign_attributes(allowed(params))
-      order_detail.reconciled_note = @bulk_reconcile_note if @bulk_reconcile_note.present?
-      order_detail.deposit_number = @bulk_deposit_number if @bulk_deposit_number.present?
-      order_detail.change_status!(OrderStatus.reconciled)
+    def update_status(order_detail, params)
+      if @order_status == "reconciled"
+        order_detail.reconciled_at = @reconciled_at
+        order_detail.assign_attributes(allowed(params))
```
giladshanan, line 59: nit pick - I think this line runs in both branches, so probably you could do this first before checking `@order_status`

app/views/facility_accounts_reconciliation/_action_row.html.haml
```diff
@@ -1,27 +1,30 @@
 .row.table-actions.form-horizontal
   .span2.select_all_none= select_all_link
+  - if show_order_status
+    = select_tag :order_status, options_for_select([['Reconciled', 'reconciled'], ['Unrecoverable', 'unrecoverable']]), class: "js--orderStatusSelect"
```
giladshanan, line 4: Double quotes please

giladshanan (approved):
A few tiny nitpicks, looks good

app/models/order_detail.rb
```diff
@@ -752,7 +762,9 @@ def cancel_reservation(canceled_by, canceled_reason: nil, order_status: OrderSta
         change_status! order_status
       end
     else
+      return false if state.to_sym == :unreconciled
```
giladshanan, line 765: Is this intended to be `:unreconciled`? 
I wonder if this is part of what is making specs fail
```suggestion
      return false if state.to_sym == :unrecoverable
```

app/views/facility_accounts_reconciliation/_action_row.html.haml
```diff
@@ -1,27 +1,30 @@
 .row.table-actions.form-horizontal
   .span2.select_all_none= select_all_link
+  - if show_order_status
```
giladshanan, line 3: I wonder if we use this "action_row" partial in other places, like Move up, that could be causing spec failures.  
```suggestion
  - if local_assigns[:show_order_status]
```

giladshanan (commented):
@LeticiaErrandonea looks like some specs are failing.

app/models/order_detail.rb
```diff
@@ -752,7 +762,9 @@ def cancel_reservation(canceled_by, canceled_reason: nil, order_status: OrderSta
         change_status! order_status
       end
     else
+      return false if state.to_sym == :unrecoverable
```
giladshanan, line 765: Why is this here?

lib/tasks/demo_seed.rake
```diff
@@ -15,6 +15,7 @@ namespace :demo do
     canceled   = OrderStatus.find_or_create_by!(name: "Canceled")
     complete   = OrderStatus.find_or_create_by!(name: "Complete")
     reconciled = OrderStatus.find_or_create_by!(name: "Reconciled")
+    abandoned  = OrderStatus.find_or_create_by!(name: "Unrecoverable")
```
giladshanan, line 18:
```suggestion
    unrecoverable  = OrderStatus.find_or_create_by!(name: "Unrecoverable")
```

spec/services/order_details/reconciler_spec.rb
```diff
@@ -12,7 +12,7 @@
     end
   end
   let(:params) { order_details.each_with_object({}) { |od, h| h[od.id.to_s] = ActionController::Parameters.new(reconciled: "1") } }
-  let(:reconciler) { described_class.new(OrderDetail.all, params, Time.current) }
+  let(:reconciler) { described_class.new(OrderDetail.all, params, Time.current, "reconciled") }
```
giladshanan, line 15: Are you adding these just to be explicit, right?  The specs would pass without this?

giladshanan (commented):
Left some questions

app/models/order_detail.rb
LeticiaErrandonea: I don't know if we'll have this case, but it made sense to prevent a reservation from being canceled if the work had already been done

spec/services/order_details/reconciler_spec.rb
LeticiaErrandonea: It's now a required param. It can either be `"reconciled"` or `nil`

giladshanan (approved):

PR #4637, opened 2024-10-11
[#162402]Change cross-core orders header

diebas:
# Release Notes

In the order's show view accessed through the "Manage Facility" menu, there's a section with a "Cross-core header". This header doesn't describe the menu's function correctly. The menu allows adding cross-core products to an order but it also allows to add products from the same facility as well and many users use it this way.

# Screenshot

Before:
Screenshot 2024-10-11 at 4 13 39 PM

After:
Screenshot 2024-10-11 at 4 13 01 PM

# Additional Context

I suggest a less specific header: "Add Product to Selected Order"

Didn't opt for a dynamic header as the header is defined in the HTML view structure. I don't think moving the logic to the javascript code to change it dynamically when selecting a facility in the menu dropdown is worth the effort as it's just a header and the action button already provides this information dynamically to the user. Let me know what you guys think about this.

### DIFF
--- a/app/views/facility_orders/show.html.haml
+++ b/app/views/facility_orders/show.html.haml
@@ -27,7 +27,7 @@
   = render "order_table", order_details: @order_details, cross_core: false, facility_id: current_facility.id
   = render "order_table_footer", order: @order, cross_core: false
 
-%h2= t("views.facility_orders.show.cross_core_orders_header")
+%h2= t("views.facility_orders.show.add_to_order_header")
 - if SettingsHelper.feature_on?(:cross_core_projects) && @cross_core_data_by_facility_id.present?
   #js--reportAnIssueModal.modal.hide.fade
 
--- a/app/views/facility_orders/show.html.haml
+++ b/app/views/facility_orders/show.html.haml
@@ -27,7 +27,9 @@
   = render "order_table", order_details: @order_details, cross_core: false, facility_id: current_facility.id
   = render "order_table_footer", order: @order, cross_core: false
 
-%h2= t("views.facility_orders.show.add_to_order_header")
+- subheader_key = (@cross_core_project.present? ? "views.facility_orders.show.add_to_cross_core_order_header" : "views.facility_orders.show.add_to_order_header")
+
+%h2= t(subheader_key)
 - if SettingsHelper.feature_on?(:cross_core_projects) && @cross_core_data_by_facility_id.present?
   #js--reportAnIssueModal.modal.hide.fade
 
--- a/config/locales/views/en.facility_orders.yml
+++ b/config/locales/views/en.facility_orders.yml
@@ -5,6 +5,7 @@ en:
         ordered_by: "Ordered By"
       show:
         add_to_order_header: "Add to Order"
+        add_to_cross_core_order_header: "Add to Cross-Core Order"
         cross_core_project_id: "Cross-Core Project ID"
         cross_core_project_total: "Cross-Core Project Total"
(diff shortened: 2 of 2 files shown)


config/locales/views/en.facility_orders.yml
```diff
@@ -4,7 +4,7 @@ en:
       index:
         ordered_by: "Ordered By"
       show:
-        cross_core_orders_header: "Cross-Core Orders"
+        add_product_order_header: "Add Product to Selected Order"
```
LeticiaErrandonea, line 7: I think I'd go with just "Add to order"

LeticiaErrandonea (commented):
Left a comment.

Thanks for the complete PR description! Much appreciated

config/locales/views/en.facility_orders.yml
diebas: sure, sounds good to me

LeticiaErrandonea (approved):

giladshanan (changes requested):
Can you add the ticket number to the PR title?

app/views/facility_orders/show.html.haml
```diff
@@ -27,7 +27,7 @@
   = render "order_table", order_details: @order_details, cross_core: false, facility_id: current_facility.id
   = render "order_table_footer", order: @order, cross_core: false
 
-%h2= t("views.facility_orders.show.cross_core_orders_header")
+%h2= t("views.facility_orders.show.add_to_order_header")
```
giladshanan, line 30: Could you check if the order is cross core, then select a different translation based on that?

giladshanan (changes requested):

app/views/facility_orders/show.html.haml
giladshanan: For example: Add to Order or Add to Cross Core Order

app/views/facility_orders/show.html.haml
```diff
@@ -27,7 +27,7 @@
   = render "order_table", order_details: @order_details, cross_core: false, facility_id: current_facility.id
   = render "order_table_footer", order: @order, cross_core: false
 
-%h2= t("views.facility_orders.show.cross_core_orders_header")
+%h2= t(@order.cross_core_project_id.present? ? "views.facility_orders.show.add_to_cross_core_order_header" : "views.facility_orders.show.add_to_order_header")
```
giladshanan, line 30: Small suggestion - I find it easier to read this code broken up into 2 lines:
```suggestion
- subheader_key = @order.cross_core_project_id.present? ? "views.facility_orders.show.add_to_cross_core_order_header" : "views.facility_orders.show.add_to_order_header"
%h2= t(subheader_key)
```

giladshanan (approved):
Looks good!  I left one suggestion, nothing blocking

app/views/facility_orders/show.html.haml
```diff
@@ -27,7 +27,9 @@
   = render "order_table", order_details: @order_details, cross_core: false, facility_id: current_facility.id
   = render "order_table_footer", order: @order, cross_core: false
 
-%h2= t("views.facility_orders.show.cross_core_orders_header")
+- subheader_key = (@order.cross_core_project_id.present? ? "views.facility_orders.show.add_to_cross_core_order_header" : "views.facility_orders.show.add_to_order_header")
```
LeticiaErrandonea, line 30: nit: `@cross_core_project` is already set and used a little bit above. Can you update this to use that var?

PR #4640, opened 2024-10-14
[#162450] Tech task: run parallel:spec on CI

joaquinco:
# Release Notes

Speed up CI tests by using parallel specs.


LeticiaErrandonea:
@giladshanan do you know why we had this gem in the Gemfile but didn't use for the CI. I think this is a great change but I wonder whether there were ever issues with this before

LeticiaErrandonea (approved):

PR #4641, opened 2024-10-14
[#162433] Tech task: fix capybara warnings on selector types

joaquinco:
# Notes

Clean up capybara spec warnings by converting ids to strings when using them as selectors in matchers.


LeticiaErrandonea (approved):
Thanks for taking care of this!

Two small requests:
- Can you add a ticket number in the title?
- We have a release script that sorts the PRs based on the title. If it includes "tech task" then it's considered as such. Can you remove the hyphen in the title?

PR #4660, opened 2024-10-16
[#162448] Tech task: Group rails gems on dependabot prs

joaquinco:
# Create Dependabot group for rails gems

This should cause all rails gems to be updated together in the same pr.
Allow only patch updates since larger updates should be done with manually.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,7 +11,10 @@ updates:
   - dependency-type: all
   groups:
     rails-gems:
-      update-types: [patch]
+      update-types:
+        - patch
+        - minor
+
       patterns:
         - rails
         - railties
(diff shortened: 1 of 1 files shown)


LeticiaErrandonea:
I think I'd rather include minor version as well. It may fail, but it's an easy way to keep track of Rails minor version bumps needed and what has to be fixed

joaquinco:
I only included patch versions so they are easy to merge. Minor version upgrades might require more work that ends up being postponed indefinitely and can make dependabot omit patch versions upgrades on its behalf.

We can always tell dependabot to ignore minor versions we don't want so it does not block it from creating patch version updates so I'm ok with including it.

LeticiaErrandonea (approved):

PR #4674, opened 2024-10-18
[Shared Dev] [#162408] Instrument Setup

LeticiaErrandonea:
# Release Notes
* Add option to set Daily Booking for instruments

# Screenshot
## New
image

## Show
image

# Accessibility
- [x] Did you add accessibility [specs](https://github.com/dequelabs/axe-core-gems/blob/develop/packages/axe-core-rspec/README.md)?


app/views/instruments/_instrument_manage_fields.html.haml
```diff
@@ -4,11 +4,18 @@
...
+- if SettingsHelper.feature_on?(:show_daily_rate_option) || !daily_booking
+  = f.input :pricing_mode
+
+- if daily_booking
+  = f.input :min_reserve_days, value_method: :to_i
+  = f.input :max_reserve_days, value_method: :to_i
+- else
+  = f.input :reserve_interval, value_method: :to_i, label: "Reservation Interval"
+  = f.input :min_reserve_mins, value_method: :to_i
+  = f.input :max_reserve_mins, default_value: "None"
```
giladshanan, line 18: If I'm reading this right, I think you could simplify a bit:
```suggestion
- if @product.daily_booking?
  - if SettingsHelper.feature_on?(:show_daily_rate_option)
    = f.input :pricing_mode
  = f.input :min_reserve_days, value_method: :to_i
  = f.input :max_reserve_days, value_method: :to_i
- else
  = f.input :pricing_mode
  = f.input :reserve_interval, value_method: :to_i, label: "Reservation Interval"
  = f.input :min_reserve_mins, value_method: :to_i
  = f.input :max_reserve_mins, default_value: "None"
```

db/migrate/20241015214235_add_reserve_days_fields_to_products.rb
```diff
@@ -0,0 +1,6 @@
+class AddReserveDaysFieldsToProducts < ActiveRecord::Migration[7.0]
```
giladshanan, line 1: magic frozen string comment

giladshanan (approved):
LGTM, left a couple of small nits

app/views/instruments/_instrument_manage_fields.html.haml
LeticiaErrandonea: We agreed only setting the value was going to be behind a Feature Flag, right? In that case, I think the `if` in line 13 is unnecessary. 

That said, I took a look at the code and this should be right:
```
# No if condition here
= f.input :pricing_mode

- if daily_booking
  = f.input :min_reserve_days, value_method: :to_i
  = f.input :max_reserve_days, value_method: :to_i
- else
  = f.input :reserve_interval, value_method: :to_i, label: "Reservation Interval"
  = f.input :min_reserve_mins, value_method: :to_i
  = f.input :max_reserve_mins, default_value: "None"
```

WDYT?

app/models/instrument.rb
```diff
@@ -103,6 +109,10 @@ def duration_pricing_mode?
     pricing_mode == "Duration"
   end
 
+  def daily_booking?
+    pricing_mode == "Schedule Rule (Daily Booking only)"
```
joaquinco, line 113: Maybe it's worth using a constant for this string, what do you think?

joaquinco (approved):

PR #4682, opened 2024-10-23
[#162409] [Shared dev] Daily rate price policies updates

joaquinco:
# Release Notes

Handle price policy for instruments with daily schedules.

## Feature changes

- Added two columns to `price_policies` table to track the usage rate per day and the usage subsidy per day.
- Hide minimum cost and cancellation cost inputs for instruments in this schedule, this applies for view and edit price policies tables. See screenshot.

## Other changes

- Improve format here and there
- Align minus sign for subsidy input
- Added constants for Instrument pricing modes

# Screenshot

Daily rate price policies view
Captura desde 2024-10-23 16-12-35

Daily rate price policies edit
Captura desde 2024-10-23 16-23-43

### DIFF
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -215,9 +215,14 @@ th.order-note {
 
 td.currency,
 th.currency,
-th.hourly_rate {
+th.hourly_rate,
+th.daily_rate {
   text-align: right;
 }
+td.daily_rate {
+  width: 120px;
+}
+
 .currency-input input {
   width: 7em;
 }
@@ -282,9 +287,17 @@ th.hourly_rate {
   }
 }
 
-.price-policy-table input[type=text],
 .half-width {
   width: 50%;
+}
+.adjustment-input::before {
+  content: '–';
+  width: 0;
+  display: inline-block;
+  overflow: visible;
+  position: relative;
+  left: -10px;
+  top: -4px;
 }
 
 .full-width {
@@ -303,6 +316,10 @@ p.per-minute {
 p.per-minute-show {
   @extend .per-minute;
   margin-top: 0;
+}
+
+.price-policy-table input[type=text] {
+  width: 80%;
 }
 
 /* Account Fields */
--- a/app/assets/stylesheets/nucore.scss
+++ b/app/assets/stylesheets/nucore.scss
@@ -290,18 +290,18 @@ td.daily_rate {
 .half-width {
   width: 50%;
 }
-.adjustment-input::before {
+
+.full-width {
+  width: 100%;
+}
+
+.negative-number::before {
   content: '–';
   width: 0;
   display: inline-block;
   overflow: visible;
   position: relative;
   left: -10px;
-  top: -4px;
-}
-
-.full-width {
-  width: 100%;
 }
 
 /* Price policies */
--- a/app/models/concerns/price_policies/usage.rb
+++ b/app/models/concerns/price_policies/usage.rb
@@ -7,8 +7,25 @@ module PricePolicies
     extend ActiveSupport::Concern
 
     included do
-      validates :usage_rate, presence: true, unless: :restrict_purchase?
-      validates :usage_rate, :usage_subsidy, :minimum_cost, numericality: { allow_blank: true, greater_than_or_equal_to: 0 }
+      validates :usage_rate,
+                presence: true,
+                unless: -> { restrict_purchase? || daily_booking? }
+      validates :usage_rate_daily,
+                presence: true,
+                unless: -> { restrict_purchase? || !daily_booking? }
+
+      validates(
+        :usage_rate,
+        :usage_rate_daily,
+        :usage_subsidy,
+        :usage_subsidy_daily,
+        :minimum_cost,
+        numericality: { allow_blank: true, greater_than_or_equal_to: 0 }
+      )
+
+      validate :daily_subsidy_less_than_rate, if: :daily_booking?
+
+      before_validation :sanitize_usage_rate
     end
 
     def has_rate?
@@ -41,6 +58,10 @@ module PricePolicies
       hourly_usage_rate - hourly_usage_subsidy
     end
 
+    def subsidized_daily_usage_cost
+      usage_rate_daily - usage_subsidy_daily
+    end
+
     def minimum_cost_subsidy
       return unless has_minimum_cost?
       minimum_cost * subsidy_ratio
@@ -64,6 +85,20 @@ module PricePolicies
 
     def subsidy_field
       :usage_subsidy
+    end
+
+    def daily_subsidy_less_than_rate
+      return if usage_subsidy_daily.blank? || usage_rate_daily.blank?
+      return if usage_subsidy_daily <= usage_rate_daily
+
+      errors.add(:usage_subsidy_daily, :subsidy_greater_than_cost)
+    end
+
+    def sanitize_usage_rate
+      if usage_rate_daily_changed? && daily_booking?
+        self.usage_rate = nil
+        self.usage_subsidy = nil
+      end
     end
 
   end
--- a/app/models/price_policy.rb
+++ b/app/models/price_policy.rb
@@ -185,6 +185,12 @@ class PricePolicy < ApplicationRecord
     Settings.price_policy_note_options.include?(note) ? note : "Other"
   end
 
+  def daily_booking?
+    return false unless association(:product).loaded?
+
+    product&.daily_booking?
+  end
+
   private
 
   # TODO: Refactor
--- a/app/models/price_policy.rb
+++ b/app/models/price_policy.rb
@@ -186,8 +186,6 @@ class PricePolicy < ApplicationRecord
   end
 
   def daily_booking?
-    return false unless association(:product).loaded?
-
     product&.daily_booking?
   end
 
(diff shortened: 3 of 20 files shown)


app/models/concerns/price_policies/usage.rb
```diff
@@ -66,6 +87,20 @@ def subsidy_field
       :usage_subsidy
     end
 
+    def daily_subsidy_less_than_rate
+      return if usage_subsidy_daily.blank? || usage_rate_daily.blank?
+      return if usage_subsidy_daily <= usage_rate_daily
+
+      errors.add(:usage_subsidy_daily, :subsidy_greater_than_cost)
```
giladshanan, line 94: How does this error appear to the user?

db/migrate/20241022135335_add_price_policy_usage_rate_daily.rb
```diff
@@ -0,0 +1,8 @@
+class AddPricePolicyUsageRateDaily < ActiveRecord::Migration[7.0]
```
giladshanan, line 1: Can you  `# frozen_string_literal: true` at the top of this file?

app/views/time_based_price_policies/_price_policy_fields.html.haml
```diff
@@ -11,6 +11,7 @@
 
 %div{ style: "display: flex; flex-direction: row-reverse" }
   - if @product.duration_pricing_mode?
+
```
giladshanan, line 14: Is this newline needed?

app/views/time_based_price_policies/_price_policy_fields.html.haml
```diff
@@ -43,7 +46,6 @@
           %td
         - if local_assigns[:cancellation]
           %td
-      - if @product.duration_pricing_mode?
```
giladshanan, line 46: Why did this get removed?

config/locales/views/admin/en.price_policies.yml
```diff
@@ -24,6 +24,7 @@ en:
       edit: "Edit Price Policies"
       remove: "Remove Price Policies"
   time_based_price_policies:
+    problem: 'There are errors in this policy:'
```
giladshanan, line 27: Can you use double quotes here for consistency?

app/assets/stylesheets/nucore.scss
```diff
@@ -282,10 +287,18 @@ th.hourly_rate {
   }
 }
 
-.price-policy-table input[type=text],
 .half-width {
   width: 50%;
 }
+.adjustment-input::before {
```
giladshanan, line 293: I would give this a more generic name like `.negative-number::before`

app/views/time_based_price_policies/_table.html.haml
```diff
@@ -78,6 +80,12 @@
                 - price_per_minute = (price_policy.subsidized_hourly_usage_cost / 60)
                 = number_to_currency price_per_minute, precision: 4
                 \/ minute
+            - elsif price_policy.usage_rate_daily.present?
+              .rate= number_to_currency price_policy.usage_rate_daily
+              - if price_policy.usage_subsidy_daily.present?
+                .subsidy= "- #{number_to_currency price_policy.usage_subsidy_daily}"
+                %strong= "= #{number_to_currency price_policy.subsidized_daily_usage_cost}"
```
giladshanan, line 87: could you use `.negative-number` here?

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -400,4 +400,115 @@
       expect(page).not_to have_content("$90.00") # Step 3 rate (removed)
     end
   end
+
+  context "Schedule Rule Daily pricing mode" do
+    let(:pricing_mode) { Instrument::Pricing::SCHEDULE_DAILY }
+
+    it "editting rates automatically fills adjustments", :js do
```
giladshanan, line 407:
```suggestion
    it "editing rates automatically fills adjustments", :js do
```

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -400,4 +400,115 @@
...
+      fill_in("note", with: "Some note about pricing")
+
+      check("price_policy_#{base_price_group.id}[can_purchase]")
+      check("price_policy_#{cancer_center.id}[can_purchase]")
+      uncheck("price_policy_#{external_price_group.id}[can_purchase]")
+
+      fill_in("price_policy_#{base_price_group.id}[usage_rate_daily]", with: "100")
+      fill_in("price_policy_#{cancer_center.id}[usage_subsidy_daily]", with: "30")
+
+      # this is autofilled by js when enabled
```
giladshanan, line 448: If you enable JS for this spec, and add an expectation for the daily rate displayed to the user to be 100, I think you could delete the spec above which starts on line 407.  what do you think?

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -400,4 +400,115 @@
...
+
+        expect { click_button("Save Rules") }.to(
+          change do
+            base_price_policy.reload.usage_rate_daily
+          end.from(
+            base_price_policy.usage_rate_daily
+          ).to(
+            new_rate
+          )
+        )
```
giladshanan, line 510: would it be possible to make assertions about the front end instead?

giladshanan (changes requested):
Nice work here!
I'm not sure but I think there is a breaking change [here](https://github.com/tablexi/nucore-open/pull/4682/files#diff-541867b7c584c0598a4495c5aa5b99c096cd4ec169e345c4730398db006844f5L46).
We should probably have a spec to protect against this regression but it looks like we don't.

app/models/price_policy.rb
```diff
@@ -185,6 +185,12 @@ def note_option
     Settings.price_policy_note_options.include?(note) ? note : "Other"
   end
 
+  def daily_booking?
+    return false unless association(:product).loaded?
```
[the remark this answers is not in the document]
joaquinco: I just don't like validations to trigger db queries. I know it better for consistency to just remove this so I'm ok without it too

app/views/time_based_price_policies/_price_policy_fields.html.haml
joaquinco: good catch! this was an error

joaquinco: Oh I remember now, it's not an error. There are two ifs with the same condition one after another so I removed the second one

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -400,4 +402,105 @@
...
+            note: "Some note about pricing",
+            usage_rate_daily: 100,
+            charge_for: InstrumentPricePolicy::CHARGE_FOR[:reservation]
+          )
+
+          price_group = price_policy.price_group
+          price_policy.usage_subsidy_daily = 10 if price_group.external? || price_group.master_internal?
+
+          price_policy.save!
+        end
```
LeticiaErrandonea, line 470: I think you should be able to create the Price policy using the factory

LeticiaErrandonea (commented):
Looking good!

Left some comments

app/models/price_policy.rb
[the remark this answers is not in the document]
joaquinco: I ended up remove this

app/views/time_based_price_policies/_adjustment_row.html.haml
```diff
@@ -8,7 +8,7 @@
 
 
 - if local_assigns[:cancellation]
-  %td
+  %td.daily_rate
```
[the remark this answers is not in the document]
joaquinco: I agree it's weird, I probably wanted to add this class to another td

spec/system/admin/instrument_price_policies_controller_spec.rb
joaquinco: makes sense

app/views/time_based_price_policies/_amount_row.html.haml
```diff
@@ -19,9 +19,15 @@
...
-%td
-  = pp.label :usage_rate, t("price_policies.rate"), class: "normal-weight"
-  = pp.text_field :usage_rate, value: display_usage_rate(price_group, price_policy),
-    size: 8,
-    class: "#{price_group.master_internal? ? 'master_usage_cost' : ''} usage_rate",
-    data: { target: ".js--usageRate" }
+- if @product.daily_booking?
+  %td.daily_rate
+    = pp.label :usage_rate_daily, t("price_policies.rate"), class: "normal-weight"
+    = pp.text_field :usage_rate_daily, value: display_rate(price_policy.usage_rate_daily),
```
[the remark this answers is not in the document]
joaquinco: The `display_usage_rate` function returns a value from params if present and it is needed due to this bugfix https://github.com/tablexi/nucore-open/pull/340 which shouldn't happen for daily rates since there's no conversion from hours to minutes.

Btw, I'm not even sure if the `display_usage_rate` logic is needed at all since I couldn't reproduce the bug when removing the fix, but that's a conversation for another day.

giladshanan (approved):
LGTM!

spec/models/instrument_price_policy_spec.rb
```diff
@@ -175,4 +175,80 @@
...
+      it_behaves_like "validates policy field", :usage_rate_daily
+    end
+
+    describe "with missing usage_rate_daily value" do
+      before { policy.usage_rate_daily = nil }
+
+      it_behaves_like "validates policy field", :usage_rate_daily
+    end
+
+    describe "validates subsidy equals to rate" do
```
LeticiaErrandonea, line 216: I think we're testing this case

```suggestion
    describe "validates subsidy greater than rate" do
```

LeticiaErrandonea (approved):
LGTM!

Thanks for the changes!

PR #4692, opened 2024-10-25
[#161850]Add ability to hide Price Groups

diebas:
# Release Notes

Added the ability to hide price groups in the admin menu. This option is only available for non-global price groups.
Once a price group is hidden, current price policies related to that group aren't visible or editable anymore. Old price policies remain visible. It also updates their can_parchase value to false

# Screenshot

Screenshot 2024-10-25 at 10 27 08 AM

Screenshot 2024-10-25 at 10 34 55 AM

# Additional Context

This is a draft PR, tests are on their way

### DIFF
--- a/app/models/price_group.rb
+++ b/app/models/price_group.rb
@@ -21,6 +21,7 @@ class PriceGroup < ApplicationRecord
   before_destroy { throw :abort if global? }
   before_destroy { price_policies.destroy_all } # Cannot be a dependent: :destroy because of ordering of callbacks
   before_create  ->(o) { o.display_order = 999 unless o.facility_id.nil? }
+  before_update  :update_hidden_price_group_policies, if: :is_hidden_changed?
 
   scope :for_facility, ->(facility) { where(facility_id: [nil, facility.id]) }
   scope :globals, -> { where(facility_id: nil) }
@@ -116,6 +117,15 @@ class PriceGroup < ApplicationRecord
   end
 
   private
+
+  def is_hidden_changed?
+    self.will_save_change_to_attribute?(:is_hidden)
+  end
+
+  def update_hidden_price_group_policies
+    policies = self.price_policies.current
+    PricePolicyUpdater.update_can_purchase(policies)
+  end
 
   # Find a global price group by name, or create it if it does not exist
   def self.find_or_create_global(name:, display_order:, is_internal: false, admin_editable: true)
--- a/app/models/price_group.rb
+++ b/app/models/price_group.rb
@@ -21,7 +21,7 @@ class PriceGroup < ApplicationRecord
   before_destroy { throw :abort if global? }
   before_destroy { price_policies.destroy_all } # Cannot be a dependent: :destroy because of ordering of callbacks
   before_create  ->(o) { o.display_order = 999 unless o.facility_id.nil? }
-  before_update  :update_hidden_price_group_policies, if: :will_save_change_to_is_hidden?
+  before_update  :update_hidden_price_group_policies, if: :will_save_change_to_is_hidden? && :is_hidden
 
   scope :for_facility, ->(facility) { where(facility_id: [nil, facility.id]) }
   scope :globals, -> { where(facility_id: nil) }
--- a/app/models/price_group.rb
+++ b/app/models/price_group.rb
@@ -21,11 +21,12 @@ class PriceGroup < ApplicationRecord
   before_destroy { throw :abort if global? }
   before_destroy { price_policies.destroy_all } # Cannot be a dependent: :destroy because of ordering of callbacks
   before_create  ->(o) { o.display_order = 999 unless o.facility_id.nil? }
-  before_update  :update_hidden_price_group_policies, if: :is_hidden_changed?
+  before_update  :update_hidden_price_group_policies, if: :will_save_change_to_is_hidden?
 
   scope :for_facility, ->(facility) { where(facility_id: [nil, facility.id]) }
   scope :globals, -> { where(facility_id: nil) }
   scope :by_display_order, -> { reorder(display_order: :asc, name: :asc) }
+  scope :visible, -> { where(is_hidden: false) }
 
   def self.base
     globals.find_by(name: Settings.price_group.name.base)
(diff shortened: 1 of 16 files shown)


config/locales/en.yml
```diff
@@ -954,6 +954,7 @@ en:
   price_groups:
     price_group_fields:
       internal_instruction: "Check this box if this price group is internal:"
+      hidden_instruction: "Check this box if you want to hide this price group:"
```
LeticiaErrandonea, line 957: Maybe we should use another voice? Something like

```suggestion
      hidden_instruction: "Check this box if this price group is no longer used:"
```

db/migrate/20241014161738_add_is_hidden_to_price_groups.rb
```diff
@@ -0,0 +1,14 @@
+class AddIsHiddenToPriceGroups < ActiveRecord::Migration[7.0]
+  def up
+    add_column :price_groups, :is_hidden, :boolean, default: false
+
+    PriceGroup.reset_column_information
+    PriceGroup.update_all(is_hidden: false)
```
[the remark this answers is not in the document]
LeticiaErrandonea: Agreed. We usually go with an empty class declared in the file. You can find examples in other migrations.

app/models/price_group.rb
```diff
@@ -117,6 +118,15 @@ def setup_skip_review_price_policies
 
   private
 
+  def is_hidden_changed?
+    self.will_save_change_to_attribute?(:is_hidden)
```
LeticiaErrandonea, line 122: nit: `self` is unnecessary

Also, do we need to wrap `will_save_change_to_attribute?`?

app/models/price_group.rb
```diff
@@ -117,6 +118,15 @@ def setup_skip_review_price_policies
...
+  end
+
+  def update_hidden_price_group_policies
+    policies = self.price_policies.current
```
LeticiaErrandonea, line 126: nit: `self` is unnecessary

app/services/price_policy_builder.rb
```diff
@@ -94,7 +94,7 @@ def editable?
 
   def get_price_policies
     return [] unless editable?
-    facility.price_groups.map do |price_group|
+    facility.price_groups.where(is_hidden: false).map do |price_group|
```
LeticiaErrandonea, line 97: Not a strong opinion but I think it'd be clearer to use a scope

LeticiaErrandonea (commented):
Looking good!

Left a few comments

app/models/price_group.rb
```diff
@@ -21,10 +21,12 @@ class PriceGroup < ApplicationRecord
   before_destroy { throw :abort if global? }
   before_destroy { price_policies.destroy_all } # Cannot be a dependent: :destroy because of ordering of callbacks
   before_create  ->(o) { o.display_order = 999 unless o.facility_id.nil? }
+  before_update  :update_hidden_price_group_policies, if: :will_save_change_to_is_hidden?
```
LeticiaErrandonea, line 24: What if I'm changing from hidden to visible? As the code is now, we'll make a call to set `can_purchase` to `false`, but that's unnecessary. I think we should avoid getting to that part of the code

db/migrate/20241014161738_add_is_hidden_to_price_groups.rb
```diff
@@ -0,0 +1,13 @@
+class AddIsHiddenToPriceGroups < ActiveRecord::Migration[7.0]
```
LeticiaErrandonea, line 1: Please add `# frozen_string_literal: true`

spec/support/shared_examples/hidden_price_groups_shared_examples.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    expect(page).to have_content(base_price_group.name)
+    expect(page).not_to have_content(price_group_to_hide.name)
+
+    click_link "Add Pricing Rules"
+
+    expect(page).to have_content(base_price_group.name)
+    expect(page).not_to have_content(price_group_to_hide.name)
+  end
+end
```
LeticiaErrandonea, line 56: nit: add newline at the end of the file

spec/support/shared_examples/hidden_price_groups_shared_examples.rb
```diff
@@ -0,0 +1,56 @@
...
+
+  it "hides price policies related to that price group" do
+    visit send("facility_#{item_type}s_path", facility, item)
+    click_link item.name
+    click_link "Pricing"
+    click_link "Add Pricing Rules"
+
+    expect(page).to have_content(price_group_to_hide.name)
+
+    if item_type == "instrument"
```
LeticiaErrandonea, line 12: nit: I wonder whether `item.is_a?(Instrument)` is easier to read

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -154,6 +154,12 @@
         expect(page).to have_no_link "Remove"
       end
     end
+
+    describe "with hidden price policies", :js do
```
LeticiaErrandonea, line 158: Do we need `:js` here?

spec/system/admin/instrument_price_policies_controller_spec.rb
```diff
@@ -399,5 +405,11 @@
       # External
       expect(page).not_to have_content("$90.00") # Step 3 rate (removed)
     end
+
+    describe "with hidden price policies", :js do
```
LeticiaErrandonea, line 409: Same as above regarding `:js`

spec/system/admin/item_price_policies_controller_spec.rb
```diff
@@ -68,6 +68,10 @@
     end
   end
 
+  describe "with hidden price policies", :js do
```
LeticiaErrandonea, line 71: Same as above regarding `:js`

LeticiaErrandonea (commented):
I really like the shared examples idea!

Left some comments

spec/system/admin/instrument_price_policies_controller_spec.rb
diebas: Well, I can't tell the exact reason behind this, but if I remove the :js flag the tests fail. Capybara isn't able to find some fields

LeticiaErrandonea (approved):
LGTM!

Thanks for the changes!

giladshanan (approved):

PR #4703, opened 2024-10-30
Fix: Price policies flaky spec

diebas:
# Release Notes

There is a flaky test failing related to hidden price groups; Some assertions are happening before loading the desired page it's not finding specific content on the page. I added some tests to make sure it loads the page correctly.


joaquinco (approved):

LeticiaErrandonea (approved):
I think we're fine merging this and seeing what happens in `master`. It's been consistently failing there, so we should see whether it works soon

PR #4704, opened 2024-10-30
[#162475][Shared dev] Daily rate: Make a reservation

joaquinco:
## Notes

Allow users to reserve a daily booking instrument. A daily reservation is created out of a start date and time
and a duration in days, then the end time is taken to be the same time with date duration days later.

The user flows changed here are the ones that use the view: `reservation/new`.

- Allow daily priced instruments to be reserved facility user
- Allow admins to make daily reservations on behalf of users
- Allow to add a reservation to an order

## Testing notes

In order to test this the instrument must have full availability for the duration of the reservation. Schedule rule for daily rate instruments will be updated in a future branch.

## Screenshot

Update `reservations/new` view to hide calendar (for now) and show fields according to daily booking requirements:
- Duration is in days
- End time is adjusted when start time or duration is changed

Captura desde 2024-10-29 18-43-12


spec/models/time_range_spec.rb
```diff
@@ -38,6 +38,45 @@
...
+
+      it { is_expected.to eq(1) }
+    end
+
+    describe "when it starts and end the following day more than 24 hours appart" do
+      let(:start_at) { Time.zone.local(2024, 10, 24, 10, 0, 0) }
+      let(:end_at) { Time.zone.local(2024, 10, 25, 19, 0, 0) }
+
+      it { is_expected.to eq(2) }
+    end
```
LeticiaErrandonea, line 77: I'm not sure what's best here. We're not supposed to have these cases. For us, a daily reservation can only use 24-hour blocks. I'm not even sure what the expected math is here.

app/assets/javascripts/reservations.js
```diff
@@ -1,6 +1,18 @@
 $(document).ready(function() {
   init_datepickers();
-  new ReservationCalendar().init($("#calendar"), $(".js--reservationForm"));
+
+  var form = {
```
LeticiaErrandonea, line 4: We usually prefer `let` and `const` over `var`

LeticiaErrandonea (commented):
Looking good!

Left some comments and I'll dev QA it tomorrow

spec/models/time_range_spec.rb
joaquinco: This is just testing the `TimeRange` which is fairly it's own abstraction I think it's better if we have a defined behavior for other than expected business logic data.

spec/factories/products.rb
```diff
@@ -138,6 +138,13 @@
 
     schedule { create :schedule, facility: facility }
 
+    after(:build) do |product, _evaluator|
+      if product.daily_booking?
+        product.max_reserve_mins = nil
+        product.min_reserve_mins = nil
```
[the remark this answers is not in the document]
joaquinco: Nope, but I'll remove this in a fix soon

spec/models/time_range_spec.rb
LeticiaErrandonea: I see your point. I agree that the current implementation works this way and it's usually good to have specs consider different (or all) cases.

For me, the problem is that this case is invalid for us*, so covering it creates an idea of "expected behavior" that isn't actually expected. If somehow we tried to calculate how many days the reservation lasted and it wasn't a natural number, I'd much rather get an error immediately than have clients ask why billing is working in an unexpected way.

*Universities can't charge users for longer than what they had the instrument available for them, so 36 hours can't be 2 days. This is the reason behind having strict 24-hour blocks.

LeticiaErrandonea (commented):
I did some QA and worked fine!

I noticed that I need to have schedule rules that cover the whole day for a Daily reservation to work. This seems like something useful to highlight in the Testing notes before releasing to Staging, as I can see clients wondering how those need to be set up.

spec/models/time_range_spec.rb
joaquinco: I'll just move the ceil call over to the reservation class, where duration_days is used. This value is computed out of validated data so if it returns an incorrect value it should have been captured elsewhere

spec/factories/price_policies.rb
```diff
@@ -5,13 +5,20 @@
...
     minimum_cost { 1 }
     can_purchase { true }
     start_date { Time.zone.now.beginning_of_day }
     expire_date { PricePolicy.generate_expire_date(start_date) }
     note { "This is note" }
+    usage_rate { 10 / 60.0 }
+    usage_subsidy { 0 }
+
+    after(:build) do |price_policy, _|
+      if price_policy.daily_booking?
```
LeticiaErrandonea, line 17: Should we also make `usage_rate` and `usage_subsidy` be `nil`?

LeticiaErrandonea (approved):
Left a small change, but it LGTM!

Thanks for the changes and answers!

spec/factories/price_policies.rb
joaquinco: I'll move the `usage_rate` and `usage_subsidy` fields specification to an else clause in this block so we allow invalid instances to be build using this factory if desired

PR #4705, opened 2024-10-30
[Fix] Enable director edit price policy flag on new price policies tests

joaquinco:
This is causing newest price policies tests to fail if the environment have that flag off.


LeticiaErrandonea (approved):
Nice catch!

PR #4712, opened 2024-10-31
[#162393] Avoid creation of service and item price policies with usage_rate

diebas:
# Release Notes

There's a service that automatically creates price_policies with a usage_rate value of 0. This is not totally real as some price_policies aren't meant to have usage_rate at all which may result in future bugs.

Change the code to reflect reality, add tests, and refactor a little to reuse some code.

### DIFF
--- a/app/services/price_policy_builder.rb
+++ b/app/services/price_policy_builder.rb
@@ -125,15 +125,13 @@ class PricePolicyBuilder
   end
 
   def self.create_price_policy_for(product, price_group)
-    usage_rate = ["Service", "Item"].include?(product.type) ? nil : 0
-
     PricePolicy.create(
       type: "#{product.type}PricePolicy",
-      product: product,
+      product:,
       start_date: 1.month.ago,
       expire_date: 75.years.from_now,
-      price_group: price_group,
-      usage_rate: usage_rate,
+      price_group:,
+      usage_rate: usage_rate_for(product),
       minimum_cost: 0,
       cancellation_cost: 0,
       usage_subsidy: 0,
@@ -143,6 +141,10 @@ class PricePolicyBuilder
       charge_for: "reservation",
       note: "Price rule automatically created because of billing mode"
     )
+  end
+
+  def self.usage_rate_for(product)
+    ["Service", "Item"].include?(product.type) ? nil : 0
   end
 
 end
--- a/app/services/price_policy_builder.rb
+++ b/app/services/price_policy_builder.rb
@@ -26,42 +26,12 @@ class PricePolicyBuilder
   def self.create_skip_review_for(product, price_groups = nil)
     groups = price_groups || PriceGroup.globals
     groups.each do |price_group|
-      PricePolicy.create(
-        type: "#{product.type}PricePolicy",
-        product:,
-        start_date: 1.month.ago,
-        expire_date: 75.years.from_now,
-        price_group:,
-        usage_rate: 0,
-        minimum_cost: 0,
-        cancellation_cost: 0,
-        usage_subsidy: 0,
-        unit_cost: 0,
-        unit_subsidy: 0,
-        can_purchase: true,
-        charge_for: "reservation",
-        note: "Price rule automatically created because of billing mode"
-      )
+      create_price_policy_for(product, price_group)
     end
   end
-
+  
   def self.create_nonbillable_for(product)
-    PricePolicy.create(
-      type: "#{product.type}PricePolicy",
-      product:,
-      start_date: 1.month.ago,
-      expire_date: 75.years.from_now,
-      price_group: PriceGroup.nonbillable,
-      usage_rate: 0,
-      minimum_cost: 0,
-      cancellation_cost: 0,
-      usage_subsidy: 0,
-      unit_cost: 0,
-      unit_subsidy: 0,
-      can_purchase: true,
-      charge_for: "reservation",
-      note: "Price rule automatically created because of billing mode"
-    )
+    create_price_policy_for(product, PriceGroup.nonbillable)
   end
 
   def initialize(product, start_date)
@@ -152,6 +122,27 @@ class PricePolicyBuilder
   def price_policies_for_start_date
     return [] if start_date.blank?
     product.price_policies.for_date(start_date)
+  end
+
+  def self.create_price_policy_for(product, price_group)
+    usage_rate = ["Service", "Item"].include?(product.type) ? nil : 0
+
+    PricePolicy.create(
+      type: "#{product.type}PricePolicy",
+      product: product,
+      start_date: 1.month.ago,
+      expire_date: 75.years.from_now,
+      price_group: price_group,
+      usage_rate: usage_rate,
+      minimum_cost: 0,
+      cancellation_cost: 0,
+      usage_subsidy: 0,
+      unit_cost: 0,
+      unit_subsidy: 0,
+      can_purchase: true,
+      charge_for: "reservation",
+      note: "Price rule automatically created because of billing mode"
+    )
   end
 
 end
--- a/spec/services/price_policy_builder_spec.rb
+++ b/spec/services/price_policy_builder_spec.rb
@@ -1,0 +1,79 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe PricePolicyBuilder do
+  let(:price_group) { create(:price_group) }
+  let!(:facility) { create(:setup_facility) }
+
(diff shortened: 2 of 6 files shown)


spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.shared_examples "creates a product with billing mode" do |product_type, billing_mode|
```
LeticiaErrandonea, line 1: nit: please add `# frozen_string_literal: true`

nit 2: please add newline at the end of the file

nit 3: the filename should be something like "create_product_with_price_rule_shared_examples.rb", right? We're not only creating services

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
```diff
@@ -0,0 +1,21 @@
...
+  let(:logged_in_user) { administrator }
+
+  it "can create a #{product_type}, which automatically creates a price rule" do
+    visit facility_products_path(facility)
+    all("a", text: product_type.capitalize)[0].click
```
LeticiaErrandonea, line 6: Why do we need to get `all` and then select the first one?

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
```diff
@@ -0,0 +1,21 @@
...
+    click_link "Add #{product_type.capitalize}"
+
+    fill_in "Name", with: "My new Product", match: :first
+    fill_in "URL Name", with: "new-service"
+    select "Required", from: "#{product_type}[user_notes_field_mode]"
+    select billing_mode, from: "#{product_type}[billing_mode]"
```
LeticiaErrandonea, line 12: Can we do `choose billing_mode`?

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    click_button "Create"
+    click_on "Pricing"
```
LeticiaErrandonea, line 15: Between these two I'd check that the product was created

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    expect(page).to have_content "#{PriceGroup.nonbillable} (#{PriceGroup.nonbillable.type_string}) $0"
```
LeticiaErrandonea, line 17: If we're only creating nonbillable, do we need the `billing_mode` passed in?

LeticiaErrandonea (commented):
Looking good!

Left some comments with questions

app/services/price_policy_builder.rb
```diff
@@ -154,4 +124,25 @@ def price_policies_for_start_date
...
+  def self.create_price_policy_for(product, price_group)
+    usage_rate = ["Service", "Item"].include?(product.type) ? nil : 0
+
+    PricePolicy.create(
+      type: "#{product.type}PricePolicy",
+      product: product,
+      start_date: 1.month.ago,
+      expire_date: 75.years.from_now,
+      price_group: price_group,
+      usage_rate: usage_rate,
```
joaquinco, line 136: nit: we can use the hash shorthand syntax here

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
diebas: The reason behind this is that when the product type is service, there's another link in the menu called "Timed Services" that interferes when calling `click_link "Services"`. With this code, I find all the links that contain the word "Services" (there are only 2) and select the first one, which is the link I need to visit.

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
diebas: I don't think so, as this is a drop-down, not a radio button

app/services/price_policy_builder.rb
```diff
@@ -154,4 +124,25 @@ def price_policies_for_start_date
     product.price_policies.for_date(start_date)
   end
 
+  def self.create_price_policy_for(product, price_group)
+    usage_rate = ["Service", "Item"].include?(product.type) ? nil : 0
```
giladshanan, line 128: I'd vote for making this a separate method and adding a simple test for it

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
diebas: We are not only just creating products with nonbillable billing_mode, we are creating skip review too.
As I understand, creating products of both billing modes generates "nonbillable" price groups (base price group with $0 cost). This code checks that the price group is being displayed correctly in the product's price tab.

`PriceGroup.nonbillable ` returns the base price group

```
  def self.nonbillable
    base
  end

  def self.base
    globals.find_by(name: Settings.price_group.name.base)
  end
```

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
LeticiaErrandonea: Oh, I think that you can use something like `click_link("Services", href: facility_services_path(facility))`

spec/support/shared_examples/create_service_with_price_rule_shared_examples.rb
LeticiaErrandonea: You're right 😅 

What about something like `select billing_mode, from: "Billing Mode"`?

spec/services/price_policy_builder_spec.rb
```diff
@@ -0,0 +1,80 @@
...
+
+require "rails_helper"
+
+RSpec.describe PricePolicyBuilder do
+  let(:price_group) { create(:price_group) }
+  let!(:facility) { create(:setup_facility) }
+
+  shared_examples "a correctly created PricePolicy" do
+    before do
+      allow_any_instance_of(PricePolicy).to receive(:validate).and_return(true)
```
LeticiaErrandonea, line 11: Why do we need to skip validations?

LeticiaErrandonea (commented):
Left a question but other than that I think this is fine 😄

LeticiaErrandonea (approved):
LGTM! Thanks for the changes!

app/services/price_policy_builder.rb
```diff
@@ -154,4 +124,27 @@ def price_policies_for_start_date
...
+      unit_cost: 0,
+      unit_subsidy: 0,
+      can_purchase: true,
+      charge_for: "reservation",
+      note: "Price rule automatically created because of billing mode"
+    )
+  end
+
+  def self.usage_rate_for(product)
+    usage_rate = ["Service", "Item"].include?(product.type) ? nil : 0
```
giladshanan, line 147:
```suggestion
    ["Service", "Item"].include?(product.type) ? nil : 0
```

spec/support/shared_examples/create_product_with_price_rule_shared_examples.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+RSpec.shared_examples "creates a product with billing mode" do |product_type, billing_mode|
```
joaquinco, line 3: I think it if it's used by a single example group we can define these shared examples with the same file

app/services/price_policy_builder.rb
```diff
@@ -154,4 +124,27 @@ def price_policies_for_start_date
     product.price_policies.for_date(start_date)
   end
 
+  def self.create_price_policy_for(product, price_group)
+    PricePolicy.create(
+      type: "#{product.type}PricePolicy",
+      product: product,
```
joaquinco, line 130: nit: hash value can be omitted

joaquinco (approved):

spec/support/shared_examples/create_product_with_price_rule_shared_examples.rb
diebas: It's used in 2 different files:

spec/system/admin/creating_a_service_spec.rb
spec/system/admin/creating_an_item_spec.rb

joaquinco: I hadn't noticed it

PR #4721, opened 2024-11-04
[#162473][Daily Rate]: Find next available reservation time

joaquinco:
# Notes

Make product's next available reservation logic to work for daily rate instruments.

> To be discussed
> What should be the reserve interval for daily rate instruments?

## Testing Notes

This does not include schedule rules changes required by daily rate instruments.
When loading the reservation page, the initial date should be an empty slot where the instrument can be reserved using the default duration.

### DIFF
--- a/app/helpers/date_helper.rb
+++ b/app/helpers/date_helper.rb
@@ -102,7 +102,7 @@ module DateHelper
   private
 
   def minute_options(step = nil)
-    step ||= 1
+    step ||= 5
     (0..59).step(step).map { |d| ["%02d" % d, d] }
   end
 
--- a/app/helpers/date_helper.rb
+++ b/app/helpers/date_helper.rb
@@ -102,7 +102,7 @@ module DateHelper
   private
 
   def minute_options(step = nil)
-    step ||= 5
+    step ||= 1
     (0..59).step(step).map { |d| ["%02d" % d, d] }
   end
 
--- a/app/models/instrument.rb
+++ b/app/models/instrument.rb
@@ -16,6 +16,7 @@ class Instrument < Product
   include EmailListAttribute
 
   RESERVE_INTERVALS = [1, 5, 10, 15, 30, 60].freeze
+  RESERVE_INTERVAL_DAILY = 1
 
   PRICING_MODES = [Pricing::SCHEDULE_RULE, Pricing::DURATION].tap do |pricing_modes|
     if SettingsHelper.feature_on?(:show_daily_rate_option)
@@ -152,7 +153,7 @@ class Instrument < Product
     if daily_booking?
       self.min_reserve_mins = nil
       self.max_reserve_mins = nil
-      self.reserve_interval = nil
+      self.reserve_interval = RESERVE_INTERVAL_DAILY
     else
       self.min_reserve_days = nil
       self.max_reserve_days = nil
--- a/app/models/instrument.rb
+++ b/app/models/instrument.rb
@@ -3,9 +3,11 @@
 class Instrument < Product
 
   module Pricing
+
     SCHEDULE_RULE = "Schedule Rule"
     SCHEDULE_DAILY = "Schedule Rule (Daily Booking only)"
     DURATION = "Duration"
+
   end
 
   include Products::RelaySupport
@@ -14,6 +16,7 @@ class Instrument < Product
   include EmailListAttribute
 
   RESERVE_INTERVALS = [1, 5, 10, 15, 30, 60].freeze
+  RESERVE_INTERVAL_DAILY = 5
   PRICING_MODES = [Pricing::SCHEDULE_RULE, Pricing::DURATION].tap do |pricing_modes|
     if SettingsHelper.feature_on?(:show_daily_rate_option)
       pricing_modes.insert(1, Pricing::SCHEDULE_DAILY)
@@ -86,7 +89,7 @@ class Instrument < Product
   def create_default_price_group_products
     PriceGroup.globals.find_each do |price_group|
       price_group_products.create!(
-        price_group: price_group,
+        price_group:,
         reservation_window: PriceGroupProduct::DEFAULT_RESERVATION_WINDOW
       )
     end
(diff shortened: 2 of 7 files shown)


spec/models/instrument_spec.rb
```diff
@@ -521,16 +521,55 @@
...
       assert_equal 10, @next_reservation.reserve_start_at.hour
     end
   end
 
+  context "next available reservation with daily booking" do
+    let!(:instrument) do
+      create(
+        :instrument,
+        facility:,
+        pricing_mode: Instrument::Pricing::SCHEDULE_DAILY,
```
diebas, line 534: Not a correction, but a question: why didn't you use the :daily_booking trait you've just created?

spec/models/instrument_spec.rb
```diff
@@ -521,16 +521,55 @@
...
+      expect(second_reservation.reserve_start_at).to_not be_before(reservation.reserve_end_at)
+    end
+  end
+
   context "available hours based on schedule rules" do
     before(:each) do
       # create instrument, min reserve time is 60 minutes, max is 60 minutes
       @instrument = FactoryBot.create(:instrument,
-                                      facility: facility,
+                                      facility:,
```
diebas, line 572: Nit: As you are touching this line of code, maybe you can get rid of the old FactoryBot syntax.

spec/services/next_available_reservation_finder_spec.rb
```diff
@@ -25,4 +28,45 @@
...
+
+    describe "with an empty schedule" do
+      it "has a duration of 1 day when no minimum" do
+        expect(reservation.duration_mins.minutes).to eq 1.day
+      end
+
+      describe "reservation duration" do
+        before { instrument.update(min_reserve_days: 4) }
+
+        it "has a correctly duration in minutes" do
```
diebas, line 49: *has a correct duration in minutes

spec/services/next_available_reservation_finder_spec.rb
```diff
@@ -25,4 +28,45 @@
...
+          expect(reservation.duration_mins.minutes).to eq instrument.min_reserve_days.days
+        end
+
+        it "has a correctly duration in minutes" do
```
diebas, line 53: *has a correct duration in days

diebas (commented):
Good job! added a few comments

spec/models/instrument_spec.rb
joaquinco: good catch, I should have used it

diebas: Great, same in line 986

diebas (approved):

spec/models/instrument_spec.rb
```diff
@@ -521,16 +521,55 @@
...
-      assert_equal (Time.zone.now + 1.day).day, @next_reservation.reserve_start_at.day
+      assert_equal 1.day.from_now.day, @next_reservation.reserve_start_at.day
       assert_equal 10, @next_reservation.reserve_start_at.hour
     end
   end
 
+  context "next available reservation with daily booking" do
+    let!(:instrument) do
+      create(
+        :setup_instrument,
```
LeticiaErrandonea, line 532: Can you also use `:always_available` here?

spec/services/next_available_reservation_finder_spec.rb
```diff
@@ -25,4 +28,45 @@
...
+        it "has a correct duration in days" do
+          expect(reservation.duration_days).to eq instrument.min_reserve_days
+        end
+      end
+
+      it "returns a schedule as soon as possible" do
+        next_available_time = 1.minute.from_now.then do |time|
+          delta = time.min % instrument.reserve_interval
+          time - delta.minutes + instrument.reserve_interval.minutes
+        end
```
LeticiaErrandonea, line 62: Can you add a comment explaining what's this doing? 

From what I understand, it's trying to get the right minute to start the reservation (ie: 15:05 if current time is 15:02). In that case, I don't think we're supposed to have that restriction for daily booking. Is allowing reservation to start at 15:02 harder?

LeticiaErrandonea (commented):
Left some questions

app/models/instrument.rb
```diff
@@ -14,6 +16,7 @@ module Pricing
   include EmailListAttribute
 
   RESERVE_INTERVALS = [1, 5, 10, 15, 30, 60].freeze
+  RESERVE_INTERVAL_DAILY = 5
```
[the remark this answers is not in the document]
joaquinco: This also used on the reservation page to populate the start time minutes select (using a default of 5) and by the logic in charge of finding a next reservation slot.

I guess I can just leave the value to nil here and add take a default of 5 when finding a next available slot. Or be able to select any minute when the instrument is daily rated. Do you have any thoughts?

spec/services/next_available_reservation_finder_spec.rb
joaquinco: Following the conversation from the other comment, I'm going to leave the reserve_interval to nil and use 1 as default interval for daily reservations if required

app/helpers/date_helper.rb
```diff
@@ -102,7 +102,7 @@ def join_time_select_values(values)
   private
 
   def minute_options(step = nil)
-    step ||= 5
+    step ||= 1
```
LeticiaErrandonea, line 105: Should we make the default depend on the instrument pricing mode? I think for the non-Daily booking instruments it should be 5

LeticiaErrandonea (commented):
Left one more small comment/question

app/helpers/date_helper.rb
joaquinco: We could add it elsewhere so this helper does not depend on instruments. I took that in mind but since non daily booking instruments have the field reserve_interval as required then this value should be set for those

LeticiaErrandonea: This is only used in the view, right? In that case, I think it's better the solution you had before, setting it in the model callback. Sorry I didn't understand it properly the first time

LeticiaErrandonea (approved):
LGTM!

Thanks for the patience and changes!

PR #4724, opened 2024-11-05
[#162415][Daily Rate] Charge for daily rated instruments

joaquinco:
# Notes

- Refactor `PricePolicies::TimeBasedPriceCalculator` by splitting the logic based on the pricing strategy
- Fix formatting on spec file
- Add tests for daily rate pricing calculation

### DIFF
--- a/spec/services/price_policies/time_based_price_calculator_spec.rb
+++ b/spec/services/price_policies/time_based_price_calculator_spec.rb
@@ -3,17 +3,18 @@
 require "rails_helper"
 
 RSpec.describe PricePolicies::TimeBasedPriceCalculator do
-
   let(:calculator) { described_class.new(price_policy) }
-  let(:price_policy) { build_stubbed(:instrument_price_policy, options.merge(product: product, price_group: price_group)) }
+  let(:price_policy) do
+    build_stubbed(:instrument_price_policy, options.merge(product:, price_group:))
+  end
   let(:options) { {} }
 
   context "when product has schedule rules pricing mode" do
     let(:product) { create(:setup_instrument, skip_schedule_rules: true) }
     let(:price_group) { create(:price_group) }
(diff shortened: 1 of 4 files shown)


spec/services/price_policies/time_based_price_calculator_spec.rb
```diff
@@ -225,6 +225,14 @@
...
             is_expected.to eq(cost: 120, subsidy: 0)
           end
+
+          it "calls the correct strategy" do
+            expect_any_instance_of(
+              PricePolicies::Strategy::PerMinute
+            ).to receive(:calculate)
+
+            subject
+          end
```
LeticiaErrandonea, line 235: Given in line 169 we're defining `subject` using `PricePolicies::Strategy::PerMinute`, is this spec useful?

spec/services/price_policies/time_based_price_calculator_spec.rb
```diff
@@ -328,4 +354,51 @@
       end
     end
   end
+
+  context "when instrument has daily rate pricing" do
```
LeticiaErrandonea, line 358: Can we add a spec for edge cases? Even when it should be the same, I think it's worth it to test a reservation that starts at 00:00

LeticiaErrandonea (commented):
Looking good!

Left some comments/questions for specs

spec/services/price_policies/time_based_price_calculator_spec.rb
joaquinco: Line 169 is just defining a let like declaration named `discount`, I had to change that due to the refactor

spec/services/price_policies/time_based_price_calculator_spec.rb
joaquinco: Sure

LeticiaErrandonea (approved):
LGTM!

Thanks for the changes!

PR #4725, opened 2024-11-05
[161578] Fix bulk note checkbox dynamic behavior

diebas:
# Release Notes

On the Credit Card Orders Reconciliation page the "Update order" button should only be available if an order is selected. 
The button wasn't behaving as expected because it was turning into active once the "Use bulk note" checkbox was checked despite no orders being selected.

Before:
Screenshot 2024-11-05 at 8 46 43 PM

After:
Screenshot 2024-11-05 at 8 45 35 PM

# Additional Context

The check box form logic was looking for any checkbox in the submit button's parent component to be checked to activate the button. I exclude the bulk note checkbox from this check

### DIFF
--- a/app/assets/javascripts/app/checkbox_form.js
+++ b/app/assets/javascripts/app/checkbox_form.js
@@ -1,6 +1,6 @@
 $(document).ready(function() {
   let submitButton = $(".js--requireValueForSubmit");
-  let checkboxes = submitButton.parents("form").find(":checkbox");
+  let checkboxes = submitButton.parents("form").find(":checkbox").not('[name="bulk_note_checkbox"]');
   let confirmationMessageEle = document.querySelector(".js--confirmationMessage");
   let reconcileAll = document.querySelector("#js--reconcileAll");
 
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/app/checkbox_form.js
```diff
@@ -1,6 +1,6 @@
 $(document).ready(function() {
   let submitButton = $(".js--requireValueForSubmit");
-  let checkboxes = submitButton.parents("form").find(":checkbox");
+  let checkboxes = submitButton.parents("form").find(":checkbox").not('[name="bulk_note_checkbox"]');
```
LeticiaErrandonea, line 3: What do you think about using a new class for the checkboxes for the rows, and then use that selector? I find it clearer than selecting them all and then exclude one

LeticiaErrandonea (approved):
LGTM!

Left a question/suggestion

LeticiaErrandonea:
@diebas can you please add the ticket number in the title? Thanks!

app/assets/javascripts/app/checkbox_form.js
diebas: Well, it's something that came into my mind, but this is used on another view. I decided to do it this way as it seemed a little intrusive to change other views just to make this one work. But I'm ok with both approaches.

joaquinco (approved):

PR #4732, opened 2024-11-07
[#162474][Daily Rate] Handle reservations that end at beginning of day

joaquinco:
# Notes

- Fix reservation checks on schedule rules when a reservation ends exactly at midnight (12:00 AM)
- Add tests to move a reservation next to unavailable day and next to another reservation when the availability ends at midnight

### DIFF
--- a/app/models/schedule_rule.rb
+++ b/app/models/schedule_rule.rb
@@ -119,8 +119,8 @@ class ScheduleRule < ApplicationRecord
   end
 
   def cover?(dt)
-    # we check the previous minute in case dt is exactly the
-    # beginning of a day in case it's the end of a reservation
+    # check the previous minute in case dt is exactly the
+    # beginning of a day and it's the end of a reservation
     # (if it's not then the next minuted checked will be uncovered).
     dt_int = dt.hour * 100 + dt.min
     dt_day = if dt_int.zero?
--- a/app/models/schedule_rule.rb
+++ b/app/models/schedule_rule.rb
@@ -119,9 +119,18 @@ class ScheduleRule < ApplicationRecord
   end
 
   def cover?(dt)
-    return false unless on_day?(dt)
-
+    # we check the previous minute in case dt is exactly the
+    # beginning of a day in case it's the end of a reservation
+    # (if it's not then the next minuted checked will be uncovered).
     dt_int = dt.hour * 100 + dt.min
+    dt_day = if dt_int.zero?
+               dt - 1.minute
+             else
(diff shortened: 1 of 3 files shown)


spec/system/editing_a_reservation_spec.rb
```diff
@@ -91,4 +91,96 @@
...
+          :purchased_reservation,
+          user:,
+          reserve_start_at: reserve_end_at - 1.day,
+          reserve_end_at:,
+          product: instrument
+        )
+
+        expect(invalid_reservation).to_not be_valid
+        expect(invalid_reservation.errors.map(&:type)).to include(error_key)
+      end
```
LeticiaErrandonea, line 137: Sorry, is this case when a reservation ends on an unavailable day? Isn't that case supposed to be valid or are we leaving that for the next ticket?

app/models/schedule_rule.rb
```diff
@@ -119,9 +119,18 @@ def on_day?(datetime)
   end
 
   def cover?(dt)
-    return false unless on_day?(dt)
-
+    # check the previous minute in case dt is exactly the
+    # beginning of a day and it's the end of a reservation
+    # (if it's not then the next minuted checked will be uncovered).
```
LeticiaErrandonea, line 124:
```suggestion
    # (if it's not then the next minute checked will be uncovered).
```

LeticiaErrandonea (commented):
Left a question

spec/system/editing_a_reservation_spec.rb
joaquinco: this is just checking that the unavailable day is actually unavailable due the reason specified on `error_key`

spec/system/editing_a_reservation_spec.rb
```diff
@@ -91,4 +91,98 @@
...
```
LeticiaErrandonea, line 137: Sorry again! 

`instrument` isn't available on `unavailable_beginning_of_day`. This reservation starts the day before at 00:01 and end on `unavailable_beginning_of_day` at 00:01, right?

LeticiaErrandonea (approved):
After talking about this sync, I understand the problem behind it and I'm fine with these changes.

I'd like to have @giladshanan's input in case "it's not a bug but a feature" 😄

joaquinco:
This change instruduces an error on another edge case which is kind of the oposite of this one: create a reservation starting at beginning of dat after an unavailable day. Closing this an created issue #162494

PR #4734, opened 2024-11-07
[#162413][Daily Rate] Monthly calendar UI for daily booking instruments

joaquinco:
# Notes

Render monthly view for daily booking instrument's reservation page

## Additional Context

- Create class `CalendarEventsPresenter` in charge of rendering calendar events for reservations and schedule rules
- Move `ScheduleRuleCalendarPresenter` to `ScheduleRules::CalendarPresenter` and add the possibility to handle month calendar view, which for now differs on how unavailable days are rendered as a full gray box.
- Allow full calendar init to receive options through html element data attributes
- Only show monthly view for daily booking instruments' schedule (Product Schedule admin tab)
- Use `as_json` instead of `to_json` for calendar event presenters which better follows rails semantics (return Hash or Array of hashes vs json formatted string)

## Testing Notes

Go to the reservation page of a daily booking instrument.
The calendar is not editable.
When changing the month on the start date the calendar renders the selected month events.

## Screenshot

Captura desde 2024-11-07 16-28-58


app/assets/javascripts/app/calendar.js
```diff
@@ -49,6 +54,26 @@ window.FullCalendarConfig = class FullCalendarConfig {
...
+    let self = this;
+    allowedKeys.forEach(function(key) {
+      let value = self.$element.data(key)
+      if (value) {
+        ret[key] = value;
+      }
+    });
+
+    console.log(this.$element.data());
+    console.log(ret);
```
LeticiaErrandonea, line 73: nit: stray `console.log`s

app/assets/javascripts/instruments.js
```diff
@@ -1,9 +1,15 @@
 $(document).ready(function() {
-  new FullCalendarConfig($("#calendar"), {
-    header: {
-      left: 'title',
-      center: '',
-      right: 'prev,next today agendaDay,agendaWeek,month',
-    }
-  }).init();
+  const calendar = $("#calendar");
+  let defaultView = calendar.data('defaultView');
```
LeticiaErrandonea, line 3: nit: I think this can be a `const`

app/assets/javascripts/instruments.js
```diff
@@ -1,9 +1,15 @@
...
+
+  const header = { left: 'title', center: '', right: '' };
+
+  if (defaultView != 'month') {
+    header.right = 'prev,next today agendaDay,agendaWeek,month';
+  }
+
+  new FullCalendarConfig(
+    calendar,
+    { header: header }
```
LeticiaErrandonea, line 13: nit: 
```suggestion
    { header }
```

app/controllers/reservations_controller.rb
```diff
@@ -16,6 +16,8 @@ class ReservationsController < ApplicationController
   include ReservationSwitch
   helper TimelineHelper
 
+  MAX_RESERVATIONS_PERIOD = 40.days
```
LeticiaErrandonea, line 19: Was this number already used somewhere?

app/helpers/calendar_helper.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module CalendarHelper
+
+  def product_calendar_view(product)
+    return "month" if product.daily_booking?
+
+    "agendaWeek"
```
LeticiaErrandonea, line 8: nit: in this cases I usually go for a regular if-else or even a ternary. When it's just to return a value, I find any of them more readable than this pattern

LeticiaErrandonea (commented):
Left a few comments. Specs are failing so I didn't go in depth

app/controllers/reservations_controller.rb
joaquinco: Nope, but if it the index endpoint is only used for calendars it should be enough

LeticiaErrandonea:
I did some Dev QA and I noticed something that I think is part of the Schedule Rule ticket, but just in case: it'd be better to have this view display the monthly calendar. Bringing this up because PR updates calendars
image

Other than that, it seemed to work fine, for both Daily booking and existing instruments

joaquinco:
Thanks for the QA.

I thought about changing the schedule rules calendar view but since they are defined for each day of the week I though they were easier to visualize the way it was already done.

LeticiaErrandonea (approved):
LGTM!

PR #4759, opened 2024-11-13
Fix scanning-in-and-out-spec failing test

diebas:
# Release Notes

A small operator change in the time_based_price_calculator resulted in failing tests on some repos as it was skipping the calculation when the start date and the end date where the same.


LeticiaErrandonea (approved):
Pending specs

PR #4762, opened 2024-11-14
[#162413][Daily rate] Instrument open hours

joaquinco:
## Notes

Add table with information about what time the instrument is available for daily rate instruments since we don't show this information in the calendar month view.

Text labels and table styles and position are open to suggestions.

## Screenshots

Phone viewport

Captura desde 2024-11-15 10-31-53

Desktop viewport

Captura desde 2024-11-15 10-31-38


app/support/schedule_rules/open_hours.rb
```diff
@@ -0,0 +1,60 @@
...
+
+          merged_rule_times(wday_rules).map do |time|
+            { weekday:, range: self.class.time_range(time[:start], time[:end]) }
+          end
+        end
+    end
+
+    private
+
+    def merged_rule_times(rules)
```
LeticiaErrandonea, line 43: Could you add a comment explaining what this is doing and what's the shape of the returned value?

spec/services/schedule_rules/open_hours_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+          end_hour: 1,
+          end_min: 0,
+        )
+      end
+
+      it "handles discontinuous rules correctly" do
+        expect(subject["Mon"]).to eq(
+          product.schedule_rules.order(:start_hour).map do |sr|
+            described_class.time_range(sr.start_time, sr.end_time)
+          end.join(", ")
```
LeticiaErrandonea, line 59: I wonder if this specs should just use the string we expect to have. Give we know the schedule rules, if we use a string it should be easier to the  next devs to understand what's expected here

spec/support/shared_examples/reservations_new_daily.rb
```diff
@@ -23,6 +23,8 @@
     expect(page).to have_content("Create Reservation")
     expect(page).to have_content("Duration days")
 
+    expect(page).to have_xpath("//table/tbody/tr/th[text()='Start time availability']")
```
LeticiaErrandonea, line 26: Can you please also add some accessibility spec?

LeticiaErrandonea (commented):
I like how this looks!

Left some comments

app/views/reservations/_open_hours.html.haml
```diff
@@ -0,0 +1,25 @@
+.hidden-phone
+  %table.table.table-striped.table-bordered
+    %tbody
+      %tr
+        %th
+          %span Day of Week
```
LeticiaErrandonea, line 6: Can you please use translations in this file?

spec/services/schedule_rules/open_hours_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    before do
+      create(
+        :schedule_rule,
+        :weekday,
+        product:,
+        start_hour: 9,
+        start_min: 0,
+        end_hour: 17,
+        end_min: 0,
```
LeticiaErrandonea, line 31: I think we don't need this as these are the default values

LeticiaErrandonea (commented):
Left a few more comments that I forgot about the last time

spec/services/schedule_rules/open_hours_spec.rb
joaquinco: I think it make sense to have these set if using a string literal as expectation value

app/support/schedule_rules/open_hours.rb
LeticiaErrandonea: Hey @joaquinco you reacted with a 👍 but I think you forgot to make the changes 😄

app/views/reservations/_open_hours.html.haml
```diff
@@ -0,0 +1,23 @@
+.hidden-phone
```
LeticiaErrandonea, line 1: I feel we have a lot of repeated code in this file but I don't see an obvious way of avoiding repetition. Leaving a comment in case the next reviewer can think of something

LeticiaErrandonea (approved):
Thanks for the changes!

I left two more comments, but nothing blocking

app/views/reservations/_open_hours.html.haml
joaquinco: The other option is to do something like this https://stackoverflow.com/a/53788026 although the result is not exactly the same with bootstrap styles

app/support/schedule_rules/open_hours.rb
joaquinco: I think you need to check the latest changes

app/views/reservations/_open_hours.html.haml
joaquinco: i ended up doing what's on the stackoverflow post

joaquinco: And now I reverted because it does not look good in safari nor firefox 🥲

PR #4775, opened 2024-11-20
[#162414] Show admin holds as unavailable

joaquinco:
## Note

Show admin reservations and offline reservations as unavailable in the calendar (with the unavailable css class). I've also refactored the reservation's calendar presenter and added some tests.

> Even though the ticket is related to daily booking, the change applies to all admin holds.

## Screenshots

Captura desde 2024-11-20 11-06-30

### DIFF
--- a/app/presenters/reservations/calendar_presenter.rb
+++ b/app/presenters/reservations/calendar_presenter.rb
@@ -3,42 +3,7 @@
 module Reservations
 
   class CalendarPresenter < DelegateClass(Reservation)
-
-    def as_calendar_object(options = {})
-      calendar_object_default.merge(reservation_options(options))
-    end
-
-    private
-
-    def reservation_options(options)
-      if order.present?
-        if options[:with_details].present?
-          {
-            title: order.user.full_name,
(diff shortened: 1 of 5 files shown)


app/presenters/reservations/calendar_presenter.rb
```diff
@@ -49,6 +14,53 @@ def calendar_object_default
...
+        email: created_by.try(:full_name),
+      )
+    end
   end
 
+  class OrderCalendarPresenter < CalendarPresenter
+    def as_calendar_object(options = {})
+      ret = super
+
+      return (ret = super) if options[:with_details].blank?
```
LeticiaErrandonea, line 42: Can we just return `ret`?

LeticiaErrandonea (approved):
LGTM!

Left a question

app/presenters/reservations/calendar_presenter.rb
joaquinco: yeap, i guess i tried to avoid the method's first line but then forgot about it 🐟

PR #4782, opened 2024-11-22
[#162490] Fix cutoff hours validation

joaquinco:
## Note

Cutoff hours weren't checked because `admin?` evaluated to true.

### DIFF
--- a/spec/system/purchasing_a_reservation_spec.rb
+++ b/spec/system/purchasing_a_reservation_spec.rb
@@ -264,7 +264,7 @@ RSpec.describe "Purchasing a reservation" do
     end
   end
 
-  context "honours cutoff hours" do
+  describe "honours cutoff hours" do
     let(:reserve_time) { 78.minutes.from_now }
 
     before do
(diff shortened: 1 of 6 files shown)


spec/system/purchasing_a_reservation_spec.rb
```diff
@@ -263,4 +263,25 @@
...
   end
+
+  describe "honours cutoff hours" do
+    let(:reserve_time) { 78.minutes.from_now }
+
+    before do
+      instrument.update!(cutoff_hours: 2)
+    end
+
+    it "cannot create a reservation before cutoff hours" do
```
LeticiaErrandonea, line 274: Should this read "... *after* cutoff hours?"

LeticiaErrandonea (approved):
LGTM!

Left a nit question

spec/system/purchasing_a_reservation_spec.rb
joaquinco: yesss

PR #4796, opened 2024-11-28
[#162417] Fix order details unrecoverable status flow

diebas:
# Release Notes

Ticket: [#162417 | Status in Order Detail page does not show Unrecoverable
](https://pm.tablexi.com/issues/162417?issue_count=45&issue_position=23&next_issue_id=162414&prev_issue_id=162472)

After changing the status of an order to Unrecoverable, then going back to the Order Detail page, the Order Status is displayed as "New".

Added the Unrecoverable status to the order details view dropdown and added permissions to set the status of completed orders to unrecoverable. 

# Screenshot

Screenshot 2024-11-28 at 3 39 25 PM

### DIFF
--- a/app/controllers/order_management/order_details_controller.rb
+++ b/app/controllers/order_management/order_details_controller.rb
@@ -101,7 +101,7 @@ class OrderManagement::OrderDetailsController < ApplicationController
     return if @order_detail.reconciled?
 
     if @order_detail.complete?
-      @order_statuses = [OrderStatus.complete, OrderStatus.canceled]
+      @order_statuses = [OrderStatus.complete, OrderStatus.canceled, OrderStatus.unrecoverable]
       @order_statuses << OrderStatus.reconciled if @order_detail.can_reconcile?
     elsif @order_detail.order_status.root == OrderStatus.canceled
       @order_statuses = OrderStatus.canceled_statuses_for_facility(current_facility)
--- a/app/controllers/order_management/order_details_controller.rb
+++ b/app/controllers/order_management/order_details_controller.rb
@@ -101,8 +101,8 @@ class OrderManagement::OrderDetailsController < ApplicationController
     return if @order_detail.reconciled?
 
     if @order_detail.complete?
-      @order_statuses = [OrderStatus.complete, OrderStatus.canceled, OrderStatus.unrecoverable]
-      @order_statuses << OrderStatus.reconciled if @order_detail.can_reconcile?
+      @order_statuses = OrderStatus.where(name: %w[complete canceled unrecoverable reconciled]).to_a
+      @order_statuses.reject! { |status| status.name == 'reconciled' } unless @order_detail.can_reconcile?
     elsif @order_detail.order_status.root == OrderStatus.canceled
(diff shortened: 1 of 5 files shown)


app/models/order_status.rb
```diff
@@ -122,6 +122,7 @@ def non_protected_statuses(facility)
       new_status.self_and_descendants
         .or(in_process.self_and_descendants)
         .or(canceled.self_and_descendants)
+        .or(unrecoverable.self_and_descendants)
```
LeticiaErrandonea, line 125: Why is this change needed? I see this method is used in a few places and I don't know if there could be any unintended side effects

LeticiaErrandonea (commented):
Looking good.

Left a question and I'll test it tomorrow

app/models/order_status.rb
diebas: On the OrderDetailsController I added the unrecoverable status to the eligible statuses for completed orders so that you can change it to unrecoverable. This is not enough, once the order is set as unrecoverable, if you open the order details view, the unrecoverable status is not visible as it's not a completed order anymore, so it appears with the "new" status. This is why I needed to add the unrecoverable status here too.

diebas: In addition, the controllers use this non_protected_statuses method to set the available status options for an order. We have logic order_detail.rb that defines the permissions needed for an order to be updated to a certain status so that if the Unrecoverable status appears as an option for an order that can't be set as unrecoverable and the user tries to do so, the app won't allow it and an error message will show up.

LeticiaErrandonea (approved):
Tested locally and looks fine!

LGTM, and thanks for the explanation!

app/controllers/order_management/order_details_controller.rb
```diff
@@ -101,7 +101,7 @@ def load_order_statuses
     return if @order_detail.reconciled?
 
     if @order_detail.complete?
-      @order_statuses = [OrderStatus.complete, OrderStatus.canceled]
+      @order_statuses = [OrderStatus.complete, OrderStatus.canceled, OrderStatus.unrecoverable]
```
joaquinco, line 104: This line is making 3 db queries and possibly one more in the following, maybe we can refactor this to make just one. I know this is not actually related to this ticket so it's not blocking at all.

app/controllers/order_management/order_details_controller.rb
```diff
@@ -101,8 +101,8 @@ def load_order_statuses
     return if @order_detail.reconciled?
 
     if @order_detail.complete?
-      @order_statuses = [OrderStatus.complete, OrderStatus.canceled, OrderStatus.unrecoverable]
-      @order_statuses << OrderStatus.reconciled if @order_detail.can_reconcile?
+      @order_statuses = OrderStatus.where(name: %w[complete canceled unrecoverable reconciled]).to_a
```
joaquinco, line 104: Note that this relies on mysql's case insensitive string comparison. I think we should set the status names to constants in the model and use a scope or something.

If you feel this is getting complex let's create another PR to tackle this

joaquinco (approved):

giladshanan (approved):

PR #4817, opened 2024-12-05
[#162477] Add "Order for" in purchase email

joaquinco:
## Notes

Add "Order For" in email with information about the user targeting a reservation when it's donde by someone else.

> Also formatted some files


app/views/reservations/new.html.haml
```diff
@@ -38,7 +38,6 @@
       - if @order.order_details.size == 1
         .span6.send-notification
           = label_tag :send_notification, class: "checkbox" do
-            = hidden_field_tag :send_notification, 0
```
LeticiaErrandonea, line 41: Is this change related to the PR? I noticed you added it in a separate commit but I'm not sure how it impacted the emails

joaquinco: Not really, I was testing this and found this was not working correctly. I added a note on the PR description.

I'll create a separete PR for this

LeticiaErrandonea (approved):
LGTM!

PR #4819, opened 2024-12-06
[#162105] Handle mime type errors

diebas:
# Release Notes

# Screenshot

Optional.

# Additional Context

# Accessibility


config/environments/development.rb
```diff
@@ -12,7 +12,7 @@
   config.eager_load = true
 
   # Show full error reports.
-  config.consider_all_requests_local = true
+  config.consider_all_requests_local = false
```
joaquinco, line 15: can we leave this flag on? it's useful to have this in development

app/controllers/application_controller.rb
```diff
@@ -118,36 +118,20 @@ def acting_as?
...
+      Rails.logger.debug("#{exception.message}: #{exception.backtrace.join("\n")}")
+      render_404
     end
-  end
 
-  rescue_from NUCore::NotPermittedWhileActingAs, with: :render_acting_error
-  def render_acting_error
-    render "/acting_error", status: 403, layout: "application", formats: formats_with_html_fallback
+    rescue_from NUCore::PermissionDenied, CanCan::AccessDenied, with: :render_403
+    rescue_from NUCore::NotPermittedWhileActingAs, with: :render_acting_error
```
joaquinco, line 134: We should leave this rescue_from for all environments

joaquinco (commented):
I'm not seeing how exceptions are mapped to errors for environments other than test. It would be nice to have a single solution for all environments btw

I might be wrong but I think we are still not catching the mime type error with this approach

app/controllers/application_controller.rb
```diff
@@ -208,4 +193,23 @@ def formats_with_html_fallback
     request.formats.map(&:ref) + [:html]
   end
 
+  def render_403(_exception)
+    byebug
```
giladshanan, line 197: I don't think this should be here - do we have a spec that would catch something like this?

app/controllers/application_controller.rb
```diff
@@ -118,36 +118,21 @@ def acting_as?
...
-  end
-
-  def render_404(_exception)
-    # Add html fallback in case the 404 is a PDF or XML so the view can be found
-    render "/404", status: 404, layout: "application", formats: formats_with_html_fallback
-  end
+  rescue_from NUCore::NotPermittedWhileActingAs, with: :render_acting_error
+  
+  # Test exception handlers
+  if Rails.env.test?
```
giladshanan, line 124: Is there a way to test without this?  I usually try to avoid changing the application code to make testing work

lib/middlewares/sanitize_headers_middleware.rb
```diff
@@ -0,0 +1,11 @@
+class SanitizeHeadersMiddleware
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    env["HTTP_ACCEPT"] = "*/*" if env["HTTP_ACCEPT"] =~ /(\.\.|{|})/
```
joaquinco, line 7: I think we should let rails decide which types are invalid and set the mime type to `text/html` as fallback, like in the `fallback_to_html_format_if_invalid_mime_type` function here: https://guides.rubyonrails.org/configuring.html#config-exceptions-app.

config/application.rb
```diff
@@ -54,6 +55,9 @@ class Application < Rails::Application
     # config.autoload_paths += Dir["#{config.root}/lib"]
     # config.eager_load_paths += Dir["#{config.root}/lib"]
 
+    config.autoload_paths << Rails.root.join("lib/middlewares")
+    config.eager_load_paths << Rails.root.join("lib/middlewares")
```
joaquinco, line 59: I'm not sure if this is needed since you're requiring the file on line 8

joaquinco (approved):
💯

LeticiaErrandonea (approved):
LGTM! Thanks for the changes!

PR #4863, opened 2024-12-18
[NUOPEN-6][Shared dev] Daily rate with fixed start time: create a reservation

joaquinco:
## Note

- Add `Instrument#fixed_start_time` boolean to tell if daily reservations should start at the beginning of the day
- Add system tests for instrument setting the attribute and reservation creation
- Don't render start time from reservation form
- Don't render Start time availability table if fixed_start_time is on


LeticiaErrandonea (commented):
I wonder if the name should be different. We have the help text in the form but the label is "Fixed Start Time" which is not so self explanatory to me. Is the same for the field in the DB. Maybe something like "start_reservations_at_midnight"?

joaquinco:
@LeticiaErrandonea What about `reservation_start_time_disabled` or just `start_time_disabled`?

LeticiaErrandonea (approved):
I think unrelated specs are failing

PR #4874, opened 2024-12-23
[NUOPEN-6][Daily Rate] Update schedule rules

joaquinco:
## Notes

- When an instrument has `start_time_disabled` on the schedule rules have full day availability so time inputs are disabled and set the time range 00:00 - 00:01
- When an existing instrument has the flag `start_time_disabled` turned on, the schedule rule's time are updated to have full availability
- Do not render the calendar ui in schedule rules page if `start_time_disabled` is on
- Set fullcalendar height to `auto` so it adjusts automatically instead of doing it ourselves

## Screenshots

Only show start time for schedule rules, hide calendar view:
Captura de pantalla 2024-12-23 a la(s) 16 25 14

Do not show time input and add note on new/edit:
Captura de pantalla 2024-12-23 a la(s) 16 25 28


app/helpers/date_helper.rb
```diff
@@ -90,8 +90,11 @@ def time_select(f, field, options_tag = {}, html_options = {})
 
   def time_select24(f, field, options = {})
     options.reverse_merge! hours: (0..23)
-    output =  f.select(:"#{field}_hour", options[:hours].to_a)
-    output += f.select(:"#{field}_min", minute_options(options[:minute_step]))
+    hours = options.delete(:hours).to_a
+    minute_step = options.delete(:minute_step)
+
+    output = f.select(:"#{field}_hour", hours, {}, options)
+    output += f.select(:"#{field}_min", minute_options(minute_step), {}, options)
```
LeticiaErrandonea, line 97: Why are these changes needed? Was it a problem to use `options` as it were?

config/locales/views/en.schedule_rules.yml
```diff
@@ -3,6 +3,7 @@ en:
     schedule_rules:
       discount: Price Group Discounts
       discount_hint: During the period covered by this schedule rule, overall cost will be discounted by the following percentages.
+      start_time_disabled_help: This instrument can only be reserved on full days
```
LeticiaErrandonea, line 6: I'm not sure about the wording.

Maybe:

```suggestion
      start_time_disabled_help: This instrument can only be reserved for complete days
```

I'm not sure about this one either

spec/system/admin/editing_an_instrument_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+      expect(page).to have_content(instrument.name)
+
+      fill_in("Minimum (days)", with: "2")
+
+      click_button("Save")
+
+      expect(page).to have_content("Instrument was successfully updated")
+    end
+
+    context "switching start_time_disabled on" do
```
LeticiaErrandonea, line 27: What happens with reservations from before this change? I guess they'll work unless they want to update them. Should we check this with clients?

LeticiaErrandonea (commented):
Looking good!

Left a few questions

app/helpers/date_helper.rb
joaquinco: This was needed to be able to disable the input with a `disabled:` argument, but I ended up not using it so I'll revert this change

spec/system/admin/editing_an_instrument_spec.rb
LeticiaErrandonea: I think that's fine, but may be worth it to add a testing note about it

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

PR #4925, opened 2025-01-15
[OSU-114] Setup sanger enabled service

joaquinco:
## Notes

- Add `sanger_sequencing_enabled` to product
- Allow admins to switch that flag on for Services
- When the flag is switched on an ExternalService pointing to Sanger is created if it does not exists

## Screenshots

Captura de pantalla 2025-01-15 a la(s) 09 50 05

### DIFF
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -41,7 +41,7 @@ class ServicesController < ProductsCommonController
     if sanger_external_service.blank?
       Service.transaction do
         external_service = UrlService.find_or_create_by(
-          location: new_sanger_sequencing_submission_path
+          location: new_sanger_sequencing_submission_url
         )
         ExternalServicePasser.create!(
           external_service:,
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -30,6 +30,8 @@ class ServicesController < ProductsCommonController
   # Ensures UrlService exists for the product
   # pointing to Sanger Submission
   def ensure_sanger_url_service
+    return unless defined? new_sanger_sequencing_submission_path
+
     sanger_external_service =
       @product
       .external_services
--- a/app/controllers/services_controller.rb
+++ b/app/controllers/services_controller.rb
@@ -1,4 +1,51 @@
 # frozen_string_literal: true
 
 class ServicesController < ProductsCommonController
+  after_action :update_sanger_external_service, only: [:create, :update]
+
+  private
+
+  def permitted_params
+    params = super
+
+    if current_facility.sanger_sequencing_enabled?
+      params += %i[sanger_sequencing_enabled]
+    end
+
+    params
+  end
+
+  def update_sanger_external_service
+    return unless @product.sanger_sequencing_enabled_previously_changed?
+
+    if @product.sanger_sequencing_enabled?
+      ensure_sanger_url_service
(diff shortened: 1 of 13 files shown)


diebas (approved):

db/migrate/20250110132703_add_product_sanger_sequencing_enabled.rb
```diff
@@ -0,0 +1,5 @@
+class AddProductSangerSequencingEnabled < ActiveRecord::Migration[7.0]
```
LeticiaErrandonea, line 1: Can you add the `# frozen_string_literal: true` comment?

app/controllers/services_controller.rb
```diff
@@ -1,4 +1,51 @@
...
+    sanger_external_service =
+      @product
+      .external_services
+      .matching_location(new_sanger_sequencing_submission_path)
+      .first
+
+    if sanger_external_service.blank?
+      Service.transaction do
+        external_service = UrlService.find_or_create_by(
+          location: new_sanger_sequencing_submission_path
```
LeticiaErrandonea, line 44: While this should still work, I think it's better to use the `url`, so the users see the complete URL (including domain)

config/locales/en.controllers.yml
```diff
@@ -214,6 +214,10 @@ en:
         error: "Our apologies, but an error occurred while importing: %{error}"
         job_is_queued: "The bulk import is being processed. A report will be sent to %{email} when complete."
 
+    services:
+      sanger_sequencing_enabled: Sanger has been enabled, make sure the Order Form is inactive
+      sanger_sequencing_disabled: Sanger has been disabled, make sure the Order Form is active
```
LeticiaErrandonea, line 219: nit: should we add the period at the end of the messages?

spec/system/admin/creating_a_service_spec.rb
```diff
@@ -62,4 +62,34 @@
   context "when billing mode is Skip Review" do
     include_examples "creates a product with billing mode", "service", "Skip Review"
   end
+
+  context "when sanger enable si checked" do
```
LeticiaErrandonea, line 59: typo: _si_ -> _is_

spec/system/admin/editing_a_service_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+
+    context "when facility is sanger enabled" do
+      before do
+        facility.update(sanger_sequencing_enabled: true)
+      end
+
+      it "can enable sanger for the service" do
+        visit edit_facility_service_path(facility, service)
+
+        check "service[sanger_sequencing_enabled]"
```
LeticiaErrandonea, line 41: Wasn't `sanger_sequencing_enabled` already checked?

LeticiaErrandonea (commented):
Left a few comments/questions

spec/system/admin/editing_a_service_spec.rb
joaquinco: in the before hook it's enabled for the facility, not the service

app/controllers/services_controller.rb
joaquinco: Yeap this work. The only reason I can think of for having the domain for internal routes is end users being accustomed to see it. I'll change it anyway

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

PR #4945, opened 2025-01-21
[OSU-116] Submission with primers

joaquinco:
## Note

- Allow to add sample primers if the sanger service allows it
- Primer Name input have autocomplete like dropdown for core primers but the allow any text
- There's a button to copy primer name for every row below which works kind of the same as the primer input in the acgt module

## Screenshots

Captura de pantalla 2025-01-21 a la(s) 15 43 01


vendor/engines/sanger_sequencing/config/locales/en.primers.yml
```diff
@@ -0,0 +1,7 @@
+en:
+  sanger_sequencing:
+    primer:
+      default_list:
+        - M13FM13RT7
+        - PromoterT7
+        - Terminator
```
LeticiaErrandonea, line 7: It may be a good idea to confirm these primers should be the default ones. I know you said these are the ones you saw, is it for both Dartmouth and OSU?

vendor/engines/sanger_sequencing/app/assets/javascripts/sanger_sequencing/submission.js
```diff
@@ -3,14 +3,84 @@
...
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/main/docs/suggestions.md
  */
+
+//= require jquery-ui/widgets/autocomplete
+
+const SubmissionPrimer = {
+  handleCopy(event) {
+    const row = $(event.target).parents("tr");
+
+    if (row.length == 0) { return; }
```
LeticiaErrandonea, line 13: nit: `===`

vendor/engines/sanger_sequencing/db/migrate/20250120141340_add_sample_primer_name.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddSamplePrimerName < ActiveRecord::Migration[4.2]
```
LeticiaErrandonea, line 3: Is there any reason to use `Migration[4.2]`? I see we have a bunch of different versions. For instance, I see we used 7.0 for `AddProductStartTimeDisabled` a month ago

vendor/engines/sanger_sequencing/db/migrate/20250120141340_add_sample_primer_name.rb
```diff
@@ -0,0 +1,15 @@
...
+  def up
+    unless ActiveRecord::Base.connection.column_exists?(:sanger_sequencing_samples, :primer_name)
```
LeticiaErrandonea, line 5: Why do we need this check in order to fix Oracle? How did it break that DB before?

LeticiaErrandonea (commented):
I left a few comments/questions that are relevant whether we keep this implementation or change it

vendor/engines/sanger_sequencing/db/migrate/20250120141340_add_sample_primer_name.rb
joaquinco: I was using the same version number as previous migrations (for this engine). This might be the reason why if_no_exists parameter didn't work and I had to manually check the column existance

vendor/engines/sanger_sequencing/db/migrate/20250120141340_add_sample_primer_name.rb
joaquinco: Initially I was using the argument `if_not_exsits: true` so it does not raise an error if this column exists already (because it was added by the acgt module). When I tried this on NU it raised the error anyway, my guess is that the oracle adapter doesn't handle this flag, but didn't research much about it and I found no related issues

vendor/engines/sanger_sequencing/config/locales/en.primers.yml
joaquinco: I'll ask

LeticiaErrandonea (approved):
This is great, LGTM!

PR #4957, opened 2025-01-24
[Fix] Acting as error handling

joaquinco:
## Notes

Acting as error handling is raising template not found error.


LeticiaErrandonea (approved):
Good catch!

PR #4958, opened 2025-01-24
[OSU-121] Manage Primers per Service

joaquinco:
## Notes

- Allow admins to manage "Core Primers" per service
- Show "Core Primers" from the db when creating a submission that accepts primers for samples

## Screenhots

Captura de pantalla 2025-01-27 a la(s) 18 41 11

Captura de pantalla 2025-01-30 a la(s) 10 10 56


vendor/engines/sanger_sequencing/app/models/sanger_sequencing/primer.rb
```diff
@@ -2,7 +2,14 @@
 
 module SangerSequencing
 
-  class Primer
+  class Primer < ApplicationRecord
+    self.table_name = "sanger_sequencing_primers"
+
+    belongs_to :sanger_product
```
LeticiaErrandonea, line 8: Why use a `belongs_to` instead of `has_many through`? 

My guess is we'll have the same primers many times in the DB, which won't be a problem because of the volume but we could have unnecessary repetition in our records.

That makes me wonder whether we should use an input like the one you created for the user to select a primer (or enter a new one) that accepts multi selection. That way users don't have to enter the same primer multiple times and only have one record per primer. Does it make sense?

LeticiaErrandonea (commented):
Left a question about the solution

vendor/engines/sanger_sequencing/app/models/sanger_sequencing/primer.rb
joaquinco: yeap makes sense

LeticiaErrandonea (commented):
I think the user experience would be better if we used Chosen for the new input, just like we do for Search inputs (ie: http://localhost:3000/facilities/example/transactions)

vendor/engines/sanger_sequencing/lib/tasks/sanger_sequencing.rake
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+namespace :sanger_sequencing do
```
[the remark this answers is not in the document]
joaquinco: sure

PR #4984, opened 2025-02-04
[OSU-123] Fix batch group filtering

joaquinco:
## Notes

- Use the same semantics as submissions group filtering, that is, nil or empty treated the same as default group
- In the Sanger navtab partial, render batch link for each group instead of hardcoding default and fragment groups links


diebas (approved):

LeticiaErrandonea (approved):
LGTM!

PR #5034, opened 2025-02-20
Update sanger sequencing csv for alternative machine

diebas:
# Release Notes

Add some changes (under a FF) to the csv sanger file generation to work on a different machine.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -163,6 +163,7 @@ feature:
   show_reconciliation_deposit_number: <%= ENV.fetch("RECONCILIATION_DEPOSIT_NUMBER", false) %>
   show_daily_rate_option: true
   sanger_enabled_service: <%= ENV.fetch("SANGER_ENABLED_SERVICE", false) %>
+  alternative_csv_format: <%= ENV.fetch("ALTERNATIVE_CSV_FORMAT", false) %>
 
 split_accounts:
   # Roles are allowed to create Split Accounts
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -163,7 +163,7 @@ feature:
   show_reconciliation_deposit_number: <%= ENV.fetch("RECONCILIATION_DEPOSIT_NUMBER", false) %>
   show_daily_rate_option: true
   sanger_enabled_service: <%= ENV.fetch("SANGER_ENABLED_SERVICE", false) %>
-  alternative_csv_format: <%= ENV.fetch("ALTERNATIVE_CSV_FORMAT", false) %>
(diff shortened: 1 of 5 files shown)


vendor/engines/sanger_sequencing/app/presenters/sanger_sequencing/well_plate_presenter.rb
```diff
@@ -58,8 +61,13 @@ def sorted_cells
       end
     end
 
-    def sample_row(sample)
-      [sample.id.to_s, sample.customer_sample_id.to_s] + additional_columns.values
+    def sample_row(sample, position)
+      if position == "A1" && sample.reserved?
```
LeticiaErrandonea, line 65: Shouldn't this also be behind the check for the new FF?

LeticiaErrandonea (commented):
Left a comment. Also, specs are failing

vendor/engines/sanger_sequencing/spec/presenters/sanger_sequencing/well_plate_presenter_spec.rb
```diff
@@ -59,5 +63,26 @@
         expect(sample_rows[8]).to match(["A03", samples[7].id.to_s, samples[7].customer_sample_id.to_s] + extra_columns)
       end
     end
+
+    context "when alternative_csv_format is enabled", feature_setting: { alternative_csv_format: true } do
+      before do
+        allow(SettingsHelper).to receive(:feature_on?).and_return(true)
```
LeticiaErrandonea, line 69: We shouldn't need to do this, just `feature_setting` should be enough. Was the Feature Flag not set?

config/settings.yml
```diff
@@ -163,6 +163,7 @@ feature:
   show_reconciliation_deposit_number: <%= ENV.fetch("RECONCILIATION_DEPOSIT_NUMBER", false) %>
   show_daily_rate_option: true
   sanger_enabled_service: <%= ENV.fetch("SANGER_ENABLED_SERVICE", false) %>
+  alternative_csv_format: <%= ENV.fetch("ALTERNATIVE_CSV_FORMAT", false) %>
```
LeticiaErrandonea, line 166: nit: maybe we can try using a different name that explains where it's used? Something like `well_plate_alternative_csv_format`?

LeticiaErrandonea (commented):
Left a couple of questions.

As a general comment, I thought we were going to use the school engine to add the new values, instead of changing the yml for Sanger engine. Is there any reason you preferred to go this way?

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

PR #5062, opened 2025-02-28
[UMASS-175] Allow Global Billing admins to update actual cost

LeticiaErrandonea:
# Release Notes
* Allow Global Billing admins to update actual cost

### DIFF
--- a/app/views/order_management/order_details/_costs.html.haml
+++ b/app/views/order_management/order_details/_costs.html.haml
@@ -1,7 +1,7 @@
 - f.object.send(:extend, PriceDisplayment)
 - if f.object.actual_cost?
   .cost-table
-    .cost= f.input :actual_cost, as: :currency
+    .cost= f.input :actual_cost, as: :currency, disabled: actual_cost_edit_disabled?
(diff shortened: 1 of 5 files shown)


stevens101:
Thanks, I've tested these code changes (https://github.com/wyeworks/nucore-open/pull/5062 and https://github.com/UMass-CORUM/nucore-umass/pull/700 feature flag), and it seems to be working.

app/views/order_management/order_details/_costs.html.haml
```diff
@@ -1,7 +1,7 @@
 - f.object.send(:extend, PriceDisplayment)
 - if f.object.actual_cost?
   .cost-table
-    .cost= f.input :actual_cost, as: :currency
+    .cost= f.input :actual_cost, as: :currency, disabled: actual_cost_edit_disabled?
```
joaquinco, line 4: Should we remove `actual_cost` from permitted params as well?

joaquinco (approved):

app/views/order_management/order_details/_costs.html.haml
stevens101: I think you are right. This change to _costs.html.haml should disable the form field to prevent changes to the value?

PR #5075, opened 2025-03-06
[OSU-124] Improve facility show loading speed

joaquinco:
## Notes

- Prevent N+1 by eager loading instruments' alerts and offline reservations on facility show. I had to move some associations from instrument to product
- Some code improvement on product display groups

### DIFF
--- a/app/models/concerns/products/scheduling_support.rb
+++ b/app/models/concerns/products/scheduling_support.rb
@@ -77,7 +77,11 @@ module Products::SchedulingSupport
   end
 
   def offline?
-    current_offline_reservations.present?
+    if current_offline_reservations.loaded?
+      current_offline_reservations.present?
+    else
+      current_offline_reservations.exists?
+    end
   end
 
   def offline_category
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -23,10 +23,20 @@ class Product < ApplicationRecord
   has_one :product_display_group_product
   has_one :product_display_group, through: :product_display_group_product
 
+  # Instrument specifc
+  has_one(
+    :alert,
+    dependent: :destroy,
+    class_name: "InstrumentAlert",
+    foreign_key: :instrument_id,
+    inverse_of: :instrument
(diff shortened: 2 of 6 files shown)


app/models/concerns/products/scheduling_support.rb
```diff
@@ -77,7 +77,11 @@ def quick_reservation_intervals
   end
 
   def offline?
-    current_offline_reservations.present?
+    if current_offline_reservations.loaded?
+      current_offline_reservations.present?
+    else
+      current_offline_reservations.exists?
+    end
```
LeticiaErrandonea, line 84: Just out of curiosity, why can't we use `exists?` in all cases?

app/models/product.rb
```diff
@@ -23,10 +23,20 @@ class Product < ApplicationRecord
...
 
+  # Instrument specifc
+  has_one(
+    :alert,
+    dependent: :destroy,
+    class_name: "InstrumentAlert",
+    foreign_key: :instrument_id,
+    inverse_of: :instrument
+  )
+  has_many :current_offline_reservations, -> { current }, class_name: "OfflineReservation"
```
LeticiaErrandonea, line 34: Why did you have to move them to this model? While I don't think it's wrong, I find it a bit weird, so I'd like to understand the reason for the change

LeticiaErrandonea (commented):
Looks _promising_ 🤞 

I left a couple of questions

app/models/product.rb
joaquinco: I had to move them to be able to call `inlcudes(:alerts, :current_offline_reservation)` on a product relation

app/models/concerns/products/scheduling_support.rb
joaquinco: if the association is loaded then `exists?` will make a db query whereas `present?` won't

LeticiaErrandonea (approved):
Thanks for the explanation!

PR #5079, opened 2025-03-07
[NU-57] Bulk orders import event log

joaquinco:
## Notes

- Add log event when an order import is created
- Fix order import form tab not being selected


LeticiaErrandonea (approved):
LGTM!

PR #5092, opened 2025-03-13
[NUOPEN-212] Project show flaky specs

joaquinco:
## Note

- Fix some `project_show_spec` flaky tests that use record ids to match page content

### DIFF
--- a/spec/system/admin/projects_show_spec.rb
+++ b/spec/system/admin/projects_show_spec.rb
@@ -42,7 +42,7 @@ RSpec.describe "Projects show" do
 
       it "navigates to order" do
         within("table.js--transactions-table") do
-          click_link active_project_order.id.to_s
+          find_link(href: facility_order_path(facility, active_project_order.id)).click
         end
 
         expect(page).to have_content(active_project_order.id.to_s)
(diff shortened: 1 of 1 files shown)


diebas (approved):
🥳

spec/system/admin/projects_show_spec.rb
```diff
@@ -93,9 +93,10 @@
           within("table.js--transactions-table") do
             click_link originating_order_facility1.id.to_s
           end
-        
 
-          expect(page).to have_content(originating_order_facility1.id.to_s)
+          within("table#order-management") do
+            expect(page).to have_content(originating_order_facility1.id.to_s)
```
LeticiaErrandonea, line 98: Is it here that it fails or is it when it tries to click? I was under the impression the problem was when trying to click, as here we're only checking that it has content but we don't have a number of matches

LeticiaErrandonea (commented):
Left a comment

spec/system/admin/projects_show_spec.rb
joaquinco: There's another failure where there can be multiple links with the same text (like this https://github.com/wyeworks/nucore-open/actions/runs/13789002261/job/38563725814) but I'm still looking into it

LeticiaErrandonea: I think I fixed a bunch of them by using the `href` option with the link I expected to be there. If I'm not mistaken, the problem is when Order and Order detail have the same id

PR #5093, opened 2025-03-13
[NUOPEN-212] Fix instrument schedule positions flaky specs

joaquinco:
## Notes

- Fix `appear_in_order` custom matcher which is recomputing the actual when showing the error message which can led to misleading error because of page state change ([for example](https://github.com/wyeworks/nucore-open/actions/runs/13767199173/job/38496530568))
- Wait for content before using some `appear_in_order` matchers for some of instrument schedule position spec expects. The occasional failure comes from the fact that appear_in_order was matching content against the previous page from time to time.


LeticiaErrandonea (approved):
Oh, this is great, thank you! 🚀

PR #5094, opened 2025-03-13
[DART-44] Update Unreconcile logic to depend on Abilities

LeticiaErrandonea:
# Release Notes
* Update Unreconcile logic to depend on Abilities, so Dartmouth can have its custom permissions


app/lib/ability.rb
```diff
@@ -31,6 +31,8 @@ def initialize(user, resource, controller = nil)
       global_billing_administrator_abilities(user, resource)
 
       account_administrator_abilities(user, resource, controller)
+
+      cannot :unreconcile, OrderDetail
```
LeticiaErrandonea, line 35: Not a bug fan of this, but for us to know whether the Facility Billing Admin can/cannot unreconcile an order detail, we need to check the facility. At the end of the day, I think we need to authorize the OrderDetail, Order or Facility, and for all of them we have a `:manage` that breaks logic. Couldn't find a better way to do this, but happy to change it to something else that feels better.

joaquinco: I'm ok having a custom action for this

joaquinco (approved):

PR #5130, opened 2025-03-25
[NUOPEN-3] Upgrade sprockets

joaquinco:
### Update sprockets gem

- Use manifest instead of `Rails.application.config.assets.precompile` which changed its default.
- Run tests with compiled assets on CI so we can know when we are missing something

## Other things

- Update ckeditor gem. Now it uses a CDN insteaed of a bundled version of the JS code
- Remove execute permissions on some assets files which were not needed

### DIFF
--- a/config/initializers/ckeditor.rb
+++ b/config/initializers/ckeditor.rb
@@ -1,6 +1,9 @@
 # frozen_string_literal: true
 
 Ckeditor.setup do |config|
+  # Ckeditor switched to a comercial license in v4.25
+  # so we use a previous version
+  #
   # //cdn.ckeditor.com/<version.number>/<distribution>/ckeditor.js
   config.cdn_url = "//cdn.ckeditor.com/4.19.1/standard/ckeditor.js"
 end


config/initializers/ckeditor.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+Ckeditor.setup do |config|
+  # //cdn.ckeditor.com/<version.number>/<distribution>/ckeditor.js
+  config.cdn_url = "//cdn.ckeditor.com/4.19.1/standard/ckeditor.js"
```
LeticiaErrandonea, line 5: I'm not sure I follow this. Are we using version from here (4.19.1) or the one in the Gemfile (5.1.3)?

joaquinco: The gem and ckeditor's versions are not really synced. Ckeditor was upgraded to version 5 in gem version 5.20. Here I've upgraded to the latest ckeditor (gem) version which supported v4.

But also I had to override the default version from the gem because ckeditor changed its license in 4.25 to a comercial one which is the latest of v4 and it's LTS. I'll leave a comment about that here

LeticiaErrandonea (approved):
🚀

PR #5165, opened 2025-04-04
[NUOPEN-224] Allow to search when there are no options

LeticiaErrandonea:
# Release Notes
* Allow to search when there are no options

# Screenshot
image

# Additional Context
## Pending
- [x] Fix specs

### DIFF
--- a/app/assets/javascripts/app/chosen_search_options.js
+++ b/app/assets/javascripts/app/chosen_search_options.js
@@ -1,3 +1,16 @@
+const forceChosenInput = (select, data, query) => {
+  if (!select || !data || data.length) return;
+
+  const optionText = query
+    ? `No results found for: ${query}`
+    : "No results found";
+  select.append(
+    `<option value="" class="no-results-option"> ${optionText} </option>`
+  );
+
+  select.trigger("chosen:updated");
+};
+
 $(function () {
   const csrfTokenElement = document.querySelector('meta[name="csrf-token"]');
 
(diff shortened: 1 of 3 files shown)


spec/system/admin/creating_an_estimate_spec.rb
```diff
@@ -2,9 +2,10 @@
 
 require "rails_helper"
 
-RSpec.describe "Creating an estimate" do
```
diebas, line 5: is the :js param necessary here?

diebas (approved):
Left just uno doubt

spec/system/admin/creating_an_estimate_spec.rb
LeticiaErrandonea: Yes. Without it the JS doesn't load and the changes in this PR use the JS code

app/assets/javascripts/app/chosen_search_options.js
```diff
@@ -1,10 +1,31 @@
...
+
+  select.trigger("chosen:updated");
+};
+
+const csrfToken = (selectField) => {
+  const isTestEnv = selectField.data("testEnv");
+
+  if (isTestEnv) {
+    return "";
+  }
```
joaquinco, line 19: why is this needed?

joaquinco: I think on the test env it'll just follow the case where `csrfTokenElement` is null

LeticiaErrandonea: The problem I had is that when `csrfTokenElement` was null, the trigger wasn't loaded. At some point this was necessary but then I did a refactor and I think you're right

joaquinco (approved):

PR #5174, opened 2025-04-08
[UMass-180] Order Detail page does not display correct Order Status for orders with custom status

stevens101:
Orders with custom status (like “MIVP Pending”) are incorrectly displayed as “Canceled” on the order detail page, this will add the missing status.


joaquinco (approved):

stevens101:
Sorry I merged in wyeworks:master by mistake

PR #5180, opened 2025-04-11
[Chore] Remove umass 167 umass 176 ff

joaquinco:
## Notes

- Remove umass-167 and umass-178 ff

### DIFF
--- a/spec/system/admin/facility_statements_spec.rb
+++ b/spec/system/admin/facility_statements_spec.rb
@@ -134,31 +134,29 @@ RSpec.describe "Facility Statement Admin" do
       end
     end
 
-    context "when ff is on", :js do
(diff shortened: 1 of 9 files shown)


spec/system/admin/facility_statements_spec.rb
```diff
@@ -134,10 +134,7 @@
       end
     end
 
-    context(
-      "when ff is on", :js,
-      feature_setting: { show_statement_reconcile_notes: true },
-    ) do
+    context "when ff is on", :js do
```
LeticiaErrandonea, line 137: nit: maybe use another description? 😄

joaquinco: makes sense

PR #5191, opened 2025-04-15
[NUOPEN-226] Add Items and Services to Estimate

LeticiaErrandonea:
# Release Notes
* Add Items and Services to Estimate

# Screenshot
## When first loading
image

## After adding the first product
image

# Additional Context
There will be another PR for showing the products in the Show page.

### DIFF
--- a/app/models/estimate_detail.rb
+++ b/app/models/estimate_detail.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class EstimateDetail < ApplicationRecord
+  belongs_to :estimate, inverse_of: :estimate_details
+  belongs_to :product
+
+  validates :quantity, presence: true, numericality: { greater_than: 0 }
+
+end
--- a/app/models/estimate_detail.rb
+++ b/app/models/estimate_detail.rb
@@ -3,7 +3,28 @@
 class EstimateDetail < ApplicationRecord
   belongs_to :estimate, inverse_of: :estimate_details
   belongs_to :product
+  belongs_to :price_policy
+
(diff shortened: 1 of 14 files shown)


diebas (approved):

app/models/product.rb
```diff
@@ -271,10 +271,23 @@ def cheapest_price_policy(order_detail, date = Time.zone.now)
...
+      price_policies.each_with_object({ cost: 999_999_999, price_policy: nil }) do |pp, memo|
+        cost = pp.estimate_cost_from_estimate_detail(detail)
+
+        if cost && cost < memo[:cost]
+          memo[:cost] = cost
+          memo[:price_policy] = pp
+        end
+
+        memo
+      end
```
joaquinco, line 290: the fact that this returns a hash or a PP depending on the type makes me wonder if there's a better way to write this logic, maybe splitting it into classes or different methods

app/models/estimate_detail.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    user.price_groups.uniq
+  end
+
+  private
+
+  def assign_price_policy_and_cost
+    pp_and_cost = product.cheapest_price_policy(self, Time.current)
+
+    return unless pp_and_cost[:price_policy]
```
joaquinco, line 25: can `pp_and_cost` be nil? Maybe unlikely but since we are listing all products when creating estimates some of them can have outdated pp

joaquinco (approved):
it's looking good! Left some comments/questions

joaquinco (approved):

PR #5198, opened 2025-04-16
[UMASS-104] Add footer logo link

stevens101:
add URL config setting and link to the footer logo

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -147,7 +147,6 @@ en:
       copyright_html: "&copy; Copyright 2010&ndash;%{to_date} Northwestern University"
       logo_alt: "Northwestern Logo"
       logo_file: "logo.jpg"
-      logo_link_to: "/"
     problem_order_details:
       assign_price_policies: Assign Price Policies
       th:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -147,7 +147,7 @@ en:
       copyright_html: "&copy; Copyright 2010&ndash;%{to_date} Northwestern University"
       logo_alt: "Northwestern Logo"
       logo_file: "logo.jpg"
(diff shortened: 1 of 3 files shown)


config/locales/en.yml
```diff
@@ -147,6 +147,7 @@ en:
       copyright_html: "&copy; Copyright 2010&ndash;%{to_date} Northwestern University"
       logo_alt: "Northwestern Logo"
       logo_file: "logo.jpg"
+      logo_link_to: ""
```
joaquinco, line 150: can we set this to "/" or leave it blank so it takes the default value specified in the link_to?

joaquinco (approved):

config/locales/en.yml
stevens101: I think that is a good idea. 
Add logo_link_to: "/" to en.yml, so the default value is clearly stated. 
then the default within the link_to in app/views/shared/_footer.html.haml could be removed?

joaquinco: sounds good

config/locales/en.yml
```diff
@@ -147,6 +147,7 @@ en:
       copyright_html: "&copy; Copyright 2010&ndash;%{to_date} Northwestern University"
       logo_alt: "Northwestern Logo"
       logo_file: "logo.jpg"
+      logo_link_to: "/"
```
joaquinco, line 150: I think we should use an entry in `config/settings.yml` for this (something like `footer_logo_link: "/"`) and read it with `Settings.footer_logo_link` in the view.

The main reason for this is because it would otherwise require changes in all schools repos that override the footer locale content.

joaquinco (approved):
Thanks for the changes!

PR #5199, opened 2025-04-16
[NUOPEN-224] Tweaks for User find

LeticiaErrandonea:
# Release Notes
* Show username in dropdown.
* Include only active accounts.

# Screenshot
image

### DIFF
--- a/app/services/user_finder.rb
+++ b/app/services/user_finder.rb
@@ -21,17 +21,17 @@ class UserFinder
     end
   end
 
-  def self.search(search_term, limit: nil, only_include_active: false)
+  def self.search(search_term, limit: nil, actives_only: false)
     if search_term.present?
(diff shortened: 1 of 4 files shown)


app/services/user_finder.rb
```diff
@@ -15,22 +15,23 @@ class UserFinder
...
     if search_term.present?
-      new(search_term, limit).result_with_count
+      new(search_term, limit:).result_with_count
     else
       [nil, nil]
     end
   end
 
-  def self.search(search_term, limit = nil)
+  def self.search(search_term, limit: nil, only_include_active: false)
```
joaquinco, line 24: what do you think of naming the argument `actives_only` or just `actives`?

joaquinco (approved):

PR #5203, opened 2025-04-17
[UMass 177] Add open contract field to purchase order payment sources

stevens101:
Adds a new open_contract field using db/migrate, and related field on the purchase order payment source create/edit form.


db/migrate/20250310172345_add_open_contract_to_accounts.rb
```diff
@@ -0,0 +1,5 @@
+class AddOpenContractToAccounts < ActiveRecord::Migration[7.0]
```
joaquinco, line 1: I wonder if this migration should be in the c2po engine. Do you see any further use of this field outside this engine?

stevens101: You are right, I will move it to the c2po engine.

joaquinco (approved):

PR #5209, opened 2025-04-21
[NOPEN-238] Update mark as unrecoverable permissions

joaquinco:
## Notes

- Use cancancan with custom action to authorize marking an order as unrecoverable
- Hide Unrecoverable option to non allowed users

### DIFF
--- a/app/controllers/order_management/order_details_controller.rb
+++ b/app/controllers/order_management/order_details_controller.rb
@@ -123,7 +123,7 @@ class OrderManagement::OrderDetailsController < ApplicationController
       @order_statuses = OrderStatus.non_protected_statuses(current_facility)
     end
 
-    if !@order_detail.unrecoverable? && cannot?(:mark_unrecoverable, OrderDetail)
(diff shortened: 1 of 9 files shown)


app/controllers/order_management/order_details_controller.rb
```diff
@@ -117,6 +122,10 @@ def load_order_statuses
     else
       @order_statuses = OrderStatus.non_protected_statuses(current_facility)
     end
+
+    if !@order_detail.unrecoverable? && cannot?(:mark_unrecoverable, OrderDetail)
```
LeticiaErrandonea, line 126: nit: I wonder if this `if` with `!` and `cannot?` could be clearer, but I don't really have a strong opinion

LeticiaErrandonea (approved):
LGTM! Left a question

app/controllers/order_management/order_details_controller.rb
joaquinco: I rewrote it with `unless`

PR #5214, opened 2025-04-22
Add filtering to estimates

diebas:
# Release Notes

Add the ability to filter estimates by user, ID, and name. Also, allow users to hide expired estimates.

# Screenshot

Screenshot 2025-04-22 at 1 27 36 PM


LeticiaErrandonea:
I didn't go through the code but based on the image you attached (thank you for that!) I have 2 thoughts:

1. I don't think we usually use the gray background for filter forms. Can we maybe use the white one, like we do for the rest?
2. I had something like the Create Journal table in mind for the red highlight. Have you seen just the red text somewhere else?
image

app/controllers/facility_estimates_controller.rb
```diff
@@ -16,6 +16,19 @@ def index
...
+    @estimates = @estimates.where(user_id: params[:user_id]) if params[:user_id].present?
+
+    @estimates = @estimates.where(expires_at: Time.current..) if params[:hide_expired] == "1"
+
+    if params[:search].present?
+      search_query = params[:search].strip
+      @estimates = if search_query.match?(/^\d+$/)
+                     @estimates.where(id: search_query)
+                   else
+                     @estimates.where("LOWER(name) LIKE ?", "%#{search_query.downcase}%")
```
joaquinco, line 29: I think we should always filter on this condition. And, if it's a string of digits we can add an or with the filter by id.

What do you think?

app/views/facility_estimates/index.html.haml
```diff
@@ -5,6 +5,32 @@
...
+      .control-group
+        %label.control-label= t(".search")
+        .controls
+          = text_field_tag :search, params[:search], placeholder: t(".search_placeholder"), class: "input-large", id: "estimate_search"
+
+      .control-group
+        %label.control-label= t(".filter_user")
+        .controls
+          = select_tag :user_id, 
+            options_for_select([[t(".all_users"), ""]] + Estimate.where(facility_id: current_facility.id).includes(:user).map { |e| [e.user.full_name, e.user_id] }.uniq,
```
joaquinco, line 20: Can we move the user options array creation code to a helper method?

config/locales/en.yml
```diff
@@ -519,6 +519,12 @@ en:
     index:
       add: Add Estimate
       no_estimates: No estimates found
+      search: Search
+      search_placeholder: Search by name or ID
+      filter_user: Filter by User
+      all_users: All Users
+      hide_expired: Hide expired estimates
+      clear_filter: Clear Filters
```
LeticiaErrandonea, line 527: ultra-nit: can you add these alphabetically?

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

joaquinco (approved):

PR #5215, opened 2025-04-22
[UMass 157] Add "Export as CSV" link to the Orders in Review page

stevens101:
Similar to the main All Transactions page.
This file would only contain orders to be reviewed.

### DIFF
--- a/app/controllers/transactions_controller.rb
+++ b/app/controllers/transactions_controller.rb
@@ -54,7 +54,6 @@ class TransactionsController < ApplicationController
     else
       @too_many_results = true
     end
-    
     @extra_date_column = :reviewed_at
     @order_detail_action = :mark_as_reviewed
     @order_detail_link = {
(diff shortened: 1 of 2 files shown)


app/controllers/transactions_controller.rb
```diff
@@ -54,6 +56,10 @@ def in_review
       display?: proc { |order_detail| order_detail.can_dispute? },
       proc: proc { |order_detail| order_order_detail_path(order_detail.order, order_detail) },
     }
+    respond_to do |format|
+      format.html { @order_details = @order_details.paginate(page: params[:page]) }
```
joaquinco, line 60: Thanks for adding pagination here, are we rendering the pagination links in the view? If not, you can render them with `= will_paginate(@order_details)`

stevens101:
I'll take a closer look. There is an existing will_paginate in app/views/shared/transactions/_table_inside.html.haml

app/controllers/transactions_controller.rb
stevens101: The pagination is already working as there is an existing will_paginate in app/views/shared/transactions/_table_inside.html.haml

joaquinco: The existing pagination is for the set of recently reviewed orders which is defined some lines above.

I think we can omit pagination for `@order_details` here, so you can just remove the block passed to the `format.html` method.

PR #5225, opened 2025-04-24
[NUOPEN-227] Add instrument and Timed Service to Estimate

LeticiaErrandonea:
# Release Notes
* Add ability to add Instrument and Timed services to Estimate

# Screenshot
## New
image

## Show
image

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2025_04_21_145232) do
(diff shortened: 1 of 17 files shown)


db/schema.rb
```diff
@@ -164,6 +164,8 @@
     t.integer "quantity", null: false
     t.datetime "created_at", null: false
     t.datetime "updated_at", null: false
+    t.integer "duration_mins"
+    t.integer "duration_days"
```
joaquinco, line 168: maybe it's too late to have this discussion, but what do you think of having a `quantity_unit` column instead of a quantity field for each unit? maybe it makes things simpler

joaquinco (approved):
Looks good!

diebas (approved):

PR #5234, opened 2025-04-28
[UMASS-142] Rename to billing log

joaquinco:
## Notes

- Rename email log events to billing log events
- Add `FacilityBillingLogEventController` which does basically the same as `BillingLogEventController` but with other permissions and layout.


LeticiaErrandonea (approved):
LGTM!

PR #5247, opened 2025-04-30
[Fix] Change spec using user price groups

joaquinco:
## Notes

User price groups is feature flagged so they are empty when ff is off.


LeticiaErrandonea (approved):
Good catch

PR #5260, opened 2025-05-02
[Dart-34] Download multiple statements

diebas:
# Release Notes

Add the ability to download multiple selected statements at the same time. When clicking the download button, it downloads all the files separately.

Put it under a FF as more than one school could be interested on this feature.

# Screenshot

Screenshot 2025-05-02 at 4 21 16 PM


app/views/statements/download_selected.js.erb
```diff
@@ -0,0 +1,52 @@
...
+          link.href = dataUrl;
+          link.download = filename;
+          link.style.display = 'none';
+          document.body.appendChild(link);
+          link.click();
+          document.body.removeChild(link);
+        } catch(e) {
+          console.error("Error downloading file at index " + index, e);
+        }
+      }, index * 1000); // 1 second delay between each download
```
LeticiaErrandonea, line 43: This code is pretty similar (or the same) as in the JS file. Is there a way of reusing it?

config/settings.yml
```diff
@@ -147,6 +147,7 @@ feature:
   price_policy_requires_note: true
   multi_facility_accounts: true
   facility_directors_can_manage_price_groups: true
+  multiple_statements_download: true
```
LeticiaErrandonea, line 150: Do we need this to be behind a FF? I don't remember any schools saying it could be a problem for them

app/assets/javascripts/app/statements_download.js
```diff
@@ -0,0 +1,110 @@
...
+    selectedIds.forEach(id => formData.append('statement_ids[]', id));
+    return formData;
+  }
+
+  function fetchStatements(formData) {
+    return fetch(form.action + '.json', {
+      method: 'POST',
+      body: formData,
+      headers: {
+        'X-CSRF-Token': document.querySelector('meta[name="csrf-token"]').content,
```
LeticiaErrandonea, line 82: `csrfToken` function should be available (is defined in another file)

LeticiaErrandonea (commented):
Looking good. Left a few comments

joaquinco (commented):
With the feature now being pure client side I think we can remove rails files with unused changes, what do you think?

app/views/statements/download_selected.js.erb
```diff
@@ -0,0 +1,52 @@
+<%# Create container for PDF data %>
```
LeticiaErrandonea: I think this file should be a `js.haml` so we're consistent

LeticiaErrandonea (approved):
It looks and works fine!

~I'm still wondering if there's a way of reusing the code in the js.erb file, but it's not blocking~

joaquinco (approved):

PR #5279, opened 2025-05-09
[UMASS-111] Hide account expiration if > 75.years.from_now

joaquinco:
## Notes

- Hide account expiration date if it's greater than 75 years in the future


spec/system/admin/managing_payment_sources_spec.rb
```diff
@@ -72,4 +72,62 @@
...
+  describe(
+    "hide far future expiration dates",
+  ) do
+    before do
+      visit facility_accounts_path(facility)
+      fill_in "search_term", with: account.account_number
+      click_on "Search"
+    end
+
+    context "when expiration is less than 75 years from now" do
```
LeticiaErrandonea, line 85: Should we have another `it` for when the FF is on?

LeticiaErrandonea (approved):

PR #5293, opened 2025-05-13
[NUOPEN-229] Add products from another facility to the estimate

LeticiaErrandonea:
# Release Notes
* Add products from another facility to the estimate

# Screenshot
## New
image

## Show
image

### DIFF
--- a/app/models/product.rb
+++ b/app/models/product.rb
@@ -73,6 +73,7 @@ class Product < ApplicationRecord
   scope :active, -> { where(is_archived: false, is_hidden: false) }
   scope :alphabetized, -> { order(Arel.sql("LOWER(products.name)")) }
   scope :archived, -> { where(is_archived: true) }
+  scope :available_for_estimates, -> { where({ type: %w[Item Service Instrument TimedService Bundle] }).not_archived }
   scope :not_archived, -> { where(is_archived: false) }
   scope :mergeable_into_order, -> { not_archived.where(type: mergeable_types) }
   scope :cross_core_available, -> { where(cross_core_ordering_available: true) }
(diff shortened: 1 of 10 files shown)


app/models/product.rb
```diff
@@ -73,6 +73,7 @@ def self.billing_modes
   scope :active, -> { where(is_archived: false, is_hidden: false) }
   scope :alphabetized, -> { order(Arel.sql("LOWER(products.name)")) }
   scope :archived, -> { where(is_archived: true) }
+  scope :available_for_estimates, -> { where({ type: %w[Item Service Instrument TimedService Bundle] }).not_archived }
```
joaquinco, line 76: should we use `active` instead of `not_archived`? I'm not sure what the hidden flag means though

app/views/facility_estimates/_estimate_detail.html.haml
```diff
@@ -5,7 +5,11 @@
     = hidden_field_tag "#{id_prefix}[duration_unit]", estimate_detail.duration_unit
   = hidden_field_tag "#{id_prefix}[_destroy]", false, class: 'destroy-field'
 
-  %td= estimate_detail.product.try(:name)
+  %td
+    - product_name = estimate_detail.product.try(:name)
+    - if estimate_detail.product.try(:facility) && estimate_detail.product.facility != current_facility
```
joaquinco, line 10: can we always assume that `estimate_details.product` will be set? I would think so

joaquinco (approved):

app/models/product.rb
LeticiaErrandonea: Hidden products are the ones that the users can't purchase by themselves but can be ordered through "Order for". I think between cores it's fine to order hidden products.

app/views/facility_estimates/_estimate_detail.html.haml
LeticiaErrandonea: I think you're right

diebas (approved):

PR #5294, opened 2025-05-13
[NUOPEN-237] Refactor OrderImport processing into a job

joaquinco:
## Note

Part of ticket [NUOPEN-237].
Refactor `OrderImport` async run to an active job

### DIFF
--- a/app/jobs/order_import_job.rb
+++ b/app/jobs/order_import_job.rb
@@ -1,0 +1,12 @@
+# frozen_string_literal: true
+
+class OrderImportJob < ApplicationJob
+  def perform(order_import, report_recipient)
+    order_import.process_upload!
+    report = Reports::OrderImport.new(order_import)
+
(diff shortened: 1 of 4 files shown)


app/jobs/order_import_job.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+class OrderImportJob < ApplicationJob
+  def perform(order_import, report_recipient)
```
LeticiaErrandonea, line 4: nit: should we name this `perform!` so it's obvious it could raise an exception?

LeticiaErrandonea (approved):
LGTM

Left a question

app/jobs/order_import_job.rb
joaquinco: this is defining active job entry point method which is inherited

PR #5298, opened 2025-05-14
[NUOPEN-243] Add validation to check Price policy is present

LeticiaErrandonea:
# Release Notes
* Add validation to check Price policy is present

# Screenshot
image

# Additional context
While adding specs, I found out they got broken when the Estimate had an error. Problem was that the quantity and duration weren't set. That I fixed. It's also a problem with User, which I didn't do and I went through just selecting it again in the spec. May be worth it to check when we change the User and Price group behavior.


app/models/estimate.rb
```diff
@@ -18,8 +19,22 @@ def total_cost
...
   end
+
+  def validate_estimate_details
+    estimate_details.each do |estimate_detail|
+      next if estimate_detail.marked_for_destruction?
+      next if estimate_detail.valid?
+
+      errors.delete(:"estimate_details.base")
+      estimate_detail.errors.each do |error|
+        errors.add(:base, "#{estimate_detail.product.name}: #{error.message}")
```
joaquinco, line 36: aren't these errors shown in the estimate details table? I think it's enough to have them there

joaquinco (approved):
What do you think of adding a spec?

joaquinco (approved):

diebas (approved):

PR #5308, opened 2025-05-16
[NUOPEN-243] Add ability to recalculate the estimate cost

LeticiaErrandonea:
# Release Notes
* Add ability to recalculate the estimate cost

# Screenshot
image

### DIFF
--- a/app/views/facility_estimates/show.html.haml
+++ b/app/views/facility_estimates/show.html.haml
@@ -48,3 +48,7 @@
       %td{colspan: 3}
       %td.currency
         %strong=number_to_currency(@estimate.total_cost)
+
+- if @estimate.estimate_details.any?
+  %p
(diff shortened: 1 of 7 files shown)


joaquinco (approved):
Looks good!

Should we add the updated_at date somewhere so users know when the price was set?

app/views/facility_estimates/show.html.haml
```diff
@@ -47,3 +48,7 @@
       %td{colspan: 3}
       %td.currency
         %strong=number_to_currency(@estimate.total_cost)
+
+- if @estimate.estimate_details.any?
+  %p
+    = t(".updated_at_notice", updated_at: @estimate.estimate_details.maximum(:updated_at).strftime('%m/%d/%Y %H:%M'))
```
joaquinco, line 54: we can use `I18n.l` to format time here, what do you think?

joaquinco (approved):

diebas (approved):

PR #5310, opened 2025-05-19
[OSU-139] Remove "print me" label on sanger submission

joaquinco:
## Notes

- Remove label to print sanger submission
- Fix sanger enable flash message regarding Order Form activation


LeticiaErrandonea (approved):
Good catch!

PR #5316, opened 2025-05-20
[NUOPEN-239] ActiveJob queue adapter per env

joaquinco:
## Notes

- Set different active job queue adapter for each env
- Change mailer specs to use mailer matchers
- Remove now useless `:active_job` spec hook


LeticiaErrandonea (approved):
LGTM!

PR #5317, opened 2025-05-20
[OSU-52] Fix translation missing condition

joaquinco:
## Note

Fix missing translation caused by order details with `problem_description_key` nil but `problem` attribute set to true.

### DIFF
--- a/app/presenters/order_detail_notice_presenter.rb
+++ b/app/presenters/order_detail_notice_presenter.rb
@@ -23,13 +23,11 @@ class OrderDetailNoticePresenter < DelegateClass(OrderDetail)
   end
 
   def warnings
-    warnings = []
-
-    problem_description_key.then do |problem_key|
-      warnings << Notice.new(problem_key, :warning) if problem_key.present? && problem?
+    if problem?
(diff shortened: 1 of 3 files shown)


LeticiaErrandonea:
Do we know how these orders get to this state of having `problem: true` but no reason?

joaquinco:
@LeticiaErrandonea the problem flag seems to be out of sync when a reservation is ended past its due date.

I guess what happens is that the recurrent task `EndReservationOnly` marks the reservation as completed once the reservation `reserve_end_at` has passed (and the problem flag is set to true because actual end time is missing), then the user manually ends the reservation and the problem goes away but the order detail never knew it.

LeticiaErrandonea:
That makes sense. In that case, should we also reset the `problem` flag to `false`? I'm wondering if we ever want to have that value set to `true` but not knowing why

joaquinco:
I ended up adding a badge for the out of sync case and create a ticket to further research. I wasn't able to reproduce the case on my previous comment.

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

PR #5326, opened 2025-05-22
[NUOPEN-239] Fix engine specs

joaquinco:
## Notes

- Some engines specs were not running in the CI. I've fixed mailer matchers on those and added to the CI.
- Use mailer matcher for email not delivered specs


LeticiaErrandonea (approved):
LGTM!

PR #5338, opened 2025-05-27
[UMASS-184] Account price group tab

joaquinco:
## Notes

Add "Price Groups" tab to account admin page where the account's price groups are listed.

## Screenshot

Captura de pantalla 2025-05-27 a la(s) 11 27 45

### DIFF
--- a/app/controllers/facility_account_price_groups_controller.rb
+++ b/app/controllers/facility_account_price_groups_controller.rb
@@ -13,6 +13,8 @@ class FacilityAccountPriceGroupsController < ApplicationController
   def show
   end
 
+  private
+
   def load_resources
     @account = Account.find(params[:account_id])
     @price_groups =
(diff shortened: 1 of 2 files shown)


app/controllers/facility_account_price_groups_controller.rb
```diff
@@ -0,0 +1,28 @@
...
+  before_action { @active_tab = "admin_users" }
+  before_action :load_resources
+  before_action :authorize_account
+
+  layout "two_column"
+
+  def show
+  end
+
+  def load_resources
```
diebas, line 16: I think load_resources and authorize_account should be private methods as they are not controller actions

config/routes.rb
```diff
@@ -296,6 +296,7 @@
       end
 
       resources :orders, controller: "facility_account_orders", only: [:index]
+      resource :price_groups, controller: "facility_account_price_groups", only: [:show, :edit, :update]
```
diebas, line 299: I would rather add edit and update routes once they are implemented

diebas (approved):
Just a few comments

PR #5345, opened 2025-05-28
[NUOPEN-235] Autocoplete username on estimates edit screen

diebas:
# Release Notes

While opening the estimate edit view for an estimate with an existing user, the username field was shown as empty. With this change, it autofills the username on the form

### DIFF
--- a/app/views/facility_estimates/_form.html.haml
+++ b/app/views/facility_estimates/_form.html.haml
@@ -13,6 +13,7 @@
       = f.label :custom_name, Estimate.human_attribute_name(:user)
       = f.input_field :custom_name,
         class: "js--user-autocomplete",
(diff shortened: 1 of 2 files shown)


app/views/facility_estimates/_form.html.haml
```diff
@@ -13,6 +13,7 @@
       = f.label :custom_name, Estimate.human_attribute_name(:user)
       = f.input_field :custom_name,
         class: "js--user-autocomplete",
+        value: (@estimate.user.try(:name) || @estimate.custom_name),
```
LeticiaErrandonea, line 16: Can we use `@estimate.user_display_name`?

LeticiaErrandonea (approved):
Left a question, but I think this is good

PR #5377, opened 2025-06-03
[NUOPEN-244] Add static 500 error view

diebas:
# Release Notes

Use a static page for 500 errors.

### DIFF
--- a/app/controllers/errors_controller.rb
+++ b/app/controllers/errors_controller.rb
@@ -8,7 +8,7 @@ class ErrorsController < ApplicationController
   end
 
   def internal_server_error
-    render_error(nil, :internal_server_error)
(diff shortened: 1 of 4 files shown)


app/controllers/errors_controller.rb
```diff
@@ -8,7 +8,7 @@ def not_found
   end
 
   def internal_server_error
-    render_error("internal_server_error", :internal_server_error)
+    render_error(nil, :internal_server_error)
```
joaquinco, line 11: what do you think of rendering the static file here (with send_static)? The `send_static_error_page` handles the case were 500.html does not exist, I think we can safely assume it exists

joaquinco (approved):

PR #5433, opened 2025-06-23
Auto stop previous reservation

diebas:
# Release Notes

Add a feature under a feature flag that automatically ends an open reservation when another user starts a new reservation, notifying the previous user. This feature only applies to timer/relay instruments

### DIFF
--- a/app/models/reservation.rb
+++ b/app/models/reservation.rb
@@ -178,6 +178,11 @@ class Reservation < ApplicationRecord
   end
 
   def start_reservation!
+    # Auto-end functionality: automatically end previous reservations when feature is enabled
+    if SettingsHelper.feature_on?(:auto_end_reservations_on_next_start)
+      AutoEndPreviousReservation.new(product, user).end_previous_reservations!
+    end
+
     # If there are any reservations running over their time on the shared schedule,
     # kick them over to the problem queue.
     product.schedule.products.flat_map(&:started_reservations).each do |reservation|
--- a/app/services/auto_end_previous_reservation.rb
+++ b/app/services/auto_end_previous_reservation.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+class AutoEndPreviousReservation
+
+  def initialize(product, current_user)
+    @product = product
+    @current_user = current_user
+  end
+
+  def end_previous_reservations!
+    return unless SettingsHelper.feature_on?(:auto_end_reservations_on_next_start)
+    return unless timer_based_instrument?
+
+    reservations_to_end = previous_reservations
+
+    reservations_to_end.each do |reservation|
+      auto_end_reservation(reservation)
+    end
+  end
+
+  private
+
+  def timer_based_instrument?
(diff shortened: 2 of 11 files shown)


app/services/auto_end_previous_reservation.rb
```diff
@@ -0,0 +1,47 @@
...
+    @product.control_mechanism == Relay::CONTROL_MECHANISMS[:timer] ||
+      @product.control_mechanism == Relay::CONTROL_MECHANISMS[:relay]
+  end
+
+  def previous_reservations
+    @product.reservations.joins(:order_detail)
+            .where(actual_end_at: nil)
+            .where(reserve_end_at: ...Time.current)
+            .where(order_details: { canceled_at: nil })
+            .where.not(actual_start_at: nil)
```
joaquinco, line 33: I  think we should limit to the most recent reservation and define a timewindow for reservations that can be considered.

A few questions that come to my mind:
- Does it make sense for this to end a reservation if it was several days ago? what about one day ago? what's the limit? 
- Does it make sense to end several reservations and setting all of them the same actual end time?

joaquinco (commented):
Looks good. I've left a question regarding the previous reservations to consider

app/models/reservation.rb
```diff
@@ -178,6 +178,11 @@ def force_dirty!
   end
 
   def start_reservation!
+    # Auto-end functionality: automatically end previous reservations when feature is enabled
+    if SettingsHelper.feature_on?(:auto_end_reservations_on_next_start)
+      AutoEndPreviousReservation.new(product, user).end_previous_reservations!
+    end
```
joaquinco, line 184: Can we move this to the controller or to an active job and call it from the controller?

I think this model method is too complex already to add this side effect.
Also, since this is a side effect of start a reservation I think any errors occurred within the new code shouldn't affect the original functionallity

app/services/auto_end_previous_reservation.rb
diebas: we already have a task called AutoExpireReservation that runs every 5 minutes and look for reservations with more than 12 hs and move them to the problem queue, so there aren't many active ones hanging around. This feature is complementary, as it handles the case when someone arrives before those 12 hours.

Multiple active reservations per instrument are rare due to validations, but if it happens (mostly admin reservations), ending them all with the same timestamp makes sense since that's when the new user actually took control.

Do you think I should add a time limit anyway?

app/services/auto_end_previous_reservation.rb
```diff
@@ -0,0 +1,48 @@
...
+
+class AutoEndPreviousReservation
+
+  def initialize(product, current_user)
+    @product = product
+    @current_user = current_user
+  end
+
+  def end_previous_reservations!
+    return unless SettingsHelper.feature_on?(:auto_end_reservations_on_next_start)
```
joaquinco, line 11: this is also checked on the instrument switcher, does it make sense to have both?

joaquinco (approved):

PR #5451, opened 2025-06-30
[NUOPEN-260] Improve facility transaction page

diebas:
# Release Notes

Improve facility transactions page performance by fixing N+1 queries and optimizing total calculations. Added polymorphic associations to the Product model to enable efficient eager loading of external service data across all product types.

Added `has_many :external_service_passers, as: :passer` association to the base `Product` model
Added `has_many :external_services, through: :external_service_passers` association
Previously, these associations were only available in the `Service` subclass, causing ActiveRecord configuration errors when trying to eager load external service data for all product types
This change allows Rails to properly include external service associations without breaking for non-Service products (Instruments, Item)

## Screenshots

Before:
Screenshot 2025-07-01 at 5 06 22 PM

After:
Screenshot 2025-07-01 at 5 04 51 PM

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -144,7 +144,7 @@ class FacilitiesController < ApplicationController
     @order_details = @search.order_details.joins(order: :user).reorder(sort_clause)
 
     if @order_details.size < 1000
-      @grand_total = @order_details.map { |od| od.actual_total || od.estimated_total }.compact.sum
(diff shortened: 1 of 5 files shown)


LeticiaErrandonea (approved):
LGTM!

Is there anything that can be removed from the Bullet "safelist" defined in the `test.rb` file?

joaquinco (approved):

PR #5467, opened 2025-07-07
[DART-153] Fix statement report for cross core facility

joaquinco:
## Notes

Statement report was failing for Cross Core Facility (All Facilities) because it's a instance created on the fly which is not serializable.

I've taken the change to refactor the report the same way other reports are done:
- Use csv report mail job to generate and send the report
- Move the report to its own file

Also:
- Clear enqueued jobs before each system example since calling `perform_enqueued_jobs` without a block seemed to be running other examples jobs

### DIFF
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -236,6 +236,10 @@ RSpec.configure do |config|
     Warden.test_reset!
   end
 
+  config.before type: :system do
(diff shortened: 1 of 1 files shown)


LeticiaErrandonea (approved):
LGTM! 👏

diebas (approved):

PR #5484, opened 2025-07-15
[DART-59] Allow business admins to change Payment Sources in bulk

LeticiaErrandonea:
# Release Notes
* Allow business admins to change Payment Sources in bulk

# Screenshot
image

image

image

image

image

### DIFF
--- a/app/controllers/concerns/movable_transactions.rb
+++ b/app/controllers/concerns/movable_transactions.rb
@@ -68,8 +68,8 @@ module MovableTransactions
 
   def bulk_reassignment_failure(reassignment_error)
     flash[:alert] = I18n.t("controllers.facilities.bulk_reassignment.move.failure",
-                               reassignment_error: reassignment_error,
-                               order_detail_id: reassignment_error.record.id)
+                           reassignment_error: reassignment_error,
+                           order_detail_id: reassignment_error.record.id)
   end
 
   def reassign_account_from_params!
--- a/app/controllers/concerns/movable_transactions.rb
+++ b/app/controllers/concerns/movable_transactions.rb
@@ -74,11 +74,6 @@ module MovableTransactions
 
   def reassign_account_from_params!
     account = Account.find(params[:account_id])
-
-    unless current_user.account_administrator_of?(account)
-      raise ActiveRecord::RecordNotFound
-    end
-
     OrderDetail.reassign_account!(account, @order_details)
   end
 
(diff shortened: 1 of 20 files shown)


app/controllers/concerns/movable_transactions.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def bulk_reassignment_failure(reassignment_error)
+    flash[:alert] = I18n.t("controllers.facilities.bulk_reassignment.move.failure",
+                           reassignment_error: reassignment_error,
+                           order_detail_id: reassignment_error.record.id)
+  end
+
+  def reassign_account_from_params!
+    account = Account.find(params[:account_id])
```
joaquinco, line 76: we should validate this account is one of the user managed accounts (when this is used from the business admin flow)

app/lib/ability.rb
```diff
@@ -308,6 +308,10 @@ def account_administrator_abilities(user, resource, controller)
       can :manage, AccountUser
       can [:show, :suspend, :unsuspend, :user_search, :user_accounts, :statements, :show_statement, :index], Statement
     end
+
+    if SettingsHelper.feature_on?(:move_transactions_account_roles) && Account.administered_by(user).any?
+      can [:movable_transactions, :reassign_chart_strings, :confirm_transactions, :move_transactions], TransactionsController
```
joaquinco, line 313: is there are reason to use this granularity instead of a single `move_transactions` or something like that?

config/locales/en.yml
```diff
@@ -983,19 +996,9 @@ en:
...
-      unmovable_transactions:
-        label: "Transactions that will not be moved:"
-
-    reassign_chart_strings:
-      account_select:
-        placeholder: "Select a !chart_string_downcase!"
-        explanation: "All !chart_strings_downcase! listed above are available to reassign to at least one of the selected transactions below."
-      head: "Move Transactions"
-      submit: Reassign !Chart_string!
+
```
diebas, line 999: Remove these blank lines

config/locales/en.yml
```diff
@@ -1348,6 +1351,10 @@ en:
     in_review: Transactions in Review
     none_in_review_found: No transactions in review were found.
     recently_reviewed: Recently Reviewed
+    movable_transactions:
+      head: "My Movable Transactions"
+      no_transactions: "You have no movable transactions."
+
```
diebas, line 1357: Also this one

diebas (approved):

app/lib/ability.rb
LeticiaErrandonea: I don't think I understood your question. You mean why using the actions or the controller?

joaquinco (approved):

PR #5507, opened 2025-07-22
[NU-12] Retry connection for relay status

LeticiaErrandonea:
# Release Notes
* Retry connection for relay status

### DIFF
--- a/app/models/power_relay.rb
+++ b/app/models/power_relay.rb
@@ -52,10 +52,19 @@ module PowerRelay
   # boolean - The current on/off status of the outlet (and secondary outlet, if configured).
   def query_status
     log_power_relay_connection(:status) do
-      relay_status = relay_connection.status(outlet)
-      if secondary_outlet
-        secondary_outlet_status = relay_connection.status(secondary_outlet)
+      relay_status = nil
+      max_retries = 2
+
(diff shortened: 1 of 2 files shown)


app/models/power_relay.rb
```diff
@@ -52,10 +52,19 @@ def activate_secondary_outlet
...
+      max_retries = 2
+
+      retry_on_error(max_retries, NetBooter::Error) do
+        relay_status = relay_connection.status(outlet)
+      end
+
       if secondary_outlet
-        secondary_outlet_status = relay_connection.status(secondary_outlet)
+        retry_on_error(max_retries, NetBooter::Error) do
+          relay_connection.status(secondary_outlet)
```
diebas, line 64: Does this method return the actual status of the outlet? If so, I don't think it's being used. Maybe you need to assign it to a variable and add it in the method's output. Does that make any sense?

LeticiaErrandonea: Before we weren't returning it. I think the idea is that both outlets should have the same status, so I'm not even sure why we read both of them. 

I could reach out to the clients and confirm the expected behavior. However, they haven't brought up having an issue with the current implementation.

diebas: I think we can remove that call then as it's not used

diebas (approved):

PR #5537, opened 2025-07-30
[NUOPEN-263] Use field for order detail status badges

joaquinco:
## Notes

Add field to store order detail notices and problems instead of computing them every time which slows down order detail listings.

There was a before save callback that set the `problem` flag which computed the order details problems from scratch. This branch extends that callback by computing the notices and saving everything. So it's a little more expensive to save but much faster to read.

Also, there are two external events (or situations) which affect order detail notices and problems which are:
- Journal closed: when this happens all order details are updated by queuing a job that updates problems and notices
- Reservation missing actuals: there was a callback that triggered order detail save already so there's no change here. I could've change this so the aforementioned job was called, I'm open to discuss


app/models/order_detail/notices.rb
```diff
@@ -0,0 +1,31 @@
...
+  def set_problem_and_notices
+    notice_service = OrderDetails::NoticesService.new(self)
+
+    self.problem_keys = notice_service.problems
+    self.notice_keys = notice_service.notices
+    self.problem = problem_keys.present?
+  end
+
+  # This must be called after updating problem field
+  def update_fulfilled_at_on_resolve
```
LeticiaErrandonea, line 23: `set_problem_order` had `true # problem might be false; we need the callback chain to continue` as the last line. Should we also do that here?

LeticiaErrandonea (commented):
Looks good, left a comment about a small change compared to the previous code.

Also, what are we doing with the existing order details? My guess is we'll want to run the job for the last N of them when we release, but I couldn't find any mention of that in the PR

app/models/order_detail/notices.rb
joaquinco: I've found no documentation on the return value affecting the callback chain. It's only interrupted by an exception or throwing :abort. https://guides.rubyonrails.org/v7.0/active_record_callbacks.html#halting-execution

db/migrate/20250729132443_add_notices_to_order_details.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class AddNoticesToOrderDetails < ActiveRecord::Migration[7.0]
+  def change
```
diebas, line 4: Should we add a rollback here?

spec/services/order_details/notices_service_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      is_expected.to eq([:in_review])
+    end
+
+    it "shows in dispute" do
+      allow(order_detail).to receive(:in_dispute?).and_return(true)
+
+      is_expected.to eq([:in_dispute])
+    end
+
+    it "shows in dispute" do
```
diebas, line 35: You have a duplicated test name here. I think you should call this one "shows global admin must resolve"

diebas (approved):
Good job! just 2 little comments

db/migrate/20250729132443_add_notices_to_order_details.rb
joaquinco: Do you mean to make this migration revertable? because it already is

spec/services/order_details/notices_service_spec.rb
joaquinco: good catch!

PR #5560, opened 2025-08-11
[NU-409] Fix sortable column order in Oracle

joaquinco:
Oracle adapter uses `first_value()` SQL function when implementing `columns_for_distinct()` method, which is used by active record, in this case, when there's an order by clause and some conditions apply on the joined tables.

We were providing several order criteria as a string which made first_value to be called with multiple arguments thus giving an error.

The fix is to provide multiple order criteria as independent entries.

Reference:
- FIRST_VALUE(): https://docs.oracle.com/en/database/oracle/oracle-database/21/sqlrf/FIRST_VALUE.html
- Oracle adapter `columns_for_distinct()`: https://github.com/rsim/oracle-enhanced/blob/1ad893df4f3c083c8c6044e57f5b06c80cad2b28/lib/active_record/connection_adapters/oracle_enhanced_adapter.rb#L652

### DIFF
--- a/app/controllers/concerns/sortable_column_controller.rb
+++ b/app/controllers/concerns/sortable_column_controller.rb
@@ -11,7 +11,7 @@ module SortableColumnController
   def sort_clause
     Array(sort_lookup_hash[sort_column]).map do |clause|
       [clause, sort_direction].join(" ")
-    end.join(", ")
+    end
   end
 
   private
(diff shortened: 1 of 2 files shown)


LeticiaErrandonea (approved):
LGTM!

diebas (approved):

PR #5565, opened 2025-08-12
[DART-154] Make accounts valid on the day they expire

LeticiaErrandonea:
# Release Notes
* Make accounts valid on the day they expire

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -46,7 +46,7 @@ class Account < ApplicationRecord
   accepts_nested_attributes_for :account_users
   has_many :log_events, as: :loggable
 
-  scope :active, -> { where("expires_at > ?", Time.current).where(suspended_at: nil) }
+  scope :active, -> { where("expires_at > ?", Time.current.beginning_of_day).where(suspended_at: nil) }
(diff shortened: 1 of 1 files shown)


app/models/account.rb
```diff
@@ -187,7 +187,7 @@ def suspended?
   end
 
   def expired?
-    expires_at && expires_at <= Time.zone.now
+    expires_at && expires_at.beginning_of_day <= Time.zone.now.beginning_of_day
```
joaquinco, line 190: Can we add a test for this?

app/models/account.rb
```diff
@@ -46,7 +46,7 @@ def price_groups
   accepts_nested_attributes_for :account_users
   has_many :log_events, as: :loggable
 
-  scope :active, -> { where("expires_at > ?", Time.current).where(suspended_at: nil) }
+  scope :active, -> { where("expires_at > ?", Time.current.beginning_of_day).where(suspended_at: nil) }
   scope :active_at, ->(time) { where("expires_at >= ?", time).where(suspended_at: nil) }
```
diebas, line 50: Shouldn't we add this change in the active_at scope too?

PR #5633, opened 2025-09-11
[UMASS-213] Add transactions price group to CSV

joaquinco:
# Release Notes

Add price group to exported transactions CSV


LeticiaErrandonea (commented):
LGTM!

I wonder if we should have asked the other schools about this, as it doesn't seem to cause any problems, and it's just more information in the CSV. I mean, I think we could even push this without the Feature Flag.

LeticiaErrandonea (approved):

PR #5634, opened 2025-09-11
[NU-26] Use account owner for pricing when feature flag is enabled

diebas:
# Release Notes

Added configurable pricing flow feature that allows price groups to be determined by payment source (account owner) instead of purchaser when enabled.

# Additional Context

This change introduces a new feature flag `pricing_flows_from_payment_source` that modifies how price groups are calculated for order details. When enabled, the system will use the account owner's price groups instead of the purchaser's price groups for pricing calculations.


LeticiaErrandonea (approved):
LGTM! Left a comment in a spec

PR #5651, opened 2025-09-18
Fix Holiday reservation specs when FF is on

diebas:
# Release Notes

Fix tests that breaks when the user_based_price_groups_exclude_purchaser FF is turned on.


joaquinco (commented):
Left a comment on one of the reservation test. Also, these reservation tests are not related to holidays so I think we should either rename the tests file or move them to other place

PR #5683, opened 2025-09-30
[NU-28] Only global admin can mass unriconcile

diebas:
# Notes
  Allow only global administrators to mass unreconcile journal orders, including accessory orders that were previously failing due to state transition conflicts. Only orders within the selected journal are affected - related orders outside the journal remain unchanged.

  ## [NU-28 | Allow mass journal transactions unreconciling](https://universe-of-universities.atlassian.net/browse/NU-28)

  ## Additional Context

  ### Problem
  Previously, mass unreconciling would fail when journals contained parent orders with accessories. When a parent order was unreconciled, it would trigger an `after_save` callback that attempted to update child orders' status, causing invalid state transitions from 'reconciled' to 'complete' using the wrong event.

  ### Solution
  - Restricted mass unreconciling feature to global administrators only
  - Added proper state machine transition `unreconcile` event to OrderDetail model
  - Modified reconciler to use the state machine's `unreconcile!` method instead of bypassing callbacks
  - Added logic to skip child order updates during unreconciliation to prevent state conflicts
  - Each order is processed independently - failures don't affect other orders
  - Added clear error messaging to indicate which specific orders failed and why

### DIFF
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -145,6 +145,10 @@ class FacilityJournalsController < ApplicationController
       raise CanCan::AccessDenied, I18n.t("controllers.facility_journals.unreconcile.feature_disabled")
     end
 
+    unless current_user.administrator?
+      raise CanCan::AccessDenied, I18n.t("controllers.facility_journals.unreconcile.access_denied")
+    end
+
     process_reconciliation(:unreconcile)
   end
 
--- a/app/controllers/facility_journals_controller.rb
+++ b/app/controllers/facility_journals_controller.rb
@@ -170,6 +170,9 @@ class FacilityJournalsController < ApplicationController
 
     if count > 0
       flash[:notice] = I18n.t("controllers.facility_journals.#{action}.success", count: count)
+      if reconciler.full_errors.any?
+        flash[:error] = reconciler.full_errors.join("<br />").html_safe
+      end
     elsif reconciler.full_errors.any?
       flash[:error] = reconciler.full_errors.join("<br />").html_safe
     else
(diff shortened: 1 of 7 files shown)


app/services/order_details/reconciler.rb
```diff
@@ -58,29 +58,26 @@ def reconcile_all
...
+
+        begin
+          order_detail.update_columns(
+            state: "complete",
+            order_status_id: OrderStatus.complete.id,
+            reconciled_at: nil,
+            deposit_number: nil,
+            reconciled_note: nil,
+            updated_at: Time.current
+          )
```
joaquinco, line 73: What's the reason to skip callbacks? I think we should use the state machine, adding a new transition if necessary

Also, the only reason I can see this can fail is due to some db error unrelated to the application

app/controllers/facility_journals_controller.rb
```diff
@@ -145,6 +145,10 @@ def unreconcile
       raise CanCan::AccessDenied, I18n.t("controllers.facility_journals.unreconcile.feature_disabled")
     end
 
+    unless current_user.administrator?
```
joaquinco, line 148: there's an `:unreconcile, OrderDetail` rule in abilities which seems to be only enabled to administrators, can we use that?

app/views/facility_journals/show.html.haml
```diff
@@ -45,7 +45,7 @@
   - has_reconciled = @journal.order_details.where(state: "reconciled").any?
   - has_unreconciled = @journal.order_details.where.not(state: "reconciled").any?
   - journal_can_reconcile = @journal.successful? && has_unreconciled
-  - journal_can_unreconcile = @journal.successful? && has_reconciled && SettingsHelper.feature_on?(:allow_mass_unreconciling)
+  - journal_can_unreconcile = @journal.successful? && has_reconciled && SettingsHelper.feature_on?(:allow_mass_unreconciling) && current_user.administrator?
```
joaquinco, line 48: we should use the ability check here

joaquinco (commented):
Looks good, thanks for the changes.

I left just one final small comment

PR #5688, opened 2025-10-01
force nonbillable price policy assignment

diebas:
# Notes

When user_based_price_groups_exclude_purchaser for some reason we need to investigate deeper the price policy is not being assigned for non-billable products.


joaquinco:
Closing this as we've fixed the spec

PR #6121, opened 2026-04-16
[NUOPEN-281] Fix order detail notices 

joaquinco:
# Notes

Fix order details notices by splitting stored and time based order detail notices.

[NUOPEN-281 | Fix order detail notices](https://universe-of-universities.atlassian.net/browse/NUOPEN-281)

### DIFF
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -820,10 +820,6 @@ class OrderDetail < ApplicationRecord
     !valid_service_meta?
   end
 
-  def cached_missing_form?
-
-  end
-
   def ready_for_statement?
     reviewed? &&
       statement_id.blank? &&
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -820,12 +820,23 @@ class OrderDetail < ApplicationRecord
     !valid_service_meta?
   end
 
+  def cached_missing_form?
+
+  end
+
   def ready_for_statement?
-    reviewed? && statement_id.blank? && Account.config.using_statements?(account.type) && !missing_form?
+    reviewed? &&
+      statement_id.blank? &&
+      Account.config.using_statements?(account.type) &&
+      problem_keys.exclude?(:missing_form)
   end
 
   def ready_for_journal?
-    reviewed? && journal_id.blank? && Account.config.using_journal?(account.type) && !reconciled? && !missing_form?
+    reviewed? &&
+      journal_id.blank? &&
+      Account.config.using_journal?(account.type) &&
+      !reconciled? &&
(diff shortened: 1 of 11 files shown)


app/presenters/order_detail_notice_presenter.rb
```diff
@@ -9,22 +9,15 @@ def notices_service
...
 
   def statuses
-    values = if SettingsHelper.feature_on?(:stored_order_notices)
-               notice_keys
-             else
-               notices_service.notices
-             end
-    values.map { |s| Notice.new(s) }
+    values = notice_keys
+    values += notices_service.time_based_notices
```
LeticiaErrandonea: nit
```suggestion
    values = notice_keys + notices_service.time_based_notices
```

app/services/order_details/notices_service.rb
```diff
@@ -13,19 +13,26 @@ def initialize(order_detail)
...
+        global_admin_must_resolve: in_dispute? && global_admin_must_resolve?,
+        missing_form: missing_form? && !problem?,
+        can_reconcile: can_reconcile_journaled?,
+        in_open_journal: in_open_journal?,
+        awaiting_payment: awaiting_payment?,
+      }.compact_blank.keys
+    end
+
+    ##
+    # Notices that cannot be stored since depend on current time
```
LeticiaErrandonea, line 33: Can we add a comment to `notices` method saying that this is used for calculating the stored notices? Or maybe it's too redundant? I think it could be good so just by seeing the file you know clearly the distinction between the two methods

app/models/order_detail.rb
```diff
@@ -602,7 +604,6 @@ def valid_reservation?
 
   def validate_service_meta
     return unless product.is_a?(Service)
-    return if being_purchased_by_admin && product.admin_skip_order_form
```
LeticiaErrandonea, line 605: I think this could make the `missing_form?` method behave differently and maybe cause the notice to appear in cases it shouldn't. `valid_service_meta?` used to be `true` this case but I think could be `false` now. Asking specially because of `OrderDetailObserver#ready_for_merge?`

LeticiaErrandonea (commented):
Left a couple of questions

app/models/order_detail.rb
joaquinco: yeas, it behaves different already so I left it that way. The attribute `being_purchased_by_admin` is only set on purchase and it's not stored, so the notice is already shown even though an order is purchased by an admin

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

diebas (approved):

PR #6127, opened 2026-04-17
[Fix] Some facility billing pages styles

joaquinco:
# Notes

There were some UI glitches to fix:

1. Reconcile Account Type page: notice overlapping filter button when there're no orders to reconcile
Before:
Captura desde 2026-04-17 09-35-37

2. Two columns layout: page content and side menu looked weird on sm and xs view ports. Make it look the same as two columns head layout: menu on top and content below, both with full width.
Before:
Captura desde 2026-04-17 10-53-54

After:
Captura desde 2026-04-17 10-54-22

3. Statement/invoice history page: filter form was narrower than others. Make it look the same as the rest. Also use chosen dynamic width, otherwise chosen inputs will have a fixed size.

Before:
Captura desde 2026-04-17 11-00-23

After:
Captura desde 2026-04-17 11-00-37

### DIFF
--- a/app/javascript/utils/activate_chosen.js
+++ b/app/javascript/utils/activate_chosen.js
@@ -4,9 +4,22 @@
  * Full docs: https://github.com/decaffeinate/decaffeinate/blob/main/docs/suggestions.md
  */
 window.ChosenActivator = class ChosenActivator {
+  static activateChosen(elem, options = null) {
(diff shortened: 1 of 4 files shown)


LeticiaErrandonea (approved):
LGTM!

diebas (approved):

PR #6128, opened 2026-04-17
[Feat] Statement show

joaquinco:
# Notes

Add Statement Show page. 

Ported from Nucore OSU.

# Screenshot

Captura desde 2026-04-17 14-37-01

### DIFF
--- a/app/views/facility_statements/show.html.haml
+++ b/app/views/facility_statements/show.html.haml
@@ -9,7 +9,7 @@
   = f.input :owner_user, input_html: { value: @statement.account.owner_user.full_name }
   = f.input :account
   = f.input :created_at, input_html: { value: I18n.l(@statement.created_at) }
-  = f.input :invoice_date, input_html: { value: I18n.l(@statement.invoice_date, format: :usa) }
+  = f.input :invoice_date, input_html: { value: I18n.l(@statement.invoice_date) }
   = f.input :status,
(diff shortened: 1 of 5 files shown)


LeticiaErrandonea (approved):
LGTM!

app/views/facility_statements/show.html.haml
```diff
@@ -1,25 +1,25 @@
...
+= content_for :h2 do
+  = "#{Statement.model_name.human} ##{@statement.invoice_number}"
 = content_for :sidebar do
-  = render :partial => 'admin/shared/sidenav_billing', :locals => { :sidenav_tab => 'statement_history' }
+  = render "admin/shared/sidenav_billing", sidenav_tab: "statement_history"
+
+= readonly_form_for(:statement) do |f|
+  = f.input :owner_user, input_html: { value: @statement.account.owner_user.full_name }
+  = f.input :account
+  = f.input :created_at, input_html: { value: I18n.l(@statement.created_at) }
```
diebas, line 11: Nit: Is it on purpose tho have both dates with different format?

diebas (approved):

app/views/facility_statements/show.html.haml
joaquinco: I've added a default format for dates so we can stop using `:usa` as a format since it does not make much sense to use locales like that

PR #6129, opened 2026-04-17
[Feat] Bulk imports

joaquinco:
# Notes

Add generic bulk import functionality. The section under Global Settings is only visible if there's a regisiterd bulk importer, which there is not.

Ported from Nucore OSU.

> [!IMPORTANT]
> This PR adds `@hotwired/{turbo-rails,stimulus}` dependencies

# Screenshot

A dummy importer "Some importer" was defined for the screenshots.

Index
Captura desde 2026-04-17 16-07-27

New
Captura desde 2026-04-24 15-56-14

Show
Captura desde 2026-04-17 16-08-32

### DIFF
--- a/db/migrate/20260312124843_create_bulk_import.rb
+++ b/db/migrate/20260312124843_create_bulk_import.rb
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateBulkImport < ActiveRecord::Migration[8.0]
+  def change
+    create_table :bulk_imports do |t|
+      t.string :import_type, null: false
+      t.string :status
+      t.string :created_by_id, null: false
+      t.text :data
+
+      t.timestamps
+    end
+
+    add_index :bulk_imports, [:created_at, :import_type]
+  end
+end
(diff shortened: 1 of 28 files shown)


LeticiaErrandonea (approved):

db/migrate/20260312124843_create_bulk_import.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class CreateBulkImport < ActiveRecord::Migration[8.0]
+  def change
+    create_table :bulk_imports do |t|
+      t.string :import_type, null: false
+      t.string :status
+      t.string :created_by_id, null: false
```
diebas: Is there a reason for this to be a string type? I think we use integer for type of values

spec/requests/bulk_imports_spec.rb
```diff
@@ -0,0 +1,225 @@
...
+      -> { post bulk_imports_path, params: }
+    end
+
+    after do
+      File.unlink(file.path) if file
+    end
+
+    it_behaves_like "require login"
+
+    context "with admin user" do
```
diebas, line 187: Does it add any value to test the non-admin scenario?

diebas (approved):

db/migrate/20260312124843_create_bulk_import.rb
joaquinco: I'll change it to int

PR #6132, opened 2026-04-20
[NUOPEN-339][Shared Dev] Schedule rule discount edit requires product pricing permission

joaquinco:
# Notes

Require "Product Pricing" permission to set schedule rules discounts. If a user can manage products but cannot edit pricing then schedule rules are editable but the discount section is not shown.

[NUOPEN-339 | Schedule rules discount permission require product_pricing](https://universe-of-universities.atlassian.net/browse/NUOPEN-339)


LeticiaErrandonea (approved):
LGTM!

diebas (approved):

PR #6140, opened 2026-04-22
[NUOPEN-330] Add read only granular permission

diebas:
# Notes                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                               
  Adds a new "Read Only" granular permission that allows a user to view a facility's Orders, Reservations, Occupancies, Products, and Projects without granting access to Billing, Admin, or Reporting areas.
                                                                                                                                                                                                                                                                               
  [NUOPEN-330 | Add Read only granular permission](https://universe-of-universities.atlassian.net/browse/NUOPEN-330)


# Screenshot

<img width="778" height="718" alt="Screenshot 2026-04-22 at 4 42 24 PM" src="https://github.com/user-attachments/assets/db665e45-69a7-4cca-8949-0dc10f038688" />

### DIFF
--- a/app/javascript/facility_user_permission_form.js
+++ b/app/javascript/facility_user_permission_form.js
@@ -1,17 +1,30 @@
 /*
  * Facility User Permission form
- * When any non-read-access permission is checked, auto-check Read Access.
- * Read Access is only unchecked when the admin does it manually.
+ * Read Access is required whenever any other permission is active.
+ * Checking another permission auto-checks Read Access and locks it
+ * (visually disabled and unclickable). When all other permissions
+ * are unchecked, Read Access becomes editable again.
  */
 $(document).ready(function() {
   const readAccessCheckbox = $(".js--readAccessCheckbox");
   const otherCheckboxes = $(".js--otherPermissionCheckbox");
 
-  if (!readAccessCheckbox.length) return;
-
-  otherCheckboxes.on("change", function() {
-    if ($(this).is(":checked")) {
+  function syncReadAccess() {
+    const anyOtherChecked = otherCheckboxes.is(":checked");
+    if (anyOtherChecked) {
       readAccessCheckbox.prop("checked", true);
+      readAccessCheckbox.attr("data-locked", "true").css("opacity", "0.5");
+    } else {
+      readAccessCheckbox.removeAttr("data-locked").css("opacity", "");
+    }
+  }
+
+  readAccessCheckbox.on("click", function(event) {
+    if ($(this).attr("data-locked") === "true") {
+      event.preventDefault();
     }
   });
+
+  otherCheckboxes.on("change", syncReadAccess);
+  syncReadAccess();
 });
--- a/app/javascript/facility_user_permission_form.js
+++ b/app/javascript/facility_user_permission_form.js
@@ -1,0 +1,17 @@
+/*
+ * Facility User Permission form
+ * When any non-read-access permission is checked, auto-check Read Access.
+ * Read Access is only unchecked when the admin does it manually.
+ */
+$(document).ready(function() {
+  const readAccessCheckbox = $(".js--readAccessCheckbox");
+  const otherCheckboxes = $(".js--otherPermissionCheckbox");
+
+  if (!readAccessCheckbox.length) return;
+
+  otherCheckboxes.on("change", function() {
+    if ($(this).is(":checked")) {
+      readAccessCheckbox.prop("checked", true);
+    }
+  });
+});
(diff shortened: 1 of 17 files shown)


db/migrate/20260422115005_add_read_only_to_facility_user_permissions.rb
```diff
@@ -0,0 +1,5 @@
+class AddReadOnlyToFacilityUserPermissions < ActiveRecord::Migration[8.0]
```
LeticiaErrandonea: nit: missing `frozen_string_literal` comment

spec/lib/ability_spec.rb
```diff
@@ -817,6 +818,31 @@
       it { is_expected.to be_allowed_to(:act_as, facility) }
       it { is_expected.to be_allowed_to(:switch, Instrument) }
     end
+
+    context "with only read_only (no other flags)" do
```
LeticiaErrandonea: Both `it_is_allowed_to ...` and `it { is_expected.to be_allowed_to ... }` are used. Is it possible to only use one way?

LeticiaErrandonea (approved):
LGTM!

Left a couple of comments

app/lib/ability.rb
```diff
@@ -477,6 +477,7 @@ def granted_permission_read_only_abilities(controller)
     can :index, [BundleProduct, ScheduleRule, ProductAccessory, ProductAccessGroup]
     can [:index], StoredFile
     can [:instrument_status, :instrument_statuses], Instrument
+    can [:index, :dashboard, :tab_counts, :show], SecureRooms::Occupancy
```
joaquinco: this should be placed in the secure rooms engine ability extension class. There're some schools that don't use secure rooms

app/javascript/facility_user_permission_form.js
```diff
@@ -0,0 +1,17 @@
+/*
+ * Facility User Permission form
+ * When any non-read-access permission is checked, auto-check Read Access.
+ * Read Access is only unchecked when the admin does it manually.
+ */
+$(document).ready(function() {
+  const readAccessCheckbox = $(".js--readAccessCheckbox");
+  const otherCheckboxes = $(".js--otherPermissionCheckbox");
+
+  if (!readAccessCheckbox.length) return;
```
joaquinco: let's let it break if not present

There're some js where early returns are performed if conditions don't apply but they are included in the application bundle so they are loaded on every page which is not the case here (and I think it's fine).

app/javascript/facility_user_permission_form.js
```diff
@@ -0,0 +1,17 @@
...
+
+  otherCheckboxes.on("change", function() {
+    if ($(this).is(":checked")) {
+      readAccessCheckbox.prop("checked", true);
+    }
```
joaquinco: what happens when read access is unchecked and there're other permissions active? in that case we end up with a user that has no read access but can write some stuff which would not work for how we handle manage facility pages

I think we should also prevent this case from happening at the model level

LeticiaErrandonea: Maybe when some other permission is already checked the Read access should be disabled

spec/lib/ability_spec.rb
```diff
@@ -817,6 +817,44 @@
...
+      it_is_not_allowed_to([:manage], FacilityUserPermission)
+      it_is_not_allowed_to([:manage, :adjust_price], OrderDetail)
+      it_is_not_allowed_to([:manage], Product)
+      it_is_not_allowed_to([:manage], OfflineReservation)
+      it { is_expected.not_to be_allowed_to(:manage_billing, facility) }
+      it { is_expected.not_to be_allowed_to(:edit, facility) }
+      it { is_expected.not_to be_allowed_to(:act_as, facility) }
+    end
+
+    context "with read_access: false but other flags true (master switch off)" do
```
LeticiaErrandonea: I think we could consider this case invalid. I don't really see any way when it could make sense

vendor/engines/secure_rooms/spec/models/ability_spec.rb
```diff
@@ -67,4 +67,24 @@
     it_is_allowed_to([:index, :dashboard, :tab_counts, :show_problems,
                       :assign_price_policies_to_problem_orders], SecureRooms::Occupancy)
   end
+
+  describe "granular permission user with read_access", feature_setting: { granular_permissions: true } do
+    let(:user) { FactoryBot.create(:user) }
```
LeticiaErrandonea: nit: can we avoid the `FactoryBot.` part?

LeticiaErrandonea (commented):
This is looking good! I left a couple more comments and I agree with Joaquín's

joaquinco (approved):
Looks good!

LeticiaErrandonea (approved):
LGTM!

PR #6167, opened 2026-05-04
[NUOPEN-344] Only allow order status update for order management permission

diebas:
# Notes                                                                                                                                                                                                                                                                    
         
  Fixes a granular-permissions bug where users with only the "Price adjustment" permission could change an order's status (e.g. mark Complete). Status changes now require the "Order management" permission.              
                                                                                                                                                                                                                                                                             
  [NUOPEN-344 — I should be able to change the status of an order only with the "Order management" permission](https://universe-of-universities.atlassian.net/browse/NUOPEN-344)

### DIFF
--- a/app/controllers/order_management/order_details_controller.rb
+++ b/app/controllers/order_management/order_details_controller.rb
@@ -15,6 +15,7 @@ class OrderManagement::OrderDetailsController < ApplicationController
 
   before_action :authorize_order_detail, except: %i(sample_results update remove_from_journal)
   before_action :authorize_order_detail_update, only: %i(update remove_from_journal)
+  before_action :authorize_change_status, only: :update
   before_action :authorize_mark_unrecoverable, only: :update
   before_action :load_accounts, only: [:edit, :update]
   before_action :load_order_statuses, only: [:edit, :update]
@@ -111,6 +112,7 @@ class OrderManagement::OrderDetailsController < ApplicationController
 
   def load_order_statuses
     return if @order_detail.reconciled?
+    return if cannot?(:change_status, @order_detail)
 
     if @order_detail.complete?
       @order_statuses = OrderStatus.by_names([
@@ -167,6 +169,17 @@ class OrderManagement::OrderDetailsController < ApplicationController
     return if @order_detail.unrecoverable? || update_params[:order_status_id].to_s != OrderStatus.unrecoverable.id.to_s
 
     authorize!(:mark_unrecoverable, OrderDetail)
+  end
+
+  def authorize_change_status
+    return unless changing_order_status?
+
+    authorize!(:change_status, @order_detail)
+  end
+
+  def changing_order_status?
+    new_status = update_params[:order_status_id]
+    new_status.present? && new_status.to_s != @order_detail.order_status_id.to_s
   end
 
 end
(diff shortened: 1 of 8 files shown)


app/controllers/order_management/order_details_controller.rb
```diff
@@ -104,13 +107,18 @@ def read_only?
     cannot?(:update, @order_detail)
   end
 
+  def metadata_edit_disabled?
+    read_only? || cannot?(:manage_order_details, @order_detail)
```
joaquinco: can we make read only imply that the user cannot :manage_order_details? I would expect read only to mean cannot edit data nor price

app/controllers/order_management/order_details_controller.rb
```diff
@@ -158,15 +166,31 @@ def init_order_detail
 
   def update_params
     raw_params = params[:order_detail] || empty_params
-    return raw_params if can?(:adjust_price, @order_detail)
+    return raw_params.slice(*PRICE_ADJUSTMENT_ATTRIBUTES) unless can?(:manage_order_details, @order_detail)
 
-    raw_params.except(:actual_cost, :actual_subsidy)
+    raw_params = raw_params.except(:actual_cost, :actual_subsidy) unless can?(:adjust_price, @order_detail)
+    raw_params
```
joaquinco: nit: let's use an if else to make the outcome more clear.

```
if cannot?(:edit)
  only_pricing_params
elsif cannot?(:adjust_price)
  only_editting_params
else
  all_params
end
```

app/controllers/order_management/order_details_controller.rb
```diff
@@ -158,15 +166,31 @@ def init_order_detail
 
   def update_params
     raw_params = params[:order_detail] || empty_params
-    return raw_params if can?(:adjust_price, @order_detail)
+    return raw_params.slice(*PRICE_ADJUSTMENT_ATTRIBUTES) unless can?(:manage_order_details, @order_detail)
 
-    raw_params.except(:actual_cost, :actual_subsidy)
+    raw_params = raw_params.except(:actual_cost, :actual_subsidy) unless can?(:adjust_price, @order_detail)
```
joaquinco: should we use PRICE_ADJUSTMENT_ATTRIBUTES here?

joaquinco (commented):
It's looking good, left some comments

app/lib/ability.rb
```diff
@@ -414,6 +414,10 @@ def facility_granted_permission_abilities(user, resource, controller)
     if permission.price_adjustment?
       can :adjust_price, OrderDetail
       can :manage, OrderDetail
+      unless permission.order_management?
+        cannot :change_status, OrderDetail
+        cannot :manage_order_details, OrderDetail
```
joaquinco: is the `change_status` permission needed? maybe we can simplify this having `can :edit, OrderDetail` and define a custom permission for price_adjustment, like: `can :adjust_price, OrderDetail`?

```ruby
if order_management?
  can :edit, OrderDetail
end
if price_adjustment?
  can :adjust_price, OrderDetail
end
```

For these permissions we should try to avoid `:manage` since it'll allow every action on a resource, I think granting `:edit` (or `:update`) will not imply `:adjust_price`. This should be preferred so we don't give unforeseen permissions on resources.

joaquinco (approved):
Looks good, thanks for the changes!

LeticiaErrandonea (approved):

PR #6168, opened 2026-05-04
[NUOPEN-343] Cover sanger with granular permissions

diebas:
# Notes
                                                                                                                                                                                                                                                                               
  Users with the `product_management` granular permission can now access the Sanger module — both the facility-level navbar tab and the product-level Sanger tab. Previously they hit a 403 because Sanger's authorization did not account for granular permissions.
                                                                                                                                                                                                                                                                               
  [NUOPEN-343 | A 403 error happens when a user with "active" granular permissions try to access the Sanger tab in the navbar](https://universe-of-universities.atlassian.net/browse/NUOPEN-343)


joaquinco (commented):
Looks good. While testing it I noticed two things:

- the sanger tab at the facility level is still visible when the user does not have access
- the sanger tab on services (requires ff sanger_enabled_service: true) is not accessible by a user with product management permission

joaquinco (approved):

PR #6179, opened 2026-05-11
[NU-444] Timeslot available notification

joaquinco:
# Notes

Allow instrument to be configured for notify users on reservation cancellation.

- Add `ProductNotifiation` model to store notifications for products
- Notify users accordingly if a reservation is canceled and the product notification is configured

> Note: Changes in views will be done in an upcomming PR.

[NU-444 | Notify users on canceled reservation](https://universe-of-universities.atlassian.net/browse/NU-444)

## Other changes

- Change `Reservation.upcoming` to return an active record relation instead of an array. It used an array filter because of an oracle issue but it seems to be from long ago (2013).

### DIFF
--- a/app/mailers/product_notification_mailer.rb
+++ b/app/mailers/product_notification_mailer.rb
@@ -2,7 +2,7 @@
 
 class ProductNotificationMailer < ApplicationMailer
   ##
-  # An future time slot has become available
+  # A future time slot has become available
   def slot_available(product, user, start_time, end_time)
     @product = product
     @user = user
--- a/app/mailers/product_notification_mailer.rb
+++ b/app/mailers/product_notification_mailer.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+class ProductNotificationMailer < ApplicationMailer
+  ##
+  # An future time slot has become available
+  def slot_available(product, user, start_time, end_time)
+    @product = product
+    @user = user
+    @time_range = TimeRange.new(start_time, end_time)
+
+    mail(
+      to: user.email,
+      subject: text(
+        "product_notification_mailer.slot_available.subject",
+        product_facility: "#{product} (#{product.facility.abbreviation})",
+        time_range: @time_range,
+      ),
+    )
+  end
+end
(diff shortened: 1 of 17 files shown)


app/mailers/product_notification_mailer.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class ProductNotificationMailer < ApplicationMailer
+  ##
+  # An future time slot has become available
```
LeticiaErrandonea: nit: 
```suggestion
  # A future time slot has become available
```

app/models/product_notification.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class ProductNotification < ApplicationRecord
+  belongs_to :product
+
+  DEFAULT_RESERVATION_DAYS = 15
```
LeticiaErrandonea, line 6: Should we confirm this default?

LeticiaErrandonea (approved):
LGTM!

config/locales/views/en.product_notification_mailer.yml
```diff
@@ -0,0 +1,16 @@
+en:
+  product_notification_mailer:
+    slot_available:
+      subject: "!app_name! - New availability %{product_facility} - %{time_range}"
+      body_html: |
+        Hello %{user_name},
+
+        Time slot %{time_range} became available for %{product} at %{facility}.
+
+        [Make a reservation](%{new_reservation_url}Make a reservation)
```
diebas: The link text is repeated.

 ` [Make a reservation](%{new_reservation_url})`

config/locales/views/en.product_notification_mailer.yml
```diff
@@ -0,0 +1,16 @@
+en:
+  product_notification_mailer:
+    slot_available:
+      subject: "!app_name! - New availability %{product_facility} - %{time_range}"
+      body_html: |
```
diebas: Is it necessary to have body_html and body inside the mailer? Other mailers use a single body section

diebas (approved):

config/locales/views/en.product_notification_mailer.yml
joaquinco: I went through the rails way initially and then noticed markdown support required the text_helpers gem and it handled plain text conversion as well. So it's not necessary, I'll change it.

PR #6186, opened 2026-05-13
[OSU-223] Handle job retries with active job

joaquinco:
# Notes

[OSU-223 | Handle Job retries](https://universe-of-universities.atlassian.net/browse/OSU-223)

Handle job retries purely with with active job. The `max_attempts` attribute previously defined in jobs did not work. This change allow us to define the retry mechanism on a per job basis. 

Delayed job retry mechanism conflicts active job's. If we use both, AJ retries will catch exceptions and reschedule the job until the maximum attempts is reached, in which case the exception bubbles up and the DJ retry mechanism kicks in.

So, in order to further decouple from delayed job I've disabled retries there.

Also, disabled `BulkImportJob` retries.

## Job failure debugging

AJ retry mechanism reschedule a new job, so from the point of view of `Delayed::Job` the attempts fields will always be 0. The information about job executions is stored in the handler field which contains active job object serialized.

### DIFF
--- a/app/jobs/application_job.rb
+++ b/app/jobs/application_job.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 class ApplicationJob < ActiveJob::Base
-
+  retry_on StandardError, attempts: 5
 end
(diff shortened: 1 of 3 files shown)


diebas (approved):

LeticiaErrandonea (approved):
LGTM

PR #6193, opened 2026-05-15
[NUOPEN-338] Split product management permission

diebas:
# Notes
  
  Splits the Product Management granular permission into two separate permissions — Product Creation and Product Edition — so facility admins can grant the right to edit products without granting the right to create them. Users who can create products can also edit them.
  
  [NUOPEN-338 | Split Product Management granular permission](https://universe-of-universities.atlassian.net/browse/NUOPEN-338)

### DIFF
--- a/app/helpers/facility_user_permissions_helper.rb
+++ b/app/helpers/facility_user_permissions_helper.rb
@@ -9,4 +9,11 @@ module FacilityUserPermissionsHelper
     end.join(", ")
   end
 
+  def permission_checkbox_classes(perm)
+    classes = [perm == :read_access ? "js--readAccessCheckbox" : "js--otherPermissionCheckbox"]
+    classes << "js--productCreationCheckbox" if perm == :product_creation
+    classes << "js--productEditionCheckbox" if perm == :product_edition
+    classes.join(" ")
+  end
+
 end
--- a/app/helpers/facility_user_permissions_helper.rb
+++ b/app/helpers/facility_user_permissions_helper.rb
@@ -10,9 +10,10 @@ module FacilityUserPermissionsHelper
   end
 
   def permission_checkbox_classes(perm)
-    classes = [perm == :read_access ? "js--readAccessCheckbox" : "js--otherPermissionCheckbox"]
-    classes << "js--productCreationCheckbox" if perm == :product_creation
-    classes << "js--productEditionCheckbox" if perm == :product_edition
+    return "js--readAccessCheckbox" if perm == :read_access
+
+    classes = ["js--otherPermissionCheckbox"]
+    classes << "js--#{perm.to_s.camelize(:lower)}Checkbox" if %i[product_creation product_edition].include?(perm)
     classes.join(" ")
   end
 
--- a/app/javascript/facility_user_permission_form.js
+++ b/app/javascript/facility_user_permission_form.js
@@ -1,30 +1,58 @@
 /*
  * Facility User Permission form
- * Read Access is required whenever any other permission is active.
- * Checking another permission auto-checks Read Access and locks it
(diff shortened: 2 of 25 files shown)


vendor/engines/sanger_sequencing/app/models/sanger_sequencing/ability.rb
```diff
@@ -11,19 +11,19 @@ def initialize(user, facility = nil)
...
+      if facility && (user.operator_of?(facility) || granted_product_edition?(user, facility))
         can [:index, :show], Submission
         can :manage, [Batch, BatchForm, Primer]
       end
     end
 
     private
 
-    def granted_product_management?(user, facility)
+    def granted_product_edition?(user, facility)
```
LeticiaErrandonea, line 22: Should `product_creation` be checked?

LeticiaErrandonea (commented):
Looking good!

I wonder if we should set `product_edition` to true when `product_creation` is set. That way I think some checks would be simplified. WDYT?

app/javascript/facility_user_permission_form.js
```diff
@@ -1,30 +1,58 @@
...
  * Facility User Permission form
- * Read Access is required whenever any other permission is active.
- * Checking another permission auto-checks Read Access and locks it
- * (visually disabled and unclickable). When all other permissions
- * are unchecked, Read Access becomes editable again.
+ *
+ * Two invariants are enforced visually here (and server-side in the model):
+ * - Read Access is required whenever any other permission is active.
+ *   Checking another permission auto-checks Read Access and locks it
+ *   (visually disabled and unclickable).
```
LeticiaErrandonea:
```suggestion
 * - Read Access is required whenever any other permission is active.
 *   Checking another permission auto-checks Read Access and locks it
 *   (visually disabled and unclickable). When all other permissions
 *   are unchecked, Read Access becomes editable again.
```

app/lib/ability.rb
```diff
@@ -375,12 +375,17 @@ def facility_granted_permission_abilities(user, resource, controller)
         OfflineReservation,
       ]
       cannot :create_daily_booking, Product
+      cannot :create, Product
```
LeticiaErrandonea: What about adding the check here and avoiding having the `cannot :create, Product` and 10 lines below `can :create, Product`?

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

Left a couple  of small comments

app/helpers/facility_user_permissions_helper.rb
```diff
@@ -7,4 +7,11 @@ def permission_summary(permission)
     active.map { |perm| I18n.t("views.facility_user_permissions.edit.permission_labels.#{perm}") }.join(", ")
   end
 
+  def permission_checkbox_classes(perm)
+    classes = [perm == :read_access ? "js--readAccessCheckbox" : "js--otherPermissionCheckbox"]
+    classes << "js--productCreationCheckbox" if perm == :product_creation
+    classes << "js--productEditionCheckbox" if perm == :product_edition
+    classes.join(" ")
```
joaquinco: this returns a single css class, right?

Maybe it can be simplified like:

```ruby
return "js--otherPermissionCheckbox" if %i[read_access product_creation product_edition].exclude?(perm)

"js--#{perm.camelize(:lower)}Checkbox"
```

app/lib/ability.rb
joaquinco: I think we should use `can :update, Product` instead of `:manage` which implies every permission over Product. Then we should be able to remove this cannot

joaquinco (approved):
Looks good!

PR #6208, opened 2026-05-21
[NUOPEN-349] Fix reservation cross origin

diebas:
# Notes
  
  Fixes the `ActionController::InvalidCrossOriginRequest` exception from the public instrument schedule across every school. The calendar events endpoint now responds as JSON instead of JS — same body, but no longer subject to Rails' cross-origin protection.
  
  [NUOPEN-349 | [ERROR] reservations#index (ActionController::InvalidCrossOriginRequest)](https://universe-of-universities.atlassian.net/browse/NUOPEN-349)

### DIFF
--- a/app/controllers/facility_accounts_controller.rb
+++ b/app/controllers/facility_accounts_controller.rb
@@ -149,7 +149,7 @@ class FacilityAccountsController < ApplicationController
 
   def accounts_available_for_order
     respond_to do |format|
-      format.js do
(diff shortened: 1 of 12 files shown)


joaquinco (approved):
Nice catch. There're some other endpoints using the js format. I think we should update them too.

LeticiaErrandonea (approved):
LGTM!

PR #6209, opened 2026-05-22
[NUOPEN-335] Project management granular permission

diebas:
# Notes
  
  Adds a new "Project Management" granular permission that lets facility staff grant non-admin users access to create, view, edit, and update projects (including the cross-core orders view) within a facility.
  
  [NUOPEN-335 — Add Managing Projects granular permission](https://universe-of-universities.atlassian.net/browse/NUOPEN-335)

### DIFF
--- a/app/lib/ability.rb
+++ b/app/lib/ability.rb
@@ -341,6 +341,10 @@ class Ability
       granted_permission_reservation_abilities(permission)
     end
 
+    if resource.is_a?(Project) && permission.project_management?
+      can [:show, :edit, :update], Project
+    end
+
     return unless resource.is_a?(Facility)
 
     granted_permission_read_access_abilities(controller)
@@ -430,6 +434,10 @@ class Ability
       can :manage, Reports::ReportsController
     end
 
+    if permission.project_management?
+      can [:create, :new, :cross_core_orders], Project
+    end
+
     granted_permission_user_management_abilities(permission, resource, controller)
   end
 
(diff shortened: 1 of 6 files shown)


spec/lib/ability_spec.rb
```diff
@@ -879,6 +868,40 @@
...
+    context "with project_management" do
+      before do
+        user.facility_user_permissions.where(facility:).update!(project_management: true)
+      end
+
+      it_is_allowed_to([:create, :new, :cross_core_orders], Project)
+
+      it "does not allow switching to another user" do
+        expect(ability).not_to be_allowed_to(:switch_to, create(:user))
+      end
```
LeticiaErrandonea: Is there any reason for testing this? I think it's unrelated

LeticiaErrandonea (approved):
Left a comment, but it LGTM

app/lib/ability.rb
```diff
@@ -341,6 +341,10 @@ def facility_granted_permission_abilities(user, resource, controller)
       granted_permission_reservation_abilities(permission)
     end
 
+    if resource.is_a?(Project) && permission.project_management?
+      can [:show, :edit, :update], Project
+    end
```
joaquinco, line 346: Can we use the facility as ability resource so the permission setup is more uniform?

then we would also add a condition on Projects: `can [...], Project, facility_id: facility.id`

joaquinco (approved):
Looks good. Just one minor suggestion

PR #6222, opened 2026-05-27
[NUOPEN-300] Refactor github action pipelines

diebas:
# Notes

TECH TASK: Refactors the GitHub Actions test suite so assets are built once and shared across jobs, and the main rspec suite runs as a 4-way sharded matrix. This cuts CI wall-clock from ~11 min to ~7.5 min and makes re-running a failed (e.g. flaky) job skip the rebuild.
  
  [NUOPEN-300 | Refactor Github actions pipelines](https://universe-of-universities.atlassian.net/browse/NUOPEN-300)
  
  # Additional Context
  
  Previously each test job re-ran the full setup independently (`yarn build` ×3, `assets:precompile` ×2) and the main rspec suite ran as a single 4-process job.
  
  ### What changed
  
  **Shared build step**
  - New `build-assets` job runs `yarn build` + `assets:precompile` once and uploads a `compiled-assets` artifact (`public/assets` + `app/assets/builds`, `include-hidden-files: true` for the sprockets manifest).
  - `run-rspec`, `run-rspec-engines`, and `run-teaspoon` now `needs: build-assets` and download the artifact instead of rebuilding. Removed `run-rspec`'s redundant `db:create db:schema:load` and `run-teaspoon`'s duplicate `yarn build`.
  - New `setup-deps` composite action (deps + config only) replaces `setup-rails`, which conflated dependency install, DB creation, and asset building.
  
  **Sharded rspec**
  - `run-rspec` is now a `strategy.matrix` of 4 nodes (`ci_node: 0..3`, `fail-fast: false`). Each node runs `parallel_test spec --type rspec -n 16 --only-group <4 groups> --group-by filesize`, giving 16-way parallelism across machines. `run-rspec` critical path drops from ~9m39 to ~6m (slowest shard). Verified the 4 shards run the full suite (sum of examples unchanged, 0 failures).
  
  **Caching cleanup**
  - Bumped `actions/cache@v3 → @v4` (v3 is deprecated) in `setup-bundler` and `setup-node`.
  - Added a bootsnap cache (`tmp/cache/bootsnap`, keyed on `Gemfile.lock`) in `setup-deps`.
  - Fixed the mislabeled `setup-node` cache key (`-bundle-` → `-node-`, a copy-paste from the bundler action).

### DIFF
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -73,6 +73,10 @@ jobs:
     needs: build-assets
     runs-on: ubuntu-latest
     timeout-minutes: 60
+    strategy:
+      fail-fast: false
+      matrix:
+        ci_node: [0, 1, 2, 3]
     services:
       mysql:
         image: mysql:8.0.28
@@ -102,7 +106,12 @@ jobs:
       # - uses: mxschmitt/action-tmate@v3
       - run: bundle exec rails parallel:create
       - run: bundle exec rails parallel:load_schema
-      - run: bundle exec rails parallel:spec
+      - name: Run rspec shard ${{ matrix.ci_node }} of 4
+        run: |
+          start=$(( ${{ matrix.ci_node }} * 4 + 1 ))
+          groups=$(seq -s, "$start" "$(( start + 3 ))")
+          echo "Shard ${{ matrix.ci_node }}: running groups $groups of 16"
+          bundle exec parallel_test spec --type rspec -n 16 --only-group "$groups" --group-by filesize
 
   run-teaspoon:
     needs: build-assets
--- a/.github/workflows/test-suite.yaml
+++ b/.github/workflows/test-suite.yaml
@@ -8,7 +8,30 @@ on:
   pull_request:
 
 jobs:
+  build-assets:
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    container:
+      image: wyeworkshub/ruby-node-chrome-pack:ruby3.4.4-node22
+    env:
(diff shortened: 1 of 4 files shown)


.github/workflows/test-suite.yaml
```diff
@@ -65,15 +97,24 @@ jobs:
...
       - run: bundle exec rails parallel:create
       - run: bundle exec rails parallel:load_schema
-      - run: bundle exec rails assets:precompile
-      - run: bundle exec rails parallel:spec
+      - name: Run rspec shard ${{ matrix.ci_node }} of 4
+        run: |
+          start=$(( ${{ matrix.ci_node }} * 4 + 1 ))
+          groups=$(seq -s, "$start" "$(( start + 3 ))")
+          echo "Shard ${{ matrix.ci_node }}: running groups $groups of 16"
+          bundle exec parallel_test spec --type rspec -n 16 --only-group "$groups" --group-by filesize
```
joaquinco, line 114: the env variable `PARALLEL_TEST_PROCESSORS` is overriden here, can we remove it?

joaquinco (approved):

LeticiaErrandonea (approved):
LGTM! 🚀

.github/workflows/test-suite.yaml
diebas: The -n 16 overrides it on that line, but parallel:create and parallel:load_schema above still read it to know how many DBs to spin up. Without it they fall back to the runner's CPU count, which happens to be 4 today but if that changes in the future it would load more DB's than necessary.

PR #6223, opened 2026-05-27
[NUOPEN-353] Fix product purchase checks with granular permissions

joaquinco:
# Notes

Update `ProductForCart` class which perform some checks before purchasing to use the ability instead of relying on fixed user roles. The permission verified to tell if someone is an operator of a facility is :act_as, Facility, which is added for all operators of fixed user roles and some granular permissions.

Also:
- Added permission over `Reservation` missed in the Granular permission refactor which added a required permission to order for (reservation creation) (https://github.com/wyeworks/nucore-open/pull/6207)
- Update `:use_test_account` spec filter to add `TestAccount` to `Account.config`

[NUOPEN-353 | Order for another user not fully working](https://universe-of-universities.atlassian.net/browse/NUOPEN-353)

### DIFF
--- a/spec/models/product_for_cart_spec.rb
+++ b/spec/models/product_for_cart_spec.rb
@@ -4,7 +4,7 @@ require "rails_helper"
 
 RSpec.describe ProductForCart do
   let(:facility) { create(:setup_facility) }
-  let(:item) { create(:setup_item, facility: facility) }
+  let(:item) { create(:setup_item, facility:) }
   let(:user) { create(:user) }
   let(:ability) { Ability.new(user, facility, ApplicationController.new) }
   let(:product_for_cart) { described_class.new(item, user, ability) }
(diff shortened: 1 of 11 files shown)


spec/models/product_for_cart_spec.rb
```diff
@@ -124,11 +122,51 @@
...
       it "sets error_message explaining that we could not find a valid payment source" do
-        product_for_cart.purchasable_by?(user, user)
+        product_for_cart.purchasable_by?(user)
         expect(product_for_cart.error_message).to match(/could not find a valid payment source/)
       end
     end
-
   end
 
+  describe "ordering for", :use_test_account do
```
LeticiaErrandonea, line 131: Does it make sense to explicitly set the FF here? I mean both ON/OFF, to be safe regarding Prod?

LeticiaErrandonea (approved):
LGTM, left a comment

diebas (approved):

PR #6239, opened 2026-06-04
[NUOPEN-281] Handle statement cancel order detail notices update

joaquinco:
# Notes

Show `awaiting payment` notice order detail as a computed notice since it's order callbacks are skipped when a statement is cancelled and it can be computed out of order attributes.

[NUOPEN-281 | Order details badges](https://universe-of-universities.atlassian.net/browse/NUOPEN-281)


LeticiaErrandonea (approved):
LGTM!

diebas (approved):

PR #6252, opened 2026-06-09
[NUOPEN-326] Fix price policy trunaction

diebas:
# Notes
  
  Adds a guard so that creating a new price policy can no longer silently truncate an existing, currently-active policy that already has orders assigned to it. Instead of leaving those orders pointing at a policy whose date range was shifted out from under them, the system now blocks the create and shows an error. Price policies are also now audited with PaperTrail.
  
  [NUOPEN-326 | Price Policy truncation result in wrong states](https://universe-of-universities.atlassian.net/browse/NUOPEN-326)

### DIFF
--- a/app/models/price_policy.rb
+++ b/app/models/price_policy.rb
@@ -4,6 +4,8 @@ class PricePolicy < ApplicationRecord
 
   include Nucore::Database::DateHelper
 
+  has_paper_trail
+
   belongs_to :price_group
   belongs_to :product, optional: true
   belongs_to :created_by, class_name: "User", optional: true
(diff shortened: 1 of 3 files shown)


app/models/price_policy.rb
```diff
@@ -4,6 +4,8 @@ class PricePolicy < ApplicationRecord
 
   include Nucore::Database::DateHelper
 
+  has_paper_trail
```
joaquinco, line 7: 👏🏽

app/models/price_policy.rb
```diff
@@ -238,16 +241,26 @@ def subsidy_less_than_rate
...
+  end
 
-    existing_policies.each do |policy|
-      policy.expire_date = (start_date - 1.day).end_of_day
-      policy.save
-    end
+  def no_truncation_of_assigned_policies
+    return unless OrderDetail.exists?(price_policy_id: policies_to_truncate.select(:id))
+
+    errors.add(:base, "cannot overlap an existing price policy that has orders assigned to it")
```
joaquinco: let's move this error message to locales

joaquinco (approved):
Looks good!

LeticiaErrandonea (approved):
LGTM!

PR #6257, opened 2026-06-10
[NUOPEN-24] Improve billing transactions page load

joaquinco:
# Notes

- General improvement: don't try to find acting user by nil id
- Transactions Project filter options: the query can take a large amount of time even though there are a handful of Projects. Simplify query to use projects from all facility orders instead of the filtered subset of orders. When the facility is cross facility then use the existing logic
- Add `orders_count` counter cache on Projects

Part of [NUOPEN-24 - Transactions page load times](https://universe-of-universities.atlassian.net/browse/NUOPEN-24)

## Project counter cache backfill

```ruby
Project.all.pluck(:id).each do |proj_id| Project.reset_counters(proj_id, :orders_count) end
```

### DIFF
--- a/app/services/projects_search/project_searcher.rb
+++ b/app/services/projects_search/project_searcher.rb
@@ -5,11 +5,24 @@ module ProjectsSearch
   class ProjectSearcher < TransactionSearch::BaseSearcher
 
     def self.key
-      :projects
+      "projects"
     end
 
     def options
-      Project.where(id: order_details.select("distinct project_id")).order(:name)
+      projects =
+        if current_facility.cross_facility?
+          Project.all
+        else
+          project_ids =
+            current_facility
+            .order_details
+            .where.not(project_id: nil)
+            .select(:project_id)
+
+          Project.where(id: project_ids)
+        end
+
+      projects.order(:name)
     end
 
     def search(params)
--- a/app/services/projects_search/project_searcher.rb
+++ b/app/services/projects_search/project_searcher.rb
@@ -5,7 +5,7 @@ module ProjectsSearch
   class ProjectSearcher < TransactionSearch::BaseSearcher
 
     def self.key
-      "projects"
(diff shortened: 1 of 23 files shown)


app/services/projects_search/project_searcher.rb
```diff
@@ -9,7 +9,20 @@ def self.key
     end
 
     def options
-      Project.where(id: order_details.select("distinct project_id")).order(:name)
+      projects =
```
LeticiaErrandonea, line 12: I think we need to add `includes(:orders)` in order to avoid N+1 (`project.name` checks `orders.any?`)

app/services/transaction_search/searcher.rb
LeticiaErrandonea: We need a facility now

app/services/projects_search/project_searcher.rb
```diff
@@ -9,7 +9,20 @@ def self.key
...
+        if current_facility.cross_facility?
+          Project.all
+        else
+          project_ids =
+            current_facility
+            .order_details
+            .where.not(project_id: nil)
+            .select(:project_id)
```
LeticiaErrandonea, line 20: I think we're missing `.distinct`

LeticiaErrandonea (commented):
Dev QA:
As a PI, I see all existing projects in `/transactions` dropdown

Image

diebas (approved):

app/services/projects_search/project_searcher.rb
joaquinco: this is used as a subquery so I don't think it's really needed

app/services/projects_search/project_searcher.rb
joaquinco: that's unexpected

LeticiaErrandonea (approved):

PR #6263, opened 2026-06-11
[NUOPEN-326] Fix truncation overlap

diebas:
# Notes

  Re-applies the guard that prevents a new price policy from silently truncating an existing, currently-active policy that already has orders assigned to it (showing an error instead). This version fixes a regression from the previously reverted PR (#6252): creation is now only blocked when the new policy's date range actually overlaps an existing policy with orders, so creating a non-overlapping policy (e.g. for a future fiscal year) works again even when older policies have orders. Price policies are also audited with PaperTrail.
  
  [NUOPEN-326 | Price Policy truncation result in wrong states](https://universe-of-universities.atlassian.net/browse/NUOPEN-326)

### DIFF
--- a/app/models/price_policy.rb
+++ b/app/models/price_policy.rb
@@ -4,6 +4,8 @@ class PricePolicy < ApplicationRecord
 
   include Nucore::Database::DateHelper
 
+  has_paper_trail
+
   belongs_to :price_group
   belongs_to :product, optional: true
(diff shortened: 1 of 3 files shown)


LeticiaErrandonea (approved):
LGTM!

joaquinco (approved):

PR #6267, opened 2026-06-12
[NU-444] Notify time slot available on reservation cancel

joaquinco:
# Notes

Add `ProductNotification` model and use it to trigger notifications on reservation cancel.

Part of [NU-444 | An automatic email that would notify user of canceled reservations.](https://universe-of-universities.atlassian.net/browse/NU-444).

### DIFF
--- a/app/controllers/order_details_controller.rb
+++ b/app/controllers/order_details_controller.rb
@@ -37,12 +37,16 @@ class OrderDetailsController < ApplicationController
     if @order_detail.reservation && @order_detail.reservation.can_cancel?
       @order_detail.transaction do
         if @order_detail.cancel_reservation(session_user)
-          CancellationMailer.notify_facility(@order_detail).deliver_later
           flash[:notice] = text("cancel.success")
         else
           flash[:error] = text("cancel.error")
           raise ActiveRecord::Rollback
         end
+      end
+
+      if @order_detail.canceled? && @order_detail.status_recently_changed
+        CancellationMailer.notify_facility(@order_detail).deliver_later
+        ProductNotifications::SlotAvailableService.from_reservation(@order_detail.reservation).notify!
       end
 
       redirect_to redirect_to_path
--- a/app/models/order_detail.rb
+++ b/app/models/order_detail.rb
@@ -81,6 +81,9 @@ class OrderDetail < ApplicationRecord
 
   alias_attribute :problem_description_keys, :problem_keys
 
+  # Used to track status changes side efects
+  attribute :status_recently_changed, :boolean, default: false
+
   def estimated_price_group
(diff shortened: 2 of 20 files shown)


app/services/product_notifications/slot_available_service.rb
```diff
@@ -0,0 +1,57 @@
...
+    end
+
+    private
+
+    def recipients
+      return User.none if product_notifications.blank?
+
+      user_ids =
+        product_notifications
+        .joins("INNER JOIN product_notifications_users")
```
LeticiaErrandonea: Missing `ON`?

app/models/order_detail.rb
```diff
@@ -81,6 +81,9 @@ class OrderDetail < ApplicationRecord
 
   alias_attribute :problem_description_keys, :problem_keys
 
+  # Used to track status changes side efects
+  attribute :status_recently_changed, :boolean, default: false
```
LeticiaErrandonea, line 85: Would `saved_change_to_order_status_id?` also work?

db/schema.rb
```diff
@@ -671,6 +671,31 @@
     t.index ["facility_id"], name: "index_product_display_groups_on_facility_id"
   end
 
+  create_table "product_notifications", charset: "utf8mb3", force: :cascade do |t|
```
LeticiaErrandonea, line 674: The last notes from a meeting in the ticket mention a subject input. Is that no longer needed?

LeticiaErrandonea (approved):
LGTM

I left a couple of comments.

Also, are you planning on adding the admin holds cancellation in another PR?

app/models/order_detail.rb
joaquinco: The places where order details are updated perform several saves so it's not possible to identify the status change by using dirty attribute methods. So I ended up adding this

db/schema.rb
joaquinco: good point

joaquinco:
@LeticiaErrandonea Yes, notifications on admin hold will come in another PR

app/controllers/order_details_controller.rb
```diff
@@ -37,14 +37,18 @@ def cancel
...
         if @order_detail.cancel_reservation(session_user)
-          CancellationMailer.notify_facility(@order_detail).deliver_later
           flash[:notice] = text("cancel.success")
         else
           flash[:error] = text("cancel.error")
           raise ActiveRecord::Rollback
         end
       end
 
+      if @order_detail.canceled? && @order_detail.status_recently_changed
```
diebas, line 47: I think that with this change, the notification is not triggered when cancellation generates a fee. Is that intended?

diebas (approved):

PR #6276, opened 2026-06-15
[NU-444] Product notifications add/edit

joaquinco:
# Notes

Add/Edit `ProductNotification`.

Part of [NU-444 | An automatic email that would notify user of canceled reservations](https://universe-of-universities.atlassian.net/browse/NU-444)

# Additional Context

# Screenshot

Captura desde 2026-06-15 15-55-37

### DIFF
--- a/app/controllers/facility_product_notifications_controller.rb
+++ b/app/controllers/facility_product_notifications_controller.rb
@@ -67,13 +67,20 @@ class FacilityProductNotificationsController < ApplicationController
   private
 
   def load_instruments
-    @instruments = current_facility.products.active.not_archived.of_type("Instrument").alphabetized
+    @instruments =
+      current_facility
+      .products
+      .active
+      .not_archived
+      .of_type(Instrument)
+      .alphabetized
   end
 
   def facility_product_notification_params
     params.require(:product_notification).permit(
       :name,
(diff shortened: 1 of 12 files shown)


app/controllers/facility_product_notifications_controller.rb
```diff
@@ -53,16 +57,37 @@ def index
...
     @product_notification.destroy
     flash[:notice] = t(".success")
-    redirect_to :index
+    redirect_to action: :index
+  end
+
+  def user_search
+    @users = UserFinder.search(
+      params[:search_term],
+      limit: 10,
```
LeticiaErrandonea, line 66: Given we're limiting the users, should we add a hint saying that if they can't find the user, they need to add the full name? I don't think it's necessary but just checking.

LeticiaErrandonea (approved):
LGTM! 🙌

app/controllers/facility_product_notifications_controller.rb
joaquinco: Since it's the same search criteria as other user searches I think it's ok without that hint. I could add a notice that only active users are searched but that's kind of expected I think

diebas (approved):

PR #6277, opened 2026-06-15
[NUOPEN-357] Allow Read-only to see price groups and Cross-core projects

LeticiaErrandonea:
# Notes
* Allow Read-only to see price groups and Cross-core projects

[NUOPEN-357 | Read-Only Permission ](https://universe-of-universities.atlassian.net/browse/NUOPEN-357)

# Additional Context
Also fixes the bug when users without access to seeing the Accounts still got a link to them.

# Screenshot
## Price groups
image

## Projects
image

## Facility Staff hiden
image


joaquinco (approved):
Looks good.

One related thing though, the Facility Staff page is shown empty when the user does not have permission. Should we hide the menu entry?

diebas (approved):

joaquinco (approved):
thanks for the change!

PR #6283, opened 2026-06-17
[NU-455] Add rack attack

diebas:
# Notes

  TECH TASK: Adds Rack::Attack as an HTTP rate-limiting layer to protect the application from abusive request volumes. Throttles requests per IP and login attempts per IP, returning HTTP 429 when limits are exceeded.
  
  [NU-455 | Receive a notification when an error happens more than X times in less than Y amount of time](https://universe-of-universities.atlassian.net/browse/NU-455)

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -24,5 +24,3 @@ class Rack::Attack
 end
 
 Rack::Attack.enabled = Settings.rack_attack&.enabled || false
-
-Rails.application.config.middleware.use Rack::Attack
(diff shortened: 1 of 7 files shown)


config/initializers/rack_attack.rb
```diff
@@ -0,0 +1,28 @@
...
+      { "Content-Type" => "text/html", "Retry-After" => retry_after.to_s },
+      [Rails.public_path.join("429.html").read],
+    ]
+  end
+
+end
+
+Rack::Attack.enabled = Settings.rack_attack&.enabled || false
+
+Rails.application.config.middleware.use Rack::Attack
```
joaquinco: the middleware is added by default when the library is required and rails is defined. See docs https://github.com/rack/rack-attack#plugging-into-the-application

joaquinco (approved):
Looks good!

LeticiaErrandonea (approved):
LGTM!

PR #6292, opened 2026-06-22
[NU-444] Notify on admin hold removal and wrap up

joaquinco:
# Notes

- Call slot available notifier on: admin hold removal, batch order update
- Allow access to Facility Operators and facility users with product management
- Run slot available notification in a job

Part of [NU-444 | An automatic email that would notify user of canceled reservations](https://universe-of-universities.atlassian.net/browse/NU-444)

### DIFF
--- a/app/services/product_notifications/slot_available_service.rb
+++ b/app/services/product_notifications/slot_available_service.rb
@@ -27,7 +27,7 @@ module ProductNotifications
     end
 
     def notify!
-      return unless should_notfy?
+      return unless should_notify?
 
       product_notifications.find_each do |product_notification|
         users =
@@ -48,7 +48,7 @@ module ProductNotifications
 
     private
 
-    def should_notfy?
+    def should_notify?
       [
         SettingsHelper.feature_on?("notifications.facility_product_notifications"),
         start_time&.future?,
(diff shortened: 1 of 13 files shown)


app/services/product_notifications/slot_available_service.rb
```diff
@@ -11,7 +11,7 @@ def self.from_reservation(reservation)
         reservation.order_detail.product,
         reservation.reserve_start_at,
         reservation.reserve_end_at,
-        exclude_user: reservation.order_detail.user,
+        exclude_user: reservation&.order_detail&.user,
```
LeticiaErrandonea: I think line 11 will fail whenever this line was going to fail. Also, reservation already belongs to a product. Can we just call it instead of going through the order detail?

joaquinco: makes sense

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

app/services/product_notifications/slot_available_service.rb
```diff
@@ -22,22 +22,44 @@ def initialize(product, start_time, end_time, exclude_user: nil)
...
-            subject: product_notification.email_subject,
+            subject: product_notification.email_subject.presence,
           ).deliver_later
         end
       end
     end
 
     private
 
+    def should_notfy?
```
diebas: *should_notfiy?

diebas (approved):
looks goood, just a nit comment

PR #6296, opened 2026-06-23
[OSU-229] Fix stepped billing external subsidies

joaquinco:
# Notes

Fix stepped billing with external subsidies using base internal rate for base external prices.
The issue was that stepped billing add up to three new rates for each price group to have in total four steps. The first rate is stored on the price group and the rest in the DurationRate model. The js that copied base rates to subsidies rates was working for the first step (rate stored in price group) but not for the others, it was taking the base rate to be the master internal. See screenshot for reference.

The rates per minute shown below the inputs fields are still incorrect. This is will be address later.

[OSU-229 | Stepped billing with external subsidies error](https://universe-of-universities.atlassian.net/browse/OSU-229)

## Screenshot

External subsidy for step 1 should be 16 - 10, but it ends up being 10 - 10.

Captura desde 2026-06-24 10-17-35

Captura desde 2026-06-24 10-10-43

### DIFF
--- a/app/javascript/price_policy.js
+++ b/app/javascript/price_policy.js
@@ -111,20 +111,25 @@ $(document).ready(function() {
     }
   });
 
+  // Duration rates
+
   $(".js--baseRate").each(function (_index, element) {
-    setRateForSubsidyPrieGroups(element);
+    setRateForSubsidyPriceGroups(element);
   });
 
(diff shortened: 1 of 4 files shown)


diebas (approved):

LeticiaErrandonea (approved):
LGTM!

PR #6326, opened 2026-07-03
[NUOPEN-273] Relay update fixes

joaquinco:
# Notes

Relay log events have some issues related to the fact that relays records are replaced on every time there's a change on the instrument relay tab.

1. Log event references relay records: the workaround was to update all log events for the destroyed relay.
2. When relay type is "Reservation Only", there's actually no relay record which causes log events to object to be blank.

The workaround in this pr is to change log events to use the product with a new log event type.

**Existing log events**: there're not many log events for relay update since it's been one year since this was enabled. To update existing records we can use:
```ruby
LogEvent.where(loggable_type: "Relay", event_type: "update").find_each do |le| le.update(loggable: le.loggable.instrument, event_type: "relay_update") if le.loggable.present? end
```

[NUOPEN-273| Relays replaced on update](https://universe-of-universities.atlassian.net/browse/NUOPEN-273)

### DIFF
--- a/app/models/concerns/products/relay_support.rb
+++ b/app/models/concerns/products/relay_support.rb
@@ -24,18 +24,18 @@ module Products::RelaySupport
   # See https://andycroll.com/ruby/be-careful-assigning-to-has-one-relations/
   def replace_relay(attributes = nil, param_control_mechanism = nil)
     self.class.transaction do
+      id = relay&.id
       relay&.destroy!
       case param_control_mechanism
       when Relay::CONTROL_MECHANISMS[:relay]
-        create_relay!(attributes)
+        create_relay!(id:, **attributes)
       when Relay::CONTROL_MECHANISMS[:timer]
-        create_relay!(type: "RelayDummy")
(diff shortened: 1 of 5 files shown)


app/models/concerns/products/relay_support.rb
```diff
@@ -24,18 +24,18 @@ def has_real_relay?
   # See https://andycroll.com/ruby/be-careful-assigning-to-has-one-relations/
   def replace_relay(attributes = nil, param_control_mechanism = nil)
     self.class.transaction do
+      id = relay&.id
       relay&.destroy!
       case param_control_mechanism
       when Relay::CONTROL_MECHANISMS[:relay]
-        create_relay!(attributes)
+        create_relay!(id:, **attributes)
```
LeticiaErrandonea: Now that we don't reference the relay, I don't think this will serve any purpose. Maybe we can drop it?

LeticiaErrandonea (approved):
LGTM!

diebas (approved):

PR #6328, opened 2026-07-06
[NUOPEN-280]  Migrate easy date inputs to native picker

diebas:
# Notes

Switches the standalone (single-date) form fields across several admin screens from the jQuery datepicker to the browser's native date picker, so dates are submitted in a consistent (ISO) format.

# Screenshots

Before:
Screenshot 2026-07-06 at 6 08 19 PM

After:
Screenshot 2026-07-06 at 6 07 41 PM

### DIFF
--- a/app/controllers/facility_accounts_reconciliation_controller.rb
+++ b/app/controllers/facility_accounts_reconciliation_controller.rb
@@ -31,7 +31,12 @@ class FacilityAccountsReconciliationController < ApplicationController
   end
 
   def update
-    reconciled_at = params[:reconciled_at].presence&.to_date&.beginning_of_day
+    reconciled_at =
+      begin
+        params[:reconciled_at].presence&.to_date&.beginning_of_day
+      rescue ArgumentError
+        nil
+      end
     reconciler = OrderDetails::Reconciler.new(
       unreconciled_details,
       params[:order_detail],
--- a/app/controllers/facility_accounts_reconciliation_controller.rb
+++ b/app/controllers/facility_accounts_reconciliation_controller.rb
@@ -31,7 +31,7 @@ class FacilityAccountsReconciliationController < ApplicationController
   end
 
   def update
-    reconciled_at = parse_usa_date(params[:reconciled_at])
+    reconciled_at = params[:reconciled_at].presence&.to_date&.beginning_of_day
     reconciler = OrderDetails::Reconciler.new(
       unreconciled_details,
       params[:order_detail],
--- a/app/controllers/facility_accounts_reconciliation_controller.rb
+++ b/app/controllers/facility_accounts_reconciliation_controller.rb
@@ -31,12 +31,7 @@ class FacilityAccountsReconciliationController < ApplicationController
   end
 
   def update
-    reconciled_at =
-      begin
-        params[:reconciled_at].presence&.to_date&.beginning_of_day
-      rescue ArgumentError
-        nil
-      end
+    reconciled_at = parse_iso_date(params[:reconciled_at])&.beginning_of_day
     reconciler = OrderDetails::Reconciler.new(
       unreconciled_details,
       params[:order_detail],
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -361,7 +361,7 @@ class OrdersController < ApplicationController
 
   def build_order_date
     if params[:order_date].present?
-      parse_usa_date(params[:order_date])
+      params[:order_date].to_date.beginning_of_day
     end
   end
 
--- a/app/controllers/orders_controller.rb
+++ b/app/controllers/orders_controller.rb
@@ -360,9 +360,7 @@ class OrdersController < ApplicationController
   end
 
   def build_order_date
-    if params[:order_date].present?
-      params[:order_date].to_date.beginning_of_day
-    end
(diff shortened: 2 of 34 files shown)


app/views/facility_accounts_reconciliation/_action_row.html.haml
```diff
@@ -17,10 +17,11 @@
...
           .col-md-4
-            = text_field_tag :reconciled_at,
-              format_usa_date(Time.current),
-              class: "datepicker__data form-control",
-              data: { min_date: unreconciled_order_details.map(&:journal_or_statement_date).min&.iso8601, max_date: Time.current.iso8601 }
+            = date_field_tag :reconciled_at,
+              Time.current.to_date.iso8601,
+              class: "form-control",
+              min: unreconciled_order_details.map(&:journal_or_statement_date).min&.to_date&.iso8601,
+              max: Time.current.to_date.iso8601
```
joaquinco: we shouldn't need to convert dates to iso format here, right?

joaquinco (commented):
I think we are missing changes to controllers and form classes where these dates are parsed (we should stop doing so)

LeticiaErrandonea:
Dev QA

In "Add to order" form, it looks too wide and also we no longer have a way of letting the user know what they're setting (no placeholder):
Before
image


After
image

LeticiaErrandonea:
Dev QA:

1. Added an item via the "Add to order" form with yesterday's date (2026-07-06).
image

2. Added another item with the pre-populated date.
3. When checking, the date is from June (2026-06-07).
image



I think it comes from https://github.com/wyeworks/nucore-open/blob/master/app/forms/add_to_order_form.rb#L204, which still parses the date.

app/models/valid_fulfilled_at_date.rb
LeticiaErrandonea, line 6: I think we don't need this anymore.

LeticiaErrandonea: nit: this should be ISO now, right?

app/controllers/facility_accounts_reconciliation_controller.rb
```diff
@@ -31,7 +31,12 @@ def index
   end
 
   def update
-    reconciled_at = parse_usa_date(params[:reconciled_at])
+    reconciled_at =
+      begin
+        params[:reconciled_at].presence&.to_date&.beginning_of_day
+      rescue ArgumentError
+        nil
+      end
```
joaquinco: can we use `parse_iso_date` here?

app/controllers/orders_controller.rb
```diff
@@ -361,7 +361,7 @@ def add_account_to_order(account)
 
   def build_order_date
     if params[:order_date].present?
-      parse_usa_date(params[:order_date])
+      params[:order_date].to_date.beginning_of_day
```
joaquinco: can we use parse_iso_date?

app/controllers/price_policies_controller.rb
```diff
@@ -143,12 +143,19 @@ def update_policies_from_params
...
       params.merge(created_by_id: current_user.id),
     )
   end
 
+  # Returns nil for blank or unparseable (e.g. "5/0/2018") input.
+  def parse_iso_date(value)
+    value.presence&.to_date
+  rescue Date::Error
+    nil
+  end
```
joaquinco: can we move this to the `DateHelper` module? then we can just replace parse_usa_date by this everywhere it's needed

app/models/journal_creation_reminder.rb
```diff
@@ -25,15 +23,26 @@ def past?
   end
 
   def starts_at=(date_string)
-    super(parse_usa_date(date_string))
+    super(coerce_date(date_string)&.beginning_of_day)
   end
 
   def ends_at=(date_string)
-    super(parse_usa_date(date_string)&.end_of_day)
+    super(coerce_date(date_string)&.end_of_day)
```
joaquinco: can we use parse_iso_date here?

app/models/valid_fulfilled_at_date.rb
```diff
@@ -32,12 +31,16 @@ def to_s
 
   # Returns nil if the date is invalid
   def to_time
-    time = parse_usa_date(@string).try(:to_date)
-    time.beginning_of_day + 12.hours if time.present?
+    return if @string.blank?
+
+    date = Date.iso8601(@string.to_s)
```
joaquinco: we should be able to use parse_iso_date here

joaquinco (commented):
Looks good, left a comment to reuse the date parsing method since I see we are using several approaches: `Date.iso8601`, `String.to_date` and a wrapper method around those those catching the parsing error

app/helpers/date_helper.rb
```diff
@@ -37,6 +37,12 @@ def parse_usa_date(date, time_string = nil)
     nil
   end
 
+  def parse_iso_date(value)
+    value.presence&.to_date
```
LeticiaErrandonea: Can we use `Date.iso8601`, so whatever isn't an ISO date is returned as `nil` (and we're consistent with the name)?

LeticiaErrandonea (approved):
Thanks for the changes!

Left a comment but LGTM

joaquinco (approved):

PR #6342, opened 2026-07-09
[NUOPEN-280] Migrate date range inputs

diebas:
# Notes

Date-range filters on Reports, Transaction Search, and Statement search now use the browser's built-in date picker instead of the old calendar widget. Behavior is unchanged for users; dates are just entered through the native picker.

[NUOPEN-280 | Standardize forms date input submission](https://universe-of-universities.atlassian.net/browse/NUOPEN-280)

# Screenshots

Screenshot 2026-07-09 at 2 08 33 PM

### DIFF
--- a/app/controllers/facilities_controller.rb
+++ b/app/controllers/facilities_controller.rb
@@ -136,7 +136,7 @@ class FacilitiesController < ApplicationController
     @search_form = TransactionSearch::SearchForm.new(
       params[:search],
       defaults: {
-        date_range_start: format_usa_date(1.month.ago.beginning_of_month),
+        date_range_start: 1.month.ago.beginning_of_month.to_date,
       },
     )
 
--- a/app/views/reports/_date_range_fields.html.haml
+++ b/app/views/reports/_date_range_fields.html.haml
@@ -5,7 +5,7 @@
 %li= form.input :date_start,
   as: :date_field,
   label: t("reports.fields.date_start"),
-  input_html: { value: @date_start&.to_date&.iso8601 }
+  input_html: { value: @date_start }
 
 %li
   .field-clarifier= t("support.array.two_words_connector")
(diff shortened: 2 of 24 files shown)


app/controllers/facilities_controller.rb
```diff
@@ -136,7 +136,7 @@ def transactions
     @search_form = TransactionSearch::SearchForm.new(
       params[:search],
       defaults: {
-        date_range_start: format_usa_date(1.month.ago.beginning_of_month),
+        date_range_start: 1.month.ago.beginning_of_month.to_date,
```
LeticiaErrandonea, line 139: I think we're missing this same change in the [app/controllers/facility_statements_controller.rb:55](https://github.com/wyeworks/nucore-open/blob/master/app/controllers/facility_statements_controller.rb#L55)

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

app/views/reports/_date_range_fields.html.haml
```diff
@@ -3,12 +3,14 @@
     = t("reports.fields.#{controller_name}.between", default: t("reports.fields.between"))
 
 %li= form.input :date_start,
+  as: :date_field,
   label: t("reports.fields.date_start"),
-  input_html: { value: format_usa_date(@date_start), class: "datepicker" }
+  input_html: { value: @date_start&.to_date&.iso8601 }
```
joaquinco: note that you can pass date objects here, so leaving `@date_start` should be fine

spec/system/admin/all_transactions_search_spec.rb
```diff
@@ -167,7 +167,7 @@ def translation_scope
     login_as director
     visit facility_transactions_path(facility)
     expected_default_date = 1.month.ago.beginning_of_month
-    expect(page).to have_field("Start Date", with: I18n.l(expected_default_date.to_date, format: :usa))
+    expect(page).to have_field("Start Date", with: expected_default_date.to_date.iso8601)
```
joaquinco: we can use date objects here

joaquinco (approved):
It's looking good. I think as a general rule, the benefit of this whole change is to use date objects as much as possible, not that it's just a format change. So there're many places I think we are formatting dates unnecessarily (including within the `DateInputField` class).

PR #6343, opened 2026-07-09
[NUOPEN-372] Refactor `PriceDisplayment`

joaquinco:
# Notes

Refactor `PriceDisplayment` module used as presenter by extending `OrderDetail` instances by moving method to `OrderDetailPresenter`.

[NUOPEN-372 | Refactor PriceDisplayment](https://universe-of-universities.atlassian.net/browse/NUOPEN-372)


app/views/order_management/order_details/_print_details.html.haml
```diff
@@ -27,10 +27,10 @@
         = banner_label @order_detail, :quantity
     .row
       - prefix = @order_detail.actual_cost? ? "actual" : "estimated"
-      - @order_detail.send(:extend, PriceDisplayment)
-      = banner_label @order_detail, :display_cost, t("activerecord.attributes.order_detail.#{prefix}_cost")
-      = banner_label @order_detail, :display_subsidy, t("activerecord.attributes.order_detail.#{prefix}_subsidy")
-      = banner_label @order_detail, :display_total, t("activerecord.attributes.order_detail.#{prefix}_total")
+      - order_detail_presenter = OrderDetailPresenter.new(@order_detail)
```
LeticiaErrandonea, line 30: I think this should go a line above, so that we use the `actual_cost?` from the presenter, which returns `true` for 0, but AR's `actual_cost?` would return `false` for that same value.

app/views/order_management/order_details/_costs.html.haml
```diff
@@ -1,4 +1,3 @@
-- f.object.send(:extend, PriceDisplayment)
```
LeticiaErrandonea, line 1: I think we could need to bring this back for the same reason as my comment on app/views/order_management/order_details/_print_details.html.haml.

LeticiaErrandonea (approved):
Left a couple of comments, but LGTM!

diebas (approved):

PR #6345, opened 2026-07-09
[TECH TASK] [NUOPEN-371] Remove `rails-observer`

LeticiaErrandonea:
# Notes
* Remove `rails-observer`

[NUOPEN-371 | Remove gem rails-observer ](https://universe-of-universities.atlassian.net/browse/NUOPEN-371)


app/controllers/reservations_controller.rb
```diff
@@ -203,7 +203,7 @@ def update
 
     render_edit && return unless changes_valid_for_update?
 
-    # merge state can change after call to #save! due to OrderDetailObserver#before_save
+    # merge state can change after call to #save! due to OrderDetail#move_to_original_merge_order
```
joaquinco, line 206: the previous logic used to check here if the order should be merged, I think we should do that here or move the check as a guard within `move_to_original_merge_order`

app/models/order_detail.rb
```diff
@@ -38,6 +38,11 @@ class OrderDetail < ApplicationRecord
   before_save :update_journal_row_amounts, if: :actual_cost_changed?
 
   after_save :update_billable_minutes_on_reservation, if: :reservation
+  after_save :reconcile_if_skip_review, if: :complete?
+
+  # Included after the callbacks above so its before_save/after_save hooks
+  # run last, matching the ordering the old OrderDetailObserver had.
```
joaquinco: let's change this comment to something like "required to be here because of callback order", so we can forget about `OrderDetailObserver`

joaquinco (approved):
It's looking good!

diebas (approved):

PR #6348, opened 2026-07-10
[NUOPEN-280] Migrate problem reservation date inputs

diebas:
# Notes

TECH TASK: Replaces the jQuery datepicker on the problem-reservation ("Fix Usage") form with the browser's native date input, which submits dates in ISO format. No change to reservation times or charges — only the date-picker widget. Also bounds the picker to today so future end dates can't be picked (previously the picker allowed them and relied on server validation).

# Screenshot

Screenshot 2026-07-13 at 12 07 24 PM

### DIFF
--- a/app/views/problem_reservations/edit.html.haml
+++ b/app/views/problem_reservations/edit.html.haml
@@ -19,13 +19,13 @@
           = f.label :actual_start_date
           .row
             .col-md-3
-              = f.input :actual_start_date, label: false, disabled: true
+              = f.input :actual_start_date, as: :date_field, label: false, input_html: { value: f.object.actual_start_at&.to_date&.iso8601, disabled: true }
             .col-md-4
               = time_select f, :actual_start, { minute_step: 1 }, disabled: true
           = f.label :actual_end_date
           .row
             .col-md-3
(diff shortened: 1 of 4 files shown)


LeticiaErrandonea (approved):
LGTM!

app/views/problem_reservations/edit.html.haml
```diff
@@ -19,13 +19,13 @@
           = f.label :actual_start_date
           .row
             .col-md-3
-              = f.input :actual_start_date, label: false, disabled: true
+              = f.input :actual_start_date, as: :date_field, label: false, input_html: { value: f.object.actual_start_at&.to_date&.iso8601, disabled: true }
```
joaquinco, line 22: This used to take the value from `Reservation.actual_start_date` which returned usa formated date, should we remove that method? or change it to just return a date object?

joaquinco (approved):

app/views/problem_reservations/edit.html.haml
diebas: Yes, you're right, but that is still used in other parts of the app, I'll clean it on another pr when finishing migrating everything

PR #6351, opened 2026-07-13
[NUOPEN-280] Migrate timeline and reservation edit dates

diebas:
# Notes

Replaces the jQuery datepicker with the browser's built-in date picker on the reservations timeline pages, the admin reservation edit form, and the Manage Order Detail form. Dates are now submitted in ISO format, removing the need for US-format date parsing on these screens.

# Screenshot

Screenshot 2026-07-13 at 12 30 55 PM
Screenshot 2026-07-13 at 12 56 32 PM

### DIFF
--- a/app/views/facility_reservations/timeline.html.haml
+++ b/app/views/facility_reservations/timeline.html.haml
@@ -14,12 +14,14 @@
   = modelless_form_for url: timeline_facility_reservations_path(current_facility), method: :get, id: "timeline_date_search" do |f|
     #reservation_date_container
       = link_to "&laquo;".html_safe,
-        timeline_facility_reservations_path(current_facility, date: format_usa_date(@display_datetime - 1.day)), id: "reservation_left"
+        timeline_facility_reservations_path(current_facility, date: (@display_datetime - 1.day).to_date), id: "reservation_left"
       #reservation_date
         = l(@display_datetime.to_date, format: :timeline_navigation)
-        = hidden_field_tag "date", format_usa_date(@display_datetime), class: "datepicker"
+        = date_field_tag "date", @display_datetime.to_date, class: "sr-only js--timelineDatePicker"
+        %button.timeline-datepicker-button.js--timelineDatePickerButton{ type: "button" }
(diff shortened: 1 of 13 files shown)


joaquinco (approved):

app/views/facility_reservations/timeline.html.haml
```diff
@@ -14,12 +14,14 @@
   = modelless_form_for url: timeline_facility_reservations_path(current_facility), method: :get, id: "timeline_date_search" do |f|
     #reservation_date_container
       = link_to "&laquo;".html_safe,
-        timeline_facility_reservations_path(current_facility, date: format_usa_date(@display_datetime - 1.day)), id: "reservation_left"
+        timeline_facility_reservations_path(current_facility, date: (@display_datetime - 1.day).to_date), id: "reservation_left"
```
LeticiaErrandonea, line 17: Should we also update the [controller](https://github.com/wyeworks/nucore-open/blob/master/app/controllers/facility_reservations_controller.rb#L163), so we use `parse_iso_date` instead of `parse_usa_date`? I think that's what we're doing in other places.

app/views/reservations/public_timeline.html.haml
```diff
@@ -9,12 +9,14 @@
   = modelless_form_for url: facility_public_timeline_path(current_facility), method: :get, id: "timeline_date_search" do |f|
     #reservation_date_container
       = link_to "&laquo;".html_safe,
-        facility_public_timeline_path(current_facility, date: format_usa_date(@display_datetime - 1.day)), id: "reservation_left"
+        facility_public_timeline_path(current_facility, date: (@display_datetime - 1.day).to_date), id: "reservation_left"
```
LeticiaErrandonea, line 12: Same question as above regarding which parser is used but for [another controller](https://github.com/wyeworks/nucore-open/blob/master/app/controllers/reservations_controller.rb#L28).

LeticiaErrandonea (approved):
I left a couple of comments but it LGTM!

PR #6356, opened 2026-07-14
[Fix] OrderDetail show when order not purchased

joaquinco:
# Notes

Fix order detail show page when order is not purchased, thus `ordered_at` attribute is nil causing the readonly form to attempt to fetch the value from the form object which is the order.

The order page can be access from the sanger sequencing submission page, before it's purchased.

### DIFF
--- a/spec/requests/order_details_spec.rb
+++ b/spec/requests/order_details_spec.rb
@@ -3,6 +3,74 @@
 require "rails_helper"
 
 RSpec.describe "order_details" do
+  describe "show" do
+    let(:user) { create(:user) }
+    let(:account) do
+      create(:setup_account).tap do |account|
+        create(:account_user, :purchaser, user:, account:)
+      end
+    end
+    let(:product) { create(:setup_item) }
+    let(:facility) { product.facility }
+    let(:action) do
+      lambda do
+        get order_order_detail_path(order, order_detail)
+      end
+    end
+
+    before { login_as user }
+
+    shared_examples "renders successfully" do
+      it "includes order detail" do
+        action.call
+
+        expect(page).to have_text(order_detail.order_number)
+      end
+
+      it "renders correctly" do
+        action.call
+
+        expect(response).to have_http_status(:ok)
+      end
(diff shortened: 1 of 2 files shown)


spec/requests/order_details_spec.rb
```diff
@@ -3,6 +3,74 @@
...
 RSpec.describe "order_details" do
+  describe "show" do
+    let(:user) { create(:user) }
+    let(:account) do
+      create(:setup_account).tap do |account|
+        create(:account_user, :purchaser, user:, account:)
+      end
+    end
+    let(:product) { create(:setup_item) }
+    let(:facility) { product.facility }
```
LeticiaErrandonea: nit: I think we can remove this, it's not used in the new spec and it's already defined in the `cancel` describe

LeticiaErrandonea (approved):
LGTM!

spec/requests/order_details_spec.rb
```diff
@@ -3,9 +3,75 @@
...
+
+    context "when order is purchased" do
+      let(:order) do
+        create(:purchased_order, product:, user:, account:)
+      end
+      let(:order_detail) { order.order_details.first }
+
+      it_behaves_like "renders successfully"
+
+      it "does not show receipt link" do
```
diebas: This description is the other way around

spec/requests/order_details_spec.rb
```diff
@@ -3,9 +3,75 @@
...
+
+    context "when ordered at is nil" do
+      let(:order) do
+        create(:setup_order, product:, ordered_at: nil, user:, account:)
+      end
+      let(:order_detail) { order.order_details.first }
+
+      it_behaves_like "renders successfully"
+
+      it "shows receipt link" do
```
diebas: Same here

diebas (approved):

spec/requests/order_details_spec.rb
joaquinco: 🙈

PR #6365, opened 2026-07-16
[NUOPEN-282] Make Duration billing available for Global admin only

LeticiaErrandonea:
# Notes
* Make Duration billing available for Global admin only

[NUOPEN-282 | Duration Billing not global admin only? ](https://universe-of-universities.atlassian.net/browse/NUOPEN-282)

# Additional Context
Also made `permitted_params` remove the `pricing_mode` from `permitted_params`. While at it, I noticed we also had `billing_mode` hidden in the UI and present in the `permitted_params` and removed it in those cases.

### DIFF
--- a/app/controllers/instruments_controller.rb
+++ b/app/controllers/instruments_controller.rb
@@ -111,20 +111,9 @@ class InstrumentsController < ProductsCommonController
       params += %i[min_reserve_days max_reserve_days start_time_disabled]
     end
 
-    params -= [:pricing_mode] if action_name == "update" && !allowed_to_set_pricing_mode?
+    params -= [:pricing_mode] if action_name == "update"
 
     params
-  end
-
-  def allowed_to_set_pricing_mode?
-    case params.dig(:instrument, :pricing_mode)
-    when Instrument::Pricing::SCHEDULE_DAILY
-      can?(:create_daily_booking, Instrument)
-    when Instrument::Pricing::DURATION
-      can?(:create_duration_billing, Instrument)
-    else
-      true
-    end
   end
 
   def start_time_disabled_changed_check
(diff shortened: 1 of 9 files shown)


app/controllers/instruments_controller.rb
```diff
@@ -107,9 +111,22 @@ def permitted_params
       params += %i[min_reserve_days max_reserve_days start_time_disabled]
     end
 
+    params -= [:pricing_mode] if action_name == "update" && !allowed_to_set_pricing_mode?
```
joaquinco: I think the conditions should be `action_name == "update" || !allowed_to_set..` since pricing mode cannot be changed once the instrument is created.

Or, we can remove the param on update action only and let it fail with the checks around line 44

app/controllers/instruments_controller.rb
```diff
@@ -40,6 +40,10 @@ def create
     if resource_params[:pricing_mode] == Instrument::Pricing::SCHEDULE_DAILY && cannot?(:create_daily_booking, Instrument)
       flash.now[:error] = t("controllers.instruments.create.daily_booking_not_authorized")
 
+      render :new
+    elsif resource_params[:pricing_mode] == Instrument::Pricing::DURATION && cannot?(:create_duration_billing, Instrument)
```
joaquinco, line 44: with the pricing mode param being filtered out if the user has no permissions, will this and the above check ever occur?

joaquinco (approved):
Looks good. Left some comments

app/controllers/instruments_controller.rb
LeticiaErrandonea: I simplified the `permitted_params` method, so this is still relevant

joaquinco (approved):

diebas (approved):

PR #6370, opened 2026-07-17
[Chore] Prevent missing template error on format not accepted

joaquinco:
# Notes

Rails already handles `ActionController::UnknownFormat` and `ActionController::MissingExactTemplate` with not acceptable HTTP error code.

Here I've changed actions in facilities controller, whose index action is the root path of the app and main cause of `MissingTemplate` exceptions, to use implicit render so `UnknownFormat` exeception is raised.


app/controllers/facilities_controller.rb
```diff
@@ -235,4 +225,12 @@ def azlist_on?
   end
   helper_method :azlist_on?
 
+  def action_layout
+    case action_name
+    when "index", "show", "list", "new" then "application"
```
LeticiaErrandonea: I think we need to add `create`, so if the form has an error, the right layout is used

LeticiaErrandonea (approved):
Left a comment, once solved it LGTM

diebas (approved):

PR #6372, opened 2026-07-17
[NUOPEN-295] Improve instrument availability loading

diebas:
# Notes

The public facility page shows a colored icon for each instrument's current availability. With that feature enabled the page had become noticeably slower — roughly doubling its load time on facilities with many instruments. This makes it load quickly regardless of how many instruments a facility has, with no change to what users see.

[NUOPEN-295 | [Investigate] Facility show page: improve Instrument availability loading](https://universe-of-universities.atlassian.net/browse/NUOPEN-295)

# Screenshot

Before:
Screenshot 2026-07-20 at 4 09 56 PM

After:
Screenshot 2026-07-20 at 4 10 06 PM

### DIFF
--- a/app/services/instruments/availability_status.rb
+++ b/app/services/instruments/availability_status.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+module Instruments
+
+  # Request-scoped instrument availability for the public facility page: two
+  # facility-wide queries instead of running walkup_available? per instrument.
+  # Build once, reuse for every instrument on the page.
+  class AvailabilityStatus
+
+    def initialize(facility)
+      @facility = facility
+    end
+
+    # Open now per its schedule rules and its schedule not currently in use.
+    def available_now?(product)
+      return false if in_use_schedule_ids.include?(product.schedule_id)
+
+      schedule_rules_by_product.fetch(product.id, []).any? { |rule| rule.cover?(Time.current) }
+    end
+
+    private
+
+    # Keyed on schedule so shared-schedule instruments block each other, matching
+    # conflicting_*_reservation's `product.schedule.product_ids`.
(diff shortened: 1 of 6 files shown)


LeticiaErrandonea (commented):
I like this idea!

There are a few cases that I think won't work like before, maybe we need to fix them or bring it up with clients:
1. This doesn't consider holidays, right?
2. We only check what's going on _now_, so if a reservation starts in 5 mins or we're close to the end-of-schedule time, they will show as available, even when you can't really create a reservation.

I think the first point is easy to fix. The second one may require some more work.

joaquinco (approved):

spec/helpers/products_helper_spec.rb
```diff
@@ -24,4 +24,83 @@
       expect(subject.to_h).to_not include(RelaySynaccessRevA)
     end
   end
+
+  describe "#public_calendar_availability_options" do
```
joaquinco, line 28: I think some of these tests are better suited under an `AvailabilityStatus` example group

joaquinco (approved):

app/services/instruments/availability_status.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    def initialize(facility)
+      @facility = facility
+    end
+
+    # Open now per its schedule rules, its schedule not currently in use, and not
+    # a holiday it restricts access on.
+    def available_now?(product)
+      return false if in_use_schedule_ids.include?(product.schedule_id)
+      return false if closed_for_holiday?(product)
```
LeticiaErrandonea: nit: I think inverting these two makes the call a little bit more efficient on holidays

LeticiaErrandonea (approved):
LGTM!

PR #6376, opened 2026-07-21
[NUOPEN-289] Refactor raw export

diebas:
# Notes

TECH TASK: Refactor the Export Raw report so large exports generate with a fraction of the database queries (e.g. ~600 → ~74 on a 1,400-row export). The CSV content does not change.

[NUOPEN-289 | Refactor RawExport](https://universe-of-universities.atlassian.net/browse/NUOPEN-289)

# Additional Context

## Performance

Two N+1s fixed in `Reports::ExportRaw` (measured on a 1,407-row export against the demo dataset; query counts include cached queries, i.e. real round-trips with varied production data):

| Scenario | Before | After |
|---|---:|---:|
| Standard export | 617 queries | **74 queries** (−88%) |
| Rows without a price policy (canceled/unpriced) | 1,408 queries / 3.3s | **71 queries** / 1.1s |

- **Bundle N+1** (546 queries): the report read `od.bundle` per row — fixed by preloading `:bundle` plus other accessed associations (`assigned_user`, `canceled_by_user`, `price_changed_by_user`, `problem_resolved_by`, `project`, `order.cross_core_project`).
- **Charge mode N+1**: rows without a price policy fell back to `Product#current_price_policies(Time.zone.now)` per row — the fresh timestamp in the SQL makes it uncacheable. Now memoized per product for the report run.
- `report_hash` was rebuilt (running every registered engine transformer) once per row — now memoized per report instance.

## Refactorings

- `initialize` uses explicit keyword args instead of the `instance_variable_set` loop; specs dropped a silently-ignored dead argument (`action_name`).
- `default_report_hash` (75 columns, `rubocop:disable Metrics/AbcSize`) split into six named sections (`order_columns`, `user_columns`, …) — the disable is gone.
- `CsvExporter#format_row`: a row that raises now pads to the full column count, so the error lands in column 1 and the row keeps the CSV width for programmatic consumers (before: a 1-cell row).
- Header line built with `CSV.generate_line` (body already used `CSV`), so a translated header containing a comma/quote can't corrupt the file.
- Dropped redundant `require "csv"` (bundled via Gemfile).

# Screenshot
Before:
Screenshot 2026-07-21 at 3 26 50 PM

After:
Screenshot 2026-07-21 at 3 26 59 PM

MiniProfiler for the same 567-row export request (job forced inline in dev):

| Before | After |
|---|---|
| 1,137 ms — 272 sql (229 cached) | 653 ms — 64 sql (20 cached) |

### DIFF
--- a/app/models/concerns/reports/csv_exporter.rb
+++ b/app/models/concerns/reports/csv_exporter.rb
@@ -62,7 +62,7 @@ module Reports
     private
 
     def csv_header
-      column_headers.join(",") + "\n"
+      CSV.generate_line(column_headers)
     end
 
     def csv_body
@@ -94,8 +94,10 @@ module Reports
       [error] + Array.new(report_hash.size - 1)
     end
 
+    # Memoized: format_row consults it for every row, and rebuilding it means
+    # re-running every registered transformer per row.
     def report_hash
-      transformers.reduce(default_report_hash) do |result, class_name|
+      @report_hash ||= transformers.reduce(default_report_hash) do |result, class_name|
         class_name.constantize.new.transform(result)
       end
     end
--- a/app/models/concerns/reports/csv_exporter.rb
+++ b/app/models/concerns/reports/csv_exporter.rb
@@ -89,7 +89,9 @@ module Reports
         end
       end
     rescue => e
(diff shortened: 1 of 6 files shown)


LeticiaErrandonea (approved):
LGTM! 👏

joaquinco (approved):

PR #6387, opened 2026-07-29
[DART-165] Allow to exclude suspended accounts on create Journal

joaquinco:
# Notes

Add filter to be able to exclude suspended accounts on the create Journal page.

Also:
- Remove css breaking checkbox inputs applied to search_form
- Remove js which handled jquery datepickers which are no longer used

[DART-165 | Ability to Filter Orders to identify and/or exclude "Suspended" payment sources from the journal](https://universe-of-universities.atlassian.net/browse/DART-165)

# Screenshot

Captura desde 2026-07-29 16-46-10

### DIFF
--- a/app/helpers/transaction_searcher_helper.rb
+++ b/app/helpers/transaction_searcher_helper.rb
@@ -1,0 +1,38 @@
+# frozen_string_literal: true
+
+module TransactionSearcherHelper
+  def searcher_input(form, searcher)
+    case searcher.input_type
+    when :transaction_chosen, :select then select_input(form, searcher)
+    when :boolean then boolean_input(form, searcher)
+    else
+      raise ArgumentError, "Unsupported transaction searcher input type: #{searcher.input_type}"
+    end
+  end
+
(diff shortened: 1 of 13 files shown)


app/helpers/transaction_searcher_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    )
+  end
+
+  def boolean_input(form, searcher)
+    form.input(
+      searcher.key,
+      as: :boolean,
+      label: false,
+      inline_label: searcher.label,
+      data_attrs: searcher.method(:data_attrs),
```
LeticiaErrandonea, line 34: nit: I think this one isn't used

LeticiaErrandonea (approved):
LGTM!

app/helpers/transaction_searcher_helper.rb
joaquinco: right. It's there to follow the searcher's api used for the other fields

PR #6389, opened 2026-07-30
[OSU-147] Stepped Billing for Timed Services

LeticiaErrandonea:
# Notes
* Stepped Billing for Timed Services

[OSU-147 | Duplicate Stepped Billing from Instruments to Timed Services ](https://universe-of-universities.atlassian.net/browse/OSU-147)

# Screenshot
## Product details
image

## Pricing
image

## Order
image

### DIFF
--- a/app/helpers/products_helper.rb
+++ b/app/helpers/products_helper.rb
@@ -31,27 +31,19 @@ module ProductsHelper
     ].compact
   end
 
-  def instrument_pricing_modes
-    Instrument::PRICING_MODES.reject do |pricing_mode|
-      (pricing_mode == Instrument::Pricing::SCHEDULE_DAILY && cannot?(:create_daily_booking, Instrument)) ||
-        (pricing_mode == Instrument::Pricing::DURATION && cannot?(:create_duration_billing, Instrument))
+  def pricing_modes_for(product_class)
+    modes = product_class::PRICING_MODES.reject do |pricing_mode|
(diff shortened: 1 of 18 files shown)


app/helpers/products_helper.rb
```diff
@@ -38,6 +38,22 @@ def instrument_pricing_modes
...
+
+    text("timed_services.timed_service_fields.pricing_modes.#{key}")
+  end
+
+  def timed_service_pricing_modes
+    modes = TimedService::PRICING_MODES
+    modes -= [TimedService::Pricing::DURATION] if cannot?(:create_duration_billing, TimedService)
+
+    modes.map { |mode| [timed_service_pricing_mode_label(mode), mode] }
+  end
```
joaquinco: Can we use the same logic for instruments pricing mode? either use locales for both or use the raw value

app/models/timed_service.rb
```diff
@@ -2,9 +2,19 @@
 
 class TimedService < Product
 
+  module Pricing
+
+    STANDARD = "Schedule Rule"
+    DURATION = "Duration"
+
+  end
```
joaquinco: maybe we can move `Instrument::Pricing` module to `Product::Pricing` and only define the allowed pricing mode in products subclasses, what do you think?

joaquinco (approved):
Looks good! Left some comments to try to unify a bit what we have in instruments and timed services

joaquinco (approved):
thanks for the changes!

PR #6391, opened 2026-07-31
Update Order Management Permissions

stevens101:
Allow the Order Management granular permission access to read the order's uploaded Service Forms, Sample Results, and zipped Results files.


stevens101:
If this shouldn't be in nucore-open, I will move it to CORUM.

PR #6394, opened 2026-08-03
[NU-491] Fix account listing filters

joaquinco:
# Notes

In the facility accounts index page, the bug was that "hide expired accounts" button was not preserved when open search box changed.
I refactored the search to be entirely server side and unified `index` and `search_result` actions.

This change also fixes pagination when account_tabs is on.

[NU-491 | Default to hide expired payment sources](https://universe-of-universities.atlassian.net/browse/NU-491)

### DIFF
--- a/app/assets/javascripts/user_accounts.js.erb
+++ b/app/assets/javascripts/user_accounts.js.erb
@@ -3,6 +3,11 @@
  */
 window.addEventListener("DOMContentLoaded", () => {
   const toggleExpiredAcctsBtn = document.querySelector(".toggle_expired_accts_btn--js");
+
+  if (!toggleExpiredAcctsBtn) {
+    return;
+  }
+
   const expiredAccts = document.querySelectorAll("tr.expired--js");
   let hideAccounts = true;
 
--- a/app/assets/javascripts/user_accounts.js.erb
+++ b/app/assets/javascripts/user_accounts.js.erb
@@ -11,7 +11,7 @@ window.addEventListener("DOMContentLoaded", () => {
   const expiredAccts = document.querySelectorAll("tr.expired--js");
   let hideAccounts = true;
 
-  if (toggleExpiredAcctsBtn) { toggleExpiredAcctsBtn.addEventListener("click", toggleExpiredAccounts); }
+  toggleExpiredAcctsBtn.addEventListener("click", toggleExpiredAccounts);
 
   toggleRowVisibility();
 
@@ -36,9 +36,7 @@ window.addEventListener("DOMContentLoaded", () => {
       acctEle.style.display = displayStyle;
     });
 
-    if (toggleExpiredAcctsBtn) {
-      toggleExpiredAcctsBtn.textContent = buttonText;
-    }
+    toggleExpiredAcctsBtn.textContent = buttonText;
   }
 
 });
--- a/app/controllers/facility_accounts_controller.rb
+++ b/app/controllers/facility_accounts_controller.rb
@@ -197,10 +197,13 @@ class FacilityAccountsController < ApplicationController
   end
 
   def filter_params
-    params
+    @filter_params ||=
+      params
       .permit(:search_term, :account_type, :suspended, :account_status)
       .to_h
       .symbolize_keys
+      .compact_blank
+      .reverse_merge(default_filters)
   end
 
   # Filter active if it's a form submission
@@ -209,5 +212,13 @@ class FacilityAccountsController < ApplicationController
   end
 
   helper_method :filters_active?
+
+  def default_filters
+    if SettingsHelper.feature_on?("accounts.account_tabs")
+      { suspended: "false" }
+    else
+      { account_status: "active" }
+    end
+  end
 
 end
--- a/app/controllers/facility_accounts_controller.rb
+++ b/app/controllers/facility_accounts_controller.rb
@@ -9,31 +9,35 @@ class FacilityAccountsController < ApplicationController
   admin_tab     :all
   before_action :check_acting_as
   before_action :init_current_facility
-  before_action :init_account, except: :search_results
+  before_action :init_account, except: :index
   before_action :build_account, only: [:new, :create]
   before_action :set_facility_accounts_for_user, only: [:accounts_available_for_order]
-  before_action :set_account_types, only: [:index, :search_results]
+  before_action :set_account_types, only: :index
 
   authorize_resource :account, except: [:accounts_available_for_order]
 
-  layout "two_column"
   before_action { @active_tab = "admin_users" }
+  layout "two_column"
 
   # GET /facilties/:facility_id/accounts
   def index
-    accounts = Account.with_orders_for_facility(current_facility)
+    searcher = AccountSearcher.new(
+      filter_params[:search_term],
+      scope: account_scope,
+      filter_params:,
+    )
 
-    if SettingsHelper.feature_on?("accounts.account_tabs")
-      accounts = if params[:suspended] == "true"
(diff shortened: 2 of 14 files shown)


app/views/facility_accounts/index.html.haml
```diff
@@ -32,16 +33,19 @@
...
 
 %hr
 
 %ul.list-inline
   - if SettingsHelper.feature_on?("accounts.edit_accounts") && current_ability.can?(:create, Account)
     %li= link_to t(".add_account"), new_account_user_search_facility_accounts_path, class: "btn-add"
-  - if !@accounts.blank? && SettingsHelper.feature_off?("accounts.account_tabs")
-    %li.toggle_expired_accts_btn--js.toggle_expired_btn.btn.btn-primary= t("views.facility_accounts.account_table.hide_text")
-#result
+  - if @accounts.present? && SettingsHelper.feature_off?("accounts.account_tabs")
```
LeticiaErrandonea: I think this should only check for the feature flag. If all the payment sources in a facility are expired and I hide them, it's not obvious how to see the full list again (this button is hidden)

app/helpers/accounts_helper.rb
```diff
@@ -2,6 +2,24 @@
 
 module AccountsHelper
 
+  def toggle_expired_account_btn
```
LeticiaErrandonea, line 5: I think this should work the other way around: by default hide the expired ones, as it was  the behavior

app/views/facility_accounts/index.html.haml
```diff
@@ -8,16 +8,17 @@
 - if SettingsHelper.feature_on?("accounts.account_tabs")
   = render "shared/accounts_tab_nav"
 
-= form_tag search_results_facility_accounts_path, id: "ajax_form", method: :get, class: "js--searchForm" do
+= form_tag facility_accounts_path, method: :get, class: "account-search-form" do
   = label_tag :search_term, text("facility_accounts.index.label.search_term")
   = hidden_field_tag :email, current_user.email, disabled: true
   = hidden_field_tag :format, params[:format], disabled: true
+  = hidden_field_tag :account_status, params[:account_status]
```
LeticiaErrandonea, line 15: We have this same field in line 33, which is only used when the feature flag is on

app/helpers/accounts_helper.rb
```diff
@@ -2,6 +2,24 @@
...
+    toggle_params = request.query_parameters.dup
```
LeticiaErrandonea: Should we clean the page param?

app/controllers/facility_accounts_controller.rb
```diff
@@ -82,46 +86,6 @@ def update
...
-  def search_results
-    account_scope = Account.for_facility(current_facility)
-    filter_params = {
-      account_type: params[:account_type],
-      suspended: params[:suspended],
-      account_status: params[:account_status],
-    }
-
-    if SettingsHelper.feature_on?("accounts.account_tabs") && params[:search_term].blank?
-      account_scope = AccountSearcher.new("", scope: account_scope, filter_params:).filtered_scope
```
LeticiaErrandonea, line 95: nit: now `filtered_scope` is unused and can be removed

app/controllers/facility_accounts_controller.rb
```diff
@@ -212,22 +176,38 @@ def set_account_types
...
+  # If we are filtering then return all accounts for facility.
+  def account_scope
+    if filters_active?
+      Account.for_facility(current_facility)
+    else
+      Account.with_orders_for_facility(current_facility)
+    end
+  end
+
+  def filter_params
```
LeticiaErrandonea, line 199: nit: should we memoize this?

app/assets/javascripts/user_accounts.js.erb
```diff
@@ -3,6 +3,11 @@
  */
 window.addEventListener("DOMContentLoaded", () => {
   const toggleExpiredAcctsBtn = document.querySelector(".toggle_expired_accts_btn--js");
+
+  if (!toggleExpiredAcctsBtn) {
```
LeticiaErrandonea, line 7: We can remove a couple `if (toggleExpiredAcctsBtn)` in this file

app/views/facility_accounts/index.html.haml
```diff
@@ -32,16 +33,19 @@
...
 %ul.list-inline
   - if SettingsHelper.feature_on?("accounts.edit_accounts") && current_ability.can?(:create, Account)
     %li= link_to t(".add_account"), new_account_user_search_facility_accounts_path, class: "btn-add"
-  - if !@accounts.blank? && SettingsHelper.feature_off?("accounts.account_tabs")
-    %li.toggle_expired_accts_btn--js.toggle_expired_btn.btn.btn-primary= t("views.facility_accounts.account_table.hide_text")
-#result
+  - if @accounts.present? && SettingsHelper.feature_off?("accounts.account_tabs")
+    %li.pull-right= toggle_expired_account_btn
+
+= turbo_frame_tag("account-search-result") do
```
LeticiaErrandonea: With this change, we no longer have the page in the URL, so refreshing always takes you back to page 1

app/models/account.rb
```diff
@@ -50,11 +50,12 @@ def account_price_groups
   accepts_nested_attributes_for :account_users
   has_many :log_events, as: :loggable
 
-  scope :active, -> { where("expires_at > ?", Time.current.beginning_of_day).where(suspended_at: nil) }
+  scope :active, -> { not_expired.not_suspended }
   scope :active_at, ->(time) { where("expires_at >= ?", time&.beginning_of_day).where(suspended_at: nil) }
   scope :suspended, -> { where.not(suspended_at: nil) }
   scope :not_suspended, -> { where(suspended_at: nil) }
   scope :expired, -> { where("expires_at < ?", Time.current.beginning_of_day) }
+  scope :not_expired, -> { where("expires_at > ?", Time.current.beginning_of_day) }
```
LeticiaErrandonea: I don't know if it's an actual bug but I'd expect expired + not_expired to have all accounts, and one that expired today at 0:00 won't be included in either

LeticiaErrandonea (commented):
Looking good, left some comments

spec/services/account_searcher_spec.rb
```diff
@@ -126,27 +126,58 @@
...
-      let(:searcher) do
-        described_class.new("", scope: Account.all, filter_params: { suspended: "true" })
+    context "with suspended false filter" do
+      let(:filter_params) do
+        { suspended: "false" }
       end
 
-      it "includes suspende account" do
-        expect(searcher.results).to match([suspended_account])
+      it "includes suspended account" do
```
LeticiaErrandonea: nit: I think this should say that _excludes_

app/helpers/accounts_helper.rb
```diff
@@ -2,6 +2,24 @@
 
 module AccountsHelper
 
+  def toggle_expired_account_btn
+    toggle_params = request.query_parameters.dup.except(:page)
+
+    if params[:account_status].blank?
+      toggle_params[:account_status] = "all"
```
LeticiaErrandonea, line 11: nit: I don't think it's relevant now, but "all" isn't defined or used anywhere else, so any word would work. Maybe we can leave a comment or something saying that?

app/views/facility_accounts/_account_table.html.haml
LeticiaErrandonea: This can be removed, right?

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

diebas (approved):

PR #6415, opened 2026-08-13
[NU-489] Invoice reconciliation

diebas:
# Notes

Facility administrators can now reconcile an entire invoice at once from the Reconcile Purchase Orders page: select an invoice from a drop-down, enter a reconciliation note and Transaction ID, and every order on that invoice is reconciled without having to check them off one by one. A Transaction ID is now required when reconciling this way. A "Display Individual Items" button opens the existing per-item page with the invoice already applied to the Statements filter, which can still be changed.

Behind the `billing.two_tier_reconciliation` feature flag, off by default — no change for schools that don't enable it.

[NU-489 | Invoice Reconciliation](https://universe-of-universities.atlassian.net/browse/NU-489)

# Screenshot

Screenshot 2026-08-13 at 9 09 18 AM

Screenshot 2026-08-13 at 9 01 44 AM

https://github.com/user-attachments/assets/b7ebdd36-fb53-438d-a165-dc07528d2e93

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,16 +4,6 @@ Because we use squash and merge, you should be able to see the changes by lookin
 at the [commit log](https://github.com/tablexi/nucore-open/commits/master). However, we have begun keeping track of breaking changes
 or optional rake tasks.
 
-### Addition of `two_tier_reconciliation` feature flag ([#6415](https://github.com/tablexi/nucore-open/pull/6415))
-
-Schools with `billing.two_tier_reconciliation: false` (the default) are not affected by this change.
-
-Schools with `billing.two_tier_reconciliation: true` get a two-tiered reconciliation flow. The reconciliation page (e.g. Reconcile Purchase Orders) opens on an invoice drop-down instead of the list of order details. Selecting an invoice and submitting reconciles every unreconciled order detail on it, applying the same reconciliation note and deposit number to all of them. A "Display Individual Items" button opens the previous per-item page with the invoice applied to the Statements filter, which can still be changed.
-
-When `billing.show_reconciliation_deposit_number` is also on, the deposit number (labeled per school through the `deposit_number_prefix` translation, e.g. "Transaction ID") is required to reconcile a whole invoice. It remains optional on the per-item page. When that flag is off, the field is not rendered and is not required, matching the previous behavior.
-
-Specs that exercise the per-item reconciliation page directly should set `feature_setting: { "billing.two_tier_reconciliation" => false }`, otherwise they will render the invoice drop-down instead.
-
 ### Addition of `PriceGroupDiscount` ([#3397](https://github.com/tablexi/nucore-open/pull/3397), [#3594](https://github.com/tablexi/nucore-open/pull/3594))
 
 Rather than one discount being set on a schedule rule, each global price group has its own discount (`PriceGroupDiscount`) for each schedule rule.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@ Because we use squash and merge, you should be able to see the changes by lookin
 at the [commit log](https://github.com/tablexi/nucore-open/commits/master). However, we have begun keeping track of breaking changes
 or optional rake tasks.
 
+### Addition of `two_tier_reconciliation` feature flag ([#6415](https://github.com/tablexi/nucore-open/pull/6415))
+
+Schools with `billing.two_tier_reconciliation: false` (the default) are not affected by this change.
+
+Schools with `billing.two_tier_reconciliation: true` get a two-tiered reconciliation flow. The reconciliation page (e.g. Reconcile Purchase Orders) opens on an invoice drop-down instead of the list of order details. Selecting an invoice and submitting reconciles every unreconciled order detail on it, applying the same reconciliation note and deposit number to all of them. A "Display Individual Items" button opens the previous per-item page with the invoice applied to the Statements filter, which can still be changed.
+
+When `billing.show_reconciliation_deposit_number` is also on, the deposit number (labeled per school through the `deposit_number_prefix` translation, e.g. "Transaction ID") is required to reconcile a whole invoice. It remains optional on the per-item page. When that flag is off, the field is not rendered and is not required, matching the previous behavior.
+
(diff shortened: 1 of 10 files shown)


app/views/facility_accounts_reconciliation/index_by_statement.html.haml
```diff
@@ -0,0 +1,45 @@
...
+      .row.table-actions.form-horizontal
+        .form-group
+          = label_tag :bulk_deposit_number, text("facility_accounts_reconciliation.index.bulk_deposit_number"), class: "control-label col-md-2"
+          .col-md-4
+            = text_field_tag :bulk_deposit_number, params[:bulk_deposit_number], class: "form-control"
+
+    .row.table-actions.form-horizontal
+      .col-md-12.text-right
+        = submit_tag t(".submit"), name: "reconcile_statement", class: "btn btn-primary", data: { disable_with: t(".submit") }
+        = submit_tag t(".display_items"), name: "display_items", class: "btn btn-default"
```
joaquinco, line 49: I think this can be a link to the index path with `show_items: true` so we don't need the redirect in the update action

joaquinco (approved):
Looks good. I think we should fix the ff `two_tier_reconciliation` on existing reconciliation specs since some might fail if that flag is off (i.e c2po system specs)

joaquinco (approved):
One more thing, should we add a notice if there're no invoices to reconcile? We are doing so in other billing pages

app/views/facility_accounts_reconciliation/index_by_statement.html.haml
diebas: I didn't use a link because the idea of this is to be dynamic. It loads the page with the "Statements filter pre-populated if an invoice is already selected" as required in the ticket description

config/settings.yml
```diff
@@ -178,6 +178,7 @@ feature:
     set_statement_search_start_date: false
     show_reconcile_credit_cards: false
     show_reconciliation_deposit_number: <%= ENV.fetch("RECONCILIATION_DEPOSIT_NUMBER", false) %>
+    two_tier_reconciliation: false
```
LeticiaErrandonea, line 181: Can you add documentation for the flag?

app/views/facility_accounts_reconciliation/index_by_statement.html.haml
```diff
@@ -0,0 +1,48 @@
...
+              options_from_collection_for_select(@statements, :id, :invoice_number, params.dig(:search, :statements)&.first),
+              include_blank: true,
+              id: :search_statements,
+              class: "form-control"
+
+      .row.table-actions.form-horizontal
+        .form-group
+          = label_tag :reconciled_at, OrderDetail.human_attribute_name(:reconciled_at), class: "control-label col-md-2"
+          .col-md-4
+            = date_field_tag :reconciled_at, Date.current, class: "form-control", max: Date.current
```
LeticiaErrandonea: I think we should have a min, that uses `journal_or_statement_date`

LeticiaErrandonea (commented):
Dev QA: 
If I use the previous flow and leave the Reconciliation date empty (or something else that prevents the order details from being reconciled), I get the error message in the new page

Image

CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ Because we use squash and merge, you should be able to see the changes by lookin
...
+
+Schools with `billing.two_tier_reconciliation: false` (the default) are not affected by this change.
+
+Schools with `billing.two_tier_reconciliation: true` get a two-tiered reconciliation flow. The reconciliation page (e.g. Reconcile Purchase Orders) opens on an invoice drop-down instead of the list of order details. Selecting an invoice and submitting reconciles every unreconciled order detail on it, applying the same reconciliation note and deposit number to all of them. A "Display Individual Items" button opens the previous per-item page with the invoice applied to the Statements filter, which can still be changed.
+
+When `billing.show_reconciliation_deposit_number` is also on, the deposit number (labeled per school through the `deposit_number_prefix` translation, e.g. "Transaction ID") is required to reconcile a whole invoice. It remains optional on the per-item page. When that flag is off, the field is not rendered and is not required, matching the previous behavior.
+
+Specs that exercise the per-item reconciliation page directly should set `feature_setting: { "billing.two_tier_reconciliation" => false }`, otherwise they will render the invoice drop-down instead.
```
joaquinco: We should be able to add a short description for the new feature flag on the feature flags document

app/views/facility_accounts_reconciliation/index_by_statement.html.haml
```diff
@@ -0,0 +1,52 @@
...
+
+- content_for :sidebar do
+  = render partial: "admin/shared/sidenav_billing", locals: { sidenav_tab: account_route }
+
+%h2= t(".head", model: account_class.model_name.human(count: 2))
+
+.col-md-12
+  - if @statements.none?
+    %p.notice= t("facility_accounts_reconciliation.index.none", model: account_class.model_name.human(count: 2).downcase)
+  - else
```
joaquinco: Can we show the form even if there're no statements?

app/views/facility_accounts_reconciliation/index_by_statement.html.haml
```diff
@@ -0,0 +1,51 @@
...
+
+- content_for :sidebar do
+  = render partial: "admin/shared/sidenav_billing", locals: { sidenav_tab: account_route }
+
+%h2= t(".head", model: account_class.model_name.human(count: 2))
+
+.col-md-12
+  - if @statements.none?
+    %p.notice= t("facility_accounts_reconciliation.index.none", model: account_class.model_name.human(count: 2).downcase)
+  = form_tag([:update, account_route.to_sym, FacilityAccount], method: :post) do
```
LeticiaErrandonea, line 10: I think all this should be in the `else` block

LeticiaErrandonea (commented):
It's looking good, I left a comment

app/views/facility_accounts_reconciliation/index_by_statement.html.haml
joaquinco: I requested the form to be shown when there're no statements (as it's done in other billing pages). Maybe we can move the notice to below the form for consistency

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

PR #6427, opened 2026-08-18
[OSU-231] Hide subsidy to final users

LeticiaErrandonea:
# Notes
* Hide subsidy to final users

[OSU-231 | Hide Subsidy Adjustments from Customer-Facing RELMS Views ](https://universe-of-universities.atlassian.net/browse/OSU-231)

# Additional Context
It was easier to implement here and turn on the feature flag than to implement in the school's repo avoiding the main app.

# Screenshot
## Admin side
image

## User side
image

## User's receipt
image

### DIFF
--- a/app/controllers/concerns/order_details_csv_export.rb
+++ b/app/controllers/concerns/order_details_csv_export.rb
@@ -14,7 +14,14 @@ module OrderDetailsCsvExport
       order_detail_ids:,
       date_range_field: @date_range_field,
       label_key_prefix: @label_key_prefix,
+      show_price_breakdown: show_price_breakdown_in_csv?,
     )
+  end
+
+  private
+
+  def show_price_breakdown_in_csv?
+    true
   end
 
 end
--- a/app/controllers/concerns/order_details_csv_export.rb
+++ b/app/controllers/concerns/order_details_csv_export.rb
@@ -21,7 +21,8 @@ module OrderDetailsCsvExport
   private
 
   def show_price_breakdown_in_csv?
-    true
+    SettingsHelper.feature_off?("pricing.hide_subsidy_from_customers") ||
+      can?(:manage_billing, current_facility)
   end
 
 end
--- a/app/helpers/orders_helper.rb
+++ b/app/helpers/orders_helper.rb
@@ -24,6 +24,10 @@ module OrdersHelper
     end
   end
 
+  def show_price_breakdown?
+    SettingsHelper.feature_off?("pricing.hide_subsidy_from_customers")
+  end
+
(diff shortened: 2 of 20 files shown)


app/helpers/orders_helper.rb
```diff
@@ -24,6 +24,10 @@ def display_cost_prefix_for_order_detail(order_detail)
     end
   end
 
+  def show_price_breakdown?
+    SettingsHelper.feature_off?("pricing.hide_subsidy_from_customers")
```
diebas, line 28: should we include acting_as? here?

app/mailers/purchase_notifier.rb
```diff
@@ -40,6 +41,7 @@ def order_receipt(args)
     @user = args[:user]
     @order = args[:order]
     @greeting = text("views.purchase_notifier.order_receipt.intro")
+    @show_price_breakdown = SettingsHelper.feature_off?("pricing.hide_subsidy_from_customers")
```
diebas, line 44: Isn't the helper accessible from here?

app/views/order_details/show.html.haml
```diff
@@ -92,18 +93,21 @@
           %td
             = order_detail_presenter.order_status_display_name
             = order_detail_status_badges(@order_detail)
+          - currency_columns = show_price_breakdown? ? (@order_detail.has_subsidies? ? 3 : 2) : 1
```
diebas, line 96: NIT: I think we can define this inside the elsif block where its used so we don't load it when it's not used

diebas (approved):
Looks good!

I also have a doubt: customers can still view the subsidy on the transaction History CSV export (TransactionsController → Reports::AccountTransactionsReport:78,123) with the flag on. I know the ticket only mentions the views, but maybe we can check if they want to hide it there too

app/helpers/orders_helper.rb
LeticiaErrandonea: Turn out we don't 😅

app/controllers/concerns/order_details_csv_export.rb
```diff
@@ -14,7 +14,14 @@ def handle_csv_search
...
       label_key_prefix: @label_key_prefix,
+      show_price_breakdown: show_price_breakdown_in_csv?,
     )
   end
 
+  private
+
+  def show_price_breakdown_in_csv?
+    true
+  end
```
joaquinco, line 26: This will make new/inprocess orders to include subsidy columns which I think they shouldn't since any staff can download them.

joaquinco (approved):

diebas (approved):

PR #6458, opened 2026-09-04
[OSU-242] Public estimates

diebas:
# Notes

Visitors can now get a cost estimate without logging in. A "Get an Estimate" link next to Support opens a page where they pick whether they are internal or external, choose a facility, and enter quantities for the products they are interested in, then see the estimated cost. Off by default behind the `public_estimates` feature flag.

[OSU-242 | Public estimates](https://universe-of-universities.atlassian.net/browse/OSU-242)

# Additional Context

**Nothing is persisted.** Phase 1 calculates and displays only. `PublicEstimatesController#show` builds an in-memory `Estimate` with `estimate_details.build` and calls `assign_price_policy_and_cost` on each detail — no new models and no migration. That method was already able to run on unsaved records; the specs pin that it resolves the policy, computes the same cost as a persisted equivalent, and writes nothing.

**Price group selection.** `PriceGroup.for_public_estimate` maps the visitor's answer to a global price group: internal → `base`, external non-profit → `secondary_external` falling back to `external`, everything else → `external`. `PriceGroup.secondary_external` is new and returns `nil` when `price_group.name.external_2` is unset, so schools without a second external group only see two options instead of three.

**Cost does not depend on the user.** Verified through `PricePolicySelector` and all three `estimate_cost_from_estimate_detail` implementations — they read only price group, quantity and duration, which is what makes an accountless estimate possible.

**Existing admin views were not reused.** `facility_estimates/show.html.haml` calls `includes` and `maximum` on the association, which hit the database and therefore return nothing for an unsaved estimate.

Products are listed from `available_for_estimates`, excluding bundles. A product with no purchaseable, in-date policy on the selected price group is shown as "No public rate available" rather than hidden, so visitors can tell the difference between a product that is unavailable and one that simply has no public rate.

# Screenshot

https://github.com/user-attachments/assets/fe832aa5-3fdd-4301-b8fd-77b9a44d0762

### DIFF
--- a/app/controllers/public_estimates_controller.rb
+++ b/app/controllers/public_estimates_controller.rb
@@ -1,0 +1,48 @@
+# frozen_string_literal: true
+
+class PublicEstimatesController < ApplicationController
+
+  skip_before_action :authenticate_user!
+
+  def show
+    @facilities = Facility.active.alphabetized
+    @facility = @facilities.find_by(id: params[:facility_id])
+    @products = @facility ? public_products : Product.none
+    @price_group = PriceGroup.for_public_estimate(
+      customer_type: params[:customer_type],
+      profit_status: params[:profit_status],
+    )
+    @estimate = build_estimate if @price_group && requested_quantities.any?
+    @total = @estimate.estimate_details.sum { |estimate_detail| estimate_detail.cost || 0 } if @estimate
+  end
+
+  private
+
+  def public_products
+    @facility.products.active.available_for_estimates.where.not(type: "Bundle").alphabetized
+  end
+
+  def requested_quantities
+    @requested_quantities ||=
+      params[:quantities].presence&.to_unsafe_h&.select { |_id, quantity| quantity.to_i.positive? } || {}
+  end
+
+  def build_estimate
+    estimate = Estimate.new(facility: @facility, price_group: @price_group)
+
+    requested_quantities.each do |product_id, quantity|
+      product = @products.find_by(id: product_id)
+      next if product.blank?
+
+      estimate.estimate_details.build(
+        product:,
+        quantity: quantity.to_i,
+        duration: params.dig(:durations, product_id).presence,
+      )
+    end
+
+    estimate.estimate_details.each(&:assign_price_policy_and_cost)
+    estimate
+  end
+
+end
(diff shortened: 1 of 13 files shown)


app/controllers/public_estimates_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  private
+
+  def customer_type_options
+    return [[t(".internal"), "internal"], [t(".external"), "external"]] if PriceGroup.secondary_external.blank?
+
+    [
+      [t(".internal"), "internal"],
+      [t(".external_for_profit"), "external"],
+      [t(".external_non_profit"), "external_non_profit"],
+    ]
```
joaquinco: can we have this predefined to base and external and allow each school override it? what do you think?

app/controllers/public_estimates_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+               .distinct.pluck(:product_id)
+  end
+
+  def public_products
+    @facility.products.active.available_for_estimates.where.not(type: "Bundle").alphabetized
+  end
+
+  def requested_quantities
+    @requested_quantities ||=
+      params[:quantities].presence&.to_unsafe_h&.select { |_id, quantity| quantity.to_i.positive? } || {}
```
joaquinco: let's use strong parameters instead of to_unsafe_h

app/controllers/public_estimates_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  def build_estimate
+    estimate = Estimate.new(facility: @facility, price_group: @price_group)
+
+    requested_quantities.each do |product_id, quantity|
+      product = @products.find_by(id: product_id)
```
joaquinco: is there a chance we can make a single db query?

config/locales/en.yml
```diff
@@ -121,10 +121,11 @@ en:
...
       customer_type: I am
-      internal: Internal
-      external: External
-      external_for_profit: External - for-profit
-      external_non_profit: External - non-profit
+      customer_types:
+        base: Internal
+        external: External
+        external_2: External - non-profit
+        cancer_center: Cancer Center
```
joaquinco, line 127: let's make it consistent with the price groups we have defined for open (without external_2)

joaquinco (approved):
Looks good!

PR #6465, opened 2026-09-09
[NUOPEN-438] Disable all journal create buttons on click

joaquinco:
# Notes

The journal creation form has several submit buttons depending on some conditions. Disable all submit buttons upon submition using disable with from Rails UJS.

This was also done for the statements create page since they share the same form.

Part of [NUOPEN-438 | Double journal creation fix](https://universe-of-universities.atlassian.net/browse/NUOPEN-438)

### DIFF
--- a/app/javascript/facility_journal.js
+++ b/app/javascript/facility_journal.js
@@ -2,16 +2,16 @@ document.addEventListener("DOMContentLoaded", function() {
   const selectAllLink = document.querySelector(".js--select_all");
   const table = document.querySelector("table.js--transactions-table");
   const submitDiv = document.querySelector(".submit");
-  const journalCreationSubmitButton = document.querySelector(".js--journal-creation__submit");
+  const journalCreationSubmitButtons = document.querySelectorAll(".js--journal-submit");
   const journalCreationHelperText = document.querySelector(".js--journal-creation__helper");
   let earliestFulfilledAtDate;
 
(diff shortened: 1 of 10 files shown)


app/javascript/facility_journal.js
```diff
@@ -45,9 +44,15 @@ document.addEventListener("DOMContentLoaded", function() {
...
   function handleSubmit(event) {
-    event.preventDefault();
-    event.target.disabled = true;
-    journalCreationHelperText.toggleAttribute("hidden");
+    disableSubmitBtns();
     document.getElementById("journals_create_form").submit();
   }
+
+  function disableSubmitBtns() {
+    $.rails.disableFormElement($(".js--journal-submit"));
```
diebas, line 52: Could we also disable the form’s submit controls here? Otherwise, users can dismiss the 90-day modal, change the date below the warning threshold, and click Create again for the same orders.

joaquinco: good catch

diebas (approved):

app/javascript/facility_journal.js
```diff
@@ -45,9 +44,16 @@ document.addEventListener("DOMContentLoaded", function() {
   }
 
   function handleSubmit(event) {
```
LeticiaErrandonea: nit: `event` is now unused

LeticiaErrandonea (approved):
LGTM!

PR #6466, opened 2026-09-09
[NUOPEN-419] Fix order files download

diebas:
# Notes                                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                               
   Users with read access can download files attached to orders in their facility, including individual results and ZIP archives.     
                                                                                                                                                                                                                                             [NUOPEN-419 | Granular permissions: cannot download order files](https://universe-of-universities.atlassian.net/browse/NUOPEN-419)

### DIFF
--- a/app/controllers/order_detail_stored_files_controller.rb
+++ b/app/controllers/order_detail_stored_files_controller.rb
@@ -7,7 +7,9 @@ class OrderDetailStoredFilesController < ApplicationController
   customer_tab  :all
 
   before_action :init_order_detail
-  authorize_resource class: OrderDetail
+  authorize_resource class: OrderDetail, except: [:sample_results, :sample_results_zip, :template_results]
+  authorize_resource class: OrderDetail, instance_name: :order_detail, id_param: :order_detail_id,
+                     only: [:sample_results, :sample_results_zip, :template_results]
 
   def sample_results
     redirect_to @order_detail.stored_files.sample_result.find(params[:id]).download_url, allow_other_host: true
--- a/app/controllers/order_detail_stored_files_controller.rb
+++ b/app/controllers/order_detail_stored_files_controller.rb
@@ -7,9 +7,7 @@ class OrderDetailStoredFilesController < ApplicationController
   customer_tab  :all
 
   before_action :init_order_detail
-  authorize_resource class: OrderDetail, except: [:sample_results, :sample_results_zip, :template_results]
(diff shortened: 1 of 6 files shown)


app/controllers/order_detail_stored_files_controller.rb
```diff
@@ -7,7 +7,9 @@ class OrderDetailStoredFilesController < ApplicationController
   customer_tab  :all
 
   before_action :init_order_detail
-  authorize_resource class: OrderDetail
+  authorize_resource class: OrderDetail, except: [:sample_results, :sample_results_zip, :template_results]
+  authorize_resource class: OrderDetail, instance_name: :order_detail, id_param: :order_detail_id,
```
joaquinco: we should always authorize on the order_detail object since these are all member actions

spec/lib/ability_spec.rb
```diff
@@ -999,6 +999,15 @@
       it_is_allowed_to(:index, StoredFile)
       it_is_allowed_to(:product_survey, StoredFile)
 
+      context "when downloading another facility's order files" do
```
joaquinco: can we add specs for allowed cases? (when the permission is granted for the order detail's facility)

app/lib/ability.rb
```diff
@@ -471,8 +471,9 @@ def cross_core_abilities(user, resource, controller)
     if resource.is_a?(OrderDetail)
       project = resource.order.cross_core_project
 
-      if project.present?
-        can [:add_accessories, :new, :show, :update, :cancel, :template_results], OrderDetail if can_manage_cross_core_orders_at?(user, project.facility)
+      if project.present? && can_manage_cross_core_orders_at?(user, project.facility)
+        can [:add_accessories, :new, :show, :update, :cancel, :template_results], OrderDetail
```
LeticiaErrandonea, line 475: Should we also add this `order: { cross_core_project_id: project.id }`?

LeticiaErrandonea (commented):
Dev QA

* As an Account owner, I can see the links to download the service's files but when I click I get Access denied:

Image

Image

* Same with a Facility Senior Staff when viewing an order (from another facility) that belongs to a project from their own facility. I don't know if any of our clients still has the feature flag off _and_ has Senior Staff.

Image

In master, they are allowed to download.

LeticiaErrandonea (approved):
LGTM, thanks for the changes!

joaquinco (approved):

PR #6468, opened 2026-09-10
[NUOPEN-438] Use mutex on journal creation

joaquinco:
# Notes

Use facility as mutex to detect duplicate open journal. Since journal row creation can take several seconds I've splitted the journal record creation within the mutex and journal rows creation after it.

Also:
- Remove truncation of long referrer header since order ids are sent in the request body
- Add basic expires_at validation for test accounts
- Add rspec filter to use test account for internal billing

[NUOPEN-438 | Prevent duplicate journal]()

### DIFF
--- a/app/models/journal.rb
+++ b/app/models/journal.rb
@@ -1,9 +1,6 @@
 # frozen_string_literal: true
 
-require "set"
-
 class Journal < ApplicationRecord
-
   class CreationError < NUCore::Error; end
 
   module Overridable
@@ -41,6 +38,8 @@ class Journal < ApplicationRecord
 
   attr_accessor :order_details_for_creation
 
+  attribute :defer_journal_row_creation, default: false
+
   belongs_to :facility, optional: true
   belongs_to :created_by_user, class_name: "User", foreign_key: :created_by
 
@@ -60,7 +59,9 @@ class Journal < ApplicationRecord
   validate :journal_date_cannot_be_before_last_fulfillment, on: :create, if: :has_order_details_for_creation?
   validates_with JournalDateMustBeAfterCutoffs, on: :create
   before_validation :set_facility_id, on: :create, if: :has_order_details_for_creation?
-  after_create :create_new_journal_rows, if: :has_order_details_for_creation?
+  after_create :create_new_journal_rows!, if: -> { has_order_details_for_creation? && !defer_journal_row_creation }
+
+  scope :pending, -> { where(is_successful: nil) }
 
   # Digs up journals pertaining to the passed in facilities
   #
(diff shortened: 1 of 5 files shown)


diebas (approved):

config/locales/en.controllers.yml
```diff
@@ -65,6 +65,7 @@ en:
       create:
         notice: "The journal file has been created successfully"
         more_errors: "There are more errors. Please resolve those above to see those that remain."
+        duplicate: "There's already a pending journal for this facility"
```
LeticiaErrandonea: Maybe use `!facility_downcase!`

app/models/journal.rb
```diff
@@ -60,7 +59,9 @@ def spreadsheet_filename
   validate :journal_date_cannot_be_before_last_fulfillment, on: :create, if: :has_order_details_for_creation?
   validates_with JournalDateMustBeAfterCutoffs, on: :create
   before_validation :set_facility_id, on: :create, if: :has_order_details_for_creation?
-  after_create :create_new_journal_rows, if: :has_order_details_for_creation?
+  after_create :create_new_journal_rows!, if: -> { has_order_details_for_creation? && !defer_journal_row_creation }
+
+  scope :pending, -> { where(is_successful: nil) }
```
LeticiaErrandonea, line 64: `Journal.facility_ids_with_pending_journals` joins the orders, so it wouldn't mark a journal as pending if for some reason (maybe while being created?) it doesn't have them. Should we consider that here as well?

app/services/journal_row_builder.rb
```diff
@@ -81,7 +81,7 @@ def valid?
 
   # Set an array of facility_ids with pending journals
   def pending_facility_ids
-    @pending_facility_ids ||= Journal.facility_ids_with_pending_journals
+    @pending_facility_ids ||= Journal.where.not(id: journal.id).facility_ids_with_pending_journals
```
LeticiaErrandonea: In which case do we need this change? I think the journal that is being created shouldn't be included in `Journal.facility_ids_with_pending_journals`.

app/models/journal.rb
```diff
@@ -41,6 +38,8 @@ def spreadsheet_filename
 
   attr_accessor :order_details_for_creation
 
+  attribute :defer_journal_row_creation, default: false
```
LeticiaErrandonea, line 41: Is there any reason for not using `attr_accessor`?

LeticiaErrandonea (commented):
Looking good, left a few questions

app/models/journal.rb
joaquinco: just to set the default. For this case it'd the equivalent indeed

app/services/journal_row_builder.rb
joaquinco: Right, this ends up not been needed

app/models/journal.rb
joaquinco: there's some code including the `facility_ids_with_pending_journals` method used to handle cross facility journals. For example, if another facility's pending journal includes an order from the current facility.

Those checks are mainly done within `JournalRowBuilder`. In practice, as far as I know, there're no cross facility journals and there's no way to create them so I think we're fine using the `is_successful` field as the only source of truth.

If there were cross facility journals, the mutex used here wouldn't prevent duplicates so we would have that problem. There's a ticket to review (and maybe remove) the cross-facility journal checks.

LeticiaErrandonea (approved):
LGTM, thanks for the changes and explanation!

PR #6488, opened 2026-09-18
[Fix] Handle nil price on estimates

joaquinco:
# Notes

Prior to time based price calculation changes in https://github.com/wyeworks/nucore-open/pull/6443, price calculations could return nil in the case all arguments where blank.

This change preservers that behavior and update time based product's estimate price to handle nils.

Part of [NUOPEN-435 | Price policies fixes](https://universe-of-universities.atlassian.net/browse/NUOPEN-435).


diebas (approved):
Should we add a test for this case?

diebas (approved):
Thanks for the changes

LeticiaErrandonea (approved):
LGTM

